Issue #73, opened 2026-05-02
OpenWRT re-architecture: e2e — fake-router container in staging compose


sameerparekh:
Replace the deleted-traffic-server e2e plan with one that exercises the new router surface.

- Add a small \`fake-router\` container to \`docker/docker-compose.yml\` (Python or shell + curl) that:
  - registers via \`POST /api/router/register\`
  - polls \`GET /api/router/policy\`
  - posts canned usage reports to \`POST /api/router/usage\`
  - emits events via \`POST /api/router/events\`
- Extend \`scripts/e2e-tests.sh\` to verify:
  - registering yields a usable token
  - usage records increment \`time_usage\` and update \`devices.last_seen_ip\`
  - hitting the daily limit causes the next \`/policy\` snapshot to mark the profile as time-limit-exceeded
  - paused profile / active schedule reflected in snapshot
  - public \`/blocked\` page renders with the right reason

Depends on the OpenWRT API + block page PRs.

sameerparekh:
**Blocked by:** #67, #68, #69, #70, #72 — fake router exercises the full chain end-to-end. Last step in the OpenWRT migration before #71 can land.

(GitHub has no strict-dependency field; tracking blockers via these comments.)

Issue #67, opened 2026-05-02
OpenWRT re-architecture: V2 migration (routers, traffic_reports, block_events, time_usage bytes)


sameerparekh:
Add migration \`api/resources/db/migration/V2__openwrt.sql\` introducing the schema needed for OpenWRT router integration.

Tables:
- \`routers (id uuid pk, name text, enrollment_token_hash text, token_hash text, last_seen_at timestamptz, last_etag text, created_at)\`
- \`traffic_reports (id, router_id fk, mac, ip, hostname, date, period_start, period_end, active_seconds int, bytes_in bigint, bytes_out bigint, unique(router_id, period_start, mac, hostname))\`
- \`block_events (id, mac, hostname, reason, ts)\`

Column additions:
- \`time_usage.bytes_in bigint default 0\`
- \`time_usage.bytes_out bigint default 0\`

Plus repo traits / impls / feature tests for the new tables. No HTTP routes yet.

Depends on #66 (spec).

PR #76, opened 2026-05-02
Bind api to LAN, keep postgres on loopback (closes #48)

### LINKED ISSUE #48: updqten the docker-compose so clients can connect from anywhere not just localhost (ie switch 127.0.0.1 to 0.0.0.0)

sameerparekh:
## Summary
- The api port mapping already uses `0.0.0.0:8080:8080` so LAN clients (including the upcoming OpenWRT router agent from #67–#73) can reach it directly — this PR keeps that and adds the missing piece.
- Postgres was previously published as `5433:5432`, which compose binds to `0.0.0.0` by default — i.e. exposed to every device on the LAN. Restrict it to `127.0.0.1:5433:5432` so it stays loopback-only.
- The `dns` container uses `network_mode: host` and connects to postgres via `127.0.0.1:5433`, so it is unaffected by the loopback restriction.

## Security
- Postgres credentials in this compose file are the staging defaults (`familydns`/`familydns`). Even though they are non-prod, exposing the DB port to the LAN is a footgun, so this PR explicitly pins it to loopback.
- No admin/internal endpoints are exposed by other services. The `traffic` service has no host port mapping.

## Test plan
- [x] `docker compose -f docker/docker-compose.yml config` shows `host_ip: 127.0.0.1` for postgres and `host_ip: 0.0.0.0` for api.
- [ ] `docker compose -f docker/docker-compose.yml up --build`, then from another LAN device: `curl http://<host-lan-ip>:8080/api/auth/login` returns 400/401 (api reachable).
- [ ] From another LAN device: `nc -vz <host-lan-ip> 5433` should fail / time out (postgres not reachable from LAN).

Closes #48

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -21,7 +21,9 @@ services:
       timeout: 2s
       retries: 30
     ports:
-      - "5433:5432"
+      # Bind to loopback only — postgres must never be reachable from the LAN.
+      # The dns container uses host networking and reaches it via 127.0.0.1.
+      - "127.0.0.1:5433:5432"
 
   api:
     build:


sameerparekh:
Closing without landing — we want to keep postgres LAN-accessible for debugging. Follow-up issue filed to rework the deploy so api+postgres ship together in a single container deploy now that dns/traffic no longer need DB access.

Issue #249, opened 2026-05-12
first_seen_mac race: device created with name='unknown' if conntrack fires before DHCP lease is written


sameerparekh:
Adjacent to #228 — diagnosed during the same debug session but separable.

## The race

In \`openwrt/files/usr/lib/lua/familydns/conntrack.lua\` \`handle_flow\`:

\`\`\`lua
if mac and not (ctx.reported_macs or {})[mac] then
  local lease = (ctx.leases or {})[mac]
  local ip, hostname
  if lease then ip = lease.ip; hostname = lease.hostname end
  batcher.add(M.build_first_seen_mac_event({
    mac = mac, ip = ip, hostname = hostname, ts = ctx.ts,
  }))
  ...
\`\`\`

\`ctx.leases\` is parsed from \`/tmp/dhcp.leases\` at the moment the *first* conntrack flow for that MAC arrives. If the kernel hasn't written the lease yet (the device just associated and the DHCPv4 exchange is still in flight), \`lease\` is \`nil\` → the event goes out with \`hostname = nil\`.

On the API side, \`api/src/routes/RouterIngestRoutes.scala\` \`applyDhcpOrFirstSeen\` does:

\`\`\`scala
case None    =>
  deviceRepo.upsertUnknown(mac, e.hostname.getOrElse("unknown"), e.ip, ts)
\`\`\`

So the device row gets \`name="unknown"\` and stays that way. There is no path that updates the device's name when a later \`dhcp_lease\` event arrives — \`applyDhcpOrFirstSeen\` only touches \`last_seen_*\` for known MACs.

## Repro recipe

1. Power off the test phone.
2. Wipe the API DB + restart the agent.
3. Power the phone on while \`logread -f | grep familydns\` is running.
4. Inspect timing: the first \`handle_flow\` for the phone's MAC arrives within ~200 ms of association, but dnsmasq's lease-script callback that writes \`/tmp/dhcp.leases\` can lag a beat behind.
5. With sufficient luck on a slow router, \`first_seen_mac\` is emitted with \`hostname=nil\` and the API ingests it as \`name="unknown"\`. The device then shows in the admin UI as 'unknown' even though DHCP later completes successfully.

## Fix candidates

Sketching, not committing to a specific approach:

[...]

Issue #130, opened 2026-05-10
CD: build and publish OpenWRT .ipk to GitHub Releases on push to main


sameerparekh:
## Context

Part of #123 (deploy architecture). `.github/workflows/openwrt-build.yml` already:
- Builds the `.ipk` on PRs touching `openwrt/`
- Creates a GitHub Release and attaches the `.ipk` on `v*` tag pushes
- (after #123 PR) Also builds on push to main (artifact only, no release)

This issue extends the workflow to also create a pre-release on push to main so the router auto-update script always has a fetchable URL, not just on tagged releases.

## Work

Options (pick one and document rationale):

**Option A — pre-release on main push**: on push to main, create/update a `latest-main` pre-release and attach the `.ipk`. The auto-update script targets this pre-release URL. Tagged releases become the stable channel.

**Option B — artifact + stable tag only**: keep the existing behavior (artifact on main push, real release on tag). The auto-update script targets `releases/latest` (only updated on tags). This means the router only auto-updates when a version tag is pushed, which is fine for now.

Recommendation: **Option B** — simpler, avoids GitHub Release churn on every commit, and tagged releases are the natural cadence for router updates.

Update `docs/deploy.md §2.2` with whichever approach is chosen.

## Acceptance

- `v*` tag push creates a GitHub Release with `familydns_<version>-1_all.ipk` attached.
- `curl -sf https://api.github.com/repos/sameerparekh/familydns/releases/latest | jsonfilter -e '@.assets[0].browser_download_url'` returns a valid `.ipk` URL.

Blocks: #123

Issue #129, opened 2026-05-10
Auto-update: systemd timer for API server Docker image


sameerparekh:
## Context

Part of #123 (deploy architecture). The API server runs as a docker compose stack on a Linux host. We want it to pick up new `latest` image pushes automatically without manual intervention.

## Work

Create and document two systemd unit files to be placed on the prod host:

**`/etc/systemd/system/familydns-update.service`**
```ini
[Unit]
Description=Pull and restart FamilyDNS API on new image
After=docker.service
Requires=docker.service

[Service]
Type=oneshot
WorkingDirectory=/opt/familydns/deploy
ExecStart=/usr/bin/docker compose -f docker-compose.prod.yml --env-file .env pull
ExecStart=/usr/bin/docker compose -f docker-compose.prod.yml --env-file .env up -d
```

**`/etc/systemd/system/familydns-update.timer`**
```ini
[Unit]
Description=FamilyDNS auto-update timer

[Timer]
OnBootSec=5min
OnUnitActiveSec=5min

[Install]
WantedBy=timers.target
```

Enable with `systemctl enable --now familydns-update.timer`.

Add these files to `deploy/systemd/` in the repo so they can be copied to the prod host, and document the setup in `docs/deploy.md` (the first-install section).

## Acceptance

- `systemctl list-timers` shows `familydns-update.timer` active.
- After pushing a new image to ghcr.io, the running container is updated within 5 minutes.
- `docker compose pull` is a no-op (no restart) when the digest hasn't changed.

Blocks: #123

PR #256, opened 2026-05-12
deploy: enable familydns-update systemd timer by default (fixes #254)

### LINKED ISSUE #254: deploy: wire familydns-update systemd timer into bootstrap so API auto-updates by default
The OpenWRT side auto-updates every 6 hours out of the box (cron entry installed by the package postinst). The API side has the same mechanism documented but **not enabled by default**: the operator has to manually install [the systemd timer/service from docs/deploy.md §1.3](docs/deploy.md) and run \`systemctl enable --now familydns-update.timer\`.

Net effect: a freshly-deployed home-server stays on whatever image it first booted with until someone \`ssh\`s in to run \`./update.sh\` — which means feature fixes (like #249 / PR #252's rename-on-dhcp-lease) don't reach production without manual intervention.

## What's needed

1. Ship the units in-tree under \`deploy/systemd/\`:
   - \`familydns-update.service\` — \`Type=oneshot\`, two \`ExecStart\` lines doing \`docker compose pull\` + \`up -d\` against \`/opt/familydns/deploy/docker-compose.prod.yml\`.
   - \`familydns-update.timer\` — \`OnBootSec=5min\` / \`OnUnitActiveSec=5min\` (matches the documented cadence).
2. Have [deploy/scripts/install.sh](deploy/scripts/install.sh) (or whichever bootstrap script the README points new deployers at) copy them into \`/etc/systemd/system/\` and \`systemctl enable --now familydns-update.timer\`.
3. Update [docs/deploy.md §1.3](docs/deploy.md) to describe this as automatic instead of opt-in, and document the disable path for operators who want manual control.
4. Test: a fresh \`./install.sh\` on a clean VM should end with \`systemctl is-active familydns-update.timer\` returning \`active\`.

## Tradeoff to consider

A 5-minute pull cadence is aggressive — every push to \`main\` reaches every deployed server within ~5 min. If we want a slower release cadence (tagged releases only, like the OpenWRT side does per #130), we should pin the deployed image tag to \`v*\` and only auto-update across tags. That decision belongs in this issue.

Discovered while reviewing the deploy story for PR #252.

sameerparekh:
## Summary

Closes #254. Today `docs/deploy.md §1.3` documents a systemd timer + service that polls ghcr.io every 5 minutes and `docker compose pull`s the latest API image, but operators have to install it manually. Net effect: fixes that merge to `main` (e.g. PR #252's rename-on-dhcp-lease for #249) never reach existing deployments until someone ssh's in to run `update.sh`.

This PR closes that deploy gap:

- Adds [`deploy/systemd/familydns-update.service`](deploy/systemd/familydns-update.service) and [`deploy/systemd/familydns-update.timer`](deploy/systemd/familydns-update.timer) as the canonical in-tree copies (matches the §1.3 documented unit content; sub-issue #129).
- [`deploy/install.sh`](deploy/install.sh) now fetches both units, installs them into `/etc/systemd/system/` (mode 0644 via `install -m 0644`, sudo when not root), and runs `systemctl daemon-reload && systemctl enable --now familydns-update.timer`. The step is idempotent — re-running `install.sh` is safe (overwrite + enable-already-enabled both succeed).
- For user-mode installs (`FAMILYDNS_PREFIX=$HOME/.familydns`), `install.sh` rewrites `WorkingDirectory=` in the service unit to the actual install dir before copying. The shipped repo copy uses `/opt/familydns/deploy` as the task spec requested.
- Updates [`docs/deploy.md §1.3`](docs/deploy.md) and [`docs/install-api.md §10`](docs/install-api.md) to describe auto-update as on-by-default and to document the disable path: `sudo systemctl disable --now familydns-update.timer`.

## Cadence

Kept at 5 min for now (matches what `docs/deploy.md` already documented and what #254 specified). Two things worth keeping an eye on, but neither felt strong enough to change cadence unilaterally:

- **ghcr.io rate limits.** Anonymous pulls are subject to GHCR rate limits; `docker compose pull` on a no-op (digest matches) still hits the manifest endpoint. At 12 polls/hour × N hosts this is plausibly fine but worth watching once we have more deployed instances.
- **Restart churn on every commit.** Today every `main` push → `latest` retag → ~5min later every deployed host restarts its API container. There's no canary; a bad merge reaches everyone within minutes. The OpenWRT side already addresses this by only auto-updating across `v*` tags (per #130); we could do the same on the API side by pinning the deployed tag to a `v*` semver. Happy to file that as a follow-up if you want.

## Test plan

- [x] `bash deploy/install.test.sh` — 12 passed, 0 failed (no regressions in tty / prompt / wait_for_api helpers).
- [x] `bash -n deploy/install.sh` — syntax clean.
- [ ] **Manual on a fresh VM** (need an env with systemd + docker; not available locally):
  - Run `curl -fsSL …/install.sh | bash` against a clean Linux host.
  - Assert `systemctl is-active familydns-update.timer` returns `active`.
  - Assert `systemctl is-enabled familydns-update.timer` returns `enabled`.
  - Assert `systemctl list-timers familydns-update.timer` shows next fire ≤ 5 min away.
  - Re-run `install.sh` (after deleting `.env` to satisfy the first-install guard, or against a second host); confirm the timer step doesn't error.
  - `sudo systemctl disable --now familydns-update.timer && systemctl is-active familydns-update.timer` → `inactive` (verifies the disable path documented in §1.3).
- [ ] `shellcheck deploy/install.sh` — not available locally; CI's `shell-tests` job will run it.
- [ ] `systemd-analyze verify deploy/systemd/familydns-update.{service,timer}` — not available locally (macOS dev box). Worth a quick run on the VM during the manual test plan above.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -465,6 +465,66 @@ else
   fi
 fi
 
+# ── 9b. Install auto-update timer ─────────────────────────────────────────
+#
+# Drops familydns-update.service + .timer into /etc/systemd/system/ and
+# enables the timer so the host pulls a fresh image every 5 minutes. This
+# is what closes the deploy gap for issue #254: without it, fixes that
+# land on `main` don't reach existing installs until someone ssh's in to
+# run update.sh by hand.
+#
+# Idempotent: `install -m 0644` overwrites cleanly, daemon-reload is a
+# no-op when nothing changed, and `systemctl enable --now` on an
+# already-enabled/active timer succeeds without error.
+step "Installing auto-update timer"
+
+SUDO=""
+if ! command -v systemctl >/dev/null 2>&1; then
+  warn "systemctl not found — skipping auto-update timer."
+  warn "Update manually with ${FAMILYDNS_INSTALL_DIR}/update.sh, or install a"
+  warn "cron entry that runs it on your preferred cadence."
+else
+  if [ "$(id -u)" -ne 0 ]; then
+    SUDO="sudo"
+  fi
+
+  UNIT_TMP="$(mktemp -d)"
+  if curl -fsSL "${REPO_RAW}/deploy/systemd/familydns-update.service" \
+        -o "${UNIT_TMP}/familydns-update.service" \
+     && curl -fsSL "${REPO_RAW}/deploy/systemd/familydns-update.timer" \
+        -o "${UNIT_TMP}/familydns-update.timer"; then
+    # The shipped unit hard-codes /opt/familydns/deploy as the
+    # WorkingDirectory (matches the documented system-wide install
+    # path). Patch it to the actual install dir so user-mode installs
+    # at $HOME/.familydns also get auto-update.
+    if [ "$FAMILYDNS_INSTALL_DIR" != "/opt/familydns" ]; then
+      sed -i.bak "s|^WorkingDirectory=/opt/familydns/deploy$|WorkingDirectory=${FAMILYDNS_INSTALL_DIR}|" \
+        "${UNIT_TMP}/familydns-update.service"
+      rm -f "${UNIT_TMP}/familydns-update.service.bak"
+    fi
+
+    install_ok=1
+    $SUDO install -m 0644 "${UNIT_TMP}/familydns-update.service" \
+      /etc/systemd/system/familydns-update.service || install_ok=0
+    $SUDO install -m 0644 "${UNIT_TMP}/familydns-update.timer" \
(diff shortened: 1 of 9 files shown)


sameerparekh:
Per follow-up: dropped both auto-update cadences to daily (API systemd timer `OnUnitActiveSec=1d`, OpenWRT cron `0 4 * * *`). Postinst on OpenWRT now rewrites any pre-existing `familydns-update` cron line on upgrade, so installed routers migrate without manual intervention. `openwrt/test/update_spec.sh` updated to assert the new expression.

Issue #286, opened 2026-05-13
screen time + sessions empty after fresh install — usage may not be flowing end-to-end


sameerparekh:
Found during round-2 UX audit (#271).

After a fresh install (clean DB, freshly enrolled router) the admin UI shows no screen-time data and no sessions — usage may not be flowing through end-to-end. #260 (sessions) and #261 (screen-time) round-1 fixes landed but the surfaces are still empty after generating real LAN traffic.

Quote from the audit: "we aren't seeing any screen time or sessions still. usage logs may not even be showing up?"

### Repro
1. Round-2 fresh install per #271 setup (API on 192.168.10.43, router on 192.168.1.1).
2. Generate traffic from at least one LAN device for several minutes.
3. Open admin UI → Sessions and Profiles → Screen Time.
4. Both surfaces show nothing.

### Diagnostic plan
**Iterate live**, not via PR-and-reinstall — the failure is somewhere in the agent→API→DB→UI pipeline and we'll need to bisect with `FAMILYDNS_DEBUG=1` + `logread`. Use:
- `scripts/deploy-api-from-branch.sh <branch>` to ship API-side patches to 192.168.10.43 without going through ghcr.
- `scripts/dev-push-router.sh push` to ship router-agent patches to 192.168.1.1 without re-packaging.

Trace order (per `familydns-live-debug` skill):
1. Router: does `logread -f | grep familydns` show usage POSTs at the configured interval (`usage_report_interval=300`)?
2. API: with `FAMILYDNS_LOG_LEVEL=DEBUG`, does `POST /api/router/usage` log each record?
3. DB: does `/api/debug/time_usage` show rows accumulating?
4. UI: does the sessions/screentime query return rows or 200-with-empty?

### Acceptance
- After generating traffic on a fresh install, Sessions surface lists actual sessions.
- Screen-time view shows non-zero `seconds_used` for the active profile.
- Root-cause writeup added to the PR description (so we don't ship a fix without understanding the regression vs round-1).

Originally surfaced in #271. Related: #260, #261.

PR #294, opened 2026-05-13
api: dedup time_usage delta per (mac, hostname) within a usage batch

sameerparekh:
## Summary
Follow-up to #291. Screen-time was over-counting because a usage batch carries one record per `(mac, dst_ip)` but `time_usage` is keyed by `(mac, hostname, date)` — and the old code called `incrementSecondsAndBytes` once per record, so every `dst_ip` in the same 5-min bucket added another `activeSeconds` to the row.

Group records by `(mac, hostname)` before applying the delta. Seconds advance by `max(activeSeconds)` per group (the bucket duration, observed at most once); bytes still sum across per-flow records.

## Root cause
After [#291](https://github.com/sameerparekh/familydns/pull/291) unblocked the usage pipeline, sameer noticed: Screen Time showed 160m for a MAC whose Sessions surface showed a single 10m session.

Numbers from the live deployment after the dedup bug:
```
time_usage.seconds_used     = 9600  (e2:5d:df:78:c5:2b / "unknown")
traffic_reports row count   = 2
SUM(traffic_reports.active_seconds) = 600
```
The MAC had 16 distinct `dst_ip` values in a 5-min bucket, all collapsing to `hostname="unknown"` (DoH / direct-IP traffic with no dnsmasq attribution). Each collapsed record added `activeSeconds=300` to `time_usage`, producing `16 * 300 = 4800` per bucket × 2 buckets = 9600.

`activeSeconds` is a bucket-duration heuristic (`render.lua`: any flow with `bytes>0` in the 5-min window counts as the full 300s). For a single `(mac, hostname)` within one bucket, those 300s should land in `time_usage` once, regardless of how many `dst_ips` mapped to that hostname. Sessions worked correctly because they derive from distinct `traffic_reports` rows (deduped by the `(router, mac, hostname, period_start, period_end)` unique key).

## Live verification
With the patched API deployed and `usage_report_interval=60s`, after one tick from a clean DB:
```
 device_mac        | domain  | seconds_used
-------------------+---------+--------------
 e2:5d:df:78:c5:2b | unknown | 300
 b0:41:6f:10:40:fd | unknown | 300
 04:d4:c4:1d:3e:98 | unknown | 300

 mac               | hostname | rows | SUM(active_seconds)
-------------------+----------+------+--------------------
 e2:5d:df:78:c5:2b | unknown  | 1    | 300
 b0:41:6f:10:40:fd | unknown  | 1    | 300
 04:d4:c4:1d:3e:98 | unknown  | 1    | 300
```
`time_usage.seconds_used` now matches `SUM(traffic_reports.active_seconds)` per MAC.

## Test plan
- [x] New unit test `usage: multiple records with same (mac, hostname) in one POST count seconds_used once`
- [x] Live deployment: Screen Time and Sessions show consistent durations for the same MAC
- [ ] Reviewer: confirm Screen Time number for a known MAC matches the sum of its session lengths

Related: #286, #291.

### DIFF
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -78,7 +78,7 @@ object Main extends ZIOAppDefault {
         deviceRepo,
         tlRepo,
         stlRepo,
-        usageRepo,
+        trafficRepo,
         extRepo,
         profileRepo,
         upRepo,
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -221,6 +221,17 @@ trait TrafficReportRepo {
    * \= Some(Nil)` returns an empty list (no devices match).
    */
   def listSessionRows(f: SessionFilter): Task[List[familydns.api.sessions.SessionRow]]
+
+  /**
+   * Minimal projection used by presence-based minute accounting (see
+   * [[familydns.api.presence.Presence]]). One row per (mac, period_start, hostname) for the given
+   * macs/date; the caller deduplicates by `(mac, period_start)` so per-hostname rows in a single
+   * bucket don't inflate total screen time.
+   */
+  def listPresenceRows(
+      macs: List[String],
+      date: LocalDate,
+  ): Task[List[familydns.api.presence.PresenceRow]]
 }
 
 trait BlockEventRepo {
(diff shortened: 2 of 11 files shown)


sameerparekh:
Pushed a follow-up commit `085ecc3` for the cross-hostname overcount: even after the (mac, hostname) dedup, a device active on multiple hostnames in the same 5-min window was still summed N× across hostnames at the read site. PolicyService and the screen-time routes now compute daily totals from `traffic_reports` presence — `COUNT(DISTINCT period_start) × 5min` per device — with the exempt-pattern filter applied per-bucket.

Design (matches what we discussed):
- `exempt_from_daily=true` → bucket excluded from the daily total only if **every** hostname in it matches an exempt pattern; per-site limit still ticks independently.
- `exempt_from_daily=false` → bucket counts toward the daily total (once, not N×) AND the per-site limit ticks.

A single non-exempt hostname in a bucket pulls the whole 5-min window into the total — actual wall-clock screen time.

New: `api/src/presence/Presence.scala` + `PresenceSpec` (9 unit tests). `TrafficReportRepo.listPresenceRows` is the new projection used by both `PolicyService.snapshot` and `TimeRoutes.buildProfile/DeviceTimeStatus`. `time_usage` rows are still maintained for byte counters and per-hostname seconds, but no longer drive the daily-total or per-site-cap read paths.

Live verification on the running deployment after deploy:
```
        mac        | presence_minutes
-------------------+------------------
 e2:5d:df:78:c5:2b | 35   (7 buckets × ~100 hostnames touched in those buckets)
 b0:41:6f:10:40:fd | 35
 04:d4:c4:1d:3e:98 | 35
 be:89:10:82:c7:4c | 30
 ...
```
`e2:5d:df:78:c5:2b` would have shown thousands of "screen-time minutes" under the old summing logic for those same 35 minutes of actual activity.

Existing `TimeApiSpec`, `RouterApiSpec`, `RouterDecisionSpec`, `RouterIngestSpec` reworked to seed via `traffic_reports` (small `seedTraffic` helper, non-overlapping 5-min buckets per hostname). All green.

Issue #346, opened 2026-05-13
VM e2e: automate enforcement + schedule validation suite


sameerparekh:
Build automated VM e2e coverage for the core enforcement scenarios we keep walking manually. Currently verified via #264 audit, #297, #317, repeated each round.

## Why

Pause, block, schedule, and time-limit enforcement are the product's reason for existing. Every PR touching `render.lua`, `policy.lua`, `PolicyService.scala`, or the snapshot schema risks breaking them. Today we catch regressions only via manual audit; automation gives PR-time signal.

## Scope — scenarios to automate

Build on the existing VM e2e harness (#148, #150, #146, #149).

### Suite A — Pause enforcement
- Pause a profile → all assigned devices' forward traffic dropped within one poll (~60s) AND HTTP/80 redirected to block page.
- Unpause → all traffic resumes within one poll.
- Assignment change while paused → newly-assigned devices inherit pause immediately on next poll; removed devices regain access.

### Suite B — Per-domain blocking (extraBlocked)
- Add a domain to extraBlocked → DNS resolves to sinkhole AND (after #329) IP-level drop blocks DoH/DoT bypass.
- Remove → access restored.
- Verify per-profile scoping (extraBlocked on Kids doesn't affect Adults).

### Suite C — Schedule enforcement (verifies #305 + #317)
- Schedule window starts → assigned-profile devices blocked within 60s; `reason=schedule` on the event.
- Schedule window ends → traffic resumes within 60s.
- Cross-midnight window (23:58–00:03) → both halves enforce correctly.
- Multiple overlapping schedules → at least one active = blocked.
- Schedule + pause precedence — pause overrides; reason flips to `paused` and back to `schedule` when unpaused mid-window.
- All-day schedule on a single weekday (verifies day-of-week match).

### Suite D — Daily time limits (verifies #15, #135, #141, #295)
- Profile with 10-min daily limit; device consumes via test client → block triggers at limit exhaustion.
- Per-app limits (when #105 lands) — exempt-vs-included accounting.
- Minute-granularity activeSeconds (verifies #295 — short bursts don't get rounded to full bucket).
- Cross-day rollover — limit resets at midnight (router's local TZ — verify expectations match docs).

### Suite E — Profile assignment changes
- Device reassigned from Profile A to Profile B → enforcement updates within one poll. Old profile's rules no longer apply; new profile's rules do.
- Device removed from all profiles → traffic permitted (per #62 "unknown device passthrough").
[...]

Issue #345, opened 2026-05-13
VM e2e: automate resilience-scenario validation suite


sameerparekh:
Build automated VM e2e coverage for the 5 resilience scenarios from `docs/resilience.md`. Today these are walked manually each round (#321 round 1, #337 round 2). Automation lets us catch regressions on every PR instead of every audit cycle.

## Why

Manual resilience audits found #328, #329, #331, #330, #334, #336 — bugs that would have been caught earlier if these scenarios ran on every PR. The audit takes ~2 hours and requires hands-on hardware actions; automation gives the signal in CI in minutes.

## Scope — five scenarios to automate

Build on the existing VM e2e harness (#148 orchestrator, #150 router VM, #146 client VM, #149 CI integration).

### Scenario 1 — Router power loss / boot-window default-deny (verifies #308)
- **Setup**: router VM running, client VM running, policy active blocking some domain.
- **Action**: `virsh destroy <router>` (or equivalent hard-kill), then `virsh start <router>`.
- **Verify**: long-running ping/curl from client VM through router fails throughout boot until `policy applied` line appears in router logread, then resumes. **Zero packets** should reach WAN during the boot window.
- **Assertion**: count packets sent from client vs. packets reaching the test echo server (running on host or 3rd VM); diff = 0 during the boot window.

### Scenario 2 — API restart (verifies #309 retry + #311 failover + #330 events)
- **Setup**: router enrolled, client active, ~30s of usage flowing.
- **Action**: `docker compose stop api`. Wait. `docker compose start api`.
- **Verify timings**:
  - 0-5 min after stop: cached snapshot enforces unchanged.
  - 5+ min after stop: per-profile failureMode kicks in (Closed=drop, Open=cached).
  - On restart: queued usage drains chronologically; queued events drain chronologically; nothing duplicated (idempotency).
- **Assertion**: API receives identical usage/event volume as if API had been up the whole time, just delayed.

### Scenario 3 — DB blip (verifies #310 + cached enforcement)
- **Setup**: router enrolled, client active.
- **Action**: `docker compose stop postgres` for 60s, then `start`.
- **Verify**: API returns 503 (not 500) on router + admin routes; agent treats 503 same as connection failure; cached snapshot continues enforcing.
- **Assertion**: `curl /api/health` returns 503 during outage; agent's poll-failure logs show 503-with-Retry-After parsing.

### Scenario 4 — Internet outage / WAN failure (verifies #311 the design call)
[...]

Issue #336, opened 2026-05-13
openwrt: agent's poll-scheduling math uses non-monotonic os.time() — backward clock jumps stall polling


sameerparekh:
Found during the #321 resilience audit walk (Scenario 5).

## Symptom
After skewing the router clock forward 15 minutes, then resetting via NTP (jumping clock backward by ~14 min), the agent stops polling for ~14 minutes — the time it takes for wall-clock to catch back up to its in-memory `last_policy_run` value.

Concrete trace (router logread, MDT):
- 11:58:34 — clock set +900s forward → wall clock reads 12:13:34.
- 12:13:36 — agent polls, persists `last_policy_run = epoch_for_12:13:36` in memory.
- 12:00:45 — operator runs `ntpd -nq`, clock snaps back to true ~12:00.
- 12:00 – 12:14 — **no polls fire** even though the agent process is running. Banner stays stale at 4500s; usage POSTs don't fire.

## Root cause
`openwrt/files/usr/sbin/familydns-agent`'s `on_tick` uses `if (now - last_policy_run) >= policy_int` to schedule polls (and the same shape for usage / activity-sample timers). All three sources of `now` come from `os.time()`. When the system clock jumps backward, `now - last_policy_run` goes negative and stays below `policy_int` until wall-clock recrosses `last_policy_run + policy_int`.

## Why it matters
- During the reset window the agent silently stops polling — usage, policy refresh, and skew correction all freeze.
- A determined-kid attack vector: jump the router clock far forward, wait for it to settle into agent memory, then snap it back. Polling is suspended for the gap → agent runs entirely on cached snapshot with no transitions.
- Less adversarially: any NTP fix-up after a long DST-aware boot, or a brief BSP that backwards-corrects sub-second drift, can cause sub-second polling stalls (harmless individually, but the design intent is "polls fire on cadence").

## Fix sketch
- Use `posix.clock_gettime(CLOCK_MONOTONIC)` for the timer math (luaposix; ship as a dep, or read `/proc/uptime` as a fallback BusyBox-friendly source).
- Keep `os.time()` only for things that need wall time (RFC3339 timestamps in usage reports, skew comparison vs API Date header).
- Defensive bandage: clamp `now - last_policy_run` to ≥ 0 and treat negative diff as "fire immediately" — handles backward jumps but not forward jumps that overshoot.

## Cross-refs
- #312 (clock skew detection — surfaces a related issue but doesn't address scheduling).
[...]

Issue #149, opened 2026-05-10
VM e2e: CI integration on self-hosted runner with KVM


sameerparekh:
Part of #106.

Wire the VM-based e2e suite into GitHub Actions so it runs on every PR that touches router-, API-, or web-relevant code paths.

## Scope

- Self-hosted runner with \`/dev/kvm\` available (bare metal or a cloud VM with nested virtualization).
- GitHub Actions workflow (e.g. \`.github/workflows/e2e-vm.yml\`) that:
  - Triggers on PRs touching \`openwrt/\`, \`api/\`, \`web/\` (and on \`workflow_dispatch\`).
  - Brings up the API stack, the router VM, and the client VM.
  - Runs \`scripts/e2e-vm.sh\`.
  - Uploads VM serial logs, API logs, and agent logs as artifacts on failure.
- Document runner provisioning: required packages (\`qemu-system-x86_64\`, \`bridge-utils\`, etc.), KVM/nested-virt requirements, secrets.

## Acceptance criteria

- VM e2e workflow runs on every PR matching the path filter.
- Workflow is green on \`main\`.
- Failure artifacts are sufficient to diagnose a scenario failure without re-running locally.
- Runner setup is reproducible from documentation alone.

## Depends on

- Router VM subissue
- Client VM subissue
- Orchestrator subissue
- Custom OpenWRT image subissue

PR #394, opened 2026-05-14
e2e-vm: orchestrator + primitive layer for VM tests (closes #148)

### LINKED ISSUE #148: VM e2e: test orchestrator — boot VMs, enroll router, run scenarios, assert API state
Part of #106.

The outer harness that ties the VM topology together: brings up the API stack on the host, boots router + client VMs, enrolls the router, runs scenarios, and asserts via API queries.

## Topology

- API server: \`docker compose\` on the host, reachable from the router VM via the host's LAN-bridge IP.
- Router VM: OpenWRT (see router-VM subissue) with familydns-agent baked in (see custom-image subissue).
- Client VM: Alpine on the LAN side (see client-VM subissue).

## Scenarios (v1)

1. **Enrollment success** — agent on the router enrolls via admin API + enrollment token; appears as an active router in the API.
2. **Allowed browsing** — client curls an allowed domain; request succeeds; \`connection_events\` row recorded with the right MAC + verdict.
3. **Blocked domain rejected** — admin blocks a domain; client request fails / hits the /blocked page; API records the block.
4. **Daily limit enforced** — per-app daily limit is set; client traffic past the limit is blocked; usage reflected in API.
5. **Usage reflected in API** — dashboard / \`/api/logs/*\` queries return the expected events for the scenario's client MAC.
6. **/blocked page rendering** — the blocked page is served and renders the expected content.

## Acceptance criteria

- \`scripts/e2e-vm.sh\` runs all scenarios green from a clean checkout on a host that satisfies the documented prerequisites.
- Each scenario is independently runnable (e.g. \`scripts/e2e-vm.sh --only blocked-domain\`).
- Snapshots are used between scenarios so total runtime stays reasonable (target: full suite < 10 min on dev hardware).
- Failures surface useful diagnostics: VM serial logs, API logs, agent logs.

## Out of scope

- CI wiring (separate subissue).
- OpnSense scenarios (deferred).

sameerparekh:
## Summary

Closes #148. Adds `scripts/e2e-vm.sh` and the `scripts/e2e/` pytest harness:

- **API stack lifecycle** (docker compose with `FAMILYDNS_DEBUG=1` so the loopback `/api/debug/*` endpoints are mounted)
- **Router VM** boot + enrollment + base snapshot, **client VM** per-scenario boot
- **Six v1 scenarios** from #148: enrollment, allowed browsing, blocked domain, daily limit, usage-in-API, /blocked page
- **Snapshot reuse** between scenarios — first one pays the boot+enroll cost, the rest restore from `e2e-base`
- **Failure diagnostics**: pytest hooks attach router serial console tail, agent `logread`, and API logs to every failed test report

## Primitive layer (for downstream consumers)

#346 (enforcement suites) and #345 (resilience suites) will build on `scripts/e2e/lib/`:

- `api_admin.AdminAPI` — login, profiles, devices, schedules, time-limits, pause toggle, logs, sessions, time_status
- `api_debug.DebugAPI` — devices(), events(limit=...), events_for_mac(mac), time_usage()
- `enrollment.enroll_router(admin, name, api_port)` — full enroll dance (the agent does NOT auto-enroll: errors out if `router_token` empty, so the orchestrator does the `/api/admin/routers` + `/api/router/register` calls itself, then `uci set` + `/etc/init.d/familydns restart`)
- `vm.{router_up,router_down,router_snapshot,router_restore,router_ssh,client_up,client_down,client_exec}`
- `wait.wait_until / wait_for_router_active / wait_for_next_poll / wait_for_etag_change` — race-free synchronization. Agent uses `CLOCK_MONOTONIC` for poll cadence (#336), so wall-clock jumps don't shift the timer; `wait_for_etag_change` is the canonical "agent fetched my new policy" gate.
- `clock.set_router_clock(when)` — sledgehammer per #346 Option 2
- `traffic.http_get / dns_query` — client-side curl/dig

## Framework choice

Python + pytest. Better assertions, fixture composition, marker-based selection, and downstream consumers can reuse `router` / `client` / `scratch_profile` / `scratch_device` fixtures rather than re-implementing the bash plumbing. VM-level primitives stay in `scripts/vm/*.sh`. The `scripts/e2e-vm.sh` entrypoint is a thin wrapper that creates a venv and shells out to pytest, with `--only <name>` mapping to pytest markers.

## Runtime

- Single scenario after enrollment: <30 s
- Full suite: ~3–5 min for scenarios 1/2/3/5/6 + ~5 min for scenario 4 (waits for usage-report cycle). Total under the 10-min target.
- Scenario 4 (daily limit) is the only slow one because it needs the agent's natural 300 s usage-report interval. Tagged so it can be skipped in tight loops via `--only` on a faster scenario, or sped up by overriding `usage_report_interval` per-test (left for the #346 enforcement suite).

## Verification status

This worktree was developed on macOS, which has no `/dev/kvm` — the harness gates on KVM and refuses to run VM-dependent tests there. The Python/bash wiring was syntax- and import-checked locally; runtime verification of the six scenarios needs a Linux host with KVM. The orchestrator follows the same shape as `scripts/e2e-router.sh` and `scripts/e2e-tests.sh` (which do run in CI today), and the VM primitives it shells out to (`scripts/vm/*.sh`) are already exercised by their own manual-verification checklists in `scripts/vm/README.md`.

## Known limitations

- OpnSense scenarios deferred (router primitives are OpenWRT-only)
- Multi-client scenarios: fixture supports `name=`/`ssh_port=` but `client-up.sh` doesn't yet auto-allocate non-overlapping SSH ports
- `/api/debug/router_status` does not exist — we use `lastSeenAt` from `GET /api/admin/routers` as the agent-poll signal

## Downstream

- #346 — enforcement suites build on this primitive layer
- #345 — resilience suites build on this primitive layer
- CI wiring tracked separately in #149

## Test plan

- [ ] On a Linux host with KVM: `scripts/vm/build-client-base.sh && scripts/vm/build-router-image.sh && scripts/e2e-vm.sh` — full suite green
- [ ] `scripts/e2e-vm.sh --only enrollment` runs scenario 1 in isolation
- [ ] `scripts/e2e-vm.sh --only blocked-domain` runs scenario 3 in isolation
- [ ] Forced failure surfaces router serial log + agent log + API log in pytest output
- [ ] `E2E_VM_KEEP=1` leaves VMs+stack up for manual poking
- [ ] `E2E_VM_SKIP_VMS=1 scripts/e2e-vm.sh` runs without VMs (sanity mode)


sameerparekh:
Pushed validation-pass commit `1a35c68` to the branch. Ran the full suite end-to-end on a Linux+KVM host; fixed a chain of issues that surfaced once the harness actually executed VMs:

**Now green**
- Scenario 1 (enrollment) — reliably
- Scenario 4 (daily-limit) — reliably when run alone

**Fixes in this commit**
- *Compose override*: `ports: !reset` was deleting the api port mapping entirely instead of replacing it — switched to `!override`. Postgres no longer exposed. `fake-router` muted (2 events/s under its own router_id was drowning real events out of any `/api/debug/events?limit=N` window).
- *Router VM*: block in `router_up` until SSH actually answers (key-auth probe, not port-open). Test key baked into `/etc/dropbear/authorized_keys`. WAN firewall allow for SSH (stock OpenWRT rejects WAN inbound). LuCI port off 8080 (collides with developer's docker stack).
- *Router image*: install `dnsmasq-full` instead of stock dnsmasq, and switch `render.lua` from `ipset=` to `nftset=` — OpenWRT 23.05's dnsmasq-full is built without `HAVE_IPSET` and crashes on the agent's `ipset=` lines. Bake dnsmasq query-log + uhttpd block-page + WAN-SSH allow into `uci-defaults/99-familydns` (was previously only set up by `openwrt/install.sh`, which the orchestrator's enrollment skips).
- *Client*: `client-exec.sh` joined argv with bare spaces, so `sh -c "curl -w 'a b'"` reached the remote as `sh -c curl -w 'a b'` and ran curl with no args — quote each arg with `printf %q`. `client_up` now waits for LAN DHCP + external-DNS resolution before returning.
- *Profile API*: `GET /api/profiles/:id` returns nested `{profile, schedules, timeLimit (object|null), siteTimeLimits}` but `PUT` requires flat with `timeLimit` as an int. The scenarios were splatting only the inner `profile` dict — every PUT 400'd on `.schedules(missing)`. New helper `AdminAPI.apply_profile_update(...)`; scenarios migrated.
- *Debug endpoints*: the loopback gate rejected the orchestrator (docker NATs source to bridge gateway 172.19.0.1). Route the calls through `docker exec` so the API sees 127.0.0.1.
- *Per-scenario lifecycle*: post-restore settle bumped from 1s → 8s.
- *Test fixes*: `test_05` was checking `s.get("deviceMac")` on `/api/time/status` rows, but the endpoint returns `ProfileTimeStatus[]` with nested `deviceSummaries`. `test_02` event-wait widened from 30s → 120s.

**Still flaky** — scenarios 2/3/5/6 in the full-suite run hit `curl HTTPCODE:000` or event-wait timeouts after live `loadvm`. The agent's `policy.fetch: unexpected status 0` immediately after restore points at QEMU SLIRP NAT state not surviving the snapshot. A clean fix likely needs cold restore (poweroff + `qemu-img snapshot -a` + boot) instead of `loadvm`, or a per-scenario fresh boot. Tracked in follow-up session.

Issue #392, opened 2026-05-14
extraBlocked / blocklist nft sets: add IPv6 (eb6_, bl6_) alongside IPv4


sameerparekh:
Follow-up to #351.

#351 lands `extraBlocked` as a per-(MAC, host) nft drop on `ip daddr @eb_<host>`, populated by dnsmasq `--ipset=/<host>/eb_<host>` at A-record resolution time. This is IPv4-only — matches the current pattern used by `siteLimits` (also v4-only) but leaves an IPv6 escape hatch: a blocked host that resolves to AAAA bypasses the drop entirely.

Once we're comfortable with the v4 path, extend `render.lua` to:

- Emit a parallel `ipset=/<host>/eb_<host>,eb6_<host>` directive. dnsmasq writes the first set on A, second on AAAA.
- Declare `set eb6_<host> { type ipv6_addr; flags dynamic,timeout; timeout 1h }`.
- Add the matching nft rule: `ether saddr <mac> ip6 daddr @eb6_<host> drop` (and the DNAT analog if we wire block-page-on-v6).
- Same for the future `bl_` / `bl6_` blocklist sets in #352.

Block-page DNAT on v6 is its own decision: there's no `::1:8081` because uhttpd binds v4. Either bind v6 as well or accept that v6-blocked HTTP just drops (no block page on v6). Probably the latter for now.

Refs: #351, #352, docs/architecture.md §0.

Issue #376, opened 2026-05-14
wire-contract: schema versioning + capability negotiation


sameerparekh:
## Architectural concern

The wire contract between API and agent (`docs/architecture.md` §0.2, §6.2; `shared/Models.scala` `PolicySnapshot`) **defines field names but has no schema-evolution policy**. There is no:

- `snapshotVersion` field on the snapshot.
- API capability set advertised to the agent.
- Agent capability set advertised to the API.
- Documented policy for handling unknown fields (forward-compat) or known-but-newer-version snapshots.

As written, the architecture reads as \"one version forever.\" This is wrong because auto-update is not instant — even after #279/#131 fully land, agents and APIs will run mismatched versions in the field for hours-to-days during rollouts. Every future shape change (collapsing the snapshot, adding unmanagedMacPolicy, etc.) needs a story for how it ships safely.

See `docs/architecture-critique.md` §4.

## Remediation steps

1. **Add to `PolicySnapshot`:**
   - `snapshotVersion: Int` — start at 1; bump on breaking-shape change only.
   - `serverCapabilities: Set[String]` — strings naming optional features (e.g. `\"unmanaged-mac-policy\"`, `\"failover-threshold-in-snapshot\"`, `\"https-block-page\"`).
2. **Agent advertises** `agentCapabilities: Set[String]` on `POST /api/router/register` (during enrollment) and on every `GET /api/router/policy` (header `X-Familydns-Capabilities` or query param). Agent also advertises its known `snapshotVersion`.
3. **API serves** the shape negotiated by the intersection: optional fields only emitted if the agent advertised the capability. `snapshotVersion` set to the highest version both ends understand.
4. **Documented policy** in `docs/architecture.md`:
   - Agents MUST ignore unknown top-level fields (forward-compat).
   - Agents MUST refuse a snapshot whose `snapshotVersion` exceeds the agent's known max (no silent partial enforcement; fall through to cached or boot-deny per resilience §1).
   - Servers MUST NOT emit fields gated on a capability the agent didn't advertise.
5. **Apply to the canonical-model migration** (the parent reason this matters): when #354's snapshot collapse lands, it becomes `snapshotVersion: 2` with `serverCapabilities` including `\"canonical-block-rules\"`. Older agents still get the v1 shape; newer agents get the v2 shape. No flag day required.

## Acceptance criteria

- `docs/architecture.md` gains an \"Evolution policy\" subsection naming the rules above as part of the wire contract.
[...]

PR #403, opened 2026-05-14
wire-contract: hostname becomes HostId tagged union (fixes #391)

### LINKED ISSUE #391: wire-contract: hostname becomes union of FQDN | IPv4 | IPv6
## Architectural concern

The architecture treats `hostname` as an unconstrained string in the wire contract:

- `usage.records[*].hostname`
- `events.dns_query.qname`
- `time_usage.domain` (schema)
- The block-reason `category:<host>`

But §7.2 establishes a fallback: when dns-tail can't attribute an IP to a hostname (direct-IP traffic, DoH-resolved domain, sidecar race), the agent puts **the IP literal in the hostname field**. So `time_usage` rows have `hostname = \"192.0.2.1\"`. Site-limit pattern matching (`*.example.com`) silently misses these rows. Presence reports show \"192.0.2.1\" as a top-host. Admin UI displays an IP as if it were a hostname.

The architecture never names \"what's the type of hostname, really?\"

See `docs/architecture-critique.md` §11.

## Design response

Promote `hostname` to a tagged union (sub-type) at the wire contract level:

```scala
sealed trait HostId
object HostId:
  case class FQDN(name: String) extends HostId      // \"youtube.com\"
  case class IPv4(addr: String) extends HostId      // \"192.0.2.1\"
  case class IPv6(addr: String) extends HostId      // \"2001:db8::1\"
```

Wire JSON form: `{\"type\": \"fqdn\", \"value\": \"youtube.com\"}` — explicit, no parsing heuristic on the receive side.

## Remediation steps

1. `shared/Models.scala`: add `HostId` ADT with JsonCodec.
2. `UsageRecord.hostname` → `host: HostId`. Same for `RouterEvent` and connection-event types.
3. `time_usage` schema: add `host_type` column (Flyway migration). Backfill existing rows: parse current `domain` value, infer type by heuristic, populate `host_type`.
4. Repos: queries that filter on hostname filter by `(host_type, host_value)` pair.
5. `PolicyService`: site-limit pattern matching skips non-FQDN rows entirely (an IP literal can never match `*.example.com`).
6. Admin UI: presence + history views render IPv4/IPv6 rows distinguishably from named hosts (small icon or label).
7. Block-reason wire encoding: `category:<host>` becomes `category:<host_type>:<host_value>` or a structured form via the canonical-model migration (#354).
8. Agent (`usage.lua`, `conntrack.lua`): emit `{type, value}` instead of bare string. Coordinates with capability negotiation (#376) so older API/agents handle the older bare-string form.

## Acceptance criteria

- Wire JSON for usage and events carries `host: {type, value}` instead of bare `hostname: string`.
- `time_usage` rows with IP-typed hosts don't pollute site-limit accounting.
- Admin UI clearly distinguishes \"direct IP traffic\" from named-host activity.
- `docs/architecture.md` §6 (wire contract) updates the type definition.

Refs #368.

sameerparekh:
## Summary

- Replaces `hostname: Hostname` (and one stringly-typed `domain: String`) with `host: HostId` — a tagged union of `Fqdn | IPv4 | IPv6` — on every wire-contract field where a "what host did the device contact" identifier flows.
- Wire form is explicit: `{"type":"fqdn"|"ipv4"|"ipv6","value":"..."}`. No heuristic on receive.
- Storage gains a `host_type` column (with CHECK constraint) alongside the renamed `host_value` on `time_usage`, `connection_events`, `block_events`, `traffic_reports` (Flyway `V12`). Doobie reads/writes the pair as a composite.
- Site-limit pattern matching (`*.example.com`) now skips IP-typed rows by type, not by string accident — pinned by new `PresenceSpec` regression cases.
- Lua agent (`usage.lua`, `conntrack.lua`) emits the new shape for `UsageRecord` and `connection_attempt` events. `dhcp_lease` / `first_seen_mac` keep `hostname` (device's DHCP-advertised name — different invariant).
- Web adds a `HostCell` component that renders FQDN rows plain and IP rows italic with an `ipv4` / `ipv6` tag so admins see at a glance which rows are unattributed direct-IP traffic (DoH / Apple Private Relay / etc.).

### Why now

Architecture §7.2 already permits the agent to fall back to the dest IP when DNS attribution misses, but it silently smuggled that IP into a field whose type implied a hostname. `Hostname.parse` already rejected IPv4 literals, so those records were either being rejected or being unsafely re-unwrapped. The HostId promotion makes the actual wire reality match the declared type.

### Migration notes

- Breaking change at the wire level — no pre-existing field deployments to coordinate with.
- The V12 backfill heuristic errs on FQDN when ambiguous (per #391 spec). Pre-prod we can wipe the DB if anything mis-classifies.
- Capability `"hostid-tagged-union"` should be registered when [#376](https://github.com/sameerparekh/familydns/issues/376) (capability negotiation) lands.
- [#392](https://github.com/sameerparekh/familydns/issues/392) (IPv6 ipsets) is downstream — it can now distinguish v4 vs v6 enforcement scopes via `HostId`.

## Test plan

- [x] `mill __.test` — all 267 tasks SUCCESS, including new `HostIdSpec` (parse + JSON round-trip) and new `PresenceSpec` cases covering: IP-row never matches FQDN pattern; mixed bucket counts only the FQDN side; IP rows never exempt from daily total.
- [x] `bash openwrt/test/run_tests.sh` — 249/249, including new Lua tests asserting connection_attempt and usage records emit `host = {type=ipv4|ipv6|fqdn, value=...}` based on attribution.
- [x] `npx vitest run` — 93/93 web tests pass.
- [x] `npx tsc --noEmit` — clean.
- [x] `mill api.compile` — clean.
- [ ] V12 migration on a fresh DB (covered by CI Flyway boot path).
- [ ] Manual smoke on `192.168.10.43` (the dev API) after merge.

Closes #391.


api/src/policy/PolicyService.scala
```diff
@@ -296,6 +296,11 @@ class PolicyServiceLive(
   private def matchesAny(domain: String, patterns: List[Hostname]): Boolean =
     patterns.exists(p => matchesDomainPattern(domain, p.value))
 
+  // Pattern matching is FQDN-only by design (#391). The decision endpoint
+  // receives `RouterDecisionRequest.hostname: Hostname`, which the type system
+  // already constrains to FQDN-shape (Hostname.parse rejects IPv4 literals).
+  // Direct-IP traffic never reaches this matcher — IP-routed flows go through
+  // the nftables enforcement path, not the DNS-time decision.
```
sameerparekh: There are no DNS-time decisions, we aren't blocking at the dns level

Issue #251, opened 2026-05-12
setup a deploy process to run the api server on render.com


sameerparekh:
file appropriate subissues for this larger task. we want to rename this project first because we don't want to have to redo our render.com account

sameerparekh:
## Need a decision before filing sub-issues: rename-first vs. deploy-now

The issue body says "rename this project first because we don't want to have to redo our render.com account." But the rename track (#365 META + #355-#364) is currently parked and the goal is to get the at-home stack on a cloud-hosted API **soon**. Three options:

### Option A — rename track first (per issue body)
Kick off the rename (#355-#364) under #365 before any render.com work. ~5-10 ordered PRs, but the render.com account/services are created once, in their final name `wifihaven`. Slowest path to cloud-up.

### Option B — render.com now under final name `wifihaven` (recommended)
Create the render.com account + services under the **wifihaven** name from the start, while the codebase + deployed image are still `familydns`. Env vars / image refs in render.com config point at `ghcr.io/sameerparekh/familydns-api` (current image). On rename day, only the image-tag URL changes inside render.com — no service rename, no account rename, no domain rename. Fastest cloud-up path; small cosmetic mismatch (code says familydns, cloud says wifihaven) during the transition window.

### Option C — render.com now under current name `familydns`, rename later
Simplest near-term, but costs a one-time rebuild of render.com config when the rename ships: service rename, domain rename, env-var rebrand. Probably 1-2 hours of redo work on rename day, and may force a downtime window.

### Recommendation: **Option B**
Cloud usage starts immediately; rename impact is minimal (one image-tag change in render.com config). It honors the spirit of the original concern (don't redo the account) without blocking on the parked rename track.

@sameerparekh — which option? I'll file sub-issues based on your pick.

sameerparekh:
Decision: **Option A** — rename track first. Putting this on hold until #365 (rename META) lands. Render.com setup will resume once the project is `wifihaven` so the account/services are created in their final name.

sameerparekh:
## Render.com deploy plan

Scope: stand up the WifiHaven API (Scala + Postgres + static React bundle, single image `ghcr.io/wifihaven/wifihaven-api`) on render.com, then cut the main-house OpenWRT agent over to it. Self-hosted box at `192.168.10.43:8080` stays running as fallback through cutover.

Out of scope (tracked separately): unified manual-gate CI (#498), home/dev topology doc (#584).

### Mapping current stack → Render primitives

[...]

PR #610, opened 2026-05-18
render: move SPA fallback to _redirects so ACME challenges pass through (#609)

### LINKED ISSUE #609: render: SPA catch-all rewrite breaks Let's Encrypt cert issuance (.well-known intercepted)
## What happened

Render Static Sites for the SPA (\`wifihaven-spa-prod\`, \`wifihaven-spa-staging\`) cannot issue Let's Encrypt certificates for their custom domains. Dashboard shows \`Certificate Error\`. DNS is correctly configured (apex hits Render's static-site IPs \`216.24.57.0/24\`, all subdomains CNAME to the right Render service hosts), and no CAA records block issuance.

## Cause

The SPA \`routes:\` in \`render.yaml\` use a greedy catch-all rewrite:

\`\`\`yaml
routes:
  - type: rewrite
    source: /*
    destination: /index.html
\`\`\`

This intercepts \`/.well-known/acme-challenge/<token>\` requests and serves \`/index.html\` instead of the ACME challenge token. Let's Encrypt's HTTP-01 validation fails. (Render's ingress is supposed to handle \`.well-known\` transparently for Static Sites, but evidently doesn't in this case — possibly because the rewrite is evaluated at the Cloudflare CDN edge, which is in the path per the DNS resolution chain.)

## Fix options (pick one)

1. **Move SPA routing to a \`_redirects\` file** under \`web/public/\` (Netlify-style format Render supports). The file is bundled into \`staticPublishPath\` and is evaluated *after* Render's ingress handles \`.well-known\`. Drop the inline \`routes:\` block from \`render.yaml\`.
2. **Add an explicit pass-through for \`.well-known\` above the catch-all**:
   \`\`\`yaml
   routes:
     - type: rewrite
       source: /.well-known/*
       destination: /.well-known/:splat
     - type: rewrite
       source: /*
       destination: /index.html
   \`\`\`
   Cheaper to write, but identity-rewrite-to-self semantics on Render are unclear; may not actually skip the catch-all.
3. **Open a Render support ticket** asking why their ingress isn't intercepting \`.well-known\` ahead of user routes. Slow.

Recommendation: **option 1**. \`_redirects\` is the documented Render pattern for SPAs that need ACME to work, and it's a one-file addition.

## Acceptance

- Certs issue successfully for \`wifihaven.net\`, \`www.wifihaven.net\`, and \`staging.wifihaven.net\` (the API services aren't affected — they don't have the SPA rewrite).
- SPA client-side routing still works on deep-link reloads.
- Doc note in \`docs/render-prod-setup.md\` (or wherever the route config is explained) calling out the \`.well-known\` hazard so this doesn't regress.

Refs: #251, #587.

sameerparekh:
## Summary

Move the SPA `/* → /index.html` fallback from `render.yaml`'s inline `routes:` block into `web/public/_redirects`, so Render can serve `/.well-known/acme-challenge/*` directly and Let's Encrypt cert issuance for the SPA custom domains stops failing.

The inline `routes:` rewrite ran ahead of Render's ingress for ACME, returning `index.html` instead of the challenge token. Render's Static Sites natively read a Netlify-style `_redirects` file at the publish root, and Vite copies `web/public/_redirects` verbatim into `dist/` at build time. ACME challenges are short-circuited by the platform before `_redirects` is consulted.

Affects both `wifihaven-spa-prod` and `wifihaven-spa-staging`.

Closes #609. Umbrella: #251.

## Test plan

- [x] `npm ci && npm run build` in `web/` — `dist/_redirects` is produced (single line `/*  /index.html  200`).
- [ ] After merge, re-trigger Blueprint sync on Render.
- [ ] Confirm Certificate Error clears for `wifihaven.net`, `www.wifihaven.net`, and `staging.wifihaven.net` (Render → Static Site → Custom Domains shows valid Let's Encrypt cert).
- [ ] Spot-check SPA deep links (e.g. `/admin/devices`) still load via client-side routing after the cert is live.


sameerparekh:
Closing — certs issued, just slow propagation. Not needed.

Issue #498, opened 2026-05-16
ci: unify api + openwrt publish behind a single manual gate; drop standalone tag-trigger


sameerparekh:
## Background

PR #497 (fixes #494) folded the openwrt `.ipk`/`.apk` build into Master CD as a reusable workflow so it's gated on the same checks as `publish-api`. That fix retained `openwrt-build.yml`'s own \`push.tags: [\"v*\"]\` trigger so ad-hoc tag releases still work standalone.

Side effect: on a real \`v*\` tag push, \`openwrt-build.yml\` fires **twice** — once standalone (ungated, as before) and once via \`master.yml\` (gated). Each run produces its own artifacts and attempts its own release. \`softprops/action-gh-release@v2\` is idempotent on the same tag so it's usually harmless, but it's wasteful CI and the ungated run defeats the point of the gate.

## Proposal

Move to a **single manual gate** in \`master.yml\` that publishes both the API image and the OpenWRT agent together as one coordinated release:

- Drop \`push.tags: [\"v*\"]\` from \`openwrt-build.yml\`. The reusable-workflow path becomes the only release path. Keep \`workflow_dispatch\` for one-off debugging builds and \`pull_request\`/\`push.branches\` for PR validation.
- Add a single manual-approval gate (GitHub \`environments: production\` or a \`workflow_dispatch\` 'release' input) before the publish jobs in \`master.yml\`, so the api image and the openwrt packages publish atomically off the same green commit.
- Consider removing the per-job path filters (\`changes.api\` / \`changes.outputs.openwrt\`) at release time — once it's a manual coordinated release, you probably want both artifacts ship together regardless of which paths changed since the last release.

Versioning (how the API image tag and the OpenWRT \`PKG_VERSION\`/\`PKG_RELEASE\` stay coordinated under the unified gate) is out of scope here — tracked separately in a follow-up issue.

## Acceptance

- One \`v*\` tag push (or one manual release dispatch) produces exactly one OpenWRT release artifact set and one API image push.
- No standalone \`openwrt-build.yml\` run on tag pushes.
- Same gating checks as today's \`publish-api\` (no regression on safety).

Refs: #494, #497.

PR #644, opened 2026-05-18
api: add /api/version + agent wireSchema logging (#597)

### LINKED ISSUE #597: api: start versioning the API wire schema (cloud + router agent ship independently)
Tracked under #251 (P4). Related: #376 (full capability-negotiation, deferred to pre-v1.0); #498 (unified manual gate).

Once the API is on Render and the OpenWRT agent is shipped separately via .ipk/.apk releases, the two artifacts can drift in time. Today we get away with tandem deploys because the API and router agents are deployed from the same commit on the same box. With cloud API + over-the-air router agents, "tandem" stops being automatic.

We don't need the full #376 capability-negotiation framework yet, but we do need:

- `/api/version` endpoint returning at minimum: API build version (git sha / semver), wire-schema version (an integer we manually bump when the policy/event JSON shape changes incompatibly).
- Router agent logs the API's wire-schema version on startup and on every drift; warns loudly if it diverges from what it was built against.
- A short doc: "compatible router agent versions" table in `docs/architecture.md` (or new `docs/api-versioning.md`).
- Note in #498's body that the unified-gate restoration should also coordinate the wire-schema bump (so a breaking-shape change can't ship the API ahead of the agent).

Out of scope: full negotiation per #376.

## Acceptance

- `curl https://api.wifihaven.app/api/version` returns JSON with `apiBuild` and `wireSchema`.
- Router agent prints the API's wire-schema version at startup; mismatched versions log a warning.
- Doc explains the bump procedure and the current compatibility matrix.

sameerparekh:
## Summary

Closes #597. Refs #498 (manual deploy gate), #376 (full capability negotiation, deferred), #251 (cloud-deploy umbrella).

Surface wire-protocol drift between the cloud API and the OpenWRT router agent before it silently breaks, without taking on the full negotiation framework from #376.

- **`GET /api/version`** (unauthenticated discovery): returns `{apiBuild, wireSchema}`. `apiBuild` is the short git sha, baked into the jar at Mill build time via a new `buildInfoSource` generated source (env override `WIFIHAVEN_BUILD_SHA`; falls back to `git rev-parse --short HEAD`; finally `"dev"`). `wireSchema` is a single integer constant in `shared/src/WireSchema.scala`, starting at `1`.
- **Router agent startup check**: after the initial policy fetch the agent hits `/api/version`, logs the response, and calls `version.compare_wire_schema(remote, log)`. Match → `wireSchema match: 1` at info. Mismatch → `wireSchema MISMATCH: agent=N, api=M — incompatible behavior possible` at warn. Agent keeps running either way — no runtime refusal.
- **`docs/api-versioning.md`**: what counts as a breaking change, the lockstep bump procedure for the two constants, and the v1 compatibility matrix. Cross-references #498 (atomic coordination) and #376 (long-term path).

## Mechanism: Scala ↔ Lua constant sync

**Manual, with a runtime safety net.** Two constants live in two build systems (Mill / OpenWRT feed); auto-generation would require either Mill to drive the OpenWRT package build (it doesn't) or a shared text file both build systems read at build time (extra moving parts). Wire-shape changes already require coordinated Scala/Lua edits, so bumping a constant next to those edits is a one-line addition. The agent's startup mismatch warning is the load-bearing safety net — a forgotten bump on either side surfaces loudly on the next deploy. Justified in detail in `docs/api-versioning.md`.

`apiBuild` is build-time generated (not manual) because it changes every commit — manual sync would always lag.

## Out of scope

- Full capability negotiation (#376).
- Refusing to operate on mismatch — warn-only by design.
- Versioning the `/api/version` endpoint itself (chicken-and-egg; treat as v0-forever).
- `.github/workflows/` changes.

## Test plan

- [x] `mill api.compile` — Mill build-info generation produces `BuildInfo.scala` with current git sha.
- [x] `mill api.test` — all 37+ feature specs pass, including new `VersionApiSpec` (status 200, `apiBuild` non-empty, `wireSchema` matches `WireSchema.Current`).
- [x] `mill api.assembly` — jar assembles cleanly.
- [x] `sh openwrt/test/run_tests.sh` — 377 successes / 0 failures; new `version_spec.lua` exercises match, mismatch, and missing-field cases.
- [ ] Compose stack: agent logs the `api version:` + `wireSchema match: 1` lines on startup. (Validate before merge.)
- [ ] `curl http://localhost:8080/api/version` from a local stack returns the expected JSON shape. (Validate before merge.)


sameerparekh:
Closing for now — parking the work. Branch `api-versioning` is preserved on the remote; will reopen / cherry-pick to a fresh PR when we're ready to land this ahead of the next breaking router-facing API change. See #597 for context.

Issue #660, opened 2026-05-18
ops: install production OpenWRT router at home, start serving live household traffic


sameerparekh:
## Scope

Physical bring-up of the **new** production OpenWRT router at home, enrolled against the cloud API (\`api.wifihaven.app\`), serving live traffic for the operator's household. The current 192.168.10.42 box stays in place as the dev router against the 192.168.10.43 dev API — see [network_topology.md](https://github.com/wifihaven/wifihaven/issues/584) for context.

This is the **launch milestone**: first real household on the cloud stack. The cloud API, SPA on CDN, enrollment docs, and self-hosted compose path are all already landed (#585, #586, #587, #589, #611, #613, #615, #620). All that's left is to actually do it.

## Pre-flight (in priority order; do not block the install on any of these unless they fail outright)

1. \`https://api.wifihaven.app/api/health\` returns 200 with \`db: ok\`.
2. \`https://wifihaven.net/\` serves the prod SPA, points at the prod API (cross-origin CORS works for login).
3. PR #658 (#590 JVM heap tuning) merged and a fresh prod deploy is healthy on Render.
4. Operator's password manager has the prod admin password.
5. Operator has the new OpenWRT hardware flashed with the current \`:latest\` agent build.

## Install steps

Follow [docs/install-openwrt.md](https://github.com/wifihaven/wifihaven/blob/main/docs/install-openwrt.md) cold (closes the dogfood loop on #589):

1. Wire the new router in front of the household's existing AP / behind the ISP modem.
2. Edit \`/etc/config/wifihaven\`: \`api_url='https://api.wifihaven.app'\`.
3. \`/etc/init.d/wifihaven restart\`.
4. Issue an enrollment token in the prod admin UI, paste it into the router CLI.
5. Confirm the first policy poll succeeds (router shows \`lastEtag\` populated, agent log shows 200 from \`/api/router/...\`).
6. Add household devices, build initial Kids / Adults profiles, verify block-page renders for a known-blocked domain from a real client.

## Acceptance

- Real household traffic flows through the new router for at least 24 h with no manual intervention.
- Admin UI on \`https://wifihaven.net/\` shows live devices + live event log from the household.
- Block decisions on real client devices match the configured policy.
- Operator has rolled back to the old non-wifihaven router setup zero times.

## What this issue is NOT

[...]

Issue #623, opened 2026-05-18
api: /api/auth/change-password returns empty body, should return JSON


sameerparekh:
## What

`POST /api/auth/change-password` returns `Response.ok` with no body ([api/src/routes/Routes.scala:58](https://github.com/wifihaven/wifihaven/blob/main/api/src/routes/Routes.scala#L58)). Every other JSON endpoint on the API returns a JSON object, so clients that assume \"200 → parseable JSON\" break here.

## Why it matters

Surfaced while writing the fake-router bootstrap rotation in #621. The existing `_post` helper does `json.loads(r.read())` on every response — fine for everything else, but blows up on this route with `Expecting value: line 1 column 1 (char 0)`. Worked around it by hand-rolling a separate urllib call ([docker/fake-router.py — `_change_password`](https://github.com/wifihaven/wifihaven/blob/main/docker/fake-router.py)), but that's a wart.

`deploy/install.sh` only inspects the HTTP status code so it doesn't notice, but any future TypeScript/Scala client (e.g. the SPA's change-password page from #607) will hit the same issue.

## Fix

Return `Response.json(...)` with a small payload — at minimum `{ \"ok\": true }`, or echo back the updated `mustChangePassword: false` so callers can refresh their session state without a round-trip.

## Scope

API only — `api/src/routes/Routes.scala` change-password handler at line ~58. Add a `ChangePasswordResponse` case class in `shared/src/Models.scala` if returning a richer shape.

Once shipped, `docker/fake-router.py::_change_password` can be simplified to use the normal `_post` helper.

## Related

- #599 / #586 — added the `must_change_password` flag and force-rotate flow
- #621 — fake-router rotation; carries the workaround
- #607 — SPA change-password page (will benefit)

sameerparekh:
Same root cause as #664 / fixed by PR #666 — `_post()` now tolerates empty bodies. The change-password handler still returns no JSON body (which #623 also wants fixed on the API side), and the hand-rolled `_change_password` workaround in `docker/fake-router.py` can be retired once #666 lands. Leaving #623 open since the API-side fix (return `{ ok: true }` or similar) hasn't been done yet.

PR #666, opened 2026-05-18
fix(e2e): stop fake-router json-decoding empty bodies (#664)

### LINKED ISSUE #664: ci: master CD e2e fails — fake-router posts usage/events errors after #659
Master CD on \`c4ca177\` (post-merge of [#659](https://github.com/wifihaven/wifihaven/pull/659)) failed in the **Live e2e against staging stack** job. No prod deploy happened.

Run: [actions/runs/26062621763](https://github.com/wifihaven/wifihaven/actions/runs/26062621763)

## Failure

\`scripts/e2e-tests.sh\` step **\"fake-router steady-state log scan\"** fails because the fake-router emits errors decoding the API response:

\`\`\`
[fake-router] usage error: Expecting value: line 1 column 1 (char 0)
[fake-router] events error: Expecting value: line 1 column 1 (char 0)
[fake-router] usage error: Expecting value: line 1 column 1 (char 0)
[fake-router] events error: Expecting value: line 1 column 1 (char 0)
...
✗ fake-router posted errors after warmup — wire contract drift?
\`\`\`

\`Expecting value: line 1 column 1 (char 0)\` is Python's \`json.loads(\"\")\` — the fake-router is calling \`.json()\` on a response that has an empty body. The API \`/api/router/usage\` and \`/api/router/events\` endpoints either:

1. Now return \`204 No Content\` (or some empty 200) where they used to return a JSON object, OR
2. The fake-router was changed in #659 to call \`.json()\` on a response that has always been empty, OR
3. The contract is correct but the fake-router's warmup check fires before posting actually settles.

Since #659 specifically touches the fake-router (\"drop legacy hostname field; add e2e event coverage\"), option 2 is the most likely.

## Scope

- Read [#659](https://github.com/wifihaven/wifihaven/pull/659)'s diff against the relevant fake-router source and the e2e test script.
- Decide whether the right fix is: (a) the fake-router should not call \`.json()\` on the response body (the API contract says empty), (b) the API should return a JSON body, or (c) the warmup scan should ignore these specific lines.
- Apply the smallest fix that gets master CD green again. We're trying to unstick the deploy pipeline; don't expand scope.
- The fundamental \"router-to-API event contract\" work belongs in #654 (Gate 2). This issue is just the immediate unblock.

## Acceptance

- Re-running master CD on \`main\` after the fix reaches the **manual approval (production gate)** step without failing in e2e.
- No regression in the actual e2e scenario coverage that #659 was adding.
- Fresh staging deploy lands so we can verify on \`api-staging.wifihaven.net\`.

sameerparekh:
## Root cause

`POST /api/router/usage` and `POST /api/router/events` both return HTTP 200 with an **empty body** (`Response.ok` in ZIO-HTTP — no JSON payload). The `_post()` helper in `docker/fake-router.py` was unconditionally calling `json.loads(r.read())` on every response, which raises Python's `"Expecting value: line 1 column 1 (char 0)"` when the body is empty.

PR #659 added three new `connection_attempt` shapes to `_post_events` and the corresponding usage records to `_post_usage`, which meant both functions were being called on **every** steady-state tick — surfacing a pre-existing latent bug on each interval and flooding the post-warmup logs with `events error` / `usage error` lines. The new log-scan step added by #659 then correctly caught those errors and failed the job.

## Exact location

**`docker/fake-router.py` line 73** — `_post()` helper:

```python
# before
return json.loads(r.read())

# after
body = r.read()
return json.loads(body) if body else {}
```

All callers that need a JSON response (`_login`, `_create_router`, `_register`) talk to endpoints that return populated bodies; `_post_usage` and `_post_events` now silently accept the empty 200.

## Impact

- Unblocks master CD immediately — the fake-router steady-state log scan will pass once this merges.
- This is a prerequisite for the production router install tracked in #660.

Closes #664

### DIFF
--- a/docker/fake-router.py
+++ b/docker/fake-router.py
@@ -70,7 +70,8 @@ def _post(path: str, payload: dict, token: str | None = None) -> dict:
         f"{API_BASE}{path}", data=data, headers=headers, method="POST"
     )
     with urllib.request.urlopen(req, timeout=10) as r:
-        return json.loads(r.read())
+        body = r.read()
+        return json.loads(body) if body else {}
 
 
 def _get(


sameerparekh:
Cross-ref: this is the same root cause as #623 (`POST /api/auth/change-password` empty-body trap). #623's body explicitly documents the `_post` helper as the underlying issue and notes that change-password was hand-rolled around it as a workaround.

Now that `_post` tolerates empty bodies, the hand-rolled `_change_password` in `docker/fake-router.py` can use the normal helper. Either fold that simplification into this PR or leave #623 open so a follow-up retires it — but #623 should NOT be closed by this PR unless the hand-roll is also removed.

PR #839, opened 2026-05-22
fix(seed): remove test blocklists from prod + gate dev fixtures (#706)

### LINKED ISSUE #706: prod cloud DB has test_ads / test_social blocklists in policy (seed data leak)
## What happened

A fresh router enrolled against the prod cloud API (`https://api.wifihaven.net`) during the #660 walkthrough pulled a policy snapshot containing two clearly-test blocklists:

```json
{
  "test_ads":    { "version": "4b6c87d2270c1c3d", "url": "/api/blocklists/test_ads" },
  "test_social": { "version": "cd721916816ca0e2", "url": "/api/blocklists/test_social" }
}
```

Both endpoints return **401** to the agent (separately filed as a logging bug). The `bl_test_ads` / `bl_test_social` nftables sets are created but empty.

## Why this is a problem

1. Every fresh router enrollment in prod inherits these placeholder blocklists in its policy snapshot. They serve no purpose for real households and create persistent log noise.
2. They are referenced from the policy but unfetchable (401), so the agent retries on policy change forever.
3. If a household admin ever picks `blockedCategories: ["test_ads"]` from the SPA, they'd be configuring against a blocklist that resolves to nothing. Either the SPA should not surface these names, or they should be removed from the prod DB.

## Likely root cause

Prod DB was initialized from a seed that included the dev-environment test blocklists. The seed step probably runs unconditionally and doesn't distinguish prod from dev.

## Suggested fix

Two parts:
1. Audit the prod DB and remove the `test_*` blocklist rows (and any references in profiles, though I doubt any prod profile references them yet).
2. Gate the test-blocklist seed on a non-prod environment flag, so future re-inits don't reintroduce them. Run the dev seed only when the API container is built/run with a `WIFIHAVEN_ENV=dev` (or similar) marker.

## Repro / evidence

- `/etc/wifihaven/policy.json` on the freshly enrolled Flint 2 (router_id `f04dd490-a8c1-431a-9aec-44f64bff3b23`) contains both entries.
- Direct curl of `https://api.wifihaven.net/api/blocklists/test_ads` and `.../test_social` both return 401 (auth required), which is correct API behavior — these shouldn't be exposed unauth — but the policy still references them.

## Related

- #660 (prod install walkthrough)
- Companion #(observability bug filed separately)

sameerparekh:
## Summary
- Adds migration `V26__remove_test_blocklists.sql` that deletes `test_ads` / `test_social` rows from `blocklist_domains` and strips them from any `profiles.blocked_categories` array. Idempotent, runs everywhere.
- Introduces `WIFIHAVEN_ENV` (values: `prod` default, `dev`). Local docker-compose and Render staging set `dev`; Render prod omits it (defaults to `prod`).
- New `DevSeeder` re-inserts the six test blocklist domains after migrations only when `cfg.isDev`. The seed previously lived in V11 unconditionally, which is the original leak source.

Closes #706.

## Before / after — policy snapshot blocklists map

Before (prod, fresh router enrollment):
```json
{
  "test_ads":    { "version": "4b6c87d2270c1c3d", "url": "/api/blocklists/test_ads" },
  "test_social": { "version": "cd721916816ca0e2", "url": "/api/blocklists/test_social" }
}
```

After (prod, post-migration, fresh enrollment): the `blocklists` map contains only categories with real domains, none of which are `test_*`. Local dev / staging still surface `test_ads` and `test_social` because `DevSeeder` re-creates the rows.

## Test plan
- [x] `mill api.test` — 190 tests pass. RouterApiSpec inserts its own `test_ads` / `test_social` rows via `BlocklistRepo.insertBatch`, so it's unaffected by V11 / V26.
- [x] Migration check (`check-migrations.sh`) passes — V26 is the next number after V25.
- [ ] **Operator post-deploy verification (prod, read-only):**
  1. After `:latest` ships, `GET https://api.wifihaven.net/api/blocklists` with admin auth — confirm no `test_ads` / `test_social` keys.
  2. Pick a known router; `GET https://api.wifihaven.net/api/admin/routers/<id>/policy-snapshot` (or equivalent) — confirm the `blocklists` map has no `test_*` entries.
  3. Re-fetch a freshly enrolled router's `/etc/wifihaven/policy.json` and confirm same.
- [ ] **Staging spot-check:** after deploy, hit staging `/api/blocklists` and confirm `test_ads` / `test_social` are still present (verifies `WIFIHAVEN_ENV=dev` path).

## Notes
- Migrations are append-only; V11 stays untouched to keep Flyway checksums valid on already-migrated DBs. The cleanup happens in V26.
- No router-side changes. No `policy_snapshot.json` contract changes.

### DIFF
--- a/api/resources/db/migration/V26__remove_test_blocklists.sql
+++ b/api/resources/db/migration/V26__remove_test_blocklists.sql
@@ -1,0 +1,17 @@
+-- #706: prod cloud DB had `test_ads` / `test_social` rows from the V11 dev
+-- seed running unconditionally. Remove the rows everywhere; dev re-seeds
+-- after migrations via the in-app DevSeeder (runs only when
+-- WIFIHAVEN_ENV=dev), so this is safe on dev too.
+
+DELETE FROM blocklist_domains
+ WHERE category IN ('test_ads', 'test_social');
+
+-- Strip any profile references too. blocked_categories is TEXT[]; no prod
+-- profile is known to reference these, but be defensive.
+UPDATE profiles
+   SET blocked_categories = array_remove(
+                              array_remove(blocked_categories, 'test_ads'),
+                              'test_social'
+                            )
+ WHERE 'test_ads'    = ANY(blocked_categories)
+    OR 'test_social' = ANY(blocked_categories);
--- a/api/src/Config.scala
+++ b/api/src/Config.scala
@@ -16,6 +16,16 @@ case class AppConfig(
   // Read from env, not HOCON, so it stays out of application.conf — debug
   // belongs to the runtime environment, not the persistent config.
   val debugEnabled: Boolean = AppConfig.envTruthy(sys.env.get("WIFIHAVEN_DEBUG"))
+
+  // #706: WIFIHAVEN_ENV gates dev-only fixtures (currently: the test_ads /
+  // test_social blocklist seed). Recognized values: "prod" (default, no
+  // fixtures), "dev" (re-seed test blocklists after migrations). Read from
+  // env, not HOCON — environment identity belongs to the deploy target, not
+  // the persistent config file. Default is "prod" so any unset deployment
+  // is safe.
+  val env: String    =
+    sys.env.get("WIFIHAVEN_ENV").map(_.trim.toLowerCase).filter(_.nonEmpty).getOrElse("prod")
+  val isDev: Boolean = env == "dev"
 }
 
 case class DbConfig(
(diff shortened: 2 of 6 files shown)


sameerparekh:
Deferring: the test_ads / test_social seed in prod is harmless until we ship blocklist management. The 401 noise loop and SPA-exposure concerns are real but low-priority; rolling this into the full blocklist-management work instead of landing it standalone.

Issue #877, opened 2026-05-22
ci: api-smoke-staging 'TimeLimit' check flaky on freshly-deployed staging


sameerparekh:
Spotted while investigating #871.

In Master CD run [26262713267](https://github.com/wifihaven/wifihaven/actions/runs/26262713267) (commit b4ec8e1, the #835 merge), the `Gate 1 — API contract smoke (staging)` → `Router API smoke against staging` step failed at:

```
▶ Usage ingest → last_seen_ip + blockReason=TimeLimit
  ✓ usage posted (90 active seconds)
  ✗ expected blocked=True reason=TimeLimit, got: blocked=False reason=None
```

This test creates a profile with \`dailyMinutes=1\`, posts 90s of usage, then expects the next router policy poll to mark the device blocked with reason \`TimeLimit\`. The same test passes consistently against the in-CI docker-compose stack and re-passed on subsequent CD runs (e.g. [26305105683](https://github.com/wifihaven/wifihaven/actions/runs/26305105683)), so this is intermittent — almost certainly a race between usage ingest visibility and the next policy GET on the freshly-deployed Render service.

Hypothesis: the staging API's usage→policy pipeline has more lag than the in-process stack (DB write visibility, cache, request-handler instance), and \`scripts/e2e-router.sh\` doesn't retry/poll for the block transition long enough.

Fix sketch: in \`scripts/e2e-router.sh\`, replace the single-shot blockReason check with a short polling loop (e.g. up to 15s, ~1s cadence) so the test tolerates the staging pipeline lag. Or push enough usage past the limit that the next single poll is unambiguous.

Severity: medium — a single bad run is fine, but combined with #871's cancel-in-progress issue it blocked the rolling artifact for ~17 hours.

PR #878, opened 2026-05-22
fix(ci): restore Master CD after #835 — split cancel-in-progress off deploy-production

### LINKED ISSUE #871: ci: Master CD has not produced a fresh openwrt-latest artifact since before #833 merged
## Problem

Since #833 merged at `2026-05-22T01:09:55Z` (commit `2492f55`), every `Master CD` workflow run on `main` has either been **cancelled** (by the next push) or **failed**:

```
2026-05-22T18:29:01Z  3f32ba2  in_progress   feat(web+api): Traffic Usage page (#846)
2026-05-22T16:37:43Z  1629c97  cancelled
2026-05-22T16:14:39Z  63f15d8  cancelled
2026-05-22T15:56:10Z  aad1b50  cancelled
2026-05-22T15:54:39Z  3d003b7  cancelled
2026-05-22T14:46:11Z  966a375  cancelled
2026-05-22T14:38:34Z  b75755b  cancelled
2026-05-22T14:37:36Z  5bbb692  cancelled
2026-05-22T01:16:57Z  b4ec8e1  failure       feat(vm): OpenWRT 25.12 / .apk router-image (#532) (#835)
2026-05-22T01:16:51Z  cf70bf1  cancelled
2026-05-22T01:12:02Z  a02becc  cancelled
2026-05-22T01:09:58Z  2492f55  cancelled     ← #833's own CD run
```

Last successful `Master CD` was `b9fc62e` at `2026-05-21T22:53:21Z` — **before** #833.

The cancel-cascade is normal: each new push to main cancels in-progress CD. The blocker is the `b4ec8e1` failure (#835, OpenWRT 25.12 .apk router-image build): subsequent pushes haven't been able to complete because they keep being superseded faster than CD takes.

Net effect: the rolling `openwrt-latest` artifact does **not** yet contain #833 or anything else merged after `b9fc62e`. So even if auto-update were working on every router, no one would have the #833 fix.

## Proposed fix

1. Investigate the `b4ec8e1` `Master CD` failure (likely in #835's new APK router-image build path).
2. Once green, the next push to main should produce a fresh artifact containing #833.
3. Consider whether `Master CD`'s cancel-policy is too aggressive — if a successful run is rare because pushes arrive faster than CD finishes, the rolling artifact will perpetually lag. May want to NOT cancel a CD run that has already passed the publish-openwrt step, so we at least get a fresh artifact even when later jobs cancel.

## Acceptance

- A successful `Master CD` run on a commit at or after `2492f55` produces a rolling `openwrt-latest` artifact whose `wifihaven` agent contains `mac_ip_tracking_rx` (the #833 marker).

## Related

- #866 (live-debug)
- #858 (downstream symptom)
- #835 (likely suspect for the original failure)

sameerparekh:
## Summary

Closes #871.

`Master CD` produced no fresh `openwrt-latest` rolling artifact for ~17h after #833 merged. Root cause is two unrelated issues stacking:

1. **The b4ec8e1 run itself failed** — not in the new APK build path from #835, but in `Gate 1 — API contract smoke (staging)` → `Router API smoke against staging`:

   ```
   ▶ Usage ingest → last_seen_ip + blockReason=TimeLimit
     ✓ usage posted (90 active seconds)
     ✗ expected blocked=True reason=TimeLimit, got: blocked=False reason=None
   ```

   Failing run: [26262713267](https://github.com/wifihaven/wifihaven/actions/runs/26262713267). This is an intermittent race between usage ingest and the next router-policy poll against the freshly-deployed Render staging service — same test passes against the in-CI docker-compose stack and passed again on a later CD run. Filed as #877 for a separate fix (poll for the block transition instead of a single-shot check). Not addressed here.

2. **Every subsequent push cancelled the run before `publish-openwrt` could complete.** This PR fixes that.

## Root cause of the rolling-artifact stall (the part this PR fixes)

`master.yml` had a workflow-level `concurrency.cancel-in-progress: true`, which cancels the **entire** run — including the deploy-production jobs — when a faster push arrives. The existing comment (lines 47–54) explicitly said we do NOT want that, and described splitting via job-level concurrency as the intended fix. That split was documented but never implemented.

With pushes arriving at < ~25 min cadence (the time the full pipeline takes), `publish-openwrt` got cancelled mid-flight on every run after b4ec8e1. The rolling release stayed pinned to the pre-#833 build.

## Change

Replace workflow-level `cancel-in-progress` with per-job concurrency on the cancellable pre-prod jobs only:

- `ci-tests`, `bootstrap-smoke`, `e2e`, `prod-stack-smoke`, `build-image`,
  `deploy-staging`, `deploy-spa-staging`, `smoke-staging`,
  `api-smoke-staging`, `gate-2-router-fake-api` → each gets its own
  job-level group with `cancel-in-progress: true`.
- `approve-production`, `retag-latest`, `publish-openwrt`,
  `deploy-prod-render`, `deploy-spa-prod` → no concurrency, so once they
  start they run to completion regardless of new pushes.

A new push still pre-empts an old run's build/staging phase (so we don't queue work), but cannot cancel an already-running prod-deploy job.

## Notes

- CD self-recovered today at 18:29Z ([run 26305105683](https://github.com/wifihaven/wifihaven/actions/runs/26305105683)) once a long enough gap between pushes let the pipeline complete. `openwrt-latest` got fresh assets at 18:54:21Z. This PR prevents the regression from recurring.
- Filed #877 for the flaky `api-smoke-staging` `TimeLimit` check.

## Test plan

- [ ] CI checks pass on this PR
- [ ] After merge, monitor next push to main — verify `publish-openwrt` is not cancelled if a subsequent push arrives while it's running
- [ ] Verify the `openwrt-latest` release timestamp advances after merge

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -48,13 +48,13 @@ on:
 # main push lands while an older run is still in build/staging, dropping
 # the older run is fine. We do NOT want cancel-in-progress applying to the
 # deploy-production jobs: cancelling mid-promotion can leave one of the
-# three sub-jobs done and the others not, producing a partial release.
-# Splitting the concurrency group by job name (via the job-level
-# `concurrency` field) is more surgical than narrowing the top-level group,
-# so we keep this group covering build/staging and let prod jobs run free.
-concurrency:
-  group: master-${{ github.ref }}
-  cancel-in-progress: true
+# three sub-jobs done and the others not, producing a partial release —
+# and in particular leaves the openwrt-latest rolling artifact stale.
+#
+# Workflow-level cancel-in-progress would cancel the whole run, including
+# deploy-production jobs. Instead we apply concurrency at the job level for
+# the cancellable build/staging jobs only; deploy-production jobs carry no
+# concurrency and always run to completion once they start. See #871.
 
 env:
   REGISTRY: ghcr.io
@@ -67,11 +67,17 @@ jobs:
   # ── 1. Full CI test matrix (unit, lint, lua, python, shell, frontend) ─────
   ci-tests:
     name: CI
+    concurrency:
+      group: master-build-${{ github.ref }}-ci-tests
+      cancel-in-progress: true
     uses: ./.github/workflows/ci.yml
 
   # ── 2. E2E + smoke tests (unconditional) ──────────────────────────────────
   bootstrap-smoke:
     name: bootstrap-host.sh smoke (Ubuntu container)
+    concurrency:
+      group: master-build-${{ github.ref }}-bootstrap-smoke
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -82,6 +88,9 @@ jobs:
 
   e2e:
     name: Live e2e against staging stack
+    concurrency:
+      group: master-build-${{ github.ref }}-e2e
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
@@ -109,6 +118,9 @@ jobs:
 
   prod-stack-smoke:
     name: deploy/docker-compose.prod.yml smoke
+    concurrency:
+      group: master-build-${{ github.ref }}-prod-stack-smoke
+      cancel-in-progress: true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


sameerparekh:
Superseded by #880 — splitting Master CD into separate router and api/ui pipelines is a more durable fix than per-job concurrency on the unified workflow. The path-filtered split means api/ui churn cannot cancel router publishes (and vice versa) at the workflow level, so cancel-in-progress isn't a concern within each pipeline.

PR #1101, opened 2026-05-27
fix(hooks): scope pre-push scalafmt to api/ + shared/

sameerparekh:
## Summary
- `scalafmt --check` with no path args recurses from the repo root, walking into sibling `.claude/worktrees/<branch>/` checkouts. An unformatted file in any concurrent worktree blocks unrelated pushes.
- Pass explicit source roots (`api/`, `shared/`, optional `mill-build/`) so the walk stays inside the current branch's tree.
- Hit while pushing [#1100](https://github.com/wifihaven/wifihaven/pull/1100) — a sibling worktree's half-formatted Scala file failed the push for a one-line workflow change.

## Test plan
- [ ] In a repo with multiple worktrees containing unformatted Scala, `git push` from the main checkout no longer trips on sibling files.
- [ ] `scalafmt --check api shared` still catches real formatting issues in the current tree.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -11,11 +11,20 @@ cd "$ROOT"
 fail=0
 
 # ── Scala: scalafmt check across the whole repo ───────────────────────────
+# Pass explicit source roots so the recursive walk doesn't descend into
+# sibling git worktrees under .claude/worktrees/ — an unformatted file
+# in another branch's worktree must not block this push.
 if command -v scalafmt >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


sameerparekh:
Closing per feedback: the right fix is to run scalafmt from inside the worktree (where pre-push already does the right thing — it cd's to `git rev-parse --show-toplevel` which inside a worktree is the worktree root). The hook itself is fine; pushing from the main checkout while sibling worktrees have dirty Scala is the operator-side problem to avoid.

Issue #1117, opened 2026-05-27
e2e: fake-API scenario asserting agent tails nflog and emits connection_attempt events for blocked MACs


sameerparekh:
Retargeted from the closed `mac_drops` counter-diff approach to the nflog-based design in #1122.

## Why

The new approach (#1122) replaces counter polling with an nflog tail in the agent. We need a targeted regression test for that wire-up — the live e2e scenario covers the end-to-end outcome but doesn't pin the agent's behavior in isolation. Fake-API tests give us cheap CI coverage and exercise the exact failure modes (kernel notification delivery, dedup, retry-queue behavior under transient API failure) without booting a real router VM.

## Scope

Boot the `wifihaven-agent` against a recording HTTP sink (matching the existing fake-API pattern in `openwrt/test/agent_spec.sh` / `test_install_spec.sh`) and assert:

1. When a snapshot marks a MAC blocked, the rendered nft ruleset contains a per-MAC drop rule with a `log group N` action AND `counter drop comment \"wh_drop:<mac>:<reason>\"`.
2. When nflog fires (simulated by an injected nflog reader returning canned notifications), the agent posts a `RouterEventsRequest` to `/api/router/events` within one tick.
3. The posted payload carries one `connection_attempt` per dropped flow with `allowed=false`, `reason=<MacBlockReason wire string>`, the right `(mac, dst_ip)`, and a `host` resolution (or absent if the dnsmasq cache had no entry).
4. Per-flow dedup at the agent collapses 1000 SYN retries within the dedup window down to a single posted event.
5. On API 5xx the events go onto the existing retry queue (the same `event_id`-based dedup that connection_attempt already uses applies on replay).
6. nflog reader is an injected dependency (matching the policy/usage tests' fn-injection pattern) — production wires it to the real nflog socket, tests wire it to a canned-notification producer.

## Out of scope

- Real OpenWRT VM (covered by the live scenario in #1122).
- API-side persistence assertions (covered by #1116).

## Related

#1122 (foundation), #1116 (fake-router counterpart).

Issue #1116, opened 2026-05-27
e2e: fake-router scenario asserting connection_events ingest for firewall-blocked MAC flows


sameerparekh:
Retargeted from the closed `mac_drops` approach to the nflog-based design in #1122.

## Why

The new approach (#1122) emits synthetic `connection_attempt` events from the agent when nflog notifications fire on per-MAC drop rules. The agent-side wire-up has unit coverage via #1117 (fake-API tests), but the **API-side contract** — does the existing `/api/router/events` ingest correctly persist these synthesized events with `allowed=false reason=<MacBlockReason>` and dedup on `event_id`? — needs an isolated e2e scenario that doesn't require a real router.

## Scope

- Extend `scripts/e2e/lib/` with a `FakeRouter` helper that POSTs `RouterEventsRequest` payloads containing one or more synthesized `connection_attempt` events with `allowed=false, reason="Paused"/"Schedule"/"TimeLimit"/"Manual"/"Unmanaged"`.
- New scenario file `scripts/e2e/scenarios/test_blocked_mac_events_ingest_fake_router.py`.
- Assertions:
  1. Authenticates the router token.
  2. Persists rows into `connection_events`.
  3. Dedups replays on `event_id` (existing behavior — verify it works for the blocked-MAC variant).
  4. Surfaces the rows on `/api/logs?mac=...&blocked=true` and `/api/connection-events/series?bucket=1h`.
  5. Reason attribution survives the round-trip (paused / schedule / time_limit / manual / unmanaged).

## Out of scope

- nflog wire-up on the router (covered by #1117).
- Anything requiring a client VM.

## Related

#1122 (foundation), #1117 (fake-API counterpart).

Issue #1115, opened 2026-05-27
mac_drops: connection-events-style groupBy/cursor/filter UI in LogsPage


sameerparekh:
## Why

#1103 surfaces `mac_drops` as a small standalone panel in [web/src/pages/LogsPage.tsx](web/src/pages/LogsPage.tsx) — flat table summed across the visible window. The Connection Events table next to it has:

- column-header groupBy toggles (device / profile / domain / app)
- per-column filter popovers (HeaderFilter)
- keyset-cursor infinite scroll
- shared FilterShelf (mac + profile selection, time-anchor calendar)
- bucket selector (1m / 10m / 1h / 12h / 1d / 1w)

The mac_drops panel should reach the same ergonomic level so an admin who notices a drop signal can drill into *when* it happened and *which profile owns it* without leaving the Logs page.

## Scope

- Promote `MacDropsPanel` to a full sibling component of `RawEventsView` / `AggregatedEventsView`.
- Add `groupBy` server-side on `/api/mac-drops/series` with `device | profile | reason` (no domain — there isn't one).
- Add keyset cursor + page envelope (`MacDropsSeriesPage.nextCursor`) the same way `ConnectionEventSeriesPage` does.
- Reuse `FilterShelf` for mac/profile filtering and `until` anchor.
- Wire the same column-header groupBy toggles + HeaderFilter popovers in the panel.
- Bucket selector — share with the existing `EventsBucket` UI.

## Tests

- `MacDropsApiSpec` extended for groupBy + cursor pagination.
- `LogsPage.test.tsx` extended for the groupBy toggle interactions on the drops panel.

## Depends on

#1114 (rollup tables) — wide buckets need the rollups in place to be cheap.

sameerparekh:
Superseded by #1122 — blocked-MAC drops now appear as native rows on the existing Connection Events table, so the Logs page UI (groupBy / cursor / HeaderFilter) already covers them. No separate panel needed.

Issue #1114, opened 2026-05-27
mac_drops: rollup tables + retention horizons matching connection_events


sameerparekh:
## Why

#1103 landed `mac_drops` as a flat per-(mac, reason, period_end) table with 30-day retention, queried on-the-fly by `/api/mac-drops/series`. That mirrors the **raw** side of `connection_events`, but not the rollup track that #809 added for `traffic_*` and that #1041 / #847 brought to the events surface.

Once a household accumulates weeks of drops, the on-the-fly `date_bin` over `mac_drops` will be the same kind of cost cliff that drove #809. We should land the rollup parity now rather than after the chart visibly stutters.

## Scope

- **Hourly rollup**: `mac_drops_hourly(router_id, mac, reason, bucket_start, packets, bytes)`. Rebuilt every 5 min by extending `RollupJobs.hourlyLoop` to also fold `mac_drops`.
- **Daily rollup**: `mac_drops_daily(router_id, mac, reason, date, packets, bytes)`. Rebuilt hourly by `dailyLoop`.
- **Retention**: 90 days hourly, 180 days daily (match `traffic_*` horizons in [api/src/usage/RetentionSweepJob.scala](api/src/usage/RetentionSweepJob.scala)).
- **Series read**: `/api/mac-drops/series` picks the narrowest table that covers the requested bucket size — raw for ≤10m buckets, hourly for 1h/12h, daily for 1d/1w. Same selection rule the connection-events series uses.
- **Repo + Flyway migration** following the V38 pattern.

## Tests

- Feature spec asserting the series endpoint reads from the right table per bucket size.
- Rollup job test: a fresh `mac_drops` insert appears in hourly within one tick.
- Retention sweep extension: rows older than horizon removed from each rollup table.

## Not in scope

- UI grouping (separate issue — see follow-up).
- Apex/app-style joins (mac_drops doesn't have a host axis).

sameerparekh:
Superseded by #1122 — the nflog-into-connection_events approach removes the separate mac_drops table, so rollup parity is no longer a separate concern (the existing connection_events rollups from #809 already cover the blocked-MAC traffic surfaced by #1122).

PR #1119, opened 2026-05-27
feat: surface MAC-level packet drops via new mac_drops table (#1103)

### LINKED ISSUE #1103: feat: surface MAC-level packet drops as a separate first-class table/series (not folded into connection_events)
## Problem

The Connection Events chart and `/api/connection-events/series` only count queries that *reach dnsmasq* — i.e., per-host blocks via NXDOMAIN/0.0.0.0 responses (categories, `extraBlocked`, blocklists).

When a device is blocked at the **MAC level** — via `@blocked_macs` from any of:
- profile paused
- bedtime/schedule window
- daily time limit exhausted
- admin manual block
- unmanaged-mac default-block (#374)

— the router enforces it with `ether saddr @blocked_macs drop` in nftables. Packets never reach dnsmasq, so **zero events are logged, blocked or otherwise**. From the admin's perspective the device just goes silent.

## User impact (live debug, 2026-05-26)

Operator noticed Kid Mac stopped appearing in Connection Events after noon. Took several minutes of API spelunking to determine: device was in `@blocked_macs` because the Kids profile's 30-min daily cap exhausted. The UI provided no signal:
- Device's `lastSeen` (DHCP/ARP) still updating
- No \"blocked\" entries in Connection Events
- No \"this device is currently blocked because X\" banner anywhere

Same opacity affects: pause toggles, bedtime onset, time-limit exhaustion, unmanaged-MAC default-block, and the daily reset window.

## Design — separate `mac_drops` table, separate series

These drops are a fundamentally different signal from connection events:

| | `connection_events` | `mac_drops` (proposed) |
|---|---|---|
| Source | dnsmasq query log (per-name) | nftables per-MAC counter |
| Cardinality | high (one row per query) | low (one row per `(mac, reason)` per polling interval) |
| Key | `(mac, host, ts)` | `(mac, reason, period_end)` |
| Question answered | \"what did this device try to reach?\" | \"what was *prevented* from leaving this device?\" |
| Trigger | DNS resolution | firewall packet drop |

Folding them into the same table would force `host` to be nullable (or carry a synthetic value), conflate \"DNS NXDOMAIN block\" with \"packet drop at firewall,\" and either explode cardinality (per-host drops in dnsmasq don't have packet counts; per-MAC drops in nftables don't have host names) or lose detail.

### Schema

```sql
CREATE TABLE mac_drops (
  id           bigserial PRIMARY KEY,
  router_id    uuid NOT NULL REFERENCES routers(id),
  mac          macaddr NOT NULL,
  reason       text NOT NULL,        -- 'paused' | 'schedule' | 'time_limit' | 'manual' | 'unmanaged'
  packets      bigint NOT NULL CHECK (packets >= 0),
  bytes        bigint NOT NULL CHECK (bytes >= 0),
  period_start timestamptz NOT NULL,
  period_end   timestamptz NOT NULL,
  UNIQUE (router_id, mac, reason, period_end)  -- idempotent ingest
);
CREATE INDEX mac_drops_ts ON mac_drops (period_end);
CREATE INDEX mac_drops_mac_ts ON mac_drops (mac, period_end);
```

### Router side

In `render.lua`, the `@blocked_macs drop` rule becomes a counter rule (or per-reason chains so we can attribute the drop). Agent polls `nft list ruleset` counters every N seconds, diffs against last snapshot, posts the deltas. Reuses the existing usage.post retry+backoff queue.

### API side

- `POST /api/router/mac-drops` — router-token auth, batch ingest, dedup on the unique key.
- `GET /api/mac-drops/series?hours=24&bucket=1h&mac=&reason=` — admin-only, mirrors `/api/connection-events/series` shape but its own endpoint.
- SPA renders the new series as a second band on the existing chart — same time axis, distinct colour/legend — or as a separate small chart panel. Decision deferred to design.

### Feature — `MacDropsApiSpec.scala` (new)

1. **Ingest**: POST a batch `[(mac, reason, packets, bytes, periodEnd), ...]` with valid router token → rows persist, idempotent on replay (unique constraint), wrong-router-token rejected, admin-token rejected.
2. **Negative-delta rejection**: posting `packets=-5` returns 400; no row inserted.
3. **Series read**: GET with `hours=24&bucket=1h&mac=...` returns hourly buckets, supports `reason=` filter, supports `mac=` repeated. Admin-only.
4. **Authorization**: parent/adult/kid roles get appropriate read access (mirrors connection events).
5. **Retention**: rows older than retention horizon are swept by the existing daily retention job (extend #811's sweep).

### Feature — `PolicySnapshotMacDropAttributionSpec.scala` (new)

For a profile in each block state (paused / schedule / time_limit / unmanaged / manual), assert the snapshot's `BlockRules.blockReason` matches what the router would tag drops as. If these drift, the chart will mis-label drops by reason — caught at API-test time.

### E2E — `scripts/e2e/scenarios/test_mac_drops_observability.py` (new)

1. **Pause**: pause Kids profile → wait one polling interval → assert `GET /api/mac-drops/series?mac=<kid>` shows rows with `reason=paused` and packets > 0 (after the test fires traffic at the device).
2. **Time limit**: inject presence to exhaust cap → traffic from the device → same assertion with `reason=time_limit`.
3. **Schedule**: open a near-now schedule window → same with `reason=schedule`.
4. **Unmanaged**: unprofiled MAC under `unmanagedMacPolicy.policy=block` (#374) → same with `reason=unmanaged`.
5. **Stop on unblock**: lift the pause / extend the cap / close the schedule → drop counters stop incrementing within one polling interval. **Absence is asserted, not just presence** — guards against zombie counters.
6. **Cross-check vs `connection_events`**: while blocked, `connection_events` for that MAC remains empty (the original symptom). Assertion: `mac_drops` is the only signal — the chart must surface it. If a future change accidentally starts logging blocked-MAC queries to `connection_events` instead, this test forces a deliberate choice.

### Web — `web/src/pages/LogsPage.test.tsx` (extend)

Mock `/api/mac-drops/series` returning drops-only for a MAC while `/api/connection-events/series` is empty for that MAC → assert the chart still renders a visible signal for that device with a \"blocked: reason\" label. Catches a regression where the SPA only consults connection events.

## Related

- See [feedback_extraallowed_beats_blocked](memory) — `extraAllowed` already carves around `@blocked_macs` at render time, but nothing surfaces the resulting drops.
- #844 (observability surface restructure) — this lives inside that effort.
- #374 (unmanaged-mac alerting/default-block) — its block path also produces silent drops today; this issue makes that path observable.
- #811 (daily retention sweep) — extend to cover `mac_drops`.
- [#1104](https://github.com/wifihaven/wifihaven/issues/1104) — UI-vs-snapshot date drift that initially masked which day's cap was being enforced.
- [#1105](https://github.com/wifihaven/wifihaven/issues/1105) — exempt-from-daily app drops that are also invisible today.
- Discovered 2026-05-26 live debug — prod Kid Mac silently blocked post-cap with no chart surface.

sameerparekh:
## Summary

- New `mac_drops` table + ingest/read endpoints surface drops from `@blocked_macs` (paused / schedule / time_limit / manual / unmanaged) that `connection_events` can't see because the firewall drops the packets before dnsmasq.
- `render.lua` converts the family-agnostic `@blocked_macs drop` into per-MAC `counter drop comment "wh_drop:<mac>:<reason>"` rules; the agent polls those counters and POSTs deltas via the existing retry-queue pattern.
- LogsPage gains a small drops panel — hidden in the steady state, lights up the moment a kid's iPad goes silent because their cap is exhausted.
- `MacBlockReason.Unmanaged` added so #374's default-block path is distinguishable from an admin's explicit Manual block at the chart.

Closes #1103.

## Test plan

- [x] `mill api.test.testOnly wifihaven.api.feature.MacDropsApiSpec` — 8 tests pass (ingest auth, idempotency, negative-delta rejection, mac/reason filters, child-role 403, retention sweep).
- [x] `mill api.test.testOnly wifihaven.api.feature.PolicySnapshotMacDropAttributionSpec` — 3 tests pass; pins the wire-string contract between API + router for all five reasons.
- [x] `mill api.test` — full suite green (192 test classes, all `0 tests failed`; pre-existing mill 1.1.5 reporter race on parallel workers, not in this change).
- [x] `mill scalafmtCheckAll` — formatted.
- [x] `npx tsc --noEmit` (web) — clean.
- [x] busted `render_spec` — 137 successes, render assertions updated for per-MAC `counter drop comment "wh_drop:..."` rules.
- [x] `LogsPage.test.tsx` — added panel-visible + panel-hidden cases (vitest couldn't be invoked in the worktree due to a local `@rolldown/binding-darwin-arm64` install hiccup unrelated to this change; tsc passes).
- [ ] `scripts/e2e/scenarios/test_mac_drops_observability.py` — written but not run from here; needs the e2e VM stack.

## Follow-ups filed

- #1114 — rollup tables (`mac_drops_hourly` / `mac_drops_daily`) + 90/180-day retention parity with `traffic_*`.
- #1115 — connection-events-style groupBy / cursor / HeaderFilter in the panel (depends on #1114 for cheap wide buckets).
- #1116 — fake-router e2e scenario asserting the `/api/router/mac-drops` wire contract under synthetic load.
- #1117 — fake-API e2e scenario asserting the agent's counter-diff + retry-queue path.

## Wire-shape notes

- Atomic API + SPA deploy. No feature flag, no shim — the router agent is the only client; old agents simply don't post to the new endpoint until upgraded.
- `MacBlockReason` parser now accepts a fifth variant (`Unmanaged`); existing snapshots emitting `Manual` for the unmanaged-block path are updated to emit `Unmanaged` (PolicySnapshotBlockedMacsSpec + RouterApiSpec updated to match).

### DIFF
--- a/api/resources/db/migration/V39__mac_drops.sql
+++ b/api/resources/db/migration/V39__mac_drops.sql
@@ -1,0 +1,29 @@
+-- #1103: surface MAC-level packet drops from nftables `@blocked_macs` (paused /
+-- schedule / time_limit / manual / unmanaged) as a first-class series.
+--
+-- `connection_events` records per-name DNS-resolved attempts. When a MAC is
+-- blocked at the firewall layer the packets never reach dnsmasq, so the
+-- chart goes silent. The router agent polls per-(mac, reason) counter rules
+-- and posts deltas; we store them here so /api/mac-drops/series can return
+-- a second band on the Logs chart.
+--
+-- Unique key is (router_id, mac, reason, period_end) — replays from the
+-- agent's retry queue collapse on ON CONFLICT DO NOTHING.
+
+CREATE TABLE mac_drops (
+  id           bigserial PRIMARY KEY,
+  router_id    uuid NOT NULL REFERENCES routers(id) ON DELETE CASCADE,
+  -- TEXT (not macaddr) to match the household-wide convention — see V4
+  -- connection_events.mac and V1 devices.mac. Mixing column types makes
+  -- joins clunky and Meta[MacAddress] is Meta[String].
+  mac          TEXT NOT NULL,
+  reason       text NOT NULL,
+  packets      bigint NOT NULL CHECK (packets >= 0),
+  bytes        bigint NOT NULL CHECK (bytes >= 0),
+  period_start timestamptz NOT NULL,
+  period_end   timestamptz NOT NULL,
+  UNIQUE (router_id, mac, reason, period_end)
+);
+
+CREATE INDEX mac_drops_ts ON mac_drops (period_end);
+CREATE INDEX mac_drops_mac_ts ON mac_drops (mac, period_end);
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -128,34 +128,35 @@ object Main extends ZIOAppDefault {
       bundledBlocklists: Map[wifihaven.shared.types.BlocklistId, BundledBlocklist],
   ) =
     for {
-      auth        <- ZIO.service[AuthService]
-      userRepo    <- ZIO.service[UserRepo]
-      upRepo      <- ZIO.service[UserProfileRepo]
-      profileRepo <- ZIO.service[ProfileRepo]
-      schedRepo   <- ZIO.service[ScheduleRepo]
-      hsRepo      <- ZIO.service[HouseholdSettingsRepo]
-      tlRepo      <- ZIO.service[TimeLimitRepo]
-      stlRepo     <- ZIO.service[SiteTimeLimitRepo]
-      deviceRepo  <- ZIO.service[DeviceRepo]
-      blRepo      <- ZIO.service[BlocklistRepo]
-      blCache     <- ZIO.service[BlocklistCache]
-      blFetcher2  <- ZIO.service[BlocklistFetcher]
-      usageRepo   <- ZIO.service[TimeUsageRepo]
-      extRepo     <- ZIO.service[TimeExtensionRepo]
-      routerRepo  <- ZIO.service[RouterRepo]
-      trafficRepo <- ZIO.service[TrafficReportRepo]
-      rollupRepo2 <- ZIO.service[RollupRepo]
-      connRepo    <- ZIO.service[ConnectionEventRepo]
-      blockEvRepo <- ZIO.service[BlockEventRepo]
-      alertRepo   <- ZIO.service[AlertRepo]
-      appRepo     <- ZIO.service[AppRepo]
-      notifier    <- ZIO.service[Notifier]
-      policy      <- ZIO.service[PolicyService]
-      timeStatus  <- ZIO.service[wifihaven.api.policy.TimeStatusService]
-      cfg         <- ZIO.service[AppConfig]
-      clock       <- ZIO.service[Clock]
-      timeCache   <- ZIO.service[TimeStatusCache]
-      xa          <- ZIO.service[Transactor[Task]]
+      auth         <- ZIO.service[AuthService]
+      userRepo     <- ZIO.service[UserRepo]
+      upRepo       <- ZIO.service[UserProfileRepo]
+      profileRepo  <- ZIO.service[ProfileRepo]
+      schedRepo    <- ZIO.service[ScheduleRepo]
+      hsRepo       <- ZIO.service[HouseholdSettingsRepo]
+      tlRepo       <- ZIO.service[TimeLimitRepo]
+      stlRepo      <- ZIO.service[SiteTimeLimitRepo]
+      deviceRepo   <- ZIO.service[DeviceRepo]
+      blRepo       <- ZIO.service[BlocklistRepo]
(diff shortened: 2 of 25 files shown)


sameerparekh:
Closing — design changed after review discussion.

The premise behind `mac_drops` as a separate table was wrong. `connection_events` is conntrack-driven, not DNS-driven, and the reason it goes silent for firewall-blocked MACs is a netfilter hook-ordering detail: dropped packets at the forward hook never reach postrouting, so `IPCT_NEW` never fires, so `conntrack -E NEW` doesn't see them. Not a DNS-layer hole.

The cleaner fix is to emit synthetic `connection_attempt` events from the agent by tailing nflog on the per-MAC drop rules — the existing Logs page then naturally surfaces blocked-MAC traffic alongside normal traffic, no parallel table, no parallel UI surface. Bytes/packets become a rollup query over `connection_events`.

Replaced by a fresh issue + PR using the nflog approach. Discussion: see follow-up comments on #1103.

Issue #1122, opened 2026-05-27
Surface firewall-blocked MAC traffic in connection_events via nflog


sameerparekh:
Implementation issue for the user-facing problem stated in #1103.

## Approach

Add `log group N` (nflog) actions to the per-MAC drop rules emitted by [render.lua](openwrt/files/usr/lib/lua/wifihaven/render.lua), so every dropped packet generates a netlink notification. Tail nflog from the agent (via `nft monitor trace` or ulogd2-style nflog socket) and synthesize `connection_attempt` events with `allowed=false` and `reason=` set to the snapshot's `BlockReason` for the MAC.

Result: blocked-MAC traffic shows up on the existing Connection Events table on the Logs page — same row shape as everything else, same filters, same rollups. No parallel table, no parallel UI surface.

## Schema impact

Zero. `connection_events` already carries `(router_id, mac, host_type, host_value, dest_ip, allowed, reason, ts)` and the wire shape (`connection_attempt`) already supports `allowed=false` with a free-form reason. The router's existing retry-queue / dedup-on-event_id path applies as-is.

## Reason wire strings

`MacBlockReason.Paused | Schedule | TimeLimit | Manual | Unmanaged`. `Unmanaged` is new (#374's default-block path was previously folded into `Manual`); needs a parallel addition to the `BlockReason` enum and snapshot `MacBlockReason` so PolicySnapshot emits the right typed value when `unmanagedMacPolicy.policy=block`.

## render.lua change

Each per-MAC drop rule gets a `log` prefix action:

```
ether saddr <mac> log group <N> counter drop comment "wh_drop:<mac>:<reason>"
```

(`log group` is non-blocking and async — drops are not delayed.) The `counter` stays for ops debugging via `nft list ruleset` but is not the read path anymore.

## Agent change

Replace the polling counter-diff path (out of scope after pivot) with an nflog tail:

- Subscribe to nflog group `N` via the Lua `lua-nflog` or by execing `nft monitor trace` and parsing.
- For each notification, extract `(mac, dst_ip, comment)` — the comment carries `wh_drop:<mac>:<reason>`.
- Look up `hname` via the existing dnsmasq query-log cache (same path conntrack.lua uses).
- Build a `connection_attempt` event with `allowed=false`, `reason=<reason>`, and post via the existing `/api/router/events` retry queue.
[...]

Issue #657, opened 2026-05-18
ci: capacity planning for self-hosted KVM runner once Gates 2+3 matrix is live


sameerparekh:
Parent: #652
Related: #532

## Goal

Decide whether to add a second self-hosted KVM runner or beef up the existing one, once Gates 2 (#654) and 3 (#655) are running their .ipk + .apk matrix in CI.

## Why this is a separate ticket

Premature to size before we know real wall times. Run the matrix serial-on-one-runner for a couple weeks first, then measure.

## Decision inputs to collect

- Wall time per matrix entry for Gate 2 (router + fake API).
- Wall time per matrix entry for Gate 3 (router + real staging).
- Queue depth: how often does the runner have multiple jobs waiting?
- Frequency of cross-job contention with other self-hosted workflows already on this runner ([keep-staging-warm.yml](../blob/main/.github/workflows/keep-staging-warm.yml), [e2e-kvm-sanity.yml](../blob/main/.github/workflows/e2e-kvm-sanity.yml)).
- Developer feedback: is "PR feedback takes N minutes" actually painful or merely noticeable?

## Options to weigh once we have data

- Stay serial on the current box (cheapest; acceptable if wall time stays modest).
- Beefier single runner (more cores → matrix entries can run as parallel qemu instances on one host).
- Add a second physical runner (true parallelism, also redundancy if one box dies).

## Out of scope

- Moving to github-hosted runners (qemu/KVM constraints make this impractical for Gates 2+3; documented in [docs/ops/kvm-runner.md](../blob/main/docs/ops/kvm-runner.md)).

sameerparekh:
Bumping this — #655 adds Gate 3a + 3b to the \`concurrency: e2e-vm\` group on the self-hosted KVM runner. Combined with Gate 2 (#686) and live-mode \`e2e-vm.yml\`, a push to main touching \`shared/contract/**\` (fires both pipelines) now queues:

- Gate 2 fake-API: 2 arms (ipk + apk)
- Gate 3a router-staging: 2 arms
- Gate 3b api-staging: 2 arms
- e2e-vm live (parallel pipeline, also on \`e2e-vm\` group)

= up to ~8 sequential qemu runs on one runner. At ~5–10 min/arm wall time that's worst-case ~80 min serialized publish gating on the router pipeline (which has Gate 3a in its needs:).

[...]

PR #1146, opened 2026-05-28
fix(openwrt): pull NFLOG kernel deps so log-group drop rules load (#1144)

### LINKED ISSUE #1144: ci: P1 — Gate 3a block-page timeout; blocked app not in staging policy snapshot (apps-migration regression)
## Symptom

Master Router CD Gate 3a (router-side staging smoke) fails on **both** ipk and apk:

```
gate3/test_smoke.py:48: in test_gate3_smoke
E   TimeoutError: timed out waiting for block page response for example.com
```

Failing run: https://github.com/wifihaven/wifihaven/actions/runs/26551394292

The router booted and enrolled fine; client DNS for `example.com` resolved to real Cloudflare IPs (`172.66.147.243`, `104.20.23.154`) — i.e. **`example.com` was never blocked**, so the block-page DNAT never fired.

## Not a test-fixture problem

`scripts/e2e/gate3/conftest.py:138-161` already provisions the block via the apps mechanism (post-#764):
```
blocked_app = admin.create_app(hosts=["example.com"])
PUT /api/apps/{aid}/policy/{pid} {mode: "blocked"}
```
So the test is using the current contract correctly. The app-based block simply isn't appearing in the router policy snapshot served by **staging**.

## Same cluster as #1138 / #1133

This is almost certainly the same root cause as:
- **#1138** — Gate 1 `#1105` step: exempt time_limited app not folded into `extraAllowed` (apps→snapshot path).
- **#1133** — dev API throws `column "extra_blocked" does not exist` SQLException on policy fetch (snapshot generation path, post-#1054 dropped columns).

All three are policy-snapshot regressions in the apps-migration wave (#1054 / #1106 / #1108). The likely culprit is the #1108 TimeStatusService snapshot refactor and/or the #1054 dropped-column fallout dropping app-derived block/allow entries from the snapshot.

## Investigation (coordinate — do NOT double-fix)

1. **Check #1138 and #1133 PRs first.** If either already repairs the snapshot's apps→block-set / apps→extraAllowed population, this issue is the *verification* of the same fix against Gate 3a — add Gate 3a green as an acceptance criterion there and do not open a competing PR touching the same Scala.
2. If the apps-derived **block set** (mode=blocked apps → snapshot block list) is a distinct gap not covered by #1138's carve fix, then this issue owns that specific fix: ensure a `mode: "blocked"` app's hosts land in the router policy snapshot's block set. Add a service-level test (TDD, failing first per AGENTS.md).
3. Reproduce against staging via the admin API (create app, assign blocked, fetch `/api/router/policy`, assert host in block set) and/or locally.

## Acceptance

- A `mode: "blocked"` app's hosts appear in the router policy snapshot served by staging.
- Gate 3a (both ipk + apk) goes green.
- If fixed under #1138/#1133, this is closed as covered-by with a backlink; if distinct, a service-level test guards the apps→block-set population.

## Priority

P1 — blocks Master Router CD (Gate 3a is a required gate).

## Related

- #1138, #1133 (same cluster), #1106/#1105, #1108, #1054/#764
- #1129 (recent gate-test fix)

sameerparekh:
## Summary

Master Router CD **Gate 3a** has been red on both ipk + apk since #1122 landed — `TimeoutError: timed out waiting for block page response for example.com`. Root cause is **not** the apps→snapshot path the issue originally suspected (that's verifiably correct — see below); it's a missing router package dependency.

- **Regression bisected to #1122.** Gate 3a was green on the immediately preceding image (`62b252ee`) and red on the first image built after #1122 (`45e08a82`). #1122 is the only render.lua change in that window.
- **Mechanism.** #1122 added `log group <N> counter drop` to every forward-chain drop rule (NFLOG → `connection_attempt` visibility for blocked flows, #1103). The nftables `log ... group` expression is NFLOG and needs `nf_log`/`nfnetlink_log` kernel support — OpenWRT package **`kmod-nft-log`**. It's absent from the default x86 image (only `kmod-nft-{core,fib,nat,offload}` ship via firewall4) and was never added to the wifihaven package deps. The agent applies the ruleset with a single **atomic `nft -f`** (`policy.lua`), so the one unsupported statement makes the **entire** wifihaven table fail to load — including the block-page DNAT chain, which has no log statement of its own. Net effect: all enforcement silently disappears. Blocked hosts resolve normally and serve real upstream with no block page — exactly the Gate 3a symptom.
- **Fix.** Add `kmod-nft-log` to all three package-dependency declarations (`Makefile` DEPENDS, `build-ipk.sh` control, `build-apk.sh` mkpkg). `PACKAGES_LIST="wifihaven ..."` in the image build resolves it in, same as the other `kmod-nft-*` deps.

## Why it slipped through

- The render unit tests only string-compare output — they assert `log group 1` is emitted but never load it into a kernel.
- `render_spec.lua` *does* have an `nft -c -f` syntax check, but it's **skipped when the `nft` binary isn't on the host** (it is on the CI lint runner → pending/skip).
- Gate 2 (qemu router that actually loads the ruleset) shares the self-hosted-runner concurrency group with Gate 3a and has been **cancelled in every run since #1122**, so it never re-validated the load.

## Not an API/snapshot bug

Verified live against staging (now serving `aa070c0`): created a `mode:blocked` app (`hosts:[example.com]`), assigned to a fresh profile + device, enrolled a throwaway router, fetched `/api/router/policy` → `extraBlocked: ['example.com']` ✅. The apps-migration cluster (#1054/#1106/#1108) was already deployed before the last green Gate 3a, so it isn't the regression. No API change needed.

## Test plan

- [x] `openwrt/test/nft_log_dep_spec.sh` fails pre-fix (red), passes post-fix (green) — committed as separate red→green commits
- [x] Full `openwrt/test/run_tests.sh` suite green (434 pass; 2 pre-existing `failover_spec` failures unrelated to this change, reproduce on `main`)
- [ ] router-image-build resolves `kmod-nft-log` into both ipk + apk images (CI)
- [ ] Gate 2 (router fake-API e2e) green — confirms the ruleset now loads
- [ ] Gate 3a green (ipk + apk) — the regression gate

> Note: Gate 2/3a require the self-hosted KVM runner and can't be run locally; CI on this PR is the validation.

## Follow-ups (separate)

- Un-skip / hard-require the `nft -c -f` ruleset syntax check so a future unsupported expression fails a normal CI job, not just the KVM gates.
- Gate 2 starvation: it never completes because it shares the concurrency group with Gate 3a (#657 capacity).

Closes #1144.

### DIFF
--- a/openwrt/Makefile
+++ b/openwrt/Makefile
@@ -13,7 +13,11 @@ define Package/wifihaven
   SECTION:=net
   CATEGORY:=Network
   TITLE:=WifiHaven router agent
-  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua
+  # kmod-nft-log: the forward-chain drop rules carry an NFLOG `log group`
+  # statement (#1122). It is not in the default image, and the agent's atomic
+  # `nft -f` rejects the whole table without it — silently dropping all
+  # enforcement, including the block-page DNAT (#1144).
+  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nft-log
   PKGARCH:=all
 endef
 
--- a/openwrt/Makefile
+++ b/openwrt/Makefile
@@ -13,11 +13,13 @@ define Package/wifihaven
   SECTION:=net
   CATEGORY:=Network
   TITLE:=WifiHaven router agent
-  # kmod-nft-log: the forward-chain drop rules carry an NFLOG `log group`
-  # statement (#1122). It is not in the default image, and the agent's atomic
-  # `nft -f` rejects the whole table without it — silently dropping all
-  # enforcement, including the block-page DNAT (#1144).
-  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nft-log
+  # NFLOG deps: the forward-chain drop rules carry an NFLOG `log group`
+  # statement (#1122) on both v4 and v6 rules in the inet table. These are not
+  # in the default image, and the agent's atomic `nft -f` rejects the whole
+  # table without them — silently dropping all enforcement, including the
+  # block-page DNAT (#1144). kmod-nfnetlink-log is the NFLOG backend;
+  # kmod-nf-log{,6} provide the netfilter log framework for each family.
+  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nfnetlink-log +kmod-nf-log +kmod-nf-log6
   PKGARCH:=all
 endef
 
--- a/openwrt/build-apk.sh
+++ b/openwrt/build-apk.sh
@@ -146,7 +146,7 @@ rm -f "$OUT_APK"
     --info "license:MIT" \
     --info "url:https://github.com/wifihaven/wifihaven" \
     --info "maintainer:WifiHaven <noreply@example.com>" \
-    --info "depends:lua libuci-lua luci-lib-jsonc conntrack curl uhttpd-mod-lua" \
+    --info "depends:lua libuci-lua luci-lib-jsonc conntrack curl uhttpd-mod-lua kmod-nft-log" \
     --script "post-install:$WORK/post-install" \
     --script "trigger:$WORK/trigger" \
     --trigger "/etc/crontabs" \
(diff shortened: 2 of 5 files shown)


sameerparekh:
Package name corrected after the first CI round: `kmod-nft-log` doesn't exist in the OpenWRT 23.05 feed (`pkg_hash_check_unresolved`). The real deps are **`kmod-nfnetlink-log`** (NFLOG `log group` backend) + **`kmod-nf-log` / `kmod-nf-log6`** (netfilter log framework for the v4/v6 rules the inet table logs).

All PR-level checks now green, including **Image Builder (x86_64)** building both ipk + apk with the deps resolved — that was the thing the first round failed on. The remaining validation (Gate 2 ruleset load + Gate 3a block-page smoke) only runs on the self-hosted KVM runner post-merge.

PR #1175, opened 2026-05-30
fix(prod): kid can reach block page during scheduled downtime (#1174)

### LINKED ISSUE #1174: bug(prod): kid can't reach block page during scheduled downtime — wrong redirect host + wifihaven.net not carving around schedule block
## Symptoms (prod, observed 2026-05-30)

Two interconnected bugs that together make the kid unable to reach the block page / parent-request flow during a scheduled-downtime window:

1. **Block-page redirect lands on `192.168.10.43`** instead of the public SPA host. `192.168.10.43` is `api.lan` / `plex.lan` — the LAN-side dev API host (operator memory `api_server_deploy`). The redirect should target `wifihaven.net/blocked?…` (and post-cutover #1171: `app.wifihaven.net`).
2. **`wifihaven.net` is blocked during scheduled downtime.** Even with the schedule active, the SPA host should be reachable (it's how the kid sees the "blocked because of schedule" page + the request-an-extension flow). It isn't.

### For (1) — wrong redirect host

Per #959 / PR #1046 the kid-side block flow is:
- Router DNATs HTTP/80 to local block-page (uhttpd on 127.0.0.1:8081 inside the router).
- The block page (or the router) constructs a redirect to the SPA's `/blocked` route with `(mac, host)` params.

`192.168.10.43` showing up suggests the redirect URL is being built from the **LAN IP of the API host the router talks to**, not from the configured public SPA host. Likely sites:

- `openwrt/files/etc/uci-defaults/setup-uhttpd-block-page.sh` (or similar) — template variable using router-side known URL.
- API-side: `api/src/routes/BlockedRoutes.scala` (#959 endpoint) and any "public SPA base URL" config (`Config.scala`). Is the prod env actually serving its `wifihaven.net` SPA base, or is it falling back to the router's API host?
- The agent's enrollment / policy snapshot may not carry a "public SPA base" — if the router only knows its enrollment URL (`api.lan` for dev, `api.wifihaven.net` for prod), and constructs the block-page redirect from THAT, prod would (correctly) point at `api.wifihaven.net` — but operator sees `192.168.10.43`, so this prod router is enrolled against the LAN API (`api.lan` resolving to `192.168.10.43`), not `api.wifihaven.net`. That's a separate misconfig: prod router pointed at the LAN/dev API.

### For (2) — `wifihaven.net` blocked under schedule

`render.yaml:154-155` configures prod:
```
- key: WIFIHAVEN_UI_ALLOWED_HOSTS
  value: "wifihaven.net,www.wifihaven.net,api.wifihaven.net"
```

Per #944 and `PolicyService.scala:546` these flow into every snapshot's `extraAllowed`. Per memory `feedback_extraallowed_beats_blocked` (#421), extraAllowed beats every block path — including `@blocked_macs` (the MAC-level drop used for paused / schedule / time-limit / manual / unmanaged). So `wifihaven.net` SHOULD survive a schedule block.

Hypotheses:

- **Env not actually set on prod Render service** (drift between `render.yaml` and the running service env). Check `wifihaven-api-prod` env vars in Render — does `WIFIHAVEN_UI_ALLOWED_HOSTS` actually have the prod value?
- **Carve missing for schedule blocks specifically.** PolicyService applies uiAllowedHosts when constructing the snapshot's `extraAllowed`. If the carve in the agent (`render.lua`) doesn't apply `ip daddr != @ea_<m>_<a>` to the `@blocked_macs` drop, schedule blocks would ignore extraAllowed. Per `render.lua` header comment lines 31-37 it explicitly does — but only "blocked MACs *with* extraAllowed are pulled out of @blocked_macs". If the snapshot's extraAllowed is empty for this MAC, the carve doesn't apply. Need to confirm the prod snapshot actually has `wifihaven.net` in extraAllowed for the kid's MAC.
- **DNS attribution gap.** `wifihaven.net` resolves via DoH? No, default resolution should work. But if dnsmasq's `--ipset=` callback isn't populating `ea_<mac>_wifihaven_net` for this MAC, the nft set is empty and the drop matches even though the rule is correct. Confirm dnsmasq sees the answer (per memory `blocking_is_traffic_layer_not_dns`).

## Investigation steps

For (1):
- Get the **actual** redirect URL the kid's browser sees on prod (operator screenshot or curl).
- Confirm: is the redirect built **on the router** (uhttpd template) or **on the API** (`/api/blocked/...` route)? Grep + read one path.
- Check prod router's `wifihaven.api_url` UCI config: does it point at `api.wifihaven.net` or `api.lan`? If the latter, the prod router was enrolled against the LAN/dev API by mistake.

For (2):
- Curl the prod API for the kid's MAC's policy snapshot: `GET /api/router/policy` as the prod router. Inspect `extraAllowed` — does it contain `wifihaven.net`? (You can't ssh prod router; do this against staging or replicate the env locally.)
- If yes, the snapshot is correct and the router/dnsmasq/nft path is the gap.
- If no, the prod env var isn't set (Render dashboard drift from `render.yaml`).

## Acceptance

- Block-page redirect lands on `https://wifihaven.net/blocked?…` (or `https://app.wifihaven.net/blocked?…` after #1171 cutover), not a LAN IP.
- `wifihaven.net` (and `www.`, `api.`) reachable from the kid's browser even when their profile is in scheduled downtime.
- A feature test asserts the snapshot for a schedule-blocked MAC carries `wifihaven.net` in `extraAllowed` when `uiAllowedHosts` is configured.
- A router-side (render_spec.lua or e2e fake) test asserts the @blocked_macs drop carves around `ea_<mac>_wifihaven_net` (the path already exists per #421 — make sure the test pins it for the prod uiAllowedHosts case).

## Priority

P1 — kid-facing UX; during a schedule the kid sees connection errors instead of the block page, and can't request an extension.

## Related

- #959 / PR #1046 — kid-side block-page redirect
- #944 — deployment-configured UI hosts in snapshot.extraAllowed
- #421 — extraAllowed beats @blocked_macs
- Memory: `feedback_extraallowed_beats_blocked.md`, `blocking_is_traffic_layer_not_dns.md`
- #1171 — domain split (`wifihaven.net` → `app.wifihaven.net`) — keep this fix consistent with the post-cutover hostnames

sameerparekh:
## Findings

**Bug 1 — block-page redirect lands on a LAN IP.** Architectural: the
router agent's block-page handler builds the kid's redirect from
\`api_url\` (the URL it polls), not from a public SPA URL. In cloud
deployments the SPA and API live on different hostnames
(\`wifihaven.net\` vs \`api.wifihaven.net\`), so even with correct
enrollment the redirect would hit the API origin — which has
\`WIFIHAVEN_SERVE_SPA=false\` and 404s non-/api paths. The observed
\`192.168.10.43\` symptom additionally requires the prod router being
enrolled against \`api.lan\` instead of \`https://api.wifihaven.net\`.

**Bug 2 — \`wifihaven.net\` blocked during scheduled downtime.**
Confirmed via SSH to the test router. Not env drift; not the API
snapshot. Root cause: **dnsmasq's \`nftset=/<domain>/...\` honors only
the LAST matching directive per domain**. \`render.lua\` was emitting
one \`nftset=\` line per (MAC, host), so on a \`wifihaven.net\`
resolution only the last MAC's \`ea_<mac>_wifihaven_net\` set was
populated; the other MACs' sets stayed empty and the
\`@blocked_macs\` carve drop fired unconditionally.

Repro on \`openwrt.lan\`: widened the dev API's
\`WIFIHAVEN_UI_ALLOWED_HOSTS\` to include \`wifihaven.net\`, watched
\`/tmp/wifihaven-dnsmasq.log\` — \`nftset add\` fired for only one of
the four kid MACs. Replaced the four split directives with one merged
\`nftset=/wifihaven.net/4#...ea_MAC1...,4#...ea_MAC2...,...\` line and
all four sets populated on the next resolution.

## Summary

- **render.lua nftset merge (bug 2 fix).** One \`nftset=\` per host,
  comma-merging every MAC's v4 + v6 ea_ set. Per-MAC scoping stays in
  the nft drop rule (\`ether saddr <mac> ... != @ea_<mac>_<host>\`),
  so the shared populator is safe.
- **Block-page SPA URL via UCI (bug 1 fix).** \`install.sh\` accepts a
  new prompt for the public SPA URL (default derived by stripping a
  leading \`api.\` from the API host), persisted as
  \`wifihaven.@wifihaven[0].spa_url\`. Agent writes it to
  \`/var/run/wifihaven/spa_url\`; \`handler.lua\` reads it and
  \`block_page.render_html\` prefers it over \`api_url\` for the
  redirect dest. Empty → falls back to \`api_url\` so single-origin
  self-hosted installs stay unchanged.
- **No PolicySnapshot/API surface change.** The SPA URL is
  per-deployment, set once at enrollment, not snapshot-shaped.
- **Schedule snapshot pin.** New
  \`PolicyScheduleGlobalAllowSpec\` asserts a schedule-blocked
  profile's snapshot carries \`uiAllowedHosts\` in \`extraAllowed\` —
  passes on main, kept as a behaviour pin so the bug 2 symptom can't
  recur from the API layer.
- New \`render_spec\` test pins the single-line-per-host nftset
  invariant.

## Operator follow-ups

- [ ] Re-enroll the prod router against
  \`https://api.wifihaven.net\` (its current
  \`uci get wifihaven.@wifihaven[0].api_url\` points at
  \`api.lan\` / \`192.168.10.43\`). When re-running install.sh, the
  new SPA URL prompt should default to \`https://wifihaven.net\`.
- [ ] After agent re-installs from this PR, confirm
  \`uci get wifihaven.@wifihaven[0].spa_url\` is set and
  \`cat /var/run/wifihaven/spa_url\` matches.
- [ ] After agent re-installs, confirm
  \`grep nftset=/wifihaven.net/ /tmp/dnsmasq.d/wifihaven.conf\`
  returns exactly one line (not one-per-MAC).

## Test plan

- [x] API feature test: \`PolicyScheduleGlobalAllowSpec\` —
  schedule-blocked profile carries uiAllowedHosts in extraAllowed
  (passes on main — behaviour pin)
- [x] Lua: \`block_page_spec\` — \`prefers spa_url over api_url\`,
  \`falls back to api_url when spa_url is empty\` (#1174)
- [x] Lua: \`render_spec\` — \`merges every MAC's ea_/ea6_ set into one
  nftset= per host (#1174)\`
- [x] Live repro confirmed on \`openwrt.lan\`: pre-fix, only one MAC
  populated; post-fix, all MACs populate from a single resolution
- [x] \`mill api.test\` clean; \`mill shared.test.testOnly
  wifihaven.shared.contract.*\` clean
- [x] \`scalafmt --check\` clean
- [ ] Post-merge: re-enroll prod router; confirm block-page redirect
  hits \`https://wifihaven.net/blocked?…\` and \`wifihaven.net\`
  reachable during bedtime

Closes #1174.

### DIFF
--- a/api/src/Config.scala
+++ b/api/src/Config.scala
@@ -69,6 +69,14 @@ case class CorsConfig(
 // Precursor to the DB-backed global profile in #937.
 case class PolicyConfig(
     uiAllowedHosts: String = "",
+    // #1174: public SPA base URL (scheme+host, e.g. https://wifihaven.net).
+    // Included in every PolicySnapshot so the router agent's block-page
+    // handler can redirect blocked kids to the SPA host instead of the API
+    // host (the API origin 404s non-/api paths in cloud deployments). Empty
+    // → snapshot.spaBaseUrl is None and the agent falls back to its enrolled
+    // api_url, preserving pre-#1174 behaviour for single-origin self-hosted
+    // installs that don't split SPA and API.
+    spaBaseUrl: String = "",
 ) {
   val uiAllowedHostsParsed: List[Hostname] =
     uiAllowedHosts
@@ -88,6 +96,8 @@ case class PolicyConfig(
           )
       }
       .toList
+
+  val spaBaseUrlOpt: Option[String] = Option(spaBaseUrl).map(_.trim).filter(_.nonEmpty)
 }
 
 object AppConfig {
--- a/api/src/Config.scala
+++ b/api/src/Config.scala
@@ -69,14 +69,6 @@ case class CorsConfig(
 // Precursor to the DB-backed global profile in #937.
 case class PolicyConfig(
     uiAllowedHosts: String = "",
-    // #1174: public SPA base URL (scheme+host, e.g. https://wifihaven.net).
-    // Included in every PolicySnapshot so the router agent's block-page
-    // handler can redirect blocked kids to the SPA host instead of the API
-    // host (the API origin 404s non-/api paths in cloud deployments). Empty
-    // → snapshot.spaBaseUrl is None and the agent falls back to its enrolled
-    // api_url, preserving pre-#1174 behaviour for single-origin self-hosted
-    // installs that don't split SPA and API.
-    spaBaseUrl: String = "",
 ) {
   val uiAllowedHostsParsed: List[Hostname] =
     uiAllowedHosts
@@ -96,8 +88,6 @@ case class PolicyConfig(
           )
       }
       .toList
-
-  val spaBaseUrlOpt: Option[String] = Option(spaBaseUrl).map(_.trim).filter(_.nonEmpty)
 }
 
 object AppConfig {
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -178,15 +178,20 @@ object Main extends ZIOAppDefault {
       clock       <- ZIO.service[Clock]
       timeCache   <- ZIO.service[TimeStatusCache]
       xa          <- ZIO.service[Transactor[Task]]
-      routerAuth    = new RouterAuthLive(routerRepo)
-      dbHealthCheck = sql"SELECT 1".query[Int].unique.transact(xa).unit
-    } yield HealthRoutes.routes(dbHealthCheck) ++
-      VersionRoutes.routes(wifihaven.api.BuildInfo.fromEnv) ++
-      AuthRoutes.routes(auth, userRepo, upRepo) ++
-      ProfileRoutes.routes(auth, profileRepo, schedRepo, tlRepo, upRepo, userRepo) ++
-      HouseholdSettingsRoutes.routes(auth, hsRepo) ++
-      DeviceRoutes.routes(auth, deviceRepo, upRepo) ++
-      TimeRoutes.routes(
+      routerAuth          = new RouterAuthLive(routerRepo)
+      dbHealthCheck       = sql"SELECT 1".query[Int].unique.transact(xa).unit
+      // #1174: the route list is split into named intermediate vals so the
+      // Scala 3 typer doesn't have to type-check one giant `++` chain in a
+      // single recursive pass. After PolicySnapshot grew a field, the chain
+      // tipped over CI's default-stack JVM recursion limit; the split keeps
+      // each chunk small enough that the typer stays under budget.
+      coreRoutes          = HealthRoutes.routes(dbHealthCheck) ++
+        VersionRoutes.routes(wifihaven.api.BuildInfo.fromEnv) ++
+        AuthRoutes.routes(auth, userRepo, upRepo) ++
+        ProfileRoutes.routes(auth, profileRepo, schedRepo, tlRepo, upRepo, userRepo) ++
+        HouseholdSettingsRoutes.routes(auth, hsRepo) ++
+        DeviceRoutes.routes(auth, deviceRepo, upRepo)
+      timeRoutes          = TimeRoutes.routes(
         auth,
         deviceRepo,
         tlRepo,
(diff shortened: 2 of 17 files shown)


sameerparekh:
Closing — starting over later. Findings preserved in #1174 comments; key takeaway is bug 2's root cause: dnsmasq honors only the LAST nftset=/<domain>/... directive per domain, so render.lua's per-(MAC, host) emission silently drops all but the last MAC. Confirmed live on openwrt.lan.

Issue #1179, opened 2026-05-30
policy+ci: migrations must be backward-compatible with the previous deployed image (durable fix for #1176-class bugs)


sameerparekh:
## Goal

Establish "migrations must be backward-compatible" as a hard invariant: **every migration must leave the schema usable by the immediately-previous API image** for as long as that image might still be serving (a rollback window). This is the durable defense against the class of bug behind #1176 (prod stuck on `aa070c0` while the DB has V40 from #1147 applied → every connection_event ingest returns BatchUpdateException) — far stronger than a runtime fail-fast guard (#1177 part B) because it keeps prod **up** through a rollback instead of taking it down.

## The invariant

For migration `V<N>` paired with the API image that introduces it ("image-N"):

- **image-(N-1) running against DB-at-V<N> must still serve all routes correctly**, not just start.
- Reads of columns image-(N-1) knows must keep returning compatible values (same type from JDBC's perspective; same wire shape on the route layer).
- Writes performed by image-(N-1) must continue to succeed against the new schema (no NOT NULL added without a default; no CHECK that rejects what image-(N-1) sends; no PK / unique key shrink that breaks an existing upsert).
- The set of "currently-running images" includes the just-promoted one AND its immediate predecessor. Older images can be excluded once a deploy has settled and observability confirms no instances of the older image remain.

Concretely, **forbidden in a single migration**:
- Direct column-type changes that change the JDBC binding (e.g. `TEXT → JSONB` — the exact #1147 case). Must be done as two migrations: first add a new column / new wire shape, dual-write, switch reads, then drop the old column in a later migration after the deploy has settled.
- `DROP COLUMN` of anything image-(N-1) writes (unless image-(N-1) tolerates "column absent").
- `NOT NULL` adds without a `DEFAULT` covering image-(N-1)'s inserts.
- `CHECK` adds that image-(N-1) doesn't enforce.
- Renames (use add + dual-write + later drop).
- Constraint tightening of unique / primary keys that image-(N-1) doesn't already match.

**Permitted**:
- Add column (nullable or with default).
- Add table.
- Add index.
- Drop column / table that image-(N-1) doesn't touch.
- Loosening constraints.

### A. Policy + checklist

- Add a section to `AGENTS.md` (and surface in PR review docs) codifying the invariant, the forbidden list, the permitted list, and the **two-migration pattern** for type/shape changes.
[...]

Issue #1176, opened 2026-05-30
bug(prod): P0 — connection_event ingest 503s with BatchUpdateException; DB schema (post-V40) ahead of API image (aa070c0); roll prod forward


sameerparekh:
## Symptom

Prod API returns `HTTP 503 {"status":"error","db":"BatchUpdateException"}` on every `POST /api/router/events` carrying a `connection_attempt`. `dhcp_lease` events on the same endpoint succeed. The router agent has been backing off on conntrack drains for at least 18 hours; no connection events have landed in `/api/logs` since 2026-05-29 23:33Z.

Reproduced live (2026-05-30):
- `POST /api/router/events` with a `connection_attempt` (valid UUID, valid host shape) → 503 BatchUpdateException.
- Same POST with `dhcp_lease` only → 200 OK.

## Root cause: image ↔ DB schema mismatch

Prod is at API image **`aa070c0`** (`/api/version` confirms). That sha is **before PR #1147** (typed BlockReason serialization), which added migration **V40 converting `connection_events.reason` and `block_events.reason` from TEXT → JSONB**.

But `/api/logs` shows the latest rows (up to 2026-05-29 23:33Z) carry `reason` as a JSON-stringified kind-tagged object (e.g. `"{\"kind\": \"allow\"}"`). That format is **only** what the post-#1147 writer emits, and only against a JSONB column. So V40 has been applied to the prod DB.

Conclusion: a post-#1147 build was promoted briefly (yesterday ~23:30Z window), ran V40, wrote some events. The API image then got rolled back to `aa070c0` while the migration stayed (Flyway is forward-only). The `aa070c0` writer binds `reason` as TEXT against a column that's now JSONB → JDBC batch fails → 503.

Dhcp_lease ingest is unaffected because it writes a different table that V40 didn't touch.

## Fix

**Forward-roll prod to current `main`.** The 9 commits ahead include #1147 (which the prod DB already expects) plus today's wave (partitioning #805/#806, e2e fixes, etc.). Manual-approval gate (operator drives the master-api-ui Cloudflare/Render promotion).

Once promoted:
- Agent stops 503'ing on conntrack drains; queued events flush within minutes.
- Per the operator's MAC `a6:05:9a:63:83:af`: connection events for the iPad start landing in `/api/logs`; the device row already exists (from a manual `dhcp_lease` POST during diagnosis); new-device alert should fire on the next batch.

## Forensic question (separate, follow-up)

[...]

Issue #1177, opened 2026-05-30
ci+api: P0 — unbreak main (compile fail) + add startup guard against image↔DB migration skew (root cause of #1176)


sameerparekh:
### (1) Master API/UI CD currently red — compile failure on main

`mill __.compile` fails on `main` with **242 errors** rooted at:

- `api/src/Main.scala:185` — `AuthRoutes.routes(auth, userRepo, upRepo) ++`
- `api/src/routes/Routes.scala` lines 1607/1621/1628 — function signatures of `requireAuth/requireAdmin/requireWriter`, with the unresolved positions at `Request`, `IO`, `Response`.

The break landed via **PR #1165** (#1160 time-used rollup) which added new repo wiring in `Main.scala` (`TimeUsedRollupJob`, `TimeUsedRollupRepo`, etc.) but apparently incomplete imports / signatures, cascading type-resolution failures across `Routes.scala`. Subsequent merges (`#1161 → c8426c6`, `#1162 branding → 37610560`) were cancelled or piled on top and the compile break stuck. The branding PR itself touched zero Scala files.

How did this merge in red? **Merge queue should have caught it.** Either the merge queue isn't gating on the `Scala Build & Test` check, or that check was passing on the PR branch (which had less merge-queue overlay) and failing only after combining with another PR's changes that landed adjacent. Diagnose and patch the gate.

**Failing run for reference:** https://github.com/wifihaven/wifihaven/actions/runs/26671755924

### (2) Prod image ↔ DB schema can diverge silently (root cause of #1176)

Prod's `/api/version` returns `aa070c0`, **before #1147**. But the prod DB has migration **V40 (#1147)** applied — confirmed by `/api/logs` rows from 2026-05-29 23:33Z carrying kind-tagged `reason` JSON (only the post-#1147 writer produces that against a JSONB column).

Timeline:

- 2026-05-29 23:37:48Z — CD successfully deployed sha `36921c3e` (post-#1147) to prod. V40 ran on startup; new image wrote kind-tagged events.
- Sometime after — image got reverted to `aa070c0` outside CD (Render dashboard rollback, or auto-revert after a container crash). Flyway is forward-only, so V40 stayed.
- Since then — every connection_event ingest returns `503 BatchUpdateException` because the old API binds `reason` as TEXT and the column is JSONB.

[...]

PR #1185, opened 2026-05-30
db: V44 add reason_text dual-write + background backfill (back-compat reshape of V40) (#1176)

sameerparekh:
## Design

- **V44** adds nullable `reason_text TEXT` to `connection_events` and `block_events`. Schema-only, fast — no inline backfill, no type change.
- Repos dual-write `reason` (JSONB, the V40 shape) AND `reason_text` (wire string via `BlockReason.asWire`).
- `ReasonTextBackfill` runs as a forked daemon on startup and populates `reason_text` from existing `reason` JSONB in id-chunked batches. Idempotent; safe to re-run; throttled with a per-batch sleep so a slow Render PG doesn't see a burst on cold start.
- Reads unchanged — still off `reason` (JSONB). No SPA-visible change.

Numbering note: the task brief named this "V41". The actual next migration slot in `api/resources/db/migration/` is V44 (V41–V43 are already in main: traffic partitioning, connection_events partitioning, time_used_daily).

## Why

With this in main, a future Render auto-rollback that drops back to a *post-V44* image still finds `reason_text` populated, so a TEXT-binding read path on a later cut-over stays valid. Combined with #1179's policy (no inline backfills, no in-place type changes), the next shape change of `reason` is rollback-safe by construction.

This PR does NOT retroactively rescue `aa070c0` — that image binds `reason` as TEXT against what is now a JSONB column. Recovering prod still requires operator-approved forward-roll once this lands.

V45 (out of scope here) will cut reads to `reason_text` and later drop the legacy JSONB column once this has settled.

## TDD progression (visible in commit history)

1. **red** — V44 + failing dual-write spec (reason_text NULL because repos don't write it yet)
2. **green** — `BlockReason.asWire` + dual-write SQL
3. **red** — backfill spec referencing not-yet-existent `ReasonTextBackfill` object (won't compile)
4. **green** — `ReasonTextBackfill` + Main wiring

## Test plan

- [x] Feature test: `insertBatch` writes both `reason` (JSONB) and `reason_text` (wire string) for every BlockReason kind
- [x] Backfill job test: NULL `reason_text` becomes populated with the matching wire string for each `BlockReason` kind
- [x] Idempotency test: two consecutive backfill runs leave no NULLs
- [x] `mill api.test` clean (14/14 suites, 0 failures)
- [x] `mill mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll __.sources` clean
- [ ] Future V45 PR — out of scope — cuts reads + drops the legacy column after this has settled

Refs #1176. Conforms to #1179's back-compat invariant.

### DIFF
--- a/api/resources/db/migration/V44__reason_text_dual_write.sql
+++ b/api/resources/db/migration/V44__reason_text_dual_write.sql
@@ -1,0 +1,24 @@
+-- #1176 / #1179: back-compat reshape of V40 (TEXT → JSONB conversion).
+--
+-- V40 changed connection_events.reason and block_events.reason from TEXT to
+-- JSONB in a single step. An image binding `reason` as TEXT cannot then read
+-- or write the column, so a Render auto-rollback after V40 applied left prod
+-- on `aa070c0` returning 503 BatchUpdateException on every connection_attempt
+-- ingest. See #1176.
+--
+-- V44 introduces a parallel TEXT column `reason_text` that mirrors the
+-- pre-V40 wire-format string (e.g. "category:porn", "blocked", "schedule").
+-- Code in this PR dual-writes it from BlockReason.asWire(); a background
+-- backfill populates it for rows inserted between V40 and V44. Reads still
+-- come from the JSONB column — no SPA-visible change here.
+--
+-- This migration is schema-only: ADD COLUMN nullable, no inline UPDATE, no
+-- type change. Conforms to #1179's back-compat invariant: an image that
+-- predates V44 ignores the new column entirely; an image that knows about
+-- reason_text degrades gracefully on NULL until the backfill completes.
+
+ALTER TABLE connection_events
+  ADD COLUMN reason_text TEXT;
+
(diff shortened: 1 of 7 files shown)


sameerparekh:
CI **Scala Build & Test** is red, but the failure is the **pre-existing main breakage from #1177 Part A** (PR #1165 left main unable to compile — `AuthRoutes.routes` line in Main.scala hits 'Recursion limit exceeded' because of incomplete `TimeUsedRollupJob` wiring, cascading 172 transitive type errors across Routes.scala). My local `mill __.compile` succeeds because the mill server has cached pre-break state.

This PR will rebase cleanly once #1177 Part A lands and main is green. Nothing in this PR contributes to the compile failure.

sameerparekh:
Follow-up cleanup tracked in [#1187](https://github.com/wifihaven/wifihaven/issues/1187) — drop the dual-write + backfill + reason_text column once V44 has settled across two prod deploys.

Issue #1189, opened 2026-05-30
ci: P0 — main red after #1185 merged; 281 cascade errors in Routes.scala


sameerparekh:
## Symptom

Master API/UI CD on main is failing after PR #1185 (V44 reason_text dual-write) merged.

- Failing run: https://github.com/wifihaven/wifihaven/actions/runs/26693420311
- Broken main HEAD: `96720e4e` (merge commit of #1185)
- Last green main: `cdc5faae` (merge of #1184 / #1177 — Routes.scala typed-chunks fix)

`CI / Scala Build & Test` fails compile with **281 errors, all in `api/src/routes/Routes.scala`**, all of the form:

```
Not found: type AuthService     // line 471
Not found: type DeviceRepo      // 472
Not found: type TimeLimitRepo   // 473
Not found: type SiteTimeLimitRepo
Not found: type TrafficReportRepo
Not found: type TimeExtensionRepo
Not found: type ProfileRepo
Not found: type UserProfileRepo
Not found: type HouseholdSettingsRepo
Not found: type Clock           // 481
… (281 total)
```

These types are all defined and unchanged on `origin/main`:
- `AuthService` — `api/src/auth/AuthService.scala`
- `*Repo` traits — `api/src/db/Repos.scala`
- `Clock` — `wifihaven.shared.Clock`

And Routes.scala imports are intact:
```scala
import wifihaven.api.auth.*
import wifihaven.api.db.*
import wifihaven.shared.*
```

That all 281 errors land in **one file** (Routes.scala) and start at `TimeRoutes.routes(` strongly suggests an upstream compile failure cascading — Scala 3 silently fails the dependent package and the consumer sees every symbol as missing. The likely culprit is in #1185's edits to `api/src/db/Repos.scala` (insertBatch dual-write changes), `shared/src/Models.scala` (`BlockReason.asWire` added), or `api/src/ReasonTextBackfill.scala`.

## Why CI passed on the #1185 PR before merge

#1185 was branched off pre-#1184 main and never rebased — its pre-merge CI ran against the pre-#1184 Main.scala/Routes.scala state. The merge resolution onto post-#1184 main produced a file that compiles in isolation in `git`'s view but fails Scala 3's resolution.

## Scope

Diagnose and fix the actual root error (not the 281 cascaded errors). Steps:

[...]

PR #1192, opened 2026-05-31
test: template-DB pattern eliminates per-test Flyway re-run (#1188)

### LINKED ISSUE #1188: ci: parallelize Scala tests — Scala Build & Test is the CI long pole
## Problem

The `Scala Build & Test` job in `.github/workflows/ci.yml` is now the long pole on every PR. As the API test suite has grown (feature tests under `api/test/src/feature/*` use embedded Postgres, which dominates wall time), CI feedback latency has gotten painful — especially during the current recovery push where every recovery PR has to round-trip through CI individually.

## Goal

Run the Scala test suite in parallel to cut wall-clock time on the `Scala Build & Test` CI job.

## Scope

Two axes to consider; pick whichever lands the bigger win first.

1. **Within-JVM parallelism (Mill / utest config).** Investigate enabling parallel test execution within a single Mill module. Each `*Spec` that spins up its own embedded Postgres via `TestDatabase.layer` is independent and could run concurrently. Watch for shared state: `Clock.TestClock` is per-spec, but anything globally registered (logger config, system properties) is a hazard.
2. **Shard across CI workers.** Split the test suite into N shards via Mill task selection (e.g. `mill api.test.testOnly 'wifihaven.api.feature.A*'`) and fan out across `matrix:` jobs in GH Actions. Reuse the existing self-hosted KVM runner pool (see #1163 / PR #1166 — multi-instance KVM runner work in flight, lands new capacity).

## Constraints

- **Embedded Postgres is the wall-time hot spot.** Each spec currently boots its own PG via `zonkyio/embedded-postgres`. Either share a single PG across specs in the same JVM (with per-spec schema reset via `TestDatabase.cleanAndMigrate`), or accept the boot cost and parallelize harder.
- Don't mock the DB — the test policy in `AGENTS.md` requires real PG. Sharing is fine; mocking is not.
- Flyway migrations must still run.
- Don't break the existing `mill __.test` local UX.

## Acceptance

- `Scala Build & Test` wall-clock on a no-op PR drops by ≥40% versus current baseline.
- All existing specs pass.
- Local `mill api.test` still works without extra flags.
- Document the chosen parallelism model (per-JVM thread count, sharding scheme) in a short note under `docs/`.

## Related

- #1163 / PR #1166 — KVM runner scale-out (provides the workers for shard fan-out).
- #1029 — CI paths-filter (masks regressions; orthogonal but related to CI hygiene).

sameerparekh:
Closes #1188.

## Summary

- `TestDatabase` now boots one `EmbeddedPostgres` per JVM, builds a
  pre-migrated `wh_template` database once, and hands every spec a
  fresh `wh_test_<n>` cloned from the template via Postgres
  `CREATE DATABASE … TEMPLATE` (single-digit ms, no Flyway).
- `cleanAndMigrate` (called per-test by `cleanDb`) becomes
  `DROP DATABASE + CREATE DATABASE … TEMPLATE wh_template` on the
  spec's own DB — eliminates the ~400–500 ms × 44-migrations cost
  that used to run on every `cleanDb` call.
- Mechanical update to 42 specs: `cleanDb` helper collapses to
  `private val cleanDb = TestDatabase.cleanAndMigrate`.
- Two specs (`RollupRepoSpec`, `RetentionSweepJobSpec`) hand-fixed —
  they were acquiring pg-advisory \"holder\" locks on
  `pg.getPostgresDatabase` (the postgres admin DB) while the repo
  under test ran against the per-spec DB. The lock was invisible to
  the repo and the test race-passed under the old shared-DB layout.
  Both now draw the holder connection from `TestDb.ds` for the same
  DB the repo runs against.
- CI workflow reverted from the ubuntu-latest-8-cores experiment
  back to `ubuntu-latest` (larger-runner label isn't available on
  free/public orgs). `MILL_JOBS=4` pinned to the runner's vCPU count.
- New `docs/testing-parallelism.md` describing the template-DB
  pattern, the per-spec-DB isolation invariant, and the
  advisory-lock gotcha (\"if you write a new contention test, draw
  the holder connection from `TestDb.ds`, not from
  `EmbeddedPostgres.getPostgresDatabase`\").

## Strategy

- **In-JVM-multi-spec parallel** was the original goal but needs a
  ZIO Test framework change to actually execute multiple `Spec`
  objects concurrently inside one JVM. Punted; the template-DB
  pattern would make it safe whenever someone wants to take that on.
- **Wider CI runner** (`ubuntu-latest-8-cores`) isn't available on
  free/public orgs — first attempt sat queued forever.
- **Hot-path fix**: the actual dominant cost was Flyway-per-test,
  not lack of parallelism. The template-DB pattern removes that
  cost across the board. Mill's existing worker-JVM parallelism
  (4 workers on 4-vCPU `ubuntu-latest`) still does the
  cross-spec concurrency work.

## Before / after

| Measurement                              | Before | After |
|------------------------------------------|--------|-------|
| `mill api.test` local, 10-core wall      | 67s    | 54s   |
| `mill api.test` local, `--jobs 1` wall   | 207s   | _pending_ |
| `Scala Build & Test` job, CI (4-core)    | 305s\* | _pending — this PR's run_ |

\* baseline = run [26693842738](https://github.com/wifihaven/wifihaven/actions/runs/26693842738) on
`a29912d1` (`ubuntu-latest`, 4 vCPU). Inside that 305s: 43s
setup/install/scalafmt, 92s compile, 170s tests.

The acceptance bar from #1188 is ≥40% reduction (≤183s). The
template-DB change should land well inside that on CI because
Flyway-migrate was a much larger fraction of work on the slower
runner than on my laptop.

## Test plan

- [ ] CI green on this PR
- [ ] `Scala Build & Test` wall-clock on this PR ≥ 40% lower than
      305s baseline
- [ ] Local `mill api.test` still works with no flags (verified —
      54s)
- [ ] `docs/testing-parallelism.md` matches what CI does

## Follow-ups (not in this PR)

- True in-JVM-multi-spec parallel (run multiple `Spec` objects
  concurrently inside one worker JVM). Now safe under the
  per-spec-DB model; needs a framework hook.
- Larger CI runner (paid plan, self-hosted KVM scale-out from
  #1166) would compound the gains.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,24 +137,27 @@ jobs:
               - '.github/actionlint.yaml'
 
   # ── Scala: format, compile, test ─────────────────────────────────────────
-  # Runs on the 8-vCPU GitHub-hosted larger runner (#1188). Mill's test
-  # parallelism is on by default and spawns up to `MILL_JOBS` worker JVMs
-  # that pull specs from a shared queue — each worker has its own
-  # EmbeddedPostgres singleton, so a wider runner directly buys more
-  # parallel JVMs without any code changes. See docs/testing-parallelism.md.
+  # The dominant test-time cost used to be the per-test Flyway re-run
+  # inside cleanAndMigrate (~400-500 ms × 44 migrations × hundreds of
+  # tests). #1188 moved that work to a one-shot template-DB init at JVM
+  # startup; per-test reset is now a sub-ms `DROP DATABASE + CREATE
+  # DATABASE … TEMPLATE wh_template`. Mill's testParallelism=true
+  # (default) is what runs worker JVMs in parallel across the runner's
+  # vCPUs — each worker has its own EmbeddedPostgres + per-spec DBs
+  # cloned from the template. See docs/testing-parallelism.md.
   scala:
     name: Scala Build & Test
     needs: changes
     if: needs.changes.outputs.scala == 'true'
-    runs-on: ubuntu-latest-8-cores
+    runs-on: ubuntu-latest
 
     env:
       WIFIHAVEN_JWT_SECRET: test-jwt-secret-for-ci-only
-      # Pass `--jobs $MILL_JOBS` to every mill invocation in this job.
-      # Match worker-JVM count to runner vCPUs — kept explicit so any
-      # change to `runs-on` stays coupled with the worker count it implies
-      # rather than silently inheriting `availableProcessors()`.
-      MILL_JOBS: "8"
+      # Pinned to the standard `ubuntu-latest` runner's 4 vCPUs (matches
+      # mill's default `availableProcessors()` but kept explicit so any
+      # change to `runs-on` stays coupled with the worker count it
+      # implies). Bump in lockstep if the runner gets wider.
+      MILL_JOBS: "4"
 
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,14 +137,24 @@ jobs:
               - '.github/actionlint.yaml'
 
   # ── Scala: format, compile, test ─────────────────────────────────────────
+  # Runs on the 8-vCPU GitHub-hosted larger runner (#1188). Mill's test
+  # parallelism is on by default and spawns up to `MILL_JOBS` worker JVMs
+  # that pull specs from a shared queue — each worker has its own
+  # EmbeddedPostgres singleton, so a wider runner directly buys more
+  # parallel JVMs without any code changes. See docs/testing-parallelism.md.
   scala:
     name: Scala Build & Test
     needs: changes
     if: needs.changes.outputs.scala == 'true'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-latest-8-cores
 
     env:
       WIFIHAVEN_JWT_SECRET: test-jwt-secret-for-ci-only
+      # Pass `--jobs $MILL_JOBS` to every mill invocation in this job.
+      # Match worker-JVM count to runner vCPUs — kept explicit so any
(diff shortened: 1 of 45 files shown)


sameerparekh:
## After-numbers

Run [26700525640](https://github.com/wifihaven/wifihaven/actions/runs/26700525640):

| Phase                          | Baseline (305s job) | This PR (225s job) | Δ      |
|--------------------------------|---------------------|--------------------|--------|
| setup + checkout + cache + scalafmt | ~43s          | ~25s               | -42%   |
| compile (`mill __.compile`)    | 92s                 | 91s                | ~0%    |
| tests (`mill shared.test` + `mill api.test`) | 170s  | **108s**           | **-36%** |
| **Job wall-clock**             | **305s**            | **225s**           | **-26%** |

Test phase is down 36% — that's the change this PR was about, and it
landed close to the local 10-core measurement (67 → 54s = 19%; CI gain
is bigger because Flyway-migrate was a larger fraction of work on
4-vCPU than on a fast laptop). Acceptance bar from #1188 was ≥40% on
the **job** wall-clock; we're at 26% because compile (92s) and
setup (~25s) now dominate the residual.

### Options to close the gap

- **Restore `out/` caching** (currently disabled by #1189 because of
  Zinc cross-SHA poisoning). A warm cache should drop compile from
  91s to near-instant on PRs that don't touch Scala sources, getting
  the total well under 183s.
- **In-JVM-multi-spec parallel** would compound the test gain but
  needs a ZIO Test framework change — the per-spec-DB groundwork is
  now in place, so it's no longer blocked on test infrastructure.
- **Larger runner** (paid plan or self-hosted KVM scale-out from
  #1166) is independent of this PR.

Want me to take a swing at the `out/` cache rehabilitation as a
follow-up, or merge this as a partial-credit win and re-open #1188
for the rest?

Issue #1194, opened 2026-05-31
ops: design + ship deploy-safety guardrails so a bad image / red main can't take prod down


sameerparekh:
## What's wrong

Today's incident chain (2026-05-30 → 31) exposed multiple gaps in how images get from a green build to prod. Symptoms:

1. **Main went red twice** within hours (#1165 cascaded compile error → #1184 fix → #1185 merged on stale main → broke again, #1189 fix in flight).
2. **Render deployed a bad image** at 2026-05-31 00:32Z while main was broken — `deploy_hook` trigger, not manual. Deploy failed `update_failed` 15min later.
3. **The previously-running container then died** at 00:53Z (Hikari pool exhaustion → instance restart → stuck in startup → 502 to users).
4. **Prod was 502 for an unknown window** until operator manually rolled back to `dep-d8cd1q0g4nts738k88kg` (image `aa070c0`).
5. **Render is configured for autoDeploy** on `:latest` push, so the next bad image push repeats this.

Each of these is "working as designed" individually. Together they form a pipeline that can take prod down from a single bad merge with no human-in-the-loop step. The 2026-05-29 V40 incident (#1176/#1179) was the previous instance of the same class of failure.

## Goal

Design and implement guardrails so a broken main, a failed build, or a bad-image push **cannot** silently take prod down. Specifically the system should ensure that:

- An image tagged `:latest` only exists for a commit that passed full CI (build + test + e2e).
- Render only deploys an image that has been independently verified (smoke-tested) post-build.
- A failed Render deploy does not destabilize the previously-running container.
- The operator is paged (or at least notified prominently) when prod is unhealthy — not discovered by clicking around the UI.

## Investigation scope

This is an investigation-and-design issue first, implementation second. Before proposing a fix, the spawned session should establish:

1. **The actual deploy-trigger chain.** What pushes `:latest`? What hits Render's deploy webhook? Is it the CD workflow on `push: main`? Which job? Is it gated on test success?
2. **Why did the 00:32Z deploy happen at all?** Main was red at that time (post-#1185 sha 96720e4e). Was the image push from #1184 (then-most-recent green build at cdc5faae)? Was the deploy_hook fired from a stale CD run? Could a queued CD have re-fired hours later?
[...]

PR #1198, opened 2026-05-31
fix(api): decouple port-bind from Postgres warm-up so slow PG doesn't fail Render port-scan (#1197)

### LINKED ISSUE #1197: ops: P0 — diagnose + fix why prod deploy at 00:32Z failed (slow PG hypothesis, no port-bind)
## Problem

Prod deploy `dep-d8do3b6rnols739i6bp0` (image sha `4661ea`, started 2026-05-31T00:32:44Z) failed `update_failed` 15 minutes later at 00:47:46Z. Render's port-scan never saw the container bind :8080.

## Hypothesis (to confirm or refute)

The container never bound the port because **Postgres connection during startup was slow** — Hikari/Doobie pool init, Flyway migration check, or some other DB-blocking step held up the JVM before `serverStarting` could run. The rollback deploy hit the same path but happened to come up under the 15-min timeout. Operator confirmed: "it took a while to load because postgres connection was slow."

## Why this matters

Prod is currently on an older rollback image (`df84684`) that predates `/api/version`. Every future deploy is vulnerable to the same timeout. We can't safely ship #1183, #1189, or any forward-roll until we understand and fix this.

## Goal

1. Confirm the root cause via actual deploy log.
2. Ship the smallest fix that lets a normal-content deploy come up under Render's port-scan window.

## Investigation steps

1. Pull `dep-d8do3b6rnols739i6bp0` stdout via Render API:
   ```bash
   RENDER_API_KEY=$(awk '/^rnd_/{print; exit}' /Users/sameer/.claude/projects/-Users-sameer-workspace-wifihaven/memory/render_api_token.md)
   curl -s -H "Authorization: Bearer $RENDER_API_KEY" -G "https://api.render.com/v1/logs" \
     --data-urlencode "ownerId=tea-d8543pmk1jcs73aqoja0" \
     --data-urlencode "resource=srv-d856kjp9rddc73a3o1t0" \
     --data-urlencode "limit=200" \
     --data-urlencode "startTime=2026-05-31T00:32:44Z" \
     --data-urlencode "endTime=2026-05-31T00:48:00Z"
   ```
   Find the last log line emitted by the JVM. That's where it hung.
2. Cross-reference with `api/src/Main.scala` to identify the blocking step (Flyway? Hikari init? template seed? household_settings ensure? something else?).
3. Confirm or refute the slow-Postgres theory. If not slow PG, find the actual cause.

## Fix candidates (pick after diagnosis)

- **Bind port before warming the connection pool.** Start the HTTP server on a stub handler that answers `/api/health` immediately, then warm Hikari/Doobie in the background. Once warm, swap in the real routes. Standard "decouple readiness from liveness" pattern.
- **Hikari `initializationFailTimeout = -1`.** Lets the pool init asynchronously; first DB call waits, but startup doesn't block.
- **Lazy Flyway.** Move `flyway.migrate()` after the port is bound. Same readiness-vs-liveness pattern.
- **Diagnose why prod Postgres is slow.** Cold start? Network? Connection limit? Render DB plan? If it's a config issue, fix that instead — but only if the fix is simple and known-safe.

Prefer the smallest change. Decoupling port-bind from DB warm-up is the most-deployable-most-quickly path.

## Constraints

- Don't ship deploy-safety guardrails. That's #1194; explicitly out of scope here.
- Don't ship a CD-pipeline change.
- Touch `api/src/Main.scala` and pool config only as needed.
- TDD per AGENTS.md: write a failing spec (e.g. assert `/api/health` responds before Flyway has run) before fixing.
- Coordinate with #1177's startup migration-version guard (already merged in #1184) — don't undo it. The guard should still run; just not on the port-bind critical path.
- `mill scalafmtCheckAll` before push.

## Acceptance

- Root cause identified and posted in the issue.
- Fix lands such that a deploy of the current main HEAD (post-#1189) comes up and binds :8080 well under 15min on prod.
- Validation method documented (how do we test this without breaking prod again?).
- New deploy succeeds. Operator confirms by rolling forward.

## Evidence

- Failed deploy: `dep-d8do3b6rnols739i6bp0` (image `4661ea`, 2026-05-31T00:32:44Z → update_failed 00:47:46Z).
- Rollback live deploy: `dep-d8doelojs32c73fom9u0` (image `df84684`, 2026-05-31T00:56:55Z).
- Prior prod state (steady, mid-degraded): image `aa070c0` since 2026-05-29.
- Hikari log from old container 2026-05-31T00:52:12Z: pool 5/5 exhausted (separate symptom, possibly related — the running container also hit DB issues).
- Render service: `srv-d856kjp9rddc73a3o1t0` (wifihaven-api-prod).

sameerparekh:
Closes #1197.

## Root cause

Prod deploy `dep-d8do3b6rnols739i6bp0` (image `4661ea`, 2026-05-31T00:32:44Z) timed out at Render's 15-min port-scan window. The container never bound :8080. The JVM logged exactly one line and then went silent:

```
00:32:54.201Z INFO Main - WifiHaven API starting on 0.0.0.0:8080
   ... <silence for 15 minutes> ...
00:47:46.334Z ==> Timed Out
```

The very next step in `Main.run` was `Database.runMigrations(cfg.db)`. Flyway never even emitted its `Schema "public" has a version …` WARN, so its own JDBC connection hung against a cold Render Postgres.

The rollback deploy hit the same code path but happened to come up at **+1m 43s** because PG was already warm. Within that 1m 43s, the slowest single step was **bundled blocklist seeding at +96s**. Every step in the warm-up chain was on the port-bind critical path.

## Fix

`Main.run` now:

1. Loads YAML templates / bundled blocklists (file-system only, no DB).
2. Constructs routes (Doobie `Transactor` + Hikari pool are lazy — no I/O).
3. Calls `Bootstrap.boot(warmup, serve)`, which:
   - Forks the **entire** DB warm-up chain (`Database.runMigrations`, `household_settings ensureDefault`, `AppTemplates.seed`, `BundledBlocklists.seed`, `RollupJobs.{hourly,daily}Loop`, `TimeUsedRollupJob.loop`, `RetentionSweepJob.start`, `ReasonTextBackfill.run`) as a daemon, catching all causes (logged, doesn't crash the process).
   - Runs `Server.serve` on the main fiber, binding :8080 immediately.

Also sets `HikariDataSource.setInitializationFailTimeout(-1)` explicitly: even the daemon's first DB call won't block waiting for an unreachable PG.

## What it preserves

- Flyway still runs at every startup (just off the critical path).
- The #1177 route-composition typing fix in `allRoutes` is untouched.
- Warm-up step ordering inside the daemon is unchanged.
- No `render.yaml` changes — fully declarative still.
- No new client / transport / framework.

## TDD

1. [test(api): red — Bootstrap.boot must bind server before warm-up](https://github.com/wifihaven/wifihaven/commit/34253983) — pins the API: serve runs before warmup completes; warmup still runs; warmup failure doesn't crash serve. Fails to compile against current main (no `Bootstrap`).
2. [fix(api): green — bind port before DB warm-up](https://github.com/wifihaven/wifihaven/commit/c23110c9) — adds `Bootstrap`, refactors `Main.run`, sets Hikari init timeout.

## Validation

- `mill api.compile` — green.
- `mill api.test` — all 52 features + the new `BootstrapSpec` pass (the 3 new tests verify serve-before-warmup, warmup-still-runs, and warmup-failure-doesn't-crash-serve).
- `scalafmt --check` — clean.
- Manual prod-shape validation not done here (can't test on prod safely). Suggested chain for the operator before forward-rolling:
  1. **Local**: bring up the API against a slow Postgres (e.g. inject a 60s delay in `TestDatabase` or `pg_ctl start` + iptables delay). Confirm port binds within seconds; `/api/health` returns 200 immediately; DB-touching endpoints block/queue or return 5xx until pool warms.
  2. **Staging** (`srv-d8549fgjo89c73buvkf0`, `wifihaven-api-staging`) — deploy via existing CD. Watch Render logs for `WifiHaven API starting` immediately followed by Render's port-scan succeeding (under 30s), then the warm-up log lines arriving asynchronously.
  3. **Prod** — only after staging shape is confirmed.

Do **not** auto-merge; operator queues manually.

## Out of scope

- Deploy-safety guardrails / pre-promotion checks — that's #1194.
- Hikari pool-size tuning — separate issue noted in the diagnosis comment (01:01Z pool-exhaustion errors on the *running* container, unrelated to the port-scan failure).

### DIFF
--- a/api/src/Bootstrap.scala
+++ b/api/src/Bootstrap.scala
@@ -1,0 +1,26 @@
+package wifihaven.api
+
+import zio.*
+
+/**
+ * Decouples the HTTP port-bind from the DB warm-up chain so that a slow Postgres at startup does
+ * not cause Render's port-scan to time out (#1197).
+ *
+ * `warmup` runs in a forked daemon: migrations, seeds, and background-job kick-off. It is allowed
+ * to fail without taking the server down — the next request to a DB-touching route surfaces the
+ * underlying error in the normal way, and operator-side restart/observability picks it up.
+ *
+ * `serve` runs on the main fiber so the process exits naturally if `Server.serve` returns.
+ */
+object Bootstrap {
+  def boot[R](
+      warmup: ZIO[R, Throwable, Unit],
+      serve: ZIO[R, Throwable, Unit],
+  ): ZIO[R, Throwable, Unit] =
+    for {
+      _ <- warmup
+        .catchAllCause(c => ZIO.logErrorCause("startup warm-up failed; server stays up", c))
+        .forkDaemon
+      _ <- serve
+    } yield ()
+}
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -25,27 +25,77 @@ object Main extends ZIOAppDefault {
 
   def run =
     (for {
-      cfg    <- ZIO.service[AppConfig]
-      _      <- ZIO.logInfo(s"WifiHaven API starting on ${cfg.http.host}:${cfg.http.port}")
-      _      <- ZIO
+      cfg       <- ZIO.service[AppConfig]
+      _         <- ZIO.logInfo(s"WifiHaven API starting on ${cfg.http.host}:${cfg.http.port}")
+      _         <- ZIO
         .logWarning(
           "WIFIHAVEN_DEBUG=1 set — /api/debug/* endpoints are MOUNTED (loopback only). " +
             "Disable in production.",
         )
         .when(cfg.debugEnabled)
-      _      <- Database.runMigrations(cfg.db)
-      _      <- ZIO.logInfo("Database migrations complete")
+      // YAML loads are file-system only (no DB) and feed route construction, so they
+      // run on the port-bind critical path. The DB-touching warm-up below forks.
+      templates <- AppTemplates.loadAll()
+      bundled   <- BundledBlocklists.loadAll()
+      templatesById = templates.map(t => t.slug -> t).toMap
+      bundledById   = bundled.map(b => b.id -> b).toMap
+      routes <- allRoutes(templatesById, bundledById)
+      withCors = Cors.wrap(routes, cfg.cors)
+      _ <- ZIO
+        .logInfo(s"CORS enabled for origins: ${cfg.cors.origins.mkString(", ")}")
+        .when(cfg.cors.origins.nonEmpty)
+      // #1017: zio-http 3.0.1's RequestStreaming.Disabled default cap is 100 KiB;
+      // /api/router/usage bodies routinely exceed that as mac_ip_tracking fills.
+      // Bump to 4 MiB — well above any realistic single-bucket payload and below
+      // Render's edge 413 threshold.
+      serverConfig = Server.Config.default
+        .port(cfg.http.port)
+        .disableRequestStreaming(4 * 1024 * 1024)
+      tz           = java.time.ZoneId.systemDefault()
+      warmup       = warmupChain(cfg, tz, templates, bundled)
+      serve        = Server
+        .serve(withCors)
+        .provide(ZLayer.succeed(serverConfig) >>> Server.live)
+      // #1197: bind :8080 immediately and run all DB warm-up in a forked daemon.
+      // Previously the entire warm-up chain (Flyway + seeds + remote blocklist
+      // fetches) ran on the critical path, so a cold Render Postgres at deploy
+      // time exceeded the 15-min port-scan window and the deploy failed.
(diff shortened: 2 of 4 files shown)


sameerparekh:
Closing per operator: the prod unblock was a one-time TRUNCATE of the two partition-target tables (traffic_reports, connection_events), which made V41/V42 run instantly — the port-bind decoupling here isn't needed. Durable guidance about slow-on-prod-data migrations will land as a docs change to AGENTS.md instead. See #1197 for the full resolution.

Issue #772, opened 2026-05-21
openwrt: staged rollout via canary channel (openwrt-canary → openwrt-latest)


sameerparekh:
Parent strategy: #652 (three-gate plan)
Depends on: #771 (agent version reporting — DONE, PR #1135)
Related: #695, #773, #774, #1170 (remove openwrt-latest bootstrap fallback), #1172/#1134/#499 (versioned semver release machinery)

> **RE-SCOPED 2026-05-31.** The original premise (rolling `openwrt-latest` daily cron) is stale. The update mechanism changed under #1172/#1134/#499. This issue is re-scoped against the current versioned-release model and now needs a **design pass first** — see open questions.

## Current state (post-#1172)

- `wifihaven-update` (cron) polls `https://api.github.com/repos/wifihaven/wifihaven/releases/latest`, which returns the highest **non-prerelease** `v<X.Y.Z>` release and re-installs when `tag_name` differs from the baked-in VERSION file (#771).
- `openwrt-latest` is published `prerelease:true`, so it's excluded from `/releases/latest`. It's now **vestigial** — only a 404 bootstrap fallback, slated for removal in #1170.
- Versioned semver releases are cut with coordinated API+OpenWRT tags (#1134/#499). Latest is `v0.2.27` (2026-05-31).
- Net: **every router is effectively on one "stable" channel today** (`/releases/latest`), and a newly-cut semver release reaches all routers on the next cron tick — still no soak window, still no blast-radius limit. The problem #772 exists to solve is unchanged; only the mechanism has moved.

## The opportunity

The canary/stable split maps onto the prerelease-vs-semver distinction that already exists:
- **stable channel** = `/releases/latest` (highest non-prerelease `v<X.Y.Z>`) — unchanged behavior.
- **canary channel** = a pre-release tag the canary cohort polls *before* the semver release is cut/promoted.

The core mechanic is **separating in time**: publish the image to a canary-visible pre-release on approval → soak → promote by cutting the non-prerelease `v<X.Y.Z>` that stable routers pull. If both are published simultaneously, canary is meaningless.

## Open questions for the design pass (resolve before coding)

1. **Who cuts the semver release and when?** Master Router CD has `publish-openwrt`; Master API/UI CD has a "Cut v<X.Y.Z> release tag (#499)" job. With coordinated semver (#1134), how are these sequenced today, and where does a soak gap fit without breaking the API↔router version coordination?
[...]

Issue #622, opened 2026-05-18
Make this work multi-tenant.


sameerparekh:
We want to create multiple tenants so other people can use the cloud installation. so we'll need a tenant table that each router, device, and profile links to. I don't think any other tables need to link to them, because all other tables link to a router, device, or profile, at a minimum. file issues to handle this, we need a tenant enrollment ui, a billing page, integration with stripe, etc.

sameerparekh:
## Phase 5 execution plan — product surfaces (beta, enrollment, billing, marketing)

The isolation substrate (#2104–#2109, all merged 2026-07-06→08) established the precondition: the invariant *household-A can never read/write household-B* is CI-pinned by `MultiTenantIsolationSpec`. Phase 5 now productizes it, per the locked pricing decisions in [`docs/design/pricing-analysis.md`](docs/design/pricing-analysis.md) and the operator's beta-gating decision (2026-07-08): **beta access is request-based with manual operator approval — no open self-serve signup.**

### Locked inputs
- Household = the unit (one subscription per home). Single price: **$10/mo / $96/yr**; founding households **$6/mo forever** (`FOUNDING`, 40% off, `duration=forever`); **self-hosted free forever**.
- **4-month free beta, cohort-wide flip date printed at signup; no card during beta** (Stripe Customer only). At flip / on dunning-lapse: **read-only grace — never brick the network** (router snapshot always serves).
- Entitlements (router cap = 1 public) resolve **per household**, never global constants. Minimal Stripe surface: 1 Product + 2 Prices + 1 coupon + Checkout + Portal + webhook — no schedules/metered/per-seat.

### Sub-issues (foundation-first)
| # | Unit | Depends on |
|---|---|---|
| #2130 | **fix**: `POST /api/users` writes into household 1 regardless of caller (cross-tenant write found in Phase-5 survey) | — |
| #2131 | **P5-1**: V66 schema-only — `beta_requests` + `household_billing` + `households.router_cap` + drop V65 `DEFAULT 1`s | #2130 |
| #2132 | **P5-2**: beta request intake + operator approval + household provisioning + invite accept (API) | #2131 |
| #2133 | **P5-3**: SPA — `/beta` form, `/welcome` invite accept, operator queue | #2132 |
| #2134 | **P5-4**: per-household `router_cap` enforcement (first entitlement) | #2131 |
[...]

PR #1215, opened 2026-05-31
feat(openwrt): updateChannel support in wifihaven-update (#772 phase a)

sameerparekh:
## Summary

Phase (a) of #772 (canary channel). Adds per-router channel selection to the
OpenWRT auto-updater — **consumer side only**, fully backward-compatible.

- New `update_channel` UCI option in `/etc/config/wifihaven`, default `stable`.
- `wifihaven-update` reads it via `uci -q get wifihaven.wifihaven.update_channel`:
  - **stable** (default): unchanged — `/releases/latest` (highest non-prerelease
    `v<X.Y.Z>`) with the existing `openwrt-latest` 404 fallback.
  - **canary**: polls the `openwrt-canary` pre-release tag. Because that tag's
    `tag_name` is the non-semver rolling string, the comparable version is
    derived from the **selected asset** (`0.3.0-1` → `0.3.0`) instead of the
    tag, so it detects real changes without reinstalling on every cron tick.
    A 404 on the canary tag **holds** on the current version (no silent drop
    back to stable / openwrt-latest).

Default `stable` means **zero behavior change for every existing router** until
an operator opts a router into canary.

### Not in this PR (later #772 phases — see the design comment on #772)
- **(b)** Producer side: publishing to `openwrt-canary` + splitting the
  canary-publish from the stable `v<X.Y.Z>` promotion to create a soak gap.
  Until (b) lands, a canary router 404s and holds — safe no-op.
- **(c)** LuCI settings channel dropdown.

## Test plan
- [x] `sh openwrt/test/update_spec.sh` — 46/46 pass, incl. 5 new canary cases
      (install, no-op, 404-hold, asset-derived version, default-stable) + 3
      static checks.
- [ ] After phase (b) ships, flip the test router (192.168.10.42) to
      `update_channel 'canary'` and confirm it pulls canary ahead of stable.

### DIFF
--- a/openwrt/files/etc/config/wifihaven
+++ b/openwrt/files/etc/config/wifihaven
@@ -8,6 +8,11 @@ config wifihaven 'wifihaven'
 	# Router UUID returned during enrollment
 	option router_id ''
 
+	# Auto-update channel (#772): 'stable' (default) tracks the highest
+	# non-prerelease v<X.Y.Z> release; 'canary' tracks the openwrt-canary
+	# pre-release, which receives builds during the pre-promotion soak window.
+	option update_channel 'stable'
+
 	# LAN subnet prefix used to identify outbound (egress) flows
 	option lan_prefix '192.168.1.'
 
--- a/openwrt/files/usr/sbin/wifihaven-update
+++ b/openwrt/files/usr/sbin/wifihaven-update
@@ -54,14 +54,42 @@ else
   exit 0
 fi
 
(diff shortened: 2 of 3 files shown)


sameerparekh:
On hold pending multi-tenant (#622) — see https://github.com/wifihaven/wifihaven/issues/772. A canary soak is pointless while the dev router (the only non-stable candidate) is barely used; resume once tenants exist so a real home router can ride canary while others stay stable. Branch `design/canary-channel` and this passing work are preserved — reopen to resume.

Issue #1265, opened 2026-05-31
db+api: connection-events rollup tier (hourly/daily) — deferred in #847, never filed


sameerparekh:
Sub-issue of #844 ("…Connection Events + Traffic Usage **with rollups**"). Tracks the connection-events rollup tier that was deferred in #847 and never filed.

## Why this exists / how it got missed

#847 (Connection Events page) shipped the aggregated view as **on-demand SQL binning over raw `connection_events`** — `ConnectionEventRepo.querySeries` runs `date_bin(make_interval(secs => bucket.seconds), ce.ts, …)` and `GROUP BY` against the base table every request. #847's body explicitly punted the storage tier:

> "Server picks the right storage source by bucket width… for short buckets compute from `query_logs` directly. **If a rollup table doesn't yet exist for this domain, file a sub-issue but don't block on it; Raw + on-demand aggregation works for now.**"

That sub-issue was never created. So the traffic side got a full rollup track (#725 → #809 tables / #813 window-width routing / #814 retention gating) while connection-events still scans raw. #844 asked implementers to "flag any gap when found" — flagging it now.

Surfaced while investigating #1262: traffic-usage's bucket grain coarsens for wide windows *because* it routes to `traffic_hourly` / `traffic_daily`; connection-events stays fine *because* it has no rollups and always bins raw. The two were supposed to share one range→bucket-width + storage-tier architecture; the events half is the missing piece.

## Scope

Mirror the traffic rollup track for connection events:

1. **Flyway migration — rollup tables** (per the schema-only migration-PR rule in AGENTS.md; backfill/runtime wiring lands in the follow-up code PR):
   - `connection_events_hourly` — pre-aggregated counts per `(router_id, mac, resolved_host, bucket_start, allowed)` (or split `count_succeeded` / `count_blocked` columns so a row is per `(…, bucket_start)` without `allowed` in the key — pick one and document). Carry `sample_count`, `rolled_at`.
   - `connection_events_daily` — same shape, `date DATE` instead of `bucket_start`.
   - Reuse the existing `rollup_runs` observability table (add `connection_events_*` job names); the admin rollup-status check should cover them.
2. **Rollup job** — periodic fiber mirroring `RollupJobs` (hourly reroll trailing N hours, daily reroll prev+current day), strictly UPSERT-keyed/idempotent.
[...]

Issue #1262, opened 2026-05-31
Traffic-usage aggregations always bucket per-day regardless of range; connection-events 1w range over-buckets to 1w


sameerparekh:
## Symptom

- **Traffic usage:** every aggregation over traffic-usage data returns **daily** buckets no matter the selected range/granularity. Shorter ranges that should show per-minute (or per-hour) buckets are collapsing to one bucket per day.
- **Connection events:** bucketing is **correct for all ranges except `1w`**, which returns a single `1w`-wide bucket instead of the expected finer (e.g. daily) buckets.

So two distinct bucket-width bugs:
1. Traffic-usage: bucket width is stuck at 1 day across the board (per-minute granularity never reached).
2. Connection-events `1w`: bucket width is mis-sized to the whole range.

## Expected

Bucket width should follow the selected time range/granularity consistently across both datasets — e.g. a short window → minute buckets, a day → hour buckets, a week → day buckets — and the two endpoints should use the **same** range→bucket-width mapping so they don't diverge.

## Likely area (to confirm during investigation)

- The API aggregation endpoints for traffic usage vs connection events and their `date_trunc` / bucket-interval selection. Traffic usage may be truncating to `day` unconditionally (note: the per-minute source is `traffic_reports` with `period_start`/`period_end`; the daily rollups feed `time_used_*` — make sure the aggregation reads the right grain for the requested range rather than always hitting a daily rollup).
- The range→bucket mapping table/function: the connection-events `1w` case suggests an off-by-one or missing branch in the bucket-width lookup that traffic-usage may share (or duplicate incorrectly).

## Repro

In the dashboard (or via the aggregation API directly), request a sub-day range for traffic usage → observe daily buckets. Request `1w` for connection events → observe a single 1w bucket; request other ranges → correct.

## Acceptance

- Traffic-usage aggregation returns buckets at the granularity implied by the requested range (down to per-minute where the source supports it).
- Connection-events `1w` returns finer buckets (matching the granularity used for adjacent ranges).
- Both endpoints share one range→bucket-width mapping (or are tested against the same expectations) so they can't drift again. Add feature/unit tests covering the range→bucket mapping per AGENTS.md TDD.

Issue #1254, opened 2026-05-31
Rollup attribution lateral pegs DB CPU — missing index on connection_events(mac, dest_ip, ts)


sameerparekh:
## Symptom

Prod DB CPU pinned at 100% **continuously** (not spiky), independent of tier: pinned at 0.1 on `basic-256mb`, then immediately pinned at 0.5 after upsizing to `basic-1gb`. The consumer expands to fill whatever CPU is available → a runaway query, not under-provisioning.

## Root cause

The byte-rollup hostname-attribution query (the `traffic_hourly` / `traffic_daily` reroll from #809) runs a **correlated `LEFT JOIN LATERAL`** with no supporting index:

```sql
SELECT tr.mac, tr.date, tr.period_start,
       CASE WHEN tr.host_type IN ('ipv4','ipv6') AND ce.resolved_host_value IS NOT NULL THEN 'fqdn' ELSE tr.host_type END,
       COALESCE(CASE WHEN tr.host_type IN ('ipv4','ipv6') THEN ce.resolved_host_value END, tr.host_value),
       tr.active_seconds, tr.bytes_in, tr.bytes_out, tr.period_start, tr.period_end
FROM traffic_reports tr
LEFT JOIN LATERAL (
  SELECT resolved_host_value
  FROM connection_events
  WHERE mac = tr.mac
    AND dest_ip = tr.host_value
    AND resolved_host_value IS NOT NULL
    AND ts >= tr.date::timestamptz
    AND ts <  (tr.date + INTERVAL '1 day')::timestamptz
  ORDER BY ...
) ce ON true
```

The only indexes on `connection_events` are `idx_conn_events_ts` (on `ts` alone) and a `(event_id, ts)` unique key. **Nothing on `(mac, dest_ip)`.** So the lateral can only use the `ts` range — a full *day* — and then filters `mac`/`dest_ip` by scan. For **every** `traffic_reports` row it effectively scans every `connection_events` row in that day.

Observed live (2026-05-31, prod): individual rollup queries running **9+ minutes**, and multiple rerolls overlapping (two identical queries active concurrently, pids 1877/1878). Because a tick takes far longer than its interval, rerolls pile up and there's always a heavy aggregation in flight → CPU never idles.

Table sizes are tiny, which proves it's a bad plan, not data volume:

| table | rows | size |
|---|---|---|
| traffic_reports_2026_22 | 69,793 | 22 MB |
[...]

Issue #811, opened 2026-05-21
db: usage retention sweep — traffic_reports & connection_events 30d, traffic_hourly 3mo, traffic_daily 6mo


sameerparekh:
Sub-issue of #725. Implements the retention sweep from [docs/design/usage-retention.md](../blob/main/docs/design/usage-retention.md).

## Scope

A daily ZIO scheduled fiber (\`RetentionSweepJob\`) running at 03:00 router-local that sweeps four tables:

1. \`traffic_reports\` — older than \`usage.rawRetentionDays\` (default **30**).
2. \`connection_events\` — older than \`usage.eventsRetentionDays\` (default **30**).
3. \`traffic_hourly\` — older than \`usage.hourlyRetentionDays\` (default **90**).
4. \`traffic_daily\` — older than \`usage.dailyRetentionDays\` (default **180**).

For each table:

- **If [#793](https://github.com/wifihaven/wifihaven/issues/793) ships partitioning for that table:** \`DROP TABLE …_YYYYMMDD\` per expired partition. Instant, releases disk, no VACUUM churn.
- **Otherwise:** batched \`DELETE WHERE <time-col> < \$cutoff\` in 100 k-row chunks; explicit \`ANALYZE\` after.

Writes a \`rollup_runs\` row per table with \`rows_deleted\` and elapsed time.

## Explicitly NOT swept

- **Sessions / session-stitching outputs** — there is no session table; sessions are derived from \`traffic_reports\` and inherit its 30-day horizon implicitly. Even when a persistent sessions table is proposed, no retention applies until #817 (sessions data-quality audit) is resolved.
- \`time_usage\` — quota-input table, not analytics. Stays unbounded.

## Configuration

\`application.conf\`:
- \`usage.rawRetentionDays\` (30)
- \`usage.eventsRetentionDays\` (30)
- \`usage.hourlyRetentionDays\` (90)
- \`usage.dailyRetentionDays\` (180)

## Dependencies

- Depends on #809 (needs \`rollup_runs\` + scheduled-fiber pattern).
- Coordinate with #793 — partition-drop branch only valid for whichever tables #793 actually partitions. Land DELETE path first; add DROP PARTITION as a follow-up commit once #793 lands.
- Sessions explicitly carved out — see #817.

## Acceptance

- Fiber visible in logs.
- Idempotent (re-running same day is a no-op).
[...]

PR #1271, opened 2026-06-01
api: connection_events reroll fibers + retention sweep (#1265, code PR 2/2)

sameerparekh:
## Summary

Code half of #1265 (connection-events rollup tier), stacked on the
schema-only migration in **#1267**. Adopts the V47
`connection_events_hourly` / `_daily` tables: re-aggregation fibers,
the repo UPSERTs, `rollup_runs` observability wiring, and the retention
sweep extension. **Read-path routing in `querySeries` is deferred** to a
follow-up PR — it needs the #1262 shared range→bucket-width mapping,
which is not yet on `main`, and the issue is emphatic about not
introducing a third divergent threshold.

> Depends on #1267 — review/merge that first. This PR is based on
> `claude/fix-1265-ce-rollups`; once #1267 merges, retarget this PR's
> base to `main`.

## What's here

- **`ConnectionEventRepo.rerollConnEventsHourly/Daily`** — UPSERT-keyed
  `INSERT … SELECT` over `connection_events`:
  - NULL-mac events fold into a `''` bucket via `COALESCE(ce.mac, '')`
    so no counts are dropped (mac is a PK column, NOT NULL in the rollup).
  - allowed/blocked split into `count_succeeded` / `count_blocked`
    (`COUNT(*) FILTER (WHERE ce.allowed)` etc.).
  - hostname is the post-resolved host
    `COALESCE(resolved_host_value, host_value)`.
  - Concurrent rerolls gated behind `pg_try_advisory_xact_lock`
    (`RollupLockKeys.ConnEventsHourly/Daily`) — multi-instance skip (#818).
- **`RollupJobs.connEvents{Hourly,Daily}Loop`** — same cadence/lookback
  as the traffic rollups, reusing the shared `runOnce` + `rollup_runs`
  observability under job names `connection_events_hourly` /
  `connection_events_daily`. `Main` forks both fibers.
- **`RetentionSweepJob`** — extends the #811 sweep with
  `connection_events_hourly` (90d) and `connection_events_daily` (180d),
  gated on `to_regclass(...)` table presence.

## Index validation (#query-explain-before-merge)

Validated against a full V1→V47 apply on Docker Postgres 16:

- **Reroll source scan** filters on raw `ce.ts >= $since` (NOT a
  function of `ts`), so it uses the existing `idx_conn_events_ts`
  (`ts DESC`) / `idx_conn_events_mac_ts` (`mac, ts DESC`) and weekly
  partition pruning — a bounded range scan per tick, never a
  full-table scan. This deliberately avoids the #1254 / V38-class
  defeat where wrapping `ts` in `(ts AT TIME ZONE 'UTC')::DATE`
  forced a seq scan. **No new source-table index needed.**
- **Rollup read-path indexes** (added in #1267, mirroring V38):
  `idx_ce_hourly_bucket_start` (`bucket_start DESC`),
  `idx_ce_hourly_mac_bucket` (`mac, bucket_start DESC`),
  `idx_ce_daily_date` (`date DESC`), `idx_ce_daily_mac_date`
  (`mac, date DESC`). The schema decision (split count columns,
  `allowed` not in the PK) is documented in V47's header.

## Test plan

- [x] `ConnectionEventRollupSpec` — V47 table/PK/index existence,
      hourly & daily reroll with split counts, UPSERT idempotency +
      late-event recompute, NULL-mac `''` bucket folding, advisory-lock
      multi-instance skip, `rollup_runs` row via `RollupJobs.runOnce`.
- [x] `RetentionSweepJobSpec` — `connection_events_hourly` (90d) and
      `connection_events_daily` (180d) horizons.
- [x] `DbFailureSpec` / `RollupShutdownSpec` / `RollupCadenceSpec` still
      green after the trait + wiring changes.
- [x] TDD red→green visible in commit history (failing specs committed
      before the implementation).
- [x] `scalafmt` clean.

### DIFF
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -145,6 +145,15 @@ object Main extends ZIOAppDefault {
         clockForJobs <- ZIO.service[Clock]
         _            <- RollupJobs.hourlyLoop(rollupRepo, appRepoForSeed, clockForJobs).forkScoped
         _ <- RollupJobs.dailyLoop(rollupRepo, appRepoForSeed, clockForJobs, tz).forkScoped
+        // #1265: connection-event rollup tier — same cadence/lookback as the
+        // traffic rollups, re-rolling the trailing windows into
+        // connection_events_hourly / _daily. forkScoped so they're interrupted
+        // before the Hikari pool closes on shutdown.
+        connRepoForJobs <- ZIO.service[ConnectionEventRepo]
+        _ <- RollupJobs.connEventsHourlyLoop(connRepoForJobs, rollupRepo, clockForJobs).forkScoped
+        _ <- RollupJobs
+          .connEventsDailyLoop(connRepoForJobs, rollupRepo, clockForJobs, tz)
+          .forkScoped
         // #1160: per-(profile, today) `used_seconds` cache. Tick aggregates today's presence into
         // a watermarked row; the read path adds a live tail of buckets after the watermark, so
         // /api/time/status/summary serves a rollup + small live aggregation instead of a full
@@ -172,7 +181,9 @@ object Main extends ZIOAppDefault {
             clockForJobs,
           )
           .forkScoped
-        _               <- ZIO.logInfo("rollup fibers forked (hourly + daily + time_used_daily)")
+        _               <- ZIO.logInfo(
+          "rollup fibers forked (hourly + daily + time_used_daily + conn_events_hourly + conn_events_daily)",
+        )
         // #1243: poll the HikariCP MXBean into the Prometheus pool gauges. forkDaemon so it lives
         // for the process and never blocks startup.
         dbPool          <- ZIO.service[Database.DbPool]
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -483,6 +483,33 @@ trait ConnectionEventRepo {
       bucketSeconds: Int,
       groupBy: Set[String],
   ): Task[List[ConnectionEventAggRow]]
+
+  /**
+   * #1265: re-aggregate the trailing N hours of `connection_events` into
+   * `connection_events_hourly`. Counterpart of [[RollupRepo.rerollHourly]] but for the event-count
+   * tier. `since` is the earliest `ts` to re-roll, inclusive; callers truncate it to an hour
+   * boundary so every bucket in the window is recomputed in full. Idempotent via UPSERT keyed
+   * `(router_id, mac, hostname, bucket_start)` — re-rolling an already-rolled bucket replaces it
+   * with the freshly summed counts.
+   *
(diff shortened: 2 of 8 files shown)


sameerparekh:
## Query-plan validation (#query-explain-before-merge)

Ran read-only `EXPLAIN` (plan only — **not** `ANALYZE`, nothing executed) of both reroll `SELECT`s against **wifihaven-pg-prod** at real prod statistics, `statement_timeout='5s'`. The rollup *target* tables don't exist on prod yet (V47 ships in #1267), so this validates the only part that touches a growth table: the `FROM connection_events` aggregation source.

**Access pattern.** Single-pass `GROUP BY` over a trailing time window of `connection_events` (weekly RANGE-partitioned on `ts`, V42/#806; capped to ~30 days by retention #811). This is *not* the #1254 shape — there is no per-row `LEFT JOIN LATERAL`; it's one bounded scan + hash aggregate.

### Hourly reroll (`WHERE ts >= NOW() - INTERVAL '2 hours'`)
```
HashAggregate  (cost=594.36..613.99 rows=785)
  ->  Append  (cost=0.15..456.97 rows=7851)
        Subplans Removed: 2                         <- partition pruning at plan time
        ->  Index Scan using connection_events_2026_23_ts_idx ...
              Index Cond: (ts >= (now() - '02:00:00'::interval))
        ->  Bitmap Index Scan on connection_events_2026_24_ts_idx ...
        ->  Bitmap Index Scan on connection_events_2026_25_ts_idx ...
        ->  Bitmap Index Scan on connection_events_2026_26_ts_idx ...
```
Fully index-backed on the existing per-partition `ts` index; irrelevant partitions pruned. No seq scan.

### Daily reroll (`WHERE ts >= (CURRENT_DATE - 2)::timestamptz`)
```
HashAggregate  (cost=2493.57..2581.89 rows=3533)
  ->  Append  (cost=0.15..1875.31 rows=35329)
        Subplans Removed: 1                         <- partition pruning at plan time
        ->  Seq Scan on connection_events_2026_22 (cost=0.00..996.95 ...)
        ->  Seq Scan on connection_events_2026_23 (cost=0.00..656.50 ...)
        ->  Index Scan using connection_events_2026_24_ts_idx ...
        ->  Index Scan using connection_events_2026_25_ts_idx ...
        ->  Index Scan using connection_events_2026_26_ts_idx ...
```
The two boundary partitions fall back to seq scan, but each is a **bounded weekly partition** (cost ~997 / ~657 — trivial) and the whole table is retention-capped to ~30 days. Total plan cost ~2.6k — bounded by partition size, not by result size or a lateral join. Runs off the startup critical path (background fiber), so even the daily-2d window can't peg the API.

**Conclusion:** no new index needed in V47 on the source table. Partition pruning + the existing `ts` indexes keep the reroll bounded; the daily seq-scan fallback is on small, retention-capped weekly partitions and is not a regression risk.

Issue #1148, opened 2026-05-29
dashboard: holistic /dashboard redesign (umbrella; supersedes piecemeal ux-audit tweaks)


sameerparekh:
## Goal

Holistically redesign `/dashboard` (`web/src/pages/DashboardPage.tsx`) as the household's at-a-glance landing surface, rather than continuing to ship the eight piecemeal ux-audit tweaks one at a time. Those tweaks (listed below) are **inputs** to this redesign, not an independent plan — several of them conflict or overlap (e.g. dropping a table vs. merging two tables vs. reordering sections), and resolving them coherently needs a single design pass.

## Current state (2026-05-28)

`DashboardPage.tsx` renders, top to bottom:

1. `<h1>Dashboard`
2. **Now** section (`NowSection`, polls every 10s) — per-profile cards → active-device rows (last-5-min activity).
3. **Top Blocked (24h)** — host + count list.
4. **Per Device (24h)** — device → total/blocked query counts.
5. **Recent Queries** — full query table.

(There is also a KPI/stat strip per #821; confirm its exact location during design.)

Terminology is stale — tables say "queries" though the data is connection events (#299), and per the architecture model DNS resolution is never the enforcement plane, so "queries" is doubly misleading.

## Design questions to resolve

- **What is the dashboard *for*?** Glanceable household status ("who's online, who's blocked, anything wrong?") vs. analytics. Analytics arguably belongs on /logs and /profiles now that those surfaces are rich. The landing page should answer "is everything OK right now?" in one screen.
- **NOW section** is the most valuable real-time element. How much vertical space should it own, and how do idle profiles / long device lists collapse? (#819 card-height cap + top-N expander, #820 collapse idle profiles, #825 single freshness timestamp.)
- **KPI strip** placement and contents (#821 — above NOW?). What are the right top-line numbers?
- **Blocking activity** — merge Top Blocked + Per Device into one "Blocking activity (24h)" panel (#822) vs. keep separate.
- **Recent Queries table** — drop entirely and link to /logs (#823)? Rename to "Connection events" (#299)? It duplicates /logs.
- **Active throughput** — show live bytes in/out (#747)?
- **Mobile/iOS parity** — the iOS v1 surface inventory (#982) names the dashboard as a primary surface; the redesign should produce an IA that the iOS app can mirror.

## Proposed approach

[...]

PR #1340, opened 2026-06-03
dashboard: "Most recently blocked" panel at the top (#1338)

### LINKED ISSUE #1338: dashboard: 'Most recently blocked' panel at top — fast diagnosis of why a site isn't working
## Ask

Add a **"Most recently blocked"** view at the top of the dashboard — a live, reverse-chronological list of the most recent blocked connection events (host, device/profile, reason, timestamp). Today the dashboard has **"Top blocked"** (aggregate counts) but not a recency view. The use case is diagnostic: when a site/app isn't working, the operator wants to immediately see *what just got dropped* (e.g. the `connectivitycheck.gstatic.com` / `ocsp` / akamai dependencies behind the Math Academy bug, or an over-broad blocklist hit) without digging into the Connection Events page.

Distinct from "Top blocked": this is **newest-first, un-aggregated, recent window** (last ~N events or last few minutes), so a just-now block shows immediately.

## Two deliverables

1. **Implement it ad-hoc now.** Add the panel at the top of the current dashboard (`web/src/pages/DashboardPage.tsx`), backed by a recent-blocked query. Reuse the existing connection-events infrastructure — there is already a blocked-events query path feeding "Top blocked" (`api/src/db/Repos.scala` + `DashboardNowRoutes.scala` / connection-events series). Add a "most recent blocked events" read (newest-first, blocked-only, small limit) if one doesn't already exist, following AGENTS.md "Adding a new API route". Auto-refresh via the existing React Query polling so it stays live.
2. **Fold it into the dashboard redesign #1148.** The holistic /dashboard redesign is in flight — update that issue's scope/spec so "Most recently blocked" is a first-class element of the redesigned layout, not lost when the ad-hoc panel is reworked. Add a comment to #1148 describing the feature + intended placement.

## Details
- Show per row: timestamp (relative, e.g. "12s ago"), host, device name + profile, block reason (the same reason text used elsewhere — daily limit / schedule / category / host-block). 
- Blocked-only filter; newest first; sensible limit (e.g. 20) with a link to the full Connection Events page for more.
- Respect the env/household scoping the rest of the dashboard uses.
- Reminder: blocked here means connection-layer drop, not DNS (a row appearing is a real drop). 

## TDD
- API: feature test that the recent-blocked read returns blocked events newest-first, limited, blocked-only (real embedded Postgres per repo philosophy).
- web: a DashboardPage test that the panel renders rows from a mocked response and updates on refetch (`nvm use 22` before vitest).

Refs: #1148 (dashboard redesign umbrella — must be updated), existing "Top blocked" in `DashboardPage.tsx`, #847 (Connection Events).

sameerparekh:
## Summary

Adds a **"Most Recently Blocked"** panel at the top of the dashboard — a live, newest-first, un-aggregated feed of the latest connection-layer drops. It's the recency complement to the existing aggregate "Top Blocked": when a site/app suddenly stops working, the operator can instantly see *what just got dropped* (the `connectivitycheck.gstatic.com` / `ocsp` / akamai dependency, an over-broad blocklist hit) without digging into the Connection Events page.

Closes #1338.

### What it shows
- Each row: relative timestamp ("12s ago"), host, `device · profile`, and block reason (same `blockReasonText` rendering used elsewhere).
- Blocked-only, newest-first, capped at 20, with a **View all →** link to `/usage/events`.
- Auto-refreshes on the dashboard's existing React Query 10s poll, so a just-now block surfaces immediately.
- Placed above the "Now" section (top of the dashboard).

### Backend — reuse, not a new query
The read the panel needs already exists: `GET /api/logs?blocked=true&limit=N` (via `ConnectionEventRepo.query`) already orders `ts DESC`, honours the limit, applies the same JWT/household scoping every other dashboard read uses, and is backed by vetted indexes. A new `useRecentBlocked` hook wraps it. This deliberately avoids adding a new, unvetted query on the unbounded-growth `connection_events` table (per the query-performance guardrail in AGENTS.md).

A row here is a real **traffic-layer** drop, not a DNS event — DNS always resolves; the block happens at the connection layer.

## Test plan
- **web** (`DashboardPage.test.tsx`, `nvm use 22`): the panel renders host / `device · profile` / reason rows from a mocked `blocked=true` response, links to the Connection Events page, shows an empty state when nothing is blocked, updates on the 10s poll, and sits above the "Now" section. `npx vitest run` → 325 passed; `type-check` + `lint` clean.
- **api** (`LogApiSpec`): new feature test (real embedded Postgres) asserts `GET /api/logs?blocked=true&limit=N` returns blocked-only events, newest-first, within the limit (allowed events excluded even when newer). Full `api.test` suite green.
- Red→green visible in history: tests committed before the implementation.

## Redesign follow-up
Commented on the dashboard redesign umbrella #1148 so "Most recently blocked" survives as a first-class element when this ad-hoc panel is reworked.

### DIFF
--- a/api/test/src/feature/LogApiSpec.scala
+++ b/api/test/src/feature/LogApiSpec.scala
@@ -180,6 +180,52 @@ object LogApiSpec
         assertTrue(logs.head.host.value == "blocked.com") &&
         assertTrue(logs.head.blocked)
     },
+    test(
+      "GET /api/logs?blocked=true&limit=N returns blocked-only events, newest-first, within the limit (#1338 recently-blocked panel)",
+    ) {
+      // Contract the dashboard "Most recently blocked" panel relies on: the
+      // blocked filter + ts-DESC ordering + limit together yield the latest
+      // connection-layer drops only (allowed events excluded), capped.
+      val base = Instant.now().minusSeconds(600)
+      for {
+        _        <- cleanDb
+        routerId <- seedRouter()
+        connRepo <- ZIO.service[ConnectionEventRepo]
+        upRepo   <- ZIO.service[UserProfileRepo]
+        auth     <- makeAuth
+        token    <- auth.login("admin", "changeme").map(_.token.value)
+        ce = (mac: String, host: String, allowed: Boolean, ts: Instant) =>
+          ConnectionEventInsert(
+            routerId,
+            Some(MacAddress.unsafe(mac)),
+            HostId.Fqdn(Hostname.unsafe(host)),
+            None,
+            allowed,
+            BlockReason.fromWire(if (allowed) "allowed" else "category:adult"),
+            ts,
+          )
+        _ <- connRepo.insertBatch(
+          List(
+            ce("aa:bb:cc:dd:ee:01", "allowed-newest.com", true, base.plusSeconds(500)),
+            ce("aa:bb:cc:dd:ee:02", "blocked-newest.com", false, base.plusSeconds(400)),
+            ce("aa:bb:cc:dd:ee:03", "blocked-middle.com", false, base.plusSeconds(300)),
+            ce("aa:bb:cc:dd:ee:04", "blocked-oldest.com", false, base.plusSeconds(100)),
+          ),
+        )
+        routes = LogRoutes.routes(auth, connRepo, upRepo)
+        resp <- getJson(routes, "/api/logs?blocked=true&limit=2", token)
+        body <- resp.body.asString
+        page <- ZIO.fromEither(body.fromJson[QueryLogPage])
+        logs = page.rows
+      } yield assertTrue(resp.status == Status.Ok) &&
+        // limit honoured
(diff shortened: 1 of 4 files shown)


sameerparekh:
Follow-up (dde1bfd): scoped the panel to the **trailing 15 minutes** rather than the newest-20-over-24h it originally used.

Surfaced in review against prod: on a quiet network a real-but-stale block (`beacons.gcp.gvt2.com`, blocked ~11h earlier on a schedule) sat at the top and read as if it had just happened. "Most Recently Blocked" should mean *just now*; the 24h aggregate is already "Top Blocked". `useRecentBlocked` now fetches a 1h window (`/api/logs` takes integer hours) and trims to 15 min client-side, with a test that an ~11h-old block is filtered out.

Separately filed #1341 for the missing host/domain search + app filter on the Connection Events / Traffic Usage pages — the gap that made this hard to diagnose from the UI.

Issue #421, opened 2026-05-15
extraAllowed: enforce at the router as per-(MAC, host) exception to blocklist/extraBlocked drops


sameerparekh:
## Background

The wire contract carries `BlockRules.extraAllowed: List[Hostname]` per device/profile, but the router agent currently does nothing with it. `render.lua` only enumerates `extraBlocked` and `blocklistIds` when emitting nft drops and dnsmasq nftset populators.

This means: today, if an admin adds `example.com` to a profile's `extraAllowed` to carve it out of a category blocklist that contains `example.com`, the carve-out has no effect — the blocklist's `bl_<id>` set still gets populated with `example.com`'s resolved IPs (via the dnsmasq nftset directive for that blocklist's hostname file), and the per-(MAC, blocklistId) drop still fires.

The API also relies on this primitive internally (per #353 design discussion): per-site time-limit exhausted → API appends to `extraBlocked`; profile-wide limit hit with per-site carve-outs → API populates `extraAllowed`. So extraAllowed enforcement at the router is load-bearing for the time-limit feature too.

## Design

Mirror the eb_/bl_ pattern with per-MAC scoping (extraAllowed is per-(MAC, host), not global — the same host can be allowed for MAC A and blocked for MAC B):

1. Tag each MAC in dnsmasq with `set:eatag_<sanmac>`.
2. For each host in that MAC's effective extraAllowed list, emit a tag-scoped nftset populator:
   `nftset=tag:eatag_<sanmac>,/<host>/4#inet#familydns#ea_<sanmac>_<sanhost>,6#inet#familydns#ea6_<sanmac>_<sanhost>`
   (per-(MAC, host) set; only the tagged MAC's resolutions land in it.)
3. In `familydns_block`, for each (MAC, host) where `host ∈ extraAllowed`: the eb_/bl_ drops that mention that MAC must NOT fire when daddr is in `@ea_<sanmac>_<sanhost>`.

In nft, "drop X unless also in Y" requires either a single rule with the negation (`ether saddr <mac> ip daddr @eb_<host> ip daddr != @ea_<mac>_<host> drop`) or a chain that jumps to accept first. Single-rule form is simpler — semantics are clear, no ordering dependence.

## Conflict resolution

**extraAllowed beats extraBlocked beats blocklists.** Rationale: explicit admin allow is the strongest signal. Explicit admin block beats a category list the admin opted into. The drop predicate becomes:

```
drop(m, d) :=
   m ∈ blocked_macs
[...]

PR #1345, opened 2026-06-03
fix(openwrt): attribute directly-queried CNAME targets to the branded host (#1344)

### LINKED ISSUE #1344: router: IP→hostname attribution records CNAME target instead of queried brand — mislabels logs and mis-classifies allowed CDN-fronted apps
## Symptom

The router's IP→hostname attribution (dns_log cache, consumed for `connection_event` /
traffic logging **and** for conntrack.lua's ea_/eb_/bl_ flow classification) records the
**final CNAME-target name** for some CDN-fronted flows instead of the branded name the user
is actually using.

Concretely, for Khan Academy / Math Academy on a kid device:

- `cdn.kastatic.org` resolves `cdn.kastatic.org → CNAME fastly.kastatic.org → CNAME prod.khan.map.fastly.net → A 199.232.65.42`
- the connection-event / traffic logs show **`prod.khan.map.fastly.net`** (Khan) and
  **`a1744.dscw154.akamai.net`** (Math Academy) instead of the branded host.

## Operator incident

During a **downtime schedule** (whole-MAC `blocked=true`) block, Math Academy and Khan
Academy stopped working. `mathacademy.com` itself showed *allowed* (its own A record
attributed correctly), but the apps' CDN-fronted asset hosts showed/were blocked. The
operator had to disable the downtime schedule to unblock the apps.

## Dual impact

The attributed hostname (`hname`) is load-bearing in two places (see
`openwrt/files/usr/lib/lua/wifihaven/conntrack.lua` header design-notes, ~lines 8-22):

1. **Logging** — `connection_events` / `traffic_reports` show the CNAME target, not the
   branded host. Hard to read; pollutes per-host usage.
2. **Flow classification** — per-host `eb_`/`bl_` and `extraAllowed` (`ea_`) decisions in
   conntrack.lua are driven off `hname` matched against `ea_hosts_by_mac` /
   `eb_hosts_by_mac` using dnsmasq suffix-match semantics. When the asset flow is
   attributed to `prod.khan.map.fastly.net` instead of a Khan-branded host, it cannot
   suffix-match the app's `kastatic.org` / `khanacademy.org` `extraAllowed` entry, so during
   a whole-MAC block it is **classified as blocked** (no carve-out).

(Per the architectural model, blocking is connection-layer nftables on resolved IPs — DNS
is never the enforcement plane. This is an attribution-naming defect; the resolved name the
flow is *labelled* with is wrong. Fixing the label does not change the enforcement model.)

## Root cause (verified against real prod router logs, dnsmasq 2.91, `--log-queries=extra`)

The originally-suspected mechanism (qid-correlation miss / `dns_log.ingest_line` falling
back to `r.name`) is **not** what happens. Captured live:

```
38053 127.0.0.1/36172 query[A] cdn.kastatic.org from 127.0.0.1
38053 127.0.0.1/36172 reply cdn.kastatic.org is <CNAME>
38053 127.0.0.1/36172 reply fastly.kastatic.org is <CNAME>
38053 127.0.0.1/36172 reply prod.khan.map.fastly.net is 199.232.65.42
```

The qid (`38053`) **is** shared across the whole chain and the `query[A]` line **is**
present — in both the forwarded and the cached path — so `ingest_line`'s qid correlation
*does* attribute `199.232.65.42 → cdn.kastatic.org` correctly for that query.

The real mechanism: the kid's Apple device (`192.168.10.159`) **re-queries the final CNAME
target directly**. Over one session it issued:

```
18×  query[...] www.khanacademy.org          (branded name)
17×  query[...] prod.khan.map.fastly.net      (CNAME target, queried DIRECTLY)
```

i.e. the device's resolver follows the CNAME chain itself and then re-resolves the target
name on its own. On the *direct* target query the qname on the wire **is**
`prod.khan.map.fastly.net`, so the cache correctly (for that query) attributes the IP to the
CNAME target — which then fails the `khanacademy.org` suffix-match.

The live dns cache (`/tmp/wifihaven-dns-cache.txt`) is consequently full of raw CDN-target
names (`a1815.dscr.akamai.net`, `e35058.api12.akamaiedge.net`,
`d2763msf1wgvw2.cloudfront.net`, `stripecdn.map.fastly.net`, …) because devices query those
targets directly. The cache is a flat, global, IP-keyed, last-writer-wins map.

Crucially, the device **did** earlier resolve the branded `www.khanacademy.org`, and
dnsmasq logged that chain (`www.khanacademy.org → … → prod.khan.map.fastly.net`). So the
target→brand relationship **was observed** and is recoverable: the router can learn
`prod.khan.map.fastly.net` is a CNAME descendant of `www.khanacademy.org` and attribute the
later direct-target query back to the branded ancestor.

## Fix (this PR — router-side only)

Teach `dns_log` to remember CNAME-target → chain-head edges from the chains it already
parses, and attribute a resolved IP to the recovered branded head when the queried name is a
known CNAME alias. This corrects:

- the logged hostname (branded, not CDN target), and
- conntrack.lua's ea_/eb_/bl_ **classification** (the flow now suffix-matches the app's
  `extraAllowed` entry).

No new wire/snapshot fields; the API's `extraAllowed` is already correct.

## Related follow-up (NOT this PR)

The **kernel** `ea_<mac>_<host>` allow-set is populated only by dnsmasq's `nftset=/host/…`
directive, which fires only for names that suffix-match the configured domain. A
*directly-queried* CNAME target (`prod.khan.map.fastly.net`) does not suffix-match
`khanacademy.org`, so its resolved IPs may be absent from the kernel `ea_` carve-out and the
flow can be **dropped** at enforcement time during a whole-MAC block — independent of the
attribution label. Restoring enforcement-time connectivity for directly-queried CNAME
targets needs a dns-tail `ea_`-set populator keyed off the same CNAME-alias memory. Filing
separately; this PR fixes the attribution lookup (logging + classification) the operator
identified as the primary defect.

## Reproduction (busted)

Feed `dns_log.ingest_line` the captured branded chain for `www.khanacademy.org` followed by
a direct `query[A] prod.khan.map.fastly.net`, then assert `lookup(<ip>)` returns the branded
ancestor, not the CNAME target.

sameerparekh:
Fixes #1344.

## Summary

The router's IP→hostname attribution (the `dns_log` cache) was recording the
**final CNAME-target name** (`prod.khan.map.fastly.net`, `a1744.dscw154.akamai.net`)
for some CDN-fronted flows instead of the branded host the user is actually on.
Because that attributed name (`hname`) drives both **logging** and conntrack.lua's
**ea_/eb_/bl_ flow classification**, the CDN target failed to suffix-match the app's
`extraAllowed` entry (`kastatic.org` / `khanacademy.org`) — so during a whole-MAC
downtime block, Khan Academy / Math Academy asset flows were mislabelled and
mis-classified as blocked. (Operator had to disable the downtime schedule to unblock.)

## Root cause — verified against real prod router logs

This was **not** a query-id correlation miss (the original hypothesis). Captured live
on `router.lan` (dnsmasq 2.91, `--log-queries=extra`):

```
38053 127.0.0.1/36172 query[A] cdn.kastatic.org from 127.0.0.1
38053 127.0.0.1/36172 reply cdn.kastatic.org is <CNAME>
38053 127.0.0.1/36172 reply fastly.kastatic.org is <CNAME>
38053 127.0.0.1/36172 reply prod.khan.map.fastly.net is 199.232.65.42
```

The qid (`38053`) **is** shared across the whole chain and the `query[A]` line **is**
present (in both the forwarded and cached paths), so qid correlation already attributes
`199.232.65.42 → cdn.kastatic.org` correctly for that query.

The real mechanism: the kid's Apple device (`192.168.10.159`) **re-queries the final
CNAME target directly** as a separate lookup. Over one session it issued:

```
18×  query[...] www.khanacademy.org          (branded name)
17×  query[...] prod.khan.map.fastly.net      (CNAME target, queried DIRECTLY)
```

On the *direct* target query the qname on the wire **is** `prod.khan.map.fastly.net`,
so the cache attributes the IP to the CDN target. The live cache
(`/tmp/wifihaven-dns-cache.txt`, 941 entries) is consequently full of raw CDN targets
(`a1815.dscr.akamai.net`, `e35058.api12.akamaiedge.net`, `d2763msf1wgvw2.cloudfront.net`,
`stripecdn.map.fastly.net`, …). But the device **did** earlier resolve the branded
`www.khanacademy.org`, and dnsmasq logged that chain — so the target→brand link was
observed and is recoverable.

## Fix (router-side only)

`dns_log` now remembers a bounded **CNAME-target → chain-head** map, built from the
chains it already parses (every non-head owner name in a chain maps to that chain's
queried head). A later direct query for a known target is attributed back to the
branded ancestor. Best-effort and TTL-bounded:

- a first direct-target query seen before any branded chain falls back to the target
  name (never worse than the IP-literal default);
- alias edges expire on the same TTL as cache entries;
- a bounded multi-hop resolver with a cycle guard handles re-queried intermediate hops.

Substitution happens at **store** time in the dns-tail sidecar, so the snapshot file the
agent reads already carries the branded name — no new wire/snapshot fields, and the API's
`extraAllowed` is already correct.

## Downstream effect

With correct attribution, `hname = cdn.kastatic.org` suffix-matches the `kastatic.org`
`extraAllowed` entry, so conntrack.lua's `#421` carve-out fires and the flow is
classified `allowed=true` instead of blocked. Proven end-to-end by a new conntrack test
that wires a **real** `dns_log` cache (fed the real prod log shapes) through
`handle_flow`, plus a contrast test showing the pre-fix mis-classification.

## Related follow-up (NOT in this PR)

The **kernel** `ea_<mac>_<host>` allow-set is populated only by dnsmasq's `nftset=/host/…`
directive, which fires only for names that suffix-match the configured domain — so a
*directly-queried* CNAME target's IPs may be absent from the kernel carve-out and the flow
can be **dropped** at enforcement time, independent of the attribution label. Restoring
enforcement-time connectivity for directly-queried CNAME targets needs a dns-tail
`ea_`-set populator keyed off the same CNAME-alias memory. Tracked in #1344 as a related
follow-up; this PR fixes the attribution lookup (logging + classification) the operator
identified as the primary defect.

## Test plan

TDD, red→green visible in history:

- `test(openwrt): failing tests for CNAME-target attribution` — 3 alias cases fail on
  current code; safe-degradation + non-rewrite cases pass.
- `fix(openwrt): attribute directly-queried CNAME targets …` — all green.

```
busted test/dns_log_spec.lua      → 36 successes / 0 failures
busted test/conntrack_spec.lua    → 94 successes / 0 failures (6 pre-existing
                                     luci.jsonc env errors in the JSON-encode path,
                                     unrelated to this change)
busted test/render_spec.lua       → 144 successes / 0 failures
```

No metric series added, so no dashboard change. Router-side only; no Scala touched.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -723,6 +723,21 @@ so a flow to `142.250.x.x` shortly after `query[A] youtube.com` is logged as
 `youtube.com`, not `youtube-ui.l.google.com` (the last CNAME hop) and not
 the IP literal.
 
+Some clients (notably Apple devices) don't stop at one query: after resolving
+the branded host they **re-query the final CNAME target directly** as a
+separate lookup. On that second query the qname on the wire *is* the CDN
+target (e.g. `query[A] prod.khan.map.fastly.net`), so the query-id correlation
+— which is working correctly — still attributes the flow to the CDN target,
+which then fails to suffix-match the app's branded `extraAllowed` entry
+(`kastatic.org`) and is mislabelled / mis-classified as blocked. To recover the
+brand, `dns_log` remembers a bounded **CNAME-target → chain-head** map built
+from the chains it already parses (every non-head owner name in a chain maps to
+that chain's queried head). A later direct query for a known target is then
+attributed back to the branded ancestor the target was first observed under.
+This is best-effort: the very first direct-target query seen before any branded
+chain falls back to the target name (never worse than the IP-literal default),
+and stale alias edges expire on the same TTL as cache entries (#1344).
+
 The per-domain `--ipset=` mechanism is still used for the `site_limits`
 enforcement chains (nftables matches `ip daddr @profileN_<domain>`), but it
 is no longer load-bearing for hostname attribution in the query log — that
--- a/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
@@ -141,6 +141,9 @@ function M.new(opts)
   local ttl           = opts.ttl_seconds or 3600
   local max_entries   = opts.max_entries or 10000
   local max_pending   = opts.max_pending or 1024
+  -- CNAME-alias map bound. Distinct CDN target names observed across chains;
+  -- shares the entry-cache order of magnitude (#1344).
+  local max_aliases   = opts.max_aliases or 10000
   local now_fn        = opts.now_fn      or os.time
 
   -- ip → { hostname, ts, seq }
@@ -149,6 +152,14 @@ function M.new(opts)
   -- qid → { qname, ts, seq }
   local pending       = {}
   local pending_count = 0
+  -- CNAME target name → { head, ts, seq } (#1344).
+  -- Learned from observed CNAME chains: every non-head owner name in a chain
+  -- maps to that chain's head (the queried name). Lets us attribute a flow
+  -- back to the branded host even when the client re-queries the final CNAME
+  -- target directly (Apple devices do this), where the qname on the wire is
+  -- the CDN target itself.
+  local aliases       = {}
+  local alias_count   = 0
   -- monotonically increasing sequence number for LRU-by-insertion eviction
   local seq           = 0
 
@@ -203,6 +214,54 @@ function M.new(opts)
     pending[qid] = { qname = qname, ts = now, seq = next_seq() }
   end
 
+  local function evict_oldest_alias()
+    local oldest_name, oldest_seq
+    for name, a in pairs(aliases) do
+      if not oldest_seq or a.seq < oldest_seq then
+        oldest_name, oldest_seq = name, a.seq
+      end
+    end
+    if oldest_name then
+      aliases[oldest_name] = nil
+      alias_count = alias_count - 1
+    end
+  end
+
+  -- resolve_head(name) → the branded chain head `name` is a CNAME alias of, or
+  -- `name` itself when it is not a known alias. Bounded multi-hop with a cycle
+  -- guard so a re-queried intermediate hop (whose own alias was learned from a
+  -- longer chain) still resolves to the ultimate head (#1344). Expired alias
+  -- edges are ignored (and lazily dropped) so a stale CDN→brand mapping can't
+  -- pin an IP to the wrong brand forever.
+  local function resolve_head(name)
+    local now  = now_fn()
+    local seen = {}
+    for _ = 1, 8 do
+      local a = aliases[name]
+      if not a then break end
+      if (now - a.ts) > ttl then
+        aliases[name] = nil
+        alias_count = alias_count - 1
+        break
+      end
+      if seen[name] then break end
+      seen[name] = true
+      name = a.head
+    end
+    return name
+  end
+
+  local function remember_alias(target, head, now)
+    if target == head then return end
+    if aliases[target] == nil then
+      if alias_count >= max_aliases then
+        evict_oldest_alias()
+      end
+      alias_count = alias_count + 1
+    end
+    aliases[target] = { head = head, ts = now, seq = next_seq() }
+  end
+
   local self = {}
 
   function self.ingest_line(line)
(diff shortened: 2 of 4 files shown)


sameerparekh:
The enforcement-side follow-up referenced above is now tracked in #1346 (kernel `ea_` allow-set population for directly-queried CNAME targets), to land after this PR.

Issue #1348, opened 2026-06-03
router: category blocklist (bl_) drop-set bypassed by directly-queried CNAME targets / CDN-anycast IP variance


sameerparekh:
Split out from #1346 (which now covers the `ea_`/`eb_` dns-tail populator path).
Related root cause: #1344 / #1345.

## Problem

Category blocklist drop-sets (`bl_<id>` / `bl6_<id>`) can be **bypassed** when a
client re-queries a CNAME target directly — the same directly-queried-CNAME-target
behaviour that motivated #1344 (verified on prod `router.lan`: Apple devices issue a
`query[A] <cdn-target>` after resolving the branded host).

But the mechanism differs from `ea_`/`eb_`, so the #1346 fix is **not** a drop-in here:

- `ea_` / `eb_` are populated per client DNS answer — `ea_` by dnsmasq's
  `nftset=/host/…` directive, `eb_` by that **plus** dns-tail's `maybe_populate_eb`
  (#515) keyed on the answered name. #1346 teaches those to consult the #1344
  CNAME-alias map.
- `bl_` is populated **agent-side**: `blocklists.lua` fetches each blocklist URL and
  resolves its member hosts itself, then adds the resolved IPs to `bl_<id>` on a
  periodic cadence — independent of what any client queries.

So the `bl_` bypass is **CDN-anycast IP variance**: the agent resolves blocklist
member `tiktok.com` → chain → some Fastly/Akamai anycast IPs and adds *those* to
`bl_<id>`. A device that re-queries the member's CNAME target directly can land on a
*different* anycast IP in the same provider range that was never added to `bl_<id>`
→ the drop rule misses → blocked content is reachable. (This is the same class of
miss the live cache showed for attribution: many distinct CDN IPs per
hostname/region, and the set only ever holds the few the agent happened to resolve.)

## Why the alias map alone doesn't fix it

The #1344 alias map (CDN target → branded head) lets dns-tail attribute a client's
*observed* answer back to a brand. `bl_` isn't driven by client answers, so there's
nothing for dns-tail to map at population time today.

## Candidate approaches (needs design)

1. **dns-tail `bl_` populator (mirror #515/#1346).** Have dns-tail scan for declared
   `bl_<id>` / `bl6_<id>` sets and, on each client `reply <name> is <ip>`, resolve
   `<name>` through the #1344 alias map; if the recovered brand matches a blocklist
[...]

Issue #1346, opened 2026-06-03
router: populate kernel ea_/eb_ sets for directly-queried CNAME targets (enforcement-side follow-up to #1344)


sameerparekh:
Follow-up to #1344 / #1345 (attribution-lookup fix).

## Problem

The **kernel** `ea_<mac>_<host>` allow-set (the nftables carve-out that lets an
`extraAllowed` host through a whole-MAC block) is populated **only** by dnsmasq's
`nftset=/<host>/…#ea_<mac>_<host>` directive. dnsmasq fires that callback only for
resolutions whose **queried name** suffix-matches the configured domain.

When an Apple device re-queries a CNAME target directly (verified on prod `router.lan`:
`192.168.10.159` issues `query[A] prod.khan.map.fastly.net` 17× after resolving
`www.khanacademy.org` 18×), the queried name is `prod.khan.map.fastly.net`, which does
**not** suffix-match `khanacademy.org` / `kastatic.org`. So dnsmasq does not add the IPs
returned by that direct query to `ea_<mac>_khanacademy_org`. Those IPs can therefore be
**absent from the kernel carve-out**, and the flow is dropped at enforcement time during a
whole-MAC (downtime-schedule) block — independent of how the flow is *labelled*.

#1345 fixes the attribution **label** (logging + conntrack.lua classification) by learning
a CNAME-target → branded-head map in `dns_log`. But it does not populate the kernel `ea_`
set for directly-queried targets, so it does not by itself restore connectivity.

## Proposed fix

dns-tail already populates `resolved_<mac>` (#505) and `eb_<sanhost>` (#515) sets out of
band by tailing the same query log. Add the symmetric **`ea_` (allow) populator**: when a
`reply <name> is <ip>` line's `<name>` resolves (via the #1344 CNAME-alias map) to a host
that has a declared `ea_<mac>_<host>` set, `nft add element` the IP to that set — inline,
in lockstep with the hostname cache flush (same first-SYN-race argument as #505/#515).

This reuses the CNAME-alias memory introduced in #1345, so it should land after that PR.

## Scope notes

- Router-side only; no wire/snapshot changes (the API's `extraAllowed` is already correct).
- Do **not** pin CDN edge hostnames (they rotate — see #1337).

---

## Scope update: the block-sets have the same gap (and it's higher severity)

The identical attribution gap applies to **extraBlocked** as well as the allow
carve-out, and there it's a **filter bypass**, not a false block — so it's the
[...]

Issue #1337, opened 2026-06-03
infra-allow: add akamai/apple CDN hosts + evaluate allowing all of Akamai


sameerparekh:
## Ask

Add two more hosts to the curated infra allowlist (`PolicyService.infraAllowHosts`, shipped in #1309 — copied into every profile's `extraAllowed` so they survive the whole-MAC block):

- `a1744.dscw154.akamai.net`
- `netcts.cdn-apple.com`

Operator hit these as transitive deps that were still getting dropped when a profile was blocked, breaking otherwise-allowed apps.

## Design question — allow ALL of Akamai?

Operator's reasoning: Akamai is a shared CDN. If we allow all akamai-hosted edge hosts (e.g. `*.akamai.net`, `*.akamaiedge.net`, `*.akamaihd.net`), it should be **safe** because a blocked *site* is gated on its own primary domain — which stays blocked — so allowing its Akamai CDN assets doesn't make the blocked site reachable. The CDN is just infra.

Evaluate this carefully before implementing the wildcard:
- **The hole:** some services are served *entirely* from an Akamai-hosted hostname with no separate origin domain we block (the akamai host effectively IS the site). A blanket Akamai allow would punch through the block for those. Enumerate whether any realistic blocked category (ads, adult, social) resolves to akamai edge hostnames that the user could hit directly. Ad/tracking CDNs in particular sometimes ride Akamai.
- **blockIpOnly interaction:** with `blockIpOnly` set, only hostname-attributed IPs are evaluated; a wildcard infra-allow must still respect that.
- **Matching mechanism:** `extraAllowed` is per-host into an `ea_` ipset populated from resolved hostnames (dnsmasq `--ipset`). A `*.akamai.net` wildcard needs a wildcard-capable attribution path (dnsmasq supports `ipset=/akamai.net/...`). Confirm the router can populate an ipset from a domain *suffix*, not just exact hosts, or the wildcard won't actually match.

**Recommendation to assess:** start by adding just the two exact hosts above (low risk, unblocks the operator now). Treat the Akamai-wildcard as a separate decision in the same PR — if the suffix-ipset path is clean and the security review finds no realistic bypass, add a curated set of Akamai edge suffixes; otherwise keep it to exact hosts and document why.

## Notes
- This is additive to the curated constant; the cleaner DB-backed/configurable version is #1315 — don't block on it.
[...]

PR #1350, opened 2026-06-03
fix(openwrt): close bl_ category bypass via dns-tail CNAME populator (#1348)

sameerparekh:
## Summary

Closes the category-blocklist (`bl_`) filter bypass in #1348: a client that
re-queries a blocklist member's CNAME target **directly** lands on a
CDN-anycast IP the router never added to `bl_<id>`, so the category drop
misses and blocked content becomes reachable. This is higher severity than the
`ea_`/`eb_` half (a silent *bypass*, not a false block).

## Design

`bl_<id>`/`bl6_<id>` are declared per blocklist id and populated at DNS resolve
time by render's `nftset=/<member>/4#…#bl_<id>,6#…#bl6_<id>` directive — which
fires only for queries whose *name* suffix-matches a member host. A direct
re-query of a member's CNAME target (verified on prod: Apple devices do this)
doesn't suffix-match the member, so the resolved anycast IP never lands in the
set.

The fix mirrors the `eb_` dns-tail populator (#515/#1346): on each client
`reply <name> is <ip>`, dns-tail resolves `<name>` back to its branded chain
head through the #1344 CNAME-alias map and, if the recovered brand is a
blocklist **member**, adds the IP to that member's `bl_` set itself.

The one new piece: dns-tail knew which `bl_` sets *exist* (from
`nft list table`) but not their **membership** (`eb_`'s set name *is* the host;
`bl_`'s is the *id*). So render exports a minimal host→set index, written by
`policy.apply` alongside the dnsmasq fragment it's derived from.

## Changes (router-side only)

- `dns_log.resolve_head` — promote the #1344 CNAME-alias resolver to a **public
  method** on the cache instance (shares the live learned alias edges; no
  second map).
- `dns_log.parse_blocklist_member_index` / `dns_log.populate_bl` — pure index
  parser and an alias-resolve + dnsmasq-subdomain-walk populator that takes an
  **injected** `add_fn` (so it's unit-testable; dns-tail wraps it with the real
  `nft add element`).
- `render.blocklist_member_index` — emit `<host>\t<bl_set>\t<bl6_set>` rows from
  the same `_blocklist_hosts` that drives the `nftset=` directives, so set names
  line up exactly.
- `policy.apply` — write the index to `paths.bl_member_index`
  (`/tmp/wifihaven-bl-members.txt`) in lockstep with `dnsmasq.conf`; non-fatal
  on write failure.
- `wifihaven-dns-tail` — load + refresh the index on the existing 30s
  set-refresh cadence and run `maybe_populate_bl` inline per reply (same
  first-SYN-race argument as #505/#515); `bl_adds` added to the heartbeat.

**No wire/snapshot changes** (the API ships `blocklistIds`; membership is
resolved on the router from the blocklist bodies it already fetches) and **no
pinned CDN hostnames/IPs** (#1337) — the populator only ever adds IPs observed
in live client answers, attributed by the rotating alias map.

## Test plan

TDD, red→green visible in history:

- `a1a847d` — failing busted tests first.
- `dff696c` — implementation.

New coverage in `openwrt/test/dns_log_spec.lua` and `render_spec.lua`:
- the core bypass: a member resolved agent-side to IP-set A, then a device
  directly querying the member's CNAME target landing on IP B — asserts B is
  added to the correct `bl_<id>` via the alias map + member index;
- v6 (AAAA → `bl6_`), dnsmasq subdomain semantics (suffix walk), a member in
  multiple blocklists, and the safe no-op when the head matches no member;
- `parse_blocklist_member_index` parsing + the `resolve_head` accessor;
- `render.blocklist_member_index` row/sanitization/empty-input output.

Local: `busted test/` → **426 successes / 2 failures / 33 errors / 1 pending**
(+12 vs base; the 2 failover failures and 33 `luci.jsonc`/`wifihaven.render`
"not found" errors are pre-existing/environmental on this host, unchanged from
base).

## Dependency / merge order

Based on **#1345** (`claude/fix-dns-attribution`, PR open) — this PR targets
that branch so the diff is isolated. The **#1346** branch isn't pushed yet;
its public `resolve_head` accessor is added here idempotently. **Reconcile the
single shared accessor when sequencing merges: #1345 → #1346 → #1348.**

Refs #1348, #1344, #1346, #515, #505, #1337.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
@@ -329,8 +329,9 @@ function M.new(opts)
   -- `name` unchanged when it is not a known alias (#1344). Exposed publicly so
   -- the wifihaven-dns-tail sidecar can reuse the SAME learned alias memory to
   -- recover the brand for a directly-queried CDN target before suffix-matching
-  -- it against the kernel ea_/eb_ sets (#1346). Internally `resolve_head` is
-  -- already used by ingest_line; this is a thin public passthrough.
+  -- it against the kernel ea_/eb_ sets (#1346) and the category bl_ sets
+  -- (#1348). Internally `resolve_head` is already used by ingest_line; this is
+  -- a thin public passthrough.
   function self.resolve_head(name)
     return resolve_head(name)
   end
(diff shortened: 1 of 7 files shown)


sameerparekh:
#1345 has merged, so this PR is now rebased onto `main` (base retargeted; the two #1344 commits dropped — main carries that content via the #1345 squash). The diff is now exactly the two #1348 commits.

Remaining merge-order note: this still adds the public `dns_log.resolve_head` accessor idempotently because **#1346 isn't merged yet**. When #1346 lands it will introduce the same accessor — reconcile the single shared definition then (sequence **#1346 then #1348**, or merge #1348 first and have #1346 reuse the accessor already here).

sameerparekh:
Rebased onto `main` now that **#1346 (PR #1349) has merged** — conflicts resolved:

- **`dns_log.resolve_head`**: dropped my idempotent copy; the branch now uses #1346's public accessor (the duplicate from the auto-merge is gone — there is exactly one definition).
- **`wifihaven-dns-tail`**: reconciled onto #1346's refactor — `refresh_nft_sets` now returns the single `s` membership table and the eb_/ea_ populators live in the new `wifihaven.dns_tail_sets` module. The bl_ populator (`maybe_populate_bl`) sits alongside them and now wires `dns_sets.nft_add_element` (the shared helper #1346 introduced) as its injected `add_fn`; the bl_ heartbeat counter is folded into the same stats line as bio/eb/ea.
- bl_ core logic stays in `dns_log.populate_bl` / `parse_blocklist_member_index` (it depends on the cache's alias map and the member→id index, not on host-name sanitization like eb_/ea_), and remains unit-tested in `dns_log_spec`.

Full suite: **442 successes / 2 failures / 33 errors / 1 pending** — the 2 failover failures and 33 `luci.jsonc`/`wifihaven.render` errors are pre-existing/environmental, unchanged from base. The PR is now MERGEABLE against `main` with no remaining #1346 overlap.

Issue #1357, opened 2026-06-03
Automate terraform apply for infra/cloudflare (remote state + master-cloudflare.yml apply pipeline)


sameerparekh:
## Goal

Add an automated `terraform apply` pipeline for `infra/cloudflare/` so changes to the Cloudflare DNS/Pages Terraform apply on merge to `main`, instead of requiring a manual operator-run `terraform apply`. Mirror the existing Grafana pipeline (`master-grafana.yml`'s `deploy-grafana` job, which `terraform apply`s `infra/grafana` on changes under `infra/grafana/**` and `deploy/grafana/dashboards/**`).

Motivation: the e2e CNAME fixture records added in #1351 (`e2e-brand/e2e-mid/e2e-edge.wifihaven.net`) currently require a manual apply before the router-VM e2e gate can resolve the chain. More generally, `infra/cloudflare` is the source of truth for the Pages projects, the API CNAMEs, and these e2e records, but has no automated apply — drift and manual steps are easy to miss.

## HARD PREREQUISITE — remote state (this is the real work)

`infra/cloudflare` uses a **local** backend and the state file is **gitignored** (`*.tfstate`; see `.gitignore` and the `TODO(#613-followup)` in `main.tf`). A CI runner is a fresh checkout with **no state file**, so a naive `terraform apply` in CI would see an empty state, try to **create all ~10 already-existing resources** (2 Pages projects, 3 Pages domains, 2 API CNAMEs, + the 3 e2e records), and fail on "already exists" — or worse, fight the operator's local state.

So the pipeline cannot be added without first moving to a **remote backend** (this is the deferred #613-followup remote-state work; fold it in here). Required sub-steps:

1. **Choose + configure a remote backend** that supports state locking. Options to weigh (pick one, justify in the PR):
   - Terraform Cloud / HCP Terraform (free tier, native locking) — least infra.
   - An S3-compatible backend on **Cloudflare R2** — keeps it in-vendor, but R2 has no native lock table; document the locking story or pair with a DynamoDB-style lock alternative.
   - Postgres/HTTP backend.
2. **Migrate the operator's existing local state** into the chosen backend (`terraform init -migrate-state`) so the ~10 live resources are tracked — do NOT recreate them. If the canonical local state has been lost, fall back to `terraform import` for each existing resource; enumerate them in the PR.
3. **Backend credentials** as GitHub secrets (declaratively documented), separate from the Cloudflare token.

## Pipeline (after remote state exists)

[...]

Issue #1334, opened 2026-06-03
bug(prod): category blocklist ('ads') not enforced — assigned hosts still reachable


sameerparekh:
## Report

On **prod**, an **'ads' blocklist is configured on the profile** for the operator's computer, but ad hosts on that list are still reachable from the device. Category-blocklist enforcement appears to be a no-op in prod despite #352 ("Router: implement category blocklist enforcement via nftables ipsets") having been closed (2026-05-14).

Filing fresh per the no-reopen-merged-issues rule. This is a regression / operational gap, not a never-built feature — enforcement code exists (`openwrt/files/usr/lib/lua/wifihaven/blocklists.lua` + `render.lua`).

## Architecture recap (target behavior)
Per AGENTS.md: category lists apply per-MAC via `blocklistIds: List[BlocklistId]` in the snapshot. The agent fetches each blocklist by URL (cached by version), resolves member hosts into a **per-blocklist nftables ipset**, and drops `(mac, dst ip ∈ ipset)` at the forward chain. DNS still resolves (blocking is connection-layer, not DNS — `memory/blocking_is_traffic_layer_not_dns.md`); the block is the nft drop / HTTP-80 DNAT to the block page. So "the ad host resolves in DNS" is NOT evidence it's unblocked — verify at the connection layer.

> Note: AGENTS.md's May-2026 "deviations" callout still claims "category blocklists are not yet applied on the router at all," which contradicts #352 being closed. Part of this issue is reconciling that — figure out whether #352 only landed part of the path (e.g. snapshot field but no ipset population), and fix the doc to match reality.

## Candidate root causes to check, in order (prod-first diagnosis)
1. **Snapshot doesn't carry it.** Fetch the deployed `/api/router/policy` snapshot for the computer's MAC and confirm the 'ads' blocklist id is actually present in that MAC's `blocklistIds` (and that `PolicyService` maps the profile's category assignment → `blocklistIds`, not just the legacy `blockedCategories`/`/api/router/decision` path #352's body flagged).
2. **Stale router agent.** Confirm the prod router is running an agent version that includes the #352 enforcement (see `memory/router_update_deadlock_recovery.md` — routers have been stuck on old versions before). `opkg info wifihaven` / agent version on the snapshot.
[...]

PR #1361, opened 2026-06-04
fix(openwrt): authenticate blocklist fetch + de-fragilize CNAME e2e (#1360)

### LINKED ISSUE #1360: Gate 2 red: CNAME direct-requery suite fails — dead origin IP (93.184.216.34) + bl_/blocked_macs enforcement timeouts
## Symptom (deterministic — both apk AND ipk arms)

Master Router CD [run 26916273450](https://github.com/wifihaven/wifihaven/actions/runs/26916273450) (sha a5f6ea61) — Gate 2 red on `scenarios_fake/test_cname_direct_requery.py` (Suite G, #1351), identically on both matrix arms:

- `test_bl_direct_requery_blocked` — timed out waiting for **block page** response for `e2e-edge.wifihaven.net`
- `test_ea_direct_requery_allowed_under_blocked_mac` — timed out waiting for mac present in **blocked_macs** (fails at SETUP precondition)
- `test_attribution_reports_branded_host_after_direct_requery` — timed out waiting for **allowed HTTP response**

(The `client-up.sh ... already running` errors on later tests are the #1286 secondary cascade off these primaries.)

This is freshly-merged feature work: #1346/#1349 (ea_/eb_ populator), #1348/#1350 (bl_ populator), #1351/#1353 (the suite), and **#1356 promoted G4 bl_ to a real gating assertion** (merged 22:12Z). main is now deterministically red, blocking all router releases — a #1194-class "merged without passing full Gate 2" event.

## Triage so far

**Ruled OUT — the documented Terraform DNS prerequisite is satisfied.** The suite docstring warns it needs `terraform apply` in infra/cloudflare/. The records DO resolve from outside CI:
```
e2e-brand.wifihaven.net CNAME e2e-mid → CNAME e2e-edge → A 93.184.216.34
```
So this is NOT the missing-DNS prereq. Failure is in the enforcement/origin step, not resolution.

**CONFIRMED contributor — the leaf origin IP is dead.** `e2e-edge.wifihaven.net A → 93.184.216.34` is the old IANA example.com IP, reassigned/decommissioned ~2025. `curl -m8 http://93.184.216.34/` → connection timed out (http_code=000); example.com now lives on 104.20.23.154 / 172.66.147.243. The suite's **allow-path** assertions need a real HTTP response from the leaf (`test_attribution_...` waits for "allowed HTTP response"), so a dead origin makes them time out. The leaf A record is managed in `infra/cloudflare/main.tf` (per the docstring).

**Still to investigate (not explained by the dead origin):**
- `test_bl_direct_requery_blocked` expects the **block page** via port-80 DNAT, which intercepts *before* the origin — a dead origin shouldn't break it. So either the bl_ kernel set isn't being populated for the directly-queried leaf (real populator bug — the "highest severity" bypass the feature targets) OR the DNAT/block-page path isn't firing in the VM.
- `test_ea_...` times out at SETUP ("mac present in blocked_macs"), i.e. before the carve-out — a policy.apply / API-sync issue, not the origin.
- Recurring across all failures: `policy.apply: smoke check failed; dnsmasq may be serving a stale config — got "nil" for tiktok.com (expected a real upstream IP)`. Determine whether this dnsmasq smoke-check failure is the common root cause (degraded in-VM resolution breaking the dns-tail pipeline) or benign noise.

## Fix (investigate, then implement)

1. **Replace the dead leaf origin.** Hardcoding an external IP is inherently fragile. Options: point `e2e-edge.wifihaven.net A` at an origin the harness controls / a stable always-up endpoint, or have fake-mode serve the origin locally so allow-path tests don't depend on third-party IP churn. Whatever's chosen, update `infra/cloudflare/main.tf` (+ note the operator `terraform apply`) and the test docstring.
2. **Root-cause the bl_ block + blocked_macs-setup timeouts** — confirm whether the dns-tail bl_ populator actually adds the directly-queried leaf IP to the kernel set in the VM, and whether the policy.apply dnsmasq smoke-check failure is breaking the pipeline. Fix the real defect (populator or harness).
3. **Decide on de-gating to unblock main.** If the real fix is non-trivial, temporarily revert the #1356 gating promotion (and/or mark Suite G xfail/skip-when-origin-unreachable) so router releases can flow — BUT only if #2 turns out to be harness/test-data, not a genuine bl_ bypass. If it's a real bl_ enforcement bug, keep it gating (it's a security carve-out) and fix forward. Call this out explicitly.

## Acceptance
- Gate 2 `test_cname_direct_requery.py` passes on both apk and ipk arms.
- Allow-path tests no longer depend on a stale third-party origin IP.
- bl_/blocked_macs failures root-caused and fixed (or consciously de-gated with rationale).

Refs #1346, #1348, #1349, #1350, #1351, #1356, #1286, #1194.

sameerparekh:
Fixes #1360. Master Router CD **Gate 2** was deterministically red on both apk and ipk arms on `scenarios_fake/test_cname_direct_requery.py` (Suite G), blocking all router releases. Root-caused from run [26916273450](https://github.com/wifihaven/wifihaven/actions/runs/26916273450). Full investigation in the [issue comment](https://github.com/wifihaven/wifihaven/issues/1360#issuecomment-4617802083).

## Root causes

| Test | Cause | Class |
|------|-------|-------|
| **G4** `test_bl_direct_requery_blocked` | Agent fetches `GET /api/blocklists/<id>` with **no `Authorization` header**, but the route is router-authenticated (`RouterRoutes.scala:93`) → **401** → `_blocklist_hosts` empty → `bl_` set never populated → no DNAT → no block page | **Real prod bug** |
| **G2** `test_ea_..._blocked_mac` | Precondition asserts `blocked_macs` membership, but per **#421** a blocked MAC *with* extraAllowed is pulled OUT of `@blocked_macs` into per-MAC ea-exception rules → can never appear there | Test bug |
| **G3** + G2-primary | Leaf origin `93.184.216.34` (legacy IANA example.com) is dead; no public IP reliably returns 2xx for an arbitrary `Host` header anymore | Harness/test-data |

The recurring `smoke check failed … got "nil" for tiktok.com` warning is **noise** — `policy.apply`'s `smoke_warn` is observation-only and never aborts the apply (`policy.lua:300-328`).

### The G4 bug is a real category-enforcement bypass

`GET /api/blocklists/<id>` requires the router bearer token in prod, but the agent's blocklist fetch never sent one — so **category (`blocklistIds`) enforcement is a silent no-op in production**, not just for directly-queried CNAME targets. Same *symptom* class as the #1334 silent-no-op, distinct *cause* (#1334 fixed arg-order; the bearer header was never added). **Fixed forward, kept hard-gating** per the issue's de-gating rule.

## Fixes

1. **Auth on blocklist fetch** — `blocklists.fetch_and_cache` takes the router token and sends `Authorization: Bearer …` (same as `policy.fetch`); the agent passes `router_token`. New busted regression tests assert the header is sent and that a missing token surfaces a fetch error.
2. **G2 precondition** — assert the per-MAC `wh_drop:<mac>` forward-chain rule (correct under #421) via a new `wait_mac_drop_rule_present` observer, instead of `blocked_macs` membership. The `ea_` populator proof (nft set membership) is reordered ahead of the HTTP step and stays **hard**.
3. **Leaf origin → RFC 5737 TEST-NET-1 `192.0.2.10`** (reserved-forever, never reassigned) in `infra/cloudflare/` (`variables.tf` default, `main.tf` comment, `terraform.tfvars.example`) + the test docstring/constant. Block-path tests (G1/G4) DNAT port 80 at prerouting **before** the dest is routed, so an unroutable leaf is fine; allow-path tests (G2-primary/G3) **`xfail` when the leaf is unreachable** rather than depending on a live third-party origin.

## Wire-contract / back-compat

The agent now sends an `Authorization` header the API route already requires — additive on the request, no wire-shape change. Older agents keep their existing (broken) behavior; newer API is unaffected.

## Operator notes

- **No manual `terraform apply` needed.** The Cloudflare apply is CI-automated on merge to main (`master-cloudflare.yml`, #1357/#1358), so the new leaf A `192.0.2.10` applies automatically.
- **Follow-up (will file):** a fully harness-served allow-path origin (fake-mode DNAT of the TEST-NET leaf → a local HTTP responder) would let G2-primary/G3 run for real instead of `xfail`.

## Validation

Cannot run Gate 2 locally (self-hosted KVM runner). Validated: `busted` — blocklists **13/13**, dns_tail_sets **14/14**, dns_log **47/47**; `terraform fmt -check` + `validate`; `py_compile`.

### DIFF
--- a/infra/cloudflare/main.tf
+++ b/infra/cloudflare/main.tf
@@ -159,14 +159,20 @@ resource "cloudflare_record" "api_staging" {
 # (#1346/#1349) end-to-end on a real router VM.
 #
 # Records are DNS-only (proxied=false) so the chain resolves as authored and
-# the leaf A record is never hidden behind Cloudflare's edge. The leaf IP is
-# 93.184.216.34 (IANA's example.com — stable, responds to HTTP/80 with a
-# non-block-page body, and is already used as the harness's reference
-# "internet-reachable" origin in scripts/e2e/lib/wait.py).
+# the leaf A record is never hidden behind Cloudflare's edge. The leaf IP
+# defaults to 192.0.2.10 (var.e2e_edge_ip) — RFC 5737 TEST-NET-1, reserved-for-
+# documentation and GUARANTEED never globally routed or reassigned (#1360). The
+# previous value 93.184.216.34 (legacy IANA example.com) was decommissioned
+# ~2025 and its HTTP/80 went dead, silently red-gating router releases. The
+# leaf is intentionally unroutable: the block-path e2e tests (G1 eb_, G4 bl_)
+# DNAT port 80 at prerouting before the dest is routed, so reachability is
+# irrelevant to them; the allow-path tests (G2/G3) xfail when the leaf is
+# unreachable rather than depending on a live third-party origin (see
+# scripts/e2e/scenarios_fake/test_cname_direct_requery.py).
 #
 # These records are applied by the CI pipeline (master-cloudflare.yml) on merge
-# to main — no manual `terraform apply` is needed going forward (#1357). The
-# ci.yml cloudflare-terraform job still validates fmt + validate on every PR
+# to main — no manual `terraform apply` is needed going forward (#1357/#1358).
+# The ci.yml cloudflare-terraform job still validates fmt + validate on every PR
 # but does NOT apply.
 
 resource "cloudflare_record" "e2e_brand" {
--- a/infra/cloudflare/terraform.tfvars.example
+++ b/infra/cloudflare/terraform.tfvars.example
@@ -4,6 +4,9 @@ zone_id    = "<paste from Cloudflare dash → wifihaven.net → right sidebar �
 api_prod_cname_target    = "<from Render → wifihaven-api-prod    → Settings → Custom Domains>"
 api_staging_cname_target = "<from Render → wifihaven-api-staging → Settings → Custom Domains>"
 
-# e2e test fixture leaf IP — default is 93.184.216.34 (IANA example.com).
-# Override only if you need a different origin for direct-requery CNAME tests (#1351).
-# e2e_edge_ip = "93.184.216.34"
+# e2e test fixture leaf IP — default is 192.0.2.10 (RFC 5737 TEST-NET-1, a
+# reserved-for-documentation address that is never globally routed/reassigned,
+# #1360). The direct-requery CNAME block-path tests DNAT port 80 before routing,
+# so the leaf need not be reachable; allow-path tests xfail when it isn't.
+# Override only if you wire up a reachable harness-served origin (#1360 follow-up).
+# e2e_edge_ip = "192.0.2.10"
--- a/infra/cloudflare/variables.tf
+++ b/infra/cloudflare/variables.tf
@@ -20,6 +20,6 @@ variable "api_staging_cname_target" {
 
 variable "e2e_edge_ip" {
   type        = string
-  description = "Leaf A record IP for e2e-edge.wifihaven.net — used by the direct-requery CNAME e2e fixture (#1351). Must be an IP that responds to HTTP/80 with a non-block-page body so the harness can distinguish 'reached real origin' from 'got DNAT block page'. Default: 93.184.216.34 (IANA example.com — stable, used by the harness as its reference internet-reachable origin)."
-  default     = "93.184.216.34"
+  description = "Leaf A record IP for e2e-edge.wifihaven.net — used by the direct-requery CNAME e2e fixture (#1351). Default 192.0.2.10 is RFC 5737 TEST-NET-1, reserved-for-documentation and GUARANTEED never globally routed or reassigned (#1360). The previous value 93.184.216.34 (legacy IANA example.com) was decommissioned ~2025 and its HTTP/80 went dead, silently red-gating router releases. The block-path e2e tests (G1/G4) DNAT port 80 at prerouting before the dest is routed, so an unroutable leaf is fine; the allow-path tests (G2/G3) xfail when the leaf is unreachable rather than depending on a live third-party origin (no public IP reliably returns 2xx for an arbitrary Host header anymore). See scripts/e2e/scenarios_fake/test_cname_direct_requery.py."
+  default     = "192.0.2.10"
 }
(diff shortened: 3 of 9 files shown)


sameerparekh:
Validated end-to-end on the self-hosted KVM runner — both `ipk-23.05` and `apk-25.12` arms now reach the assertions and Suite G G1/G4 are green:

```
G1 test_eb_direct_requery_blocked                                  PASSED
G2 test_ea_direct_requery_allowed_under_blocked_mac                XFAIL  (leaf unroutable, as designed)
G3 test_attribution_reports_branded_host_after_direct_requery      XFAIL  (leaf unroutable, as designed)
G4 test_bl_direct_requery_blocked                                  PASSED
```

Getting there required three follow-ups, all in the new commit `90ad2a4`:

**1. `mkdir -p` the blocklist cache dir (`openwrt/.../blocklists.lua`)** — closes #1363. The 401 fix lifted the symptom one layer up; on a fresh image `/etc/wifihaven/blocklists/` doesn't exist, so the atomic tmp-write ENOENT-fails, `bl_member_index` stays empty, and dns-tail's bl_ populator has no member to match on. G4 stayed red until this landed.

**2. `scripts/e2e/scenarios_fake/conftest.py` — pin dnsmasq upstream to 1.1.1.1 / 8.8.8.8 with `noresolv=1`.** Without this, the router VM's dnsmasq forwards via QEMU SLIRP → host `/etc/resolv.conf` → operator's LAN resolver. On the self-hosted KVM runner that resolver doesn't return the `e2e-{brand,mid,edge}.wifihaven.net` chain (the CD #1358 records are public, but the local resolver doesn't pick them up). Going straight to a public resolver sees the records.

**3. Same fixture, same uci batch — `rebind_domain=wifihaven.net`.** OpenWRT ships `rebind_protection=1`; dnsmasq's "private range" set includes the **RFC 5737 documentation ranges**, so the `192.0.2.10` (TEST-NET-1) leaf this PR chose gets dropped as a "possible DNS-rebind attack". Without the carve-out the leaf never resolves and G1..G4 all time out at the `dig` precondition. A narrow `rebind_domain` whitelist is preferable to disabling the protection wholesale.

Auth fix is now provable end-to-end: zero `(status=401)` log lines from the blocklist fetch, `GET /api/blocklists/e2e-cname-bl → 200`, `bl_adds` goes non-zero in dns-tail telemetry, the leaf IP lands in `bl_<id>`, and HTTP/80 to the leaf hits the block page.

(Findings #2 and #3 are environmental but real — #3 in particular would have red-gated CI just as it red-gated the local runner, since dnsmasq's rebind-protection check runs in the router VM regardless of host.)

PR #1371, opened 2026-06-04
fix(vm): make orphan-reaper pid-age probe tolerant of vanished sibling qemus (#1369)

### LINKED ISSUE #1369: vm: orphan-reaper TOCTOU race aborts router-up.sh silently (rc=1, empty output) under concurrency — total Gate 2 outage
## Symptom

Master Router CD Gate 2 intermittently goes **fully red on both arms**: every scenario test ERRORs at session setup with `RuntimeError: command failed (rc=1): scripts/vm/router-up.sh` and **completely empty stdout AND stderr**. Example: [run 26958383261](https://github.com/wifihaven/wifihaven/actions/runs/26958383261) (d8a4e84d). It's intermittent and concurrency-only — solo runs pass.

## Root cause (reproduced + traced on the KVM runner)

The #1225 orphan-reaper (`wh_reap_orphan_bridges` in `scripts/vm/lib.sh`) has a **TOCTOU race** that trips `set -euo pipefail`:

```sh
# reaper, ~line 237:
age="$(wh_proc_age_secs "${pid}")"
[[ "${age}" =~ ^[0-9]+$ ]] || continue   # guard intended to skip empty age

# helper, ~line 165:
wh_proc_age_secs() {
  local pid="$1"
  [[ "${pid}" =~ ^[0-9]+$ ]] || return 1
  ps -o etimes= -p "${pid}" 2>/dev/null | tr -d '[:space:]'
}
```

The reaper enumerates qemu PIDs on **sibling** bridges (via pgrep), then calls `wh_proc_age_secs` → `ps -o etimes= -p <pid>`. When a **concurrent run's qemu exits between the pgrep and the ps** (routine when a sibling arm finishes), `ps` exits non-zero. Under **`pipefail`** the helper's pipeline returns 1, so `wh_proc_age_secs` returns 1, so the command substitution `age="$(...)"` fails, and under **`set -e`** router-up.sh **aborts at line 237 — before the line-238 guard can run**. The abort is silent: `ps` stderr is `/dev/null`, `set -e` prints nothing, and the #1225 EXIT trap (`_release_reservation_on_fail`) is also silent. → rc=1, empty output, every test ERRORs (router fixture is session-scoped).

(The `|| return 1` for a non-numeric pid has the same set-e-abort exposure at the call site.)

### Reproduction (on the api.lan KVM runner)
Running `router-up.sh` **solo** with the exact failing image (`openwrt-wifihaven-apk-25.12.img`) → exit 0, boots fine. Running **3–4 in parallel** (distinct WH_RUN_ID/WH_PORT_BASE) reliably makes ONE fail with rc=1 and empty output — matching CI. `bash -x` trace of the failing instance ends exactly at:
```
++ wh_proc_age_secs 242181
++ ps -o etimes= -p 242181        # pid already gone → ps rc=1
++ tr -d '[:space:]'
+ age=                            # pipefail propagates rc=1 → set -e aborts here
```

This is why it looked like a regression from the CNAME train: those PRs added longer e2e suites → more concurrent VM churn → sibling qemus exit during scans more often → higher hit rate. The image/agent and router-up.sh boot path are fine.

## Fix (surgical)

Make age-probing tolerant of a vanished pid so a dead sibling can never abort the reaper:

- `wh_proc_age_secs`: swallow the failure and always return 0 with (possibly empty) output, e.g.
  ```sh
  wh_proc_age_secs() {
    local pid="$1" out
    [[ "${pid}" =~ ^[0-9]+$ ]] || return 0     # empty, not rc=1
    out="$(ps -o etimes= -p "${pid}" 2>/dev/null)" || out=""
    printf '%s' "${out}" | tr -d '[:space:]'
  }
  ```
  The caller's `[[ "${age}" =~ ^[0-9]+$ ]] || continue` then correctly skips the vanished pid.
- Audit EVERY command substitution in `wh_reap_orphan_bridges` / the pick path for the same `set -e`+`pipefail` exposure (any `x="$(fn_that_may_rc1)"` where `fn` probes a possibly-dead pid/tap). Make them `|| true`/empty-safe.

## Acceptance
- Concurrent `router-up.sh` invocations no longer fail when a sibling run's qemu exits mid-scan (repro: N parallel router-up.sh with a dying sibling → all succeed or fail only on genuine exhaustion).
- A unit/bats test: `wh_proc_age_secs <nonexistent-pid>` returns rc 0 + empty string; the reaper `continue`s past a vanished pid without aborting.
- `shellcheck` clean on `scripts/vm/lib.sh`.

## Related
- #1225 introduced the reaper (the regression source).
- Observability follow-up (separate): router-up.sh / the gate should surface the real failure (serial console / set -e abort line) instead of empty output — this class was undebuggable from CI logs and required runner access to root-cause.

Refs run 26958383261, #1225, #1286.

sameerparekh:
## Problem

Master Router CD **Gate 2** intermittently went **fully red on both arms** under concurrency: every scenario test ERRORed at session setup with `RuntimeError: command failed (rc=1): scripts/vm/router-up.sh` and **completely empty stdout AND stderr** (e.g. [run 26958383261](https://github.com/wifihaven/wifihaven/actions/runs/26958383261)). Solo runs always passed — it was concurrency-only.

## Root cause — a TOCTOU race in the #1225 orphan-reaper

`wh_reap_orphan_bridges` (in [`scripts/vm/lib.sh`](scripts/vm/lib.sh)) enumerates qemu pids on **sibling** bridges via `pgrep`, then probes each with `wh_proc_age_secs(pid)` → `ps -o etimes= -p <pid>`.

When a **concurrent run's qemu exits between the `pgrep` and the `ps`** (routine when a sibling arm finishes), `ps` exits non-zero. Under **`pipefail`** the helper's pipeline returned 1, so the command substitution `age="$(wh_proc_age_secs "${pid}")"` failed, and under **`set -e`** `router-up.sh` **aborted at the assignment — before the very next line's numeric guard** `[[ "${age}" =~ ^[0-9]+$ ]] || continue` could skip the vanished pid.

The abort was silent: `ps` stderr → `/dev/null`, `set -e` prints nothing, and the #1225 EXIT trap is silent too → **rc=1, empty output, every (session-scoped) fixture ERRORs**. The CNAME-train PRs didn't regress anything — they just added longer e2e suites → more concurrent VM churn → siblings exit mid-scan more often → higher hit rate.

```
++ wh_proc_age_secs 242181
++ ps -o etimes= -p 242181        # pid already gone → ps rc=1
++ tr -d '[:space:]'
+ age=                            # pipefail propagates rc=1 → set -e aborts here
```

## Fix (surgical)

Stop transient-pid / vanished-tap probes from aborting the reaper under `set -e`, **without** weakening the genuine pool-exhaustion `die` or the real age-reaping logic:

- **`wh_proc_age_secs`** — never returns non-zero. A non-numeric pid returns `0` (empty) instead of `rc 1`, and the `ps` failure is swallowed with `|| out=""` so `pipefail` can't propagate it. The caller's numeric guard then skips the vanished pid as designed.
- **Audited the rest of the pick/reap path** for the same `set -e`+`pipefail` exposure when a sibling vanishes mid-scan, and made each empty/`|| true`-safe:
  - `wh_bridge_reserved` — `awk` on a reservation file that can be cleared after the `-f` test.
  - `wh_pick_lan_bridge` — `ls .../brif | wc -l` (`brif` can vanish; `pipefail` propagated `ls`'s rc even though `wc` had already emitted the count) and the reservation-file `awk`.

`wh_qemu_pids_on_bridge` / `wh_any_qemu_pids_on_bridge` already end in `|| true`; the `headroom` `printf|sort|tail` and `wh_delete_orphan_taps`'s process-substitution `ls` are not exposed.

## Test

Extended [`scripts/vm/test-lan-bridge-pick.sh`](scripts/vm/test-lan-bridge-pick.sh) with two `flock`-free cases (so they exercise on dev laptops too, not just the KVM runner):

- **case 4** — `wh_proc_age_secs <vanished / non-numeric pid>` → rc `0` + empty string.
- **case 5** — `wh_reap_orphan_bridges` with a numeric-but-dead pid on a non-headroom bridge **completes (rc 0)** instead of aborting (the caller's guard skips it).

Both go **red** against the pre-fix `lib.sh` and **green** with the fix (verified locally by running the suite against `origin/main`'s `lib.sh`).

## Validation
- `shellcheck scripts/vm/lib.sh` — no new findings (the two pre-existing `SC1091`/`SC2012` info notes are unchanged from `main`).
- `bash -n` clean on both scripts.
- Test suite: `passed=6 failed=0`.

Gate 2 itself needs the self-hosted KVM runner and could not be run locally.

Closes #1369

### DIFF
--- a/scripts/vm/lib.sh
+++ b/scripts/vm/lib.sh
@@ -142,7 +142,11 @@ wh_bridge_reserved() {
   local f="${WH_BRIDGE_RESERVATION_DIR}/${br}.reservation"
   [[ -f "${f}" ]] || return 1
   local pid
-  pid="$(awk -F= '$1=="pid"{print $2; exit}' "${f}" 2>/dev/null)"
+  # `|| true`: the marker can vanish between the `-f` test above and this awk
+  # (a sibling run clears its own reservation mid-scan); awk then exits non-zero
+  # and, under `set -e`, would abort this assignment. Treat a vanished marker as
+  # "not reserved" (fall through to the rm/return 1 below).
+  pid="$(awk -F= '$1=="pid"{print $2; exit}' "${f}" 2>/dev/null || true)"
   if [[ -n "${pid}" && "${pid}" =~ ^[0-9]+$ ]] && kill -0 "${pid}" 2>/dev/null; then
     return 0
   fi
@@ -162,10 +166,19 @@ wh_bridge_reserved() {
 # legitimate run.
 
 # Wall-clock lifetime (whole seconds) of a pid, empty if not running.
+#
+# MUST NOT return non-zero: callers use `age="$(wh_proc_age_secs "${pid}")"`
+# under `set -euo pipefail`, so any non-zero rc — from a non-numeric pid, OR
+# (the #1369 TOCTOU) from the pid exiting between the pgrep and our `ps`, which
+# makes `ps` rc=1 and `pipefail` propagate it — aborts the caller at the
+# assignment, BEFORE its `[[ "${age}" =~ ^[0-9]+$ ]] || continue` guard can
+# skip the vanished pid. Always return 0 with (possibly empty) output and let
+# the guard do its job.
 wh_proc_age_secs() {
-  local pid="$1"
-  [[ "${pid}" =~ ^[0-9]+$ ]] || return 1
-  ps -o etimes= -p "${pid}" 2>/dev/null | tr -d '[:space:]'
+  local pid="$1" out
+  [[ "${pid}" =~ ^[0-9]+$ ]] || return 0
+  out="$(ps -o etimes= -p "${pid}" 2>/dev/null)" || out=""
+  printf '%s' "${out}" | tr -d '[:space:]'
 }
 
 # Longest wall-clock lifetime a *CI* VM can legitimately reach. The VM gates
@@ -322,10 +335,16 @@ wh_pick_lan_bridge() {
   local br
   while IFS= read -r br; do
     local attached
-    attached=$(ls "${WH_SYS_CLASS_NET}/${br}/brif" 2>/dev/null | wc -l)
+    # `|| true`: if `brif` vanishes mid-scan (a sibling tearing down its bridge),
+    # `ls` exits non-zero and `pipefail` makes the whole pipeline non-zero even
+    # though `wc -l` already emitted "0" — which, under `set -e`, would abort
+    # this assignment. `wc` has already written the count, so swallowing the rc
+    # leaves `attached` correct (empty/0 ⇒ treated as free).
+    attached=$(ls "${WH_SYS_CLASS_NET}/${br}/brif" 2>/dev/null | wc -l || true)
     local res="none"
     if [[ -f "${WH_BRIDGE_RESERVATION_DIR}/${br}.reservation" ]]; then
-      res="$(awk -F= '$1=="pid"{print $2; exit}' "${WH_BRIDGE_RESERVATION_DIR}/${br}.reservation" 2>/dev/null)"
+      # `|| true`: same vanish-after-test TOCTOU as wh_bridge_reserved's awk.
+      res="$(awk -F= '$1=="pid"{print $2; exit}' "${WH_BRIDGE_RESERVATION_DIR}/${br}.reservation" 2>/dev/null || true)"
     fi
     log "pick-debug ${br}: attached=${attached} reservation_pid=${res}"
     if (( attached == 0 )) && ! wh_bridge_reserved "${br}"; then
--- a/scripts/vm/test-lan-bridge-pick.sh
+++ b/scripts/vm/test-lan-bridge-pick.sh
@@ -1,15 +1,26 @@
 #!/usr/bin/env bash
-# Smoke test for scripts/vm/lib.sh::wh_pick_lan_bridge (#895).
+# Smoke test for scripts/vm/lib.sh::wh_pick_lan_bridge (#895) and the
+# orphan-reaper's TOCTOU tolerance (#1369).
 #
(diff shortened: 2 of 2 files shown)


sameerparekh:
## Follow-up: this PR also wires `scripts/vm/**` into Master Router CD

Noticed Master Router CD didn't run for this PR. Two reasons, and the second is a real gap I've now closed in this PR:

1. **By design at PR time.** [`master-router.yml`](.github/workflows/master-router.yml) is a **CD** pipeline — `on:` is `push` to `main` + `workflow_dispatch`, never `pull_request`. No PR triggers it. PR-time validation of this change came from **Image Builder** (`router-image-build.yml`, which *is* PR-gated on `scripts/vm/**`) and **Shell Tests** (ran the new case 4/5 — green).

2. **Gap, now fixed.** Master Router CD's `on.push.paths` listed `openwrt/**`, `shared/contract/**`, `scripts/e2e/**`, and the workflow files — **but not `scripts/vm/**`**, the very harness Gate 2 / Gate 3a boot the router image with. So even post-merge, a fix to the reaper wouldn't re-run the gates that depend on it. Added `scripts/vm/**` to the path scope (commit `369c3455`). The manual production-approval gate still sits before `publish-openwrt`, so a harness-only change runs the gates without auto-shipping a new `openwrt-latest`.

**On validating the fix via a live Gate 2 run:** the bug is *concurrency-only* (a sibling qemu must exit between the reaper's `pgrep` and its `ps`), so a single dispatched Gate 2 run would pass with or without the fix and wouldn't prove anything. The regression proof is the shell test (case 4/5), which is red against `origin/main`'s `lib.sh` and green here. Happy to `workflow_dispatch` Gate 2 on this branch as a boot smoke if you'd like, but it won't reproduce the race on its own.

Issue #1322, opened 2026-06-02
test: global policy layer + default-deny coverage (feature + router)


sameerparekh:
Part of the global policy layer design ([#1308](https://github.com/wifihaven/wifihaven/issues/1308)). See [`docs/design/global-policy-layer.md`](docs/design/global-policy-layer.md) §8 (item 7). Follows the PolicyService + router issues (TDD: tests land with the code adopting the schema, per the migration-isolation rule — not in the schema-only migration PR).

## Matrix
**Feature tests (`api/test/`):**
- `global.extraAllowed` carves out a `blocked=true` MAC (every block reason).
- A global block (`global.extraBlocked` / `global.blocklistIds` / `global.blocked`) beats a per-MAC `extraAllowed`, but `global.extraAllowed` beats the global block.
- A default-deny profile collapses to `blocked=true` + `blockReason=DefaultDeny`, `extraAllowed` = allow list, `extraBlocked`/`blocklistIds` omitted.
- Default-deny + `blockIpOnly`: only allowed-and-locally-resolved hosts reachable.
- Reason precedence: Paused > Schedule > TimeLimit > DefaultDeny.

**Router tests (`openwrt/test/`, OPNsense parity):**
- `@global_allow` host IP not dropped for a MAC in `@blocked_macs`.
- `@global_block` host IP dropped even when in a per-MAC `@ea_` set, but NOT dropped when in `@global_allow`.
- `blockIpOnly` composes (allowed/global-allowed hosts resolved into `resolved_<m>`).

Issue #1316, opened 2026-06-02
shared: add PolicySnapshot.global (BlockRules) + MacBlockReason.DefaultDeny


sameerparekh:
Part of the global policy layer design ([#1308](https://github.com/wifihaven/wifihaven/issues/1308)). See [`docs/design/global-policy-layer.md`](docs/design/global-policy-layer.md) §3 and §8 (item 1).

## Scope
`shared/src/Models.scala` only — wire shape, no router/SPA/PolicyService logic.

- Add `global: BlockRules` to `PolicySnapshot` (reuse the existing `BlockRules` shape; **no** bespoke struct, **no** wire `version` field — the snapshot `etag` already covers change detection).
- Add `case object DefaultDeny extends MacBlockReason` (+ `asString`/`parse`/codec/`BlockReason.toJson` arms). It is the lowest-precedence reason.
- Update `BlockRules.allowAll` usages / snapshot constructors and the ETag input (`PolicyService.computeEtag`) so `global` participates.

## Notes
- `global.extraAllowed` = always-reachable hosts (UI, block page, connectivity, PKI); `global.extraBlocked`/`global.blocklistIds` = global blocks; `global.blocked` = network lockdown; `global.blockIpOnly` = network-wide strict mode. Fields describe router behaviour, not rationale.
- Pre-v1.0: no compat shims; change all sides in tandem.

Issue #1105, opened 2026-05-27
bug: time_limited apps with exemptFromDaily=true must be in extraAllowed when budget remains, so they carve around @blocked_macs


sameerparekh:
## Problem

In [api/src/policy/PolicyService.scala](api/src/policy/PolicyService.scala), \`computeBlockRules\` builds the snapshot's \`extraAllowed\` from:

\`\`\`scala
extraAllowed = (p.extraAllowed ++ in.appExtraAllowed ++ uiAllowedHosts).distinct
\`\`\`

Only \`AppMode.Allowed\` apps contribute to \`appExtraAllowed\` ([:99-105](api/src/policy/PolicyService.scala:99)). \`AppMode.TimeLimited\` apps are turned into \`SiteTimeLimit\` rows ([:114-122](api/src/policy/PolicyService.scala:114)) and only ever contribute to \`extraBlocked\` — and only once their per-app budget is **exhausted** ([:516-520](api/src/policy/PolicyService.scala:516)).

The \`exemptFromDaily\` flag is preserved into the \`SiteTimeLimit\` but only affects the *tally* — \`totalMinutesUsed\` excludes exempt-flagged hosts ([:143-146](api/src/policy/PolicyService.scala:143)). It does **not** affect rule emission.

Net effect: when a profile hits its daily cap, the snapshot sets \`blocked=true\` → MAC goes into \`@blocked_macs\` → \`render.lua\` drops every packet from that MAC. An exempt-from-daily app with remaining budget has no host-level carve-out, so its hosts are dropped at the firewall along with everything else.

## Operator intent

\`exemptFromDaily=true\` is configured precisely so the kid can still use that app after the profile cap is hit (e.g. \"Khan Academy doesn't count toward your daily limit\" implies the daily limit also doesn't gate Khan). The current implementation only delivers half the contract.

## Repro (prod, 2026-05-26)

- Kids profile: \`dailyMinutes=30\`, Khan Academy app assignment with \`mode=time_limited\`, \`dailyMinutes=30\`, \`exemptFromDaily=true\`.
- Kid burns 30 min of non-Khan traffic by noon → profile cap exhausted → MAC in \`@blocked_macs\`.
- Kid tries Khan → packet dropped at firewall, no DNS, no event, no Khan access.
- Workaround: flip the app to \`mode=allowed\`.

## Fix

[...]

Issue #1069, opened 2026-05-26
feat: household-scoped reusable named schedules ('Bedtime', 'School hours', ...)


sameerparekh:
## Goal

Promote schedules to a **household-scoped named primitive** that anything time-bound can reference: profiles, app policy assignments, blocklist subscriptions, anything we add later. Defining "bedtime" once and reusing it everywhere is significantly better than re-editing 4 separate `{ days, from, to }` entries every time the operator wants to push bedtime back by 15 minutes.

## Today

Schedules live attached to profiles (the `profile_schedules` table or equivalent). They aren't shareable across profiles, much less across other primitives. Each app's `time_limited` mode has no schedule concept at all — it just caps daily minutes. Blocklists (#1067, just filed) want schedule-driven activation but have to refer to an existing concept.

### Schema

New `schedules` table — household-scoped, named:

```sql
CREATE TABLE schedules (
  id BIGSERIAL PRIMARY KEY,
  household_id BIGINT NOT NULL REFERENCES households(id),
  name TEXT NOT NULL,               -- "Bedtime", "School hours", "Afternoon"
  description TEXT,                  -- optional, for the operator's notes
  windows JSONB NOT NULL,            -- [{ days: ["mon", "tue", ...], from: "HH:MM", to: "HH:MM" }, ...]
  created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
  updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
  UNIQUE (household_id, name)
);
```

`windows` is a JSONB array so a single schedule can cover compound time blocks (e.g. "school hours: weekdays 8-15, Fri 8-12 early dismissal").

### References

Migrate the existing per-profile schedule into the new model:

- `profiles.schedule_id NULLABLE REFERENCES schedules(id)` — profile's active-window schedule (replaces today's inline arrangement).
- `app_policy_assignments.schedule_id NULLABLE` — apps can be schedule-gated (compose with `time_limited` budget).
- `blocklist_subscriptions.schedule_id NULLABLE` — for #1067.
- Future references: e.g. unmanaged-mac policy schedule, request-flow auto-deny windows, etc.

If `schedule_id IS NULL` → "always" (the existing semantics).

### Seed schedules

Ship a small starter set with the API on a fresh household:

- **Bedtime** — daily 20:00–07:00 (operator can adjust).
- **School hours** — weekdays 08:00–15:00.
[...]

PR #1377, opened 2026-06-04
docs: design for per-app schedules (#1376)

### LINKED ISSUE #1376: Epic: per-app schedules (allowed/blocked time windows per app) — API-logic only, no snapshot change
## Epic

Add **per-app schedules** so individual apps can carry their own time window on top of the profile's general schedule:
- Some apps should remain **allowed during general scheduled downtime** (e.g. an educational app reachable during bedtime), while others are not.
- An individual app can have its own schedule expressing either **"allowed during this time"** or **"blocked during this time"**.

This extends the existing **schedule abstraction** to the per-app layer (today schedules are profile-wide downtime windows).

## Hard constraint — API logic only, NO snapshot/wire change

This is a **`PolicyService`-only** feature. It MUST collapse into the **existing** snapshot `BlockRules` fields at snapshot-compute time — **no new wire fields, no router change.** The router stays a dumb applier (AGENTS.md invariant). Concretely, the planner should confirm the mapping:
- App **"allowed during window W"**: while the window is active AND the MAC is otherwise blocked (profile downtime / schedule), the app's hosts go into **`extraAllowed`** (which beats every block path per #421). Outside W, the app is not carved out → blocked with everything else.
- App **"blocked during window W"**: while W is active, the app's hosts go into **`extraBlocked`**. Outside W, normal.
- Everything resolves server-side into `blocked` / `extraBlocked` / `extraAllowed` exactly as today. The enforcement pipeline never learns about schedules.

## What the planning session must produce
A design doc (`docs/design/per-app-schedules.md`) + an architecture.md note + a list of implementation sub-issues (to file later, not now). It must cover:

1. **Data model.** How an app assignment gains a schedule + mode (allowed-during / blocked-during). Reuse the existing `Schedule` representation (overnight wrapping, day-of-week, household-local TZ via injected `Clock` — never `java.time` directly). Relate to #1069 (household-scoped reusable named schedules) — per-app schedules should be able to reference a named schedule, not only inline windows. DB migration shape (new table or columns on the app-assignment).
2. **Precedence / interaction.** Define exactly how a per-app schedule composes with: the profile's general schedule (the headline case — app allowed during downtime), daily time limits (#1105 exempt apps), per-app modes (Allowed / Blocked / time_limited), manual block, default-deny (#1316–#1322), and category blocklists. Produce a precedence table. Resolve the case where an app is "allowed during W" but the profile *also* hit its daily time limit — does the app stay reachable? (Decide and justify.)
3. **Snapshot-freshness implications.** Per-app schedules add more time-boundary transitions at which the per-MAC snapshot's `extraAllowed`/`extraBlocked` must change. Document how/when the snapshot is recomputed so the router polls a current decision at each app-window boundary (same mechanism profile schedules already rely on — confirm it scales to per-app granularity; flag if the boundary set gets large).
4. **SPA surface.** How the per-app schedule is configured in the app/profile editor (out of scope to build here, but specify the shape so a web sub-issue can pick it up). Note autosave preference (#995/#423).
5. **Sub-issue breakdown:** DB migration, PolicyService eval, tests (feature + schedule-boundary unit), web UI — list them; don't file yet.

## Constraints
- Honor AGENTS.md invariants: server-side decision only, per-MAC enforcement, router dumb applier, `extraAllowed` beats every block (#421), Clock injected.
- TDD when implementation lands: schedule-boundary unit tests (exact on/off, overnight wrap, DST) + feature tests proving the snapshot carves out / blocks the app's hosts at the right times.

Related: #1069 (reusable named schedules), #1067 (schedule-driven blocklist activation), #937 (global/household always-on rules), #1316–#1322 (global policy + default-deny — per-app schedules must compose with default-deny).

sameerparekh:
## Summary

Design + sub-issue breakdown for the **per-app schedules** epic (#1376) — an individual app can carry its own "allowed during W" / "blocked during W" time window on top of the profile's general schedule (headline case: an educational app stays reachable during bedtime). **Docs only** — no implementation, no sub-issues filed.

Adds:
- `docs/design/per-app-schedules.md` — the full design.
- A note in `docs/architecture.md` under the "Architectural model" / freshness section, consistent with the existing dumb-applier invariant.

## The hard constraint holds (validated)

This is a **`PolicyService`-only** feature. **No snapshot/wire field, no router change.** At snapshot-compute time the window collapses into the *existing* per-MAC `BlockRules`:
- `allowed_during W` active → app hosts in **`extraAllowed`** (beats every whole-MAC block per #421).
- `blocked_during W` active → app hosts in **`extraBlocked`**.
- Outside the window, the app falls back to its base mode, exactly as today.

The router never learns schedules exist. Confirmed there is **no case that forces a wire change** — the doc documents this explicitly rather than proposing a new field.

## What the doc covers

- **Data model** — reuses the existing `Schedule` repr (overnight wrap, DOW, household-local TZ via injected `Clock`, never `java.time`); supports both inline windows and #1069 named-schedule references; new `app_policy_schedules` child table of `app_policy_assignments`.
- **DB migration shape** — schema-only PR per the migration-isolation rule; bounded (non-growth) table, safe on the startup path.
- **Precedence/interaction table** — profile schedule, daily time limits (incl. #1105 exempt apps), per-app modes, manual block, default-deny (#1316–#1322), category blocklists. Per-app windows resolve to a single *effective disposition* (window overrides base mode) before bucketing.
- **Resolved decision (row 9):** an app "allowed during W" **stays reachable even when the profile hit its daily time limit** — justified as both the wire-forced outcome and the defensible default (with the `time_limited`-mode escape hatch for the cap-respecting case).
- **Snapshot-freshness** — freshness is pull-based and on-demand; per-app schedules need no new mechanism. Flags that the boundary set only matters for a *future* server-push channel.
- **SPA surface** — assignment-editor Schedules section, autosave (#995/#423); specified, not built.
- **Sub-issue breakdown** — DB migration, PolicyService eval, schedule-boundary unit + feature tests, web UI (listed, not filed).

Closes #1376 once the listed sub-issues ship (this PR is design only).

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -453,10 +453,11 @@ the profile's general schedule. This is a `PolicyService`-only feature with
 **no wire or router change**: at snapshot-compute time the window is evaluated
 (reusing `scheduleActiveAt`, household-local zone via the injected `Clock`) and
 collapsed into the *existing* fields — `allowed_during` while active → the app's
-hosts in `extraAllowed` (which beats every whole-MAC block per #421, including
-TimeLimit), `blocked_during` while active → `extraBlocked`. The agent never
-learns that per-app schedules exist; it still sees only the per-MAC
-`BlockRules`. Freshness rides the same pull-based path as profile schedules —
+hosts in `extraAllowed` (which beats Schedule/Paused/Manual whole-MAC blocks per
+#421; whether it also beats the *daily time limit* is a server-side choice keyed
+on the app's `exemptFromDaily` flag, not a router behaviour), `blocked_during`
+while active → `extraBlocked`. The agent never learns that per-app schedules
+exist; it still sees only the per-MAC `BlockRules`. Freshness rides the same pull-based path as profile schedules —
 the next poll after a window edge recomputes against the current instant
 (worst-case one poll interval). A precomputed boundary scheduler is needed only
 if the future server-push channel above is built. Full design:
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -446,6 +446,22 @@ The architectural invariant: **every new policy concept lands in
 `BlockRules`**. Adding decision logic to the agent (schedule evaluation,
 time accounting, category lookup, role-based defaults) is a bug.
 
+**Per-app schedules ([#1376](https://github.com/wifihaven/wifihaven/issues/1376))
+are an instance of this invariant, not an exception to it.** An app can carry
+its own time window meaning "allowed during W" or "blocked during W" on top of
+the profile's general schedule. This is a `PolicyService`-only feature with
+**no wire or router change**: at snapshot-compute time the window is evaluated
+(reusing `scheduleActiveAt`, household-local zone via the injected `Clock`) and
+collapsed into the *existing* fields — `allowed_during` while active → the app's
+hosts in `extraAllowed` (which beats every whole-MAC block per #421, including
+TimeLimit), `blocked_during` while active → `extraBlocked`. The agent never
+learns that per-app schedules exist; it still sees only the per-MAC
+`BlockRules`. Freshness rides the same pull-based path as profile schedules —
+the next poll after a window edge recomputes against the current instant
+(worst-case one poll interval). A precomputed boundary scheduler is needed only
+if the future server-push channel above is built. Full design:
+[`docs/design/per-app-schedules.md`](design/per-app-schedules.md).
+
 ## 6. Router HTTP API
 
 All endpoints below are served by the existing API process under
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Sub-issues filed under the epic (Schedules)

Updated the design to **land named schedules (#1069) first** — per-app schedules now reference the #1069 `schedules` table directly (no parallel inline-window table; a one-off is just an ad-hoc named schedule). The only addition over #1069 is the *mode* + multi-rule dimension via a small `app_policy_schedule_rules` child table.

Filed the implementation pieces as native sub-issues of #1376, Epic = `Schedules`, all **Blocked on #1069**:

1. #1378 — DB migration for `app_policy_schedule_rules` (schema-only PR)
2. #1379 — PolicyService eval + models/repo + tests (TDD)
3. #1380 — web UI — assignment schedule rules editor

Dependency order: #1069 → #1378 → #1379 → #1380.

Issue #1363, opened 2026-06-04
Agent's blocklists.fetch_and_cache fails to write cache when /etc/wifihaven/blocklists/ is missing


sameerparekh:
## Summary

After #1360 unblocked the bl_ fetch (Bearer auth), the agent's blocklist
fetch now reaches a downstream write that fails when the cache directory
doesn't exist on disk. Surfaced during PR #1361 validation in the VM e2e
fake-mode run (both ipk-23.05 and apk-25.12 arms):

```
user.warn wifihaven: blocklist fetch: blocklists: write failed for e2e-cname-bl:
   /etc/wifihaven/blocklists/e2e-cname-bl-2026.06.01.txt.tmp:
   No such file or directory
```

Sequence on the e2e router image:

1. Agent's `GET /api/blocklists/e2e-cname-bl` returns **200** (auth fix
   from #1360 works).
2. `blocklists.fetch_and_cache` tries to write the body to
   `/etc/wifihaven/blocklists/<id>-<version>.txt.tmp` and rename into place.
3. Open-for-write fails with ENOENT because the directory doesn't exist
   on this image — the package's install-time mkdir didn't run, or the
   path isn't pre-created.

Net effect: even when the fetch succeeds, the hosts never reach disk →
`load_cached` returns empty → the bl_ ipset stays empty → category
enforcement silently no-ops (same end-state as the #1360 401, different
cause).

## Why this is now visible

Before #1360 the fetch 401'd before reaching the write path, so this
latent bug was masked. The fix unblocked the fetch and exposed it. It is
not a #1361 regression — it predates it — but #1361 makes it the new
gating failure on any image that doesn't have the directory pre-created.

## Repro

Build the router image from main, boot the qemu router VM, register an
e2e blocklist on the fake API, and watch the agent's logread. The 200
fetch is followed immediately by the `write failed: ... .tmp: No such
file or directory` warn line, and `bl_<id>` stays empty.

## Suggested fix

In `openwrt/files/usr/lib/lua/wifihaven/blocklists.lua`,
`fetch_and_cache` should ensure `cache_dir` exists before writing:

```lua
-- Best-effort mkdir -p on the cache dir; ENOENT after this is a real
-- write error, not a missing-parent error.
os.execute(string.format("mkdir -p %q", cache_dir))
```

Or — preferred — thread a `mkdir_fn` through the injected `fs` table so
the unit tests can mock it, and call it once at the top of
[...]

PR #1421, opened 2026-06-05
fix(openwrt): populate category blocklists at startup + periodic timer (#1412)

### LINKED ISSUE #1412: bug(prod): blocklist still not enforced — taboola.com reachable despite being on profile's 'ads' list
## Report
On **prod**, the operator's profile has the **'ads' blocklist** assigned, yet **taboola.com is reachable from the laptop** on that profile. Category-blocklist enforcement is still not working from the device's perspective.

This is a REGRESSION / continuation after #1334 (closed, fixed in PR #1335 - fetch_and_cache URL resolution) and #1348 (CNAME-target bypass, PR #1350). Filing fresh per the no-reopen rule.

## Most likely causes (check in order - prod-first)
1. **Deployed agent is stale.** The #1335/#1350 fixes may be merged but NOT yet running on the operator's prod router (routers have stuck before - `memory/router_update_deadlock_recovery.md`). Check the agent version actually running vs the version containing the fixes. PRIME SUSPECT given #1334 was just 'fixed'.
2. **bl_ ipset empty / not populated.** Even on a current agent: is the 'ads' blocklist fetched and the bl_ set populated on the router? (`nft list set inet wifihaven bl_<...>`). #705/#709 (fetch status logging + timer refresh) are open and could leave it empty/stale.
3. **taboola.com not in the 'ads' list source**, or only some taboola hosts are. Confirm taboola.com (apex + the hosts the laptop actually hit) are members of the configured 'ads' blocklist.
4. **CDN / anycast IP variance** - taboola is ad-tech served via CDNs; the #1348 class of bypass may still let some flows through. Check whether the laptop reached taboola via an IP not in bl_.
5. **Snapshot doesn't carry the blocklistId** for this profile/MAC (PolicyService mapping).

## Division of labor
Claude drives API/prod inspection (prod creds out-of-band, never paste). Operator runs router/device-side: agent version, `nft list ruleset`, `nft list set ... bl_*`, curl to taboola from the laptop.

## Reminder
Blocking is connection-layer, not DNS - taboola resolving in DNS is NOT evidence it's unblocked (`memory/blocking_is_traffic_layer_not_dns.md`).

## TDD / acceptance
- Leave a regression test proving a blocklist member host is dropped for an assigned MAC (router enforcement test if router-side; PolicyService snapshot test if server-side).
- Acceptance: taboola.com (and other 'ads' members) dropped at the connection layer for the assigned profile; a control host stays reachable.
- If the cause is purely 'agent not deployed', say so and drive the deploy (wifihaven-update) rather than a code change - but add visibility so a stale enforcement agent is detectable (relates to #1033).

Refs: #1334 (prior fix, PR #1335), #1348 (CNAME bypass, PR #1350), #705 #709, #958, #1325.

sameerparekh:
Fixes #1412.

## Live root cause (router-side, reproduced)

Server-side is fully correct — verified against **prod** (`api.wifihaven.net`):

- The operator's profile ("Sameer") snapshot ships `blocklistIds: [malware, ads, ads-extended]` for the laptop MAC.
- `snapshot.blocklists.ads = {version: e60fd5ae…, url: /api/blocklists/ads}`; the router-facing fetch returns the body with **taboola.com** present (36/36 hosts).
- The deployed agent reports **v0.3.0**, which already contains #1334 / #1350 / #1360 / #1363. Stale-agent ruled out.

The failure is in the agent's fetch orchestration, reproduced on the test router (also v0.3.0):

- `/etc/wifihaven/blocklists/` **did not exist**, the dnsmasq fragment had **0 `bl_` nftset directives** (50 `ea_`, 0 `bl_`), and the last blocklist-fetch attempt was **hours stale** despite a 5 s poll.
- The agent ran `fetch_and_cache` **only inside the policy-poll HTTP-200 branch, and not at startup at all.** In steady state the poll returns 304, so the only thing that ever triggered a blocklist fetch was an admin policy edit. A household with a stable snapshot therefore **never populated the `bl_` sets** → render emitted no `nftset=/<member>/...#bl_<id>` directives → category enforcement was a silent no-op. Any transient first-fetch failure (a 401 during re-enrollment, a network blip) had the same permanent effect.

> Per the architecture model, this is connection-layer: taboola resolving in DNS is not evidence it's unblocked. The bug is that the per-`(member host)` `bl_` ipset is never populated, so the nftables forward-drop has nothing to match.

## Fix

- **Startup fetch.** Fetch + cache blocklists and attach `_blocklist_hosts` **before** the initial apply, so the boot ruleset already carries the `bl_` populators.
- **Periodic refresh timer** (`blocklist_refresh_interval`, default 1 h) against the last good snapshot, **independent of 200/304**, re-applying only when `blocklists.hosts_differ` reports the loaded host set actually changed. This self-heals a failed boot fetch and matches the "periodic cadence" the architecture already documents. `fetch_and_cache` is a cheap cached read when the `(id,version)` file already exists, so steady state does no real work and no dnsmasq churn.
- Both paths route through one `refresh_blocklists` helper that folds in the `blocklist_fetch_failures_total` metric + error logging.

## Tests (TDD, red→green)

- `test(openwrt)` commit adds failing unit tests for the two new seams.
- `fix(openwrt)` commit implements `blocklists.refresh` + `blocklists.hosts_differ`.
- `openwrt/test/blocklists_spec.lua` (27) and `render_spec.lua` (147) pass. `render_spec` already proves a `bl_` member yields both the `nftset=` populator and the `ether saddr <mac> ip daddr @bl_<id> … drop` rule for an assigned MAC; these commits close the orchestration gap that left the set empty.

## Wire contract

No wire change. `blocklist_refresh_interval` is a UCI key read and written by the same agent build (not part of the router↔API contract). No DB migration, no Scala.

## Acceptance

After this ships in the next agent release and the prod router updates, a freshly-booted router populates `bl_ads` with taboola's resolved IPs from boot, and the forward-drop matches — taboola.com is dropped for the assigned profile while a non-member control host stays reachable.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
@@ -229,6 +229,76 @@ function M.load_cached(snapshot, fs, cache_dir)
   return result
 end
 
+-- refresh(snapshot, http_get_fn, fs, cache_dir, base_url, auth_token)
+--   One-call fetch_and_cache → gc → load_cached. Returns:
+--     { hosts = {[id]={host,...}}, failures = {...}, errors = {...} }
+--
+-- This is the entry point the agent calls BOTH at startup and on the periodic
+-- blocklist timer, so the per-blocklist cache (and therefore the bl_ ipset
+-- nftset= directives render emits from snapshot._blocklist_hosts) is populated
+-- independent of whether the policy poll returned 200 or 304.
+--
+-- #1412: previously the agent only fetched/cached blocklists inside the
+-- policy-poll HTTP-200 branch, and not at startup at all. A household whose
+-- snapshot is stable (the steady state — every poll is a 304) therefore never
+-- populated the cache: render emitted zero `nftset=/<member>/...#bl_<id>`
+-- directives, the bl_ sets stayed empty, and an assigned list member (e.g.
+-- taboola.com ∈ 'ads') was reachable despite correct policy. Surfacing the
+-- failures back to the caller lets the agent emit blocklist_fetch_failures_total
+-- and simply retry on the next cadence rather than no-op until the next snapshot
+-- change (which on a stable household may be hours/days away).
+function M.refresh(snapshot, http_get_fn, fs, cache_dir, base_url, auth_token)
+  local fc    = M.fetch_and_cache(snapshot, http_get_fn, fs, cache_dir, base_url, auth_token)
+  M.gc(snapshot, fs, cache_dir)
+  local hosts = M.load_cached(snapshot, fs, cache_dir)
+  return { hosts = hosts, failures = fc.failures, errors = fc.errors }
+end
+
+-- hosts_differ(a, b) → boolean
+--   Order-insensitive comparison of two {[id]={host,...}} maps. Nil is treated
+--   as the empty map. Returns true iff the set of ids differs, or any id's host
+--   membership differs.
+--
+-- The agent calls this after a periodic refresh to decide whether the cached
+-- host set actually changed before re-applying. A blocklist host change alters
+-- the dnsmasq fragment (the nftset= directives) and so forces a dnsmasq reload;
+-- gating the re-apply on a real change keeps a steady state from churning
+-- dnsmasq every cadence while still catching the empty→populated cold-start
+-- transition (#1412).
+function M.hosts_differ(a, b)
+  a = a or {}
+  b = b or {}
+
+  local function host_set(list)
+    local set, n = {}, 0
+    for _, h in ipairs(list or {}) do
+      if not set[h] then
+        set[h] = true
+        n = n + 1
+      end
+    end
+    return set, n
+  end
+
+  -- Every id in `a` must exist in `b` with identical host membership.
+  -- Iterating both directions catches ids present in only one side.
+  local function subset_match(x, y)
+    for id, xlist in pairs(x) do
+      local xset, xn = host_set(xlist)
+      local yset, yn = host_set(y[id])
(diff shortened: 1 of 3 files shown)


sameerparekh:
**Converting to draft — this is based on the wrong root cause and would regress prod. Do not merge.**

Read-only inspection of the prod router (`RobotSpaceship`) shows the real cause is **the agent OOM-killing on the StevenBlack "extended" lists**, not a fetch/startup-gating gap:

- The blocklist cache is fully populated (`ads-…txt` contains taboola.com); the fetch works.
- The Kids profile assigns ads-extended (83k) + adult-extended (76k) + social-extended (3.2k) = **164,149 hosts**. The agent loads all of them into `_blocklist_hosts` and renders ~164k `nftset=` directives in one config string → **~700 MB RSS on a 1 GB router → `Out of memory: Killed process … (wifihaven-agent)`**. The apply never completes, so `bl_member_index` is 0 bytes, 0 `bl_` directives are live, `bl_ads` is empty, and the ~hourly "smoke check failed / stale config" restart is the OOM/respawn loop.

This PR's changes (fetch + attach `_blocklist_hosts` at startup; periodic re-render) would trigger the same 164k-host OOM **at boot and every cadence** — strictly worse. The real fix is to bound/curate large-list enforcement so the agent doesn't load/render 160k+ per-domain directives. Tracking that separately.

sameerparekh:
Closing in favor of #1434, which adds the render byte-cap first (the load-bearing OOM fix) and carries these helpers (refresh/hosts_differ) + startup/periodic population safely on top. Standalone, these changes would have triggered the 164k-host OOM at boot — see #1412.

Issue #1418, opened 2026-06-05
feat: profile-level 'hard pause' that blocks even allowlisted/global hosts


sameerparekh:
## Request

Add a profile-level **"hard pause"** distinct from the existing pause.

- **Today's pause** (soft): blocks all forwarded traffic for the profile's
  MACs *except* `extraAllowed` hosts. Per #421/#1413, an explicitly-allowed app
  (and the global infra allowlist, #1307) stays reachable through a pause. This
  is the intended behavior for "pause everything but the school app."
- **Hard pause** (new): block **everything**, including `extraAllowed` and the
  globally-allowed infra hosts. A true "off switch" — no carve-outs.

## Why

Soft pause is the right default ("dinner time, but homework app still works").
But sometimes the operator wants an unconditional cut — discipline, a lost/stolen
device, an emergency — where even allowlisted apps must go dark.

## Suggested shape (no new wire field)

Per the wire-shape rules in `AGENTS.md` ("express new policy via existing
functional fields — never add a field for the concept itself"), hard pause does
**not** need a new snapshot field or a new `MacBlockReason` semantics change for
enforcement:

- Hard pause = `blocked = true` **with `extraAllowed` emptied** for that MAC in
  the snapshot. The router's `@blocked_macs` carve-out is driven purely by the
  presence of `extraAllowed`; ship an empty list and the MAC drops into the
  family-agnostic unconditional-drop set with no `ip daddr != @ea_...` exception.
  Pure functional data — the router stays oblivious to "hard vs soft."
- All the policy lives server-side in `PolicyService.computeBlockRules`: when the
  profile is hard-paused, skip the `appExtraAllowed ++ … ++ infraAllowHosts`
  union (or override it to `Nil`). `blockReason` can stay `Paused` for block-page
  copy, or gain a distinct variant *only* if the block page needs different text.
- DB/UI: a per-profile `pause_mode` (`soft` | `hard`) or a boolean
  `hard_paused`. DB schema change ships in its own migration PR per the
  migration-isolation rule; the code adopting it follows.

Open questions to settle in design:
- Does hard pause still spare the **block page / UI hosts** (`uiAllowedHosts`)?
  Probably yes, so the kid sees *why* they're cut off and the admin SPA still
  loads on the household LAN — but that's a deliberate call to confirm.
- Interaction with `blockIpOnly` (already not carved by `ea_`).
- UI affordance: a second toggle vs. a long-press/confirm on the existing pause.

[...]

Issue #1307, opened 2026-06-01
Prod: allowed app (Math Academy) blocked when Kids profile hit daily time limit — extraAllowed not winning


sameerparekh:
## Symptom (prod)

On the **Kids** profile, **Math Academy** is listed as an app under **Allow**. When the profile's daily **time limit ran out**, Math Academy got **blocked** anyway. Expected: an allowed app stays reachable even when the whole profile is blocked by a time limit — `extraAllowed` beats `blocked` (#421, `feedback_extraallowed_beats_blocked`).

## What is already known to be correct (don't re-litigate)

- `PolicyService.computeBlockRules` (`api/src/policy/PolicyService.scala:549`) sets `extraAllowed = (appExtraAllowed ++ appExemptAllowedHosts ++ uiAllowedHosts).distinct` **unconditionally** — it does NOT drop allowed-app hosts when `blocked=true` (time limit). `appExtraAllowed` is built from `AppMode.Allowed` assignments expanded to hosts (PolicyService.scala:118-125, 143).
- The router's `render.lua` (L21-32) documents that `extraAllowed` (`@ea_<m>_<a>`) carves out of the `@blocked_macs` / `eb_` / `bl_` drops — so a time-limit MAC block should still let allowed-host IPs through.

So given those, the snapshot *should* keep Math Academy reachable. The bug is in one of the gaps below — investigate in order:

## Candidate root causes (investigate against PROD state)

1. **App host-list completeness.** Math Academy's real traffic may hit hostnames/CDNs/API domains NOT in `appRepo.getHosts(mathAcademyAppId)`. If a destination IP isn't attributed to an allowed hostname, the `@ea_` carve-out can't match it. Check the actual host list for the Math Academy app row in prod and compare to where Math Academy actually connects.
2. **`blockIpOnly`.** If the Kids profile (or device) has `blockIpOnly=true`, IP-direct / DoH / hardcoded-IP traffic to Math Academy is dropped with **no allowlist carve-out by design** (architecture: blockIpOnly has no extraAllowed exemption). Confirm whether blockIpOnly is set on the Kids profile.
[...]

PR #1424, opened 2026-06-05
db: migration for profiles.pause_mode — hard-pause foundation (#1418)

sameerparekh:
## Part of #1418 (schema only; code in follow-up)

Schema foundation for a profile-level **hard pause** — a true off-switch that, unlike today's soft pause, blocks even the otherwise-allowlisted hosts.

### What this adds

A single per-profile column on `profiles`:

```sql
ALTER TABLE profiles
  ADD COLUMN pause_mode TEXT NOT NULL DEFAULT 'soft'
    CHECK (pause_mode IN ('soft', 'hard'));
```

`'soft'` (the default) is today's behavior: a paused profile drops all forwarded traffic **except** its `extraAllowed` hosts, so an allowed-mode app + the #1307 global infra allowlist stay reachable through the pause (#421/#1413). `'hard'` will be the opposite — no carve-outs.

### Why a `text` enum column (not `hard_paused boolean`)

Mirrors the existing enum-as-text columns on this table (`failure_mode`, `cross_device_overlap_mode`): self-documenting, CHECK-constrained, and extensible if pause ever grows a third mode. Functionally equivalent to a boolean for now.

### Design choices being settled (implemented in the follow-up)

- **No new wire/snapshot field.** Per the wire-shape rules in `CLAUDE.md` ("express new policy via existing functional fields — never add a field for the concept itself"), hard pause is expressed *functionally*: when a profile is hard-paused, `PolicyService.computeBlockRules` ships `blocked = true` with `extraAllowed` emptied of the app/exempt/infra hosts for that MAC. The router's `@blocked_macs` carve-out is driven purely by the presence of `extraAllowed`, so an empty list ⇒ unconditional drop with no `ip daddr != @ea_…` exception. The router stays oblivious to "hard vs soft".
- **Block page / admin UI is spared.** Hard pause keeps the deployment's `uiAllowedHosts` so the kid still sees *why* they're cut off and the admin SPA still loads on the LAN — only the app/infra allowlists go dark. `blockReason` stays `Paused` (block-page copy only).

### Why this is its own PR

Per the migration-isolation rule (`CLAUDE.md` §migrations-back-compat), a PR that adds a Flyway migration may contain **only** the migration + docs. This PR is exactly that (the `V49` SQL plus an `docs/architecture.md` paragraph). The `PolicyService` / `ProfileRepo` / route / SPA code that reads and writes the column lands in the **stacked follow-up PR**, which is gated on this one merging and deploying first.

### Backwards compatibility

- `NOT NULL DEFAULT 'soft'` ⇒ image-(N-1), which never names `pause_mode` in any INSERT/UPDATE, keeps working unchanged; existing and newly-written rows are `'soft'` (today's behavior).
- `profiles` is a small metadata table (not an unbounded-growth event table), so the constant-default `ADD COLUMN` is metadata-only in PG 11+ — no table rewrite, fast on the Flyway startup path at prod volume.

### Test plan

The existing `api.test` feature suite is the back-compat gate. `TestDatabase` applies **every** migration on the classpath — including `V49` — before any test runs, and the fixtures exercise image-(N-1)'s production code paths in `api/src/**`. Green ⇒ the new schema is backward-compatible with the currently-deployed image. No source/test changes ship here (they'd bypass the gate), so the column is inert until the follow-up.

### DIFF
--- a/api/resources/db/migration/V49__add_pause_mode.sql
+++ b/api/resources/db/migration/V49__add_pause_mode.sql
@@ -1,0 +1,38 @@
+-- V49__add_pause_mode.sql
+-- #1418: per-profile "hard pause" — a true off-switch that blocks even the
+-- otherwise-allowlisted hosts (an allowed-mode app's hosts, the #1105 exempt
+-- carve-out, and the #1307 global infra allowlist).
+--
+-- Today's pause is "soft": a paused profile drops all forwarded traffic EXCEPT
+-- its `extraAllowed` hosts, so an allowed app + the infra allowlist stay
+-- reachable through a pause (#421/#1413). "Hard pause" is the opposite — no
+-- carve-outs.
+--
+-- No new wire field: hard pause is expressed *functionally* in the snapshot per
+-- the wire-shape rules in CLAUDE.md ("express new policy via existing functional
+-- fields — never add a field for the concept itself"). When a profile is
+-- hard-paused, PolicyService.computeBlockRules ships `blocked = true` with
+-- `extraAllowed` emptied (only the block-page / admin-UI hosts survive) for that
+-- MAC. The router's @blocked_macs carve-out is driven purely by the presence of
+-- `extraAllowed`, so an empty list ⇒ the MAC drops unconditionally with no
+-- `ip daddr != @ea_…` exception. The router stays oblivious to "hard vs soft".
+--
+-- This is a SCHEMA-ONLY migration (per the migration-isolation rule in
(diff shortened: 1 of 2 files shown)


sameerparekh:
Follow-up code PR (stacked on this branch): wifihaven/wifihaven#1426 — adopts the new `pause_mode` column in PolicyService / ProfileRepo / route / SPA. It must merge **after** this schema PR is merged and deployed; it will be retargeted from this branch to `main` once this lands.

Issue #1406, opened 2026-06-04
Alerting: migrate infra/grafana to an HCP Terraform remote backend (prereq)


sameerparekh:
Prerequisite for all Grafana alerting work (design: `docs/design/alerting.md` §3.1; parent #1381).

`infra/grafana` is **stateless by design** today — the CD job runs on an ephemeral runner with no persisted state, and each `grafana_dashboard` is upserted by stable `uid` with `overwrite = true`. That trick works only for dashboards. Alerting resources (`grafana_folder`, `grafana_contact_point`, `grafana_notification_policy`, `grafana_rule_group`) are stateful: a *create* against fresh-every-run empty state either 409s on a fixed identifier or accumulates duplicates, and the notification policy is a singleton root tree with no upsert-by-uid. So Terraform must own its prior state.

## Scope
- Migrate `infra/grafana` onto a **remote backend on HCP Terraform**, mirroring what #1357 did for `infra/cloudflare` (org `wifihaven`, Local execution mode, CI applies on merge to main).
- Wire `master-grafana.yml`'s `init` to the backend (the `paths:` filter already covers `infra/grafana/**`).
- Confirm the existing dashboards reconcile cleanly under managed state.

## Validation
- A second `terraform apply` is a no-op (no duplicate-dashboard churn).

## Blocks
Contact points/routing, the critical rule set, and the warning rule set all depend on this.

Refs: #1381 · #1357 · #1368

Issue #1381, opened 2026-06-04
Design the alerting strategy (rules, routing, on-call) for WifiHaven


sameerparekh:
## Why

WifiHaven has **dashboards but no alerting at all**. `infra/grafana/` provisions only `grafana_dashboard` resources — no `grafana_rule_group`, no contact points, no notification policy. A failure-mode metric that only feeds an unwatched panel is not observability: it costs cardinality but never reaches an operator's eyes until an incident.

This was the lesson of #1365 / #1368: 100% of router-metrics ingests failed for a long window (`router_metrics_batches_total{status="malformed"}` climbing, zero `status="ok"`) and nobody noticed until the empty router-fleet dashboard was hand-investigated. The dashboard side of that (an at-a-glance success-ratio stat panel) shipped in #1373; the **alerting** side was intentionally pulled out to be designed properly here rather than bolted on one metric at a time.

The "new functionality ships with metrics + a dashboard" rule (AGENTS.md) should arguably extend to **"failure-mode metrics ship with an alert"** — but only once we have a coherent strategy for *where alerts live, how they're routed, and who they page*.

## Scope — design, not (yet) implementation

Produce a short design doc (e.g. `docs/design/alerting.md`) + a tracked rollout, covering:

1. **Tooling / where alerts live.** Grafana-managed alert rules (`grafana_rule_group`) declaratively in `infra/grafana/` is the obvious default (keeps "declarative config over dashboard toggles"). Decide and document. Note: this almost certainly requires moving `infra/grafana` off its current **stateless** local-state design to a **remote backend** (HCP Terraform, like `infra/cloudflare` did in #1357) — a `grafana_folder` / `grafana_contact_point` create is not idempotent against a fresh-every-run empty state. Capture that as an explicit prerequisite.
2. **Routing & contact points.** What channel(s) do alerts page — email, Slack, PagerDuty? Single contact point vs. severity-tiered routing via a notification policy. Avoid inventing secrets; decide the declarative shape and what the operator must supply.
3. **Severity model.** `critical` (page now) vs. `warning` (notify) vs. `info`. Map to routing.
4. **The first alert set.** Catalog the failure-mode metrics that warrant an alert and the condition for each. Known candidates (all currently raw-rate panels only):
   - `router_metrics_batches_total{status}` — ingest success ratio < 0.95 (the #1365 case; condition already worked out in #1373's reverted `alerts.tf` — reuse it).
   - `auth_failures_total{reason}`
[...]

Issue #1368, opened 2026-06-04
Router metrics ingest failures produce no alert — silent 100%-malformed went unnoticed (#1365)


sameerparekh:
## What happened

For the entire window that #1365 was live, **100% of router metrics ingests were failing** — prod exposed `router_metrics_batches_total{status="malformed"}` at 872 and climbing with **zero** `status="ok"`. Nobody noticed until the downstream symptom (empty router-fleet dashboard) was investigated by hand.

## The metric and a panel already exist — that's not the gap

`deploy/grafana/dashboards/api-self-metrics.json` already has a **"Router metrics batch ingest rate (#1205)"** panel:

```promql
sum by (status) (rate(router_metrics_batches_total[5m]))   # status ∈ {ok, malformed, router_mismatch}
```

So the failure *was* plottable. It went unnoticed anyway because:

1. **There is no alerting in the repo at all.** `infra/grafana/` provisions only `grafana_dashboard` resources — no `grafana_rule_group`, no contact points, and the panel's `alert` is `null`. A passive timeseries nobody is staring at cannot catch a silent outage.
2. **The panel shape hides a total failure.** When `ok` is simply *absent* (zero successful pushes ever), there's no red line — just a lone `malformed` series climbing on an otherwise-quiet graph. A raw `sum by (status) rate` makes "some failures" visible but makes "everything is failing" look the same as "low traffic."

## What we want

Make a router-metrics ingest failure **page someone**, not just be theoretically graphable:

1. **Add an alert rule** (in `infra/grafana/` Terraform, alongside the dashboards) that fires when ingest is unhealthy — e.g. `sum(rate(router_metrics_batches_total{status!="ok"}[10m])) > 0` sustained, and/or a **success-ratio** alert: `sum(rate(...{status="ok"}[10m])) / sum(rate(...[10m])) < 0.95`. The ratio form is what catches "ok=0, malformed>0" cleanly.
2. **Add an explicit failure/health signal to the dashboard** — a success-ratio stat panel with thresholds (green/red), so a glance shows "ingest is broken" rather than requiring the reader to notice an absent `ok` line.
3. Consider the same treatment for the other ingest/auth failure counters that today only live as raw rates (`auth_failures_total`, `traffic_reports_filtered_zero_bytes_total`, `metrics_rejected_total`) — a silent spike in any of them has the same blind spot.

[...]

Issue #1365, opened 2026-06-04
bug: router-fleet Grafana dashboard empty despite successful router metrics ingest


sameerparekh:
## Symptom

Grafana shows **successful router metrics ingest** (the API is receiving `POST /api/router/metrics` batches — #1205/#1298 transport works, and the agent is pushing per #1303/#1206), but the **router-fleet dashboard is empty / "No data"**.

So the failure is somewhere between "API received the batch" and "the series are queryable in Grafana Cloud." Filing fresh (not reopening #1205/#1279).

## Most likely root cause (check first)

**The API ingests router metrics into service state but does not RE-EXPOSE them on its Prometheus `/metrics` endpoint that Alloy scrapes.** Transport (#1205) folds the pushed counters into `RouterMetricsService`, but unless those are registered as zio-metrics series surfaced on `/metrics`, Alloy (`deploy/alloy/config.alloy`) never scrapes them and Grafana has nothing — even though ingest "succeeds." Verify the bridge: ingested router series (`snapshot_poll_total`, `agent_uptime_seconds`, `agent_version`, and any others in `shared/contract/router-to-api/router_metrics_batch.json`) must appear in the API's `/metrics` output with the labels the dashboard expects (e.g. `routerId`/`env`).

## Other candidates (work in order)
1. **Series name / label mismatch** — what `openwrt/files/usr/lib/lua/wifihaven/metrics.lua` emits → what the API re-exposes → what `deploy/grafana/dashboards/router-fleet.json` queries. Diff all three; PromQL with a wrong metric name or a `job=`/`env=` filter the series doesn't carry yields empty panels.
2. **env label** — router-fleet panels may filter `env="prod"`/`env="staging"` but the re-exposed series may be untagged or tagged differently (recall both Alloy scrape jobs add an `env` label to APP series; router metrics flow through ingest, not scrape, so they may lack it).
3. **Dashboard deployed but querying the wrong datasource** — confirm the datasource var resolves to the Grafana Cloud Prom datasource that Alloy writes to. (#1288/#1292 added env-picker + datasource defaults to the other dashboards — router-fleet may have been missed.)
4. **Dashboard not actually provisioned** — `infra/grafana/main.tf` (#1270 deploy path). Operator says the dashboard renders but empty, so it IS provisioned — but confirm it points at the merged `router-fleet.json`.

## Verification path
[...]

PR #1425, opened 2026-06-05
feat(grafana): alert on router-metrics ingest failure (#1368)

sameerparekh:
## What

The concrete alert that starts the alerting thread (#1368). A router-metrics ingest failure now **pages the operator** instead of only being graphable — closing the gap that let the #1365 silent 100%-malformed outage go unnoticed.

This implements alert **C4** from the shipped alerting strategy ([`docs/design/alerting.md`](../blob/main/docs/design/alerting.md) §7.1, from #1381), plus the shared contact-point / notification-policy plumbing the rest of that catalog extends — coordinated with #1381 so its C1–C7 / W1–W5 rules *add more rule groups* against the same plumbing rather than replacing it.

### The rule
- Reuses the **exact** success-ratio PromQL from the shipped `api-self-metrics` ingest panel (#1373/#1384) — no invented series; `router_metrics_batches_total{status}` is grep-confirmed emitted at `api/src/metrics/Metrics.scala:240`.
- `(sum(rate(...{status="ok"}[10m])) or vector(0)) / sum(rate(...[10m]))` with a **zero-traffic guard** `sum(rate(...[10m])) > 0`.
- Fires `ratio < 0.95 for 15m`, severity `critical`.
- **Total-outage nuance (the #1384 fix):** the `or vector(0)` coercion makes `ok=0, malformed>0` read as `0` (→ `< 0.95`, fires) instead of empty/No-data. The guard + `no_data_state = "OK"` keep a genuinely idle env from false-firing.
- **`for: 15m`** rationale: matches the 10m rate window plus margin, so a single bad batch during a deploy doesn't page.
- A **staging twin** at `severity=warning` (staging exercises this path but must never page).

### The plumbing (§4)
- `grafana_folder.alerts` (a rule group needs a `folder_uid`).
- `grafana_contact_point` ×3 — `wifihaven-critical` / `-warning` / `-staging`, **all email** to the single household operator. No invented pager/Slack transport with no consumer; distinct resources so `critical` can later re-point at a real pager with zero rule churn.
- `grafana_notification_policy` — singleton root tree, routes by `severity`/`env` (staging matched first → notify, never page), with household-friendly grouping/throttle.
- New vars: `operator_email` (required, sensitive) and `prometheus_datasource_uid` (default `grafanacloud-prom`).

## ⚠️ Stacked on #1406 — do not merge first

These resources are **stateful** (contact points, notification policy, folder, rule group) and have no `overwrite=true`/upsert-by-uid escape hatch the way `grafana_dashboard` does. Against the current **stateless** Grafana CD (empty-every-run state) they'd 409 or accumulate duplicates — the exact reason `main.tf` already refuses to manage a folder for dashboards.

Per [`docs/design/alerting.md`](../blob/main/docs/design/alerting.md) §3.1/§9, `infra/grafana` must first migrate to an **HCP Terraform remote backend** — the blocking prerequisite **#1406** (mirrors #1357 for `infra/cloudflare`). This PR:
- is **CI-lint-clean today** (`terraform fmt -check` + `validate -backend=false` in the `grafana-terraform` job);
- must **land after #1406** so the live `terraform apply` in `master-grafana.yml` (with the `cloud {}` backend, workflow wiring, and the operator's one-time HCP `grafana` workspace + `OPERATOR_EMAIL` secret) can apply it idempotently.

## Validation
- `terraform fmt -check` ✅ and `terraform validate` ✅ (provider schema-checked) in `infra/grafana`.
- Rule expression validated by **reuse**: byte-identical to the already-deployed, already-reviewed success-ratio panel (`deploy/grafana/dashboards/api-self-metrics.json`), against a grep-confirmed emitted series. A forced-condition firing (point a throwaway `curl` at `POST /api/router/metrics` with a malformed batch, per `alerting.md` §10) becomes possible once #1406 makes the rule live.

Refs #1368, #1381, #1373, #1384, #1365, #1406

### DIFF
--- a/infra/grafana/README.md
+++ b/infra/grafana/README.md
@@ -70,6 +70,59 @@ templated `${datasource}` variable resolved at view time, so the same JSON
 loads in both Grafana Cloud and a self-hosted provisioned Grafana, #1207),
 and the GitHub Actions secrets.
 
+## Alerting (#1368 — first rule of the #1381 strategy)
+
+[`alerting.tf`](alerting.tf) adds the first **Grafana-managed alert**: a
+router-metrics ingest failure (alert **C4** in
+[`docs/design/alerting.md`](../../docs/design/alerting.md) §7.1) now pages the
+operator instead of only being graphable. It reuses the exact success-ratio
+PromQL from the `api-self-metrics` ingest panel (#1373/#1384), adds a
+zero-traffic guard, fires `< 0.95 for 15m`, and routes it `critical`. It ships
+with the shared plumbing the rest of the §7 catalog extends:
+
+- `grafana_folder.alerts` — holds the rule group (a rule group requires a
+  `folder_uid`).
+- `grafana_contact_point` ×3 — `wifihaven-critical`, `wifihaven-warning`,
+  `wifihaven-staging`, all **email** to the single household operator. Distinct
+  resources so `critical` can later be re-pointed at a real pager with zero rule
+  churn (§4 — no invented transport).
+- `grafana_notification_policy` — the singleton root tree routing by
+  `severity` / `env` (staging matched first → notify, never page).
+- `grafana_rule_group.router_metrics_ingest` — C4 (prod, critical) + its
+  staging twin (warning).
+
+### Hard prerequisite: remote Terraform state (#1406)
+
+**These alerting resources cannot be applied by the current stateless CD.**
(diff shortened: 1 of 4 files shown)


sameerparekh:
Closing — this is subsumed by #1404 (the warning/critical rule-set implementation), which carries the C4 router-metrics ingest alert as part of the broader #1381 catalog. Superseded rather than landed independently.

Issue #1411, opened 2026-06-05
infra-allow: add gvt2.com domains


sameerparekh:
## Ask
Add **gvt2.com** (and its subdomains) to the curated infra allowlist `PolicyService.infraAllowHosts` (shipped in #1309/#1339; copied into every profile's extraAllowed so it survives the whole-MAC block).

gvt2.com is Google's infrastructure domain (connectivity checks, app/asset downloads, Play services) — another transitive dependency that gets dropped during a whole-MAC block, breaking otherwise-allowed apps. Same class as the akamai/apple hosts added in #1337/#1339.

## Implementation
- Add `gvt2.com` to `infraAllowHosts`. Since the ea_ ipset is populated from resolved hostnames via dnsmasq, decide whether to use the suffix form so all `*.gvt2.com` subdomains are covered (confirm the router can populate an ipset from a domain suffix — the same open question raised in #1337; if suffix-ipset isn't wired, enumerate the specific gvt2.com hosts seen in connection events instead).
- gvt2.com is pure Google infra, not a site users would be blocked from reaching directly — low bypass risk.

## TDD
- Snapshot feature test: gvt2.com appears in every profile's extraAllowed (extend the existing infra-allow spec).
- If suffix matching lands: router render_spec test that a gvt2.com subdomain is carved out.

Refs: #1337/#1339 (akamai/apple infra hosts), #1307/#1309 (infra-allow origin), #1316-#1319 (where this moves to global.infraAllow).

Issue #1319, opened 2026-06-02
router: global composition (@global_allow / @global_block) on OpenWRT + OPNsense


sameerparekh:
Part of the global policy layer design ([#1308](https://github.com/wifihaven/wifihaven/issues/1308)). See [`docs/design/global-policy-layer.md`](docs/design/global-policy-layer.md) §5.2 and §8 (item 4). Depends on the shared-models + PolicyService issues (router needs the `global` field emitted first; it can ignore it until then — additive).

## Scope (`openwrt/` render.lua + agent; `opnsense/` parity)
Apply `snapshot.global` (a `BlockRules`) to every MAC alongside the per-MAC rules, with fixed precedence:

```
ga(d)       = d in G.extraAllowed
gblock(d)   = G.blocked or d in G.extraBlocked or d in union(ipset(G.blocklistIds))
rblock(m,d) = R.blocked or d in R.extraBlocked or d in union(ipset(R.blocklistIds))

drop(m,d) <=>
      not ga(d) and ( gblock(d) or ( d not in R.extraAllowed and rblock(m,d) ) )
   or (G.blockIpOnly or R.blockIpOnly) and d not in resolved_<m>
```

- Add fleet-wide ipsets `@global_allow` (`ga` — suppresses **every** drop, all MACs) and `@global_block` (`gblock` hosts/categories; plus a global blocked flag), layered on the existing `@ea_`/`@blocked_macs`/`@eb_`/`@bl_`.
- Key asymmetry: per-MAC `R.extraAllowed` suppresses per-MAC drops only; **only** `@global_allow` suppresses `@global_block`.
- Populate the global ipsets via dnsmasq `nftset=` callbacks; `@global_allow` hosts land in `resolved_<m>` so they pass `blockIpOnly` for free.
- Router stays a dumb applier — no profile/schedule/tier awareness; it just holds one more `BlockRules`.

## Tests
Lua tests in `openwrt/test/`; OPNsense tests in `opnsense/test/`.

Issue #1318, opened 2026-06-02
PolicyService: assemble global section, relocate uiAllowedHosts, default-deny eval


sameerparekh:
Part of the global policy layer design ([#1308](https://github.com/wifihaven/wifihaven/issues/1308)). See [`docs/design/global-policy-layer.md`](docs/design/global-policy-layer.md) §4, §5, §8 (item 3). Depends on the shared-models issue and the DB migration issue.

## Scope (`api/src/policy/PolicyService.scala`)
- Assemble `snapshot.global` (a `BlockRules`) from the new `global_allow` / `global_blocks` tables.
- **Stop** copying `uiAllowedHosts` into every profile's `extraAllowed` (`computeBlockRules`, ~`:549`) and out of the unmanaged-device block path (~`:169`); relocate those hosts to `global.extraAllowed`.
- Evaluate per-profile `default_deny`: collapse to `blocked=true` + `blockReason=DefaultDeny` (lowest precedence), `extraAllowed` = profile/device allow list, omit `extraBlocked`/`blocklistIds` (redundant under block-all).
- Resolve loosenable global *defaults* (e.g. a household default `blockIpOnly`) server-side into per-MAC `BlockRules` — distinct from the enforced `global.blockIpOnly` floor; never emit defaults as "global" on the wire.
- Produce merged device overrides (device replaces profile, as today; only the server-side computation changes).

## Metrics (per the "new functionality ships with metrics" rule)
- Counter for global-section change events / size; gauge for default-deny profile count. Route via `AppMetrics`/`MetricGuard`, bounded label keys only.
- Ship the Grafana panel in the same PR (`deploy/grafana/dashboards/`), per the dashboard rule.

## Tests
Feature tests live here (this PR is code, not the migration PR). See the dedicated tests issue for the matrix.

sameerparekh:
Implemented and merged in #1430 (merge commit 14d1a08a, now on main; Scala Build & Test + Contract Tests green).

All scope items from §4/§5/§8 satisfied:
- **Global section assembled once** from `global_allow` / `global_blocks` / `global_blocklists` + household global flags via the new `GlobalPolicyRepo.get`, populating `PolicySnapshot.global : BlockRules` — applied flat to every MAC by the router (#1319), not as a third merge tier.
- **`uiAllowedHosts` relocated** to `global.extraAllowed` (unioned in `computeSnapshot`); no longer copied into each profile's `extraAllowed` nor into the unmanaged-device block path.
[...]

Issue #1311, opened 2026-06-02
Docs: clarify that the policy snapshot is a minimal functional shape, not a policy model


sameerparekh:
## Why

While shipping the #1307 global infra allowlist, an initial attempt added a top-level \`PolicySnapshot.infraAllow\` field to carry the curated host set (and even reasons) to the router. That was the wrong instinct: it pushed a *policy concept* into the wire shape. The router is a dumb applier — it only needs the **functional** data required to enforce (hosts in \`extraAllowed\` / \`extraBlocked\`, the per-MAC \`blocked\` flag, etc.). It does **not** need to know *why* a host is always-allowed, nor does it need a separate \"infra\" channel: an always-allowed host is functionally just another \`extraAllowed\` entry.

We need the docs to state this design principle explicitly so future changes don't re-introduce policy-shaped fields into the snapshot.

## What to document (in \`docs/architecture.md\`, snapshot section)

1. **The snapshot is the minimal functional shape needed to enforce.** Each field exists because the router must act on it. If a new policy concept can be expressed via an existing functional field (e.g. always-allow → \`extraAllowed\`), it MUST be, rather than adding a new field.
2. **No policy *reasons* on the wire except where functionally required.** \`blockReason\` is the one intentional exception (block-page copy only, never enforcement). Do not add \"why\" metadata for allow/deny decisions — the server resolves policy into functional data.
3. **Server-side is where policy lives.** All composition (global / profile / device, schedules, time limits, categories, app modes, the infra allowlist) collapses server-side in \`PolicyService\` into per-MAC \`BlockRules\`. New policy concepts land in \`PolicyService\` and present to the router as one of the existing functional fields.
4. **Redundancy vs. shape are separate concerns.** Copying a shared set (like the infra allowlist) into every profile's \`extraAllowed\` is acceptable for the wire *shape*; the redundancy it creates is a separate optimization tracked in #1308 (global policy layer). Reducing redundancy must not come at the cost of teaching the router about policy tiers.

## Scope

Docs only. The actual global-policy-layer wire-shape design (carry-global-once + override model) is #1308; this issue is just the guardrail prose so the principle is written down. Cross-link #1307 (ships infra allowlist by copying into \`extraAllowed\`) and #1308 (removes the redundancy).

PR #1459, opened 2026-06-05
policy(#1321): retire the #1307 per-profile infra-allow copy

### LINKED ISSUE #1321: policy: retire the #1307 per-profile infra-allow copy once global.extraAllowed ships
Part of the global policy layer design ([#1308](https://github.com/wifihaven/wifihaven/issues/1308)). See [`docs/design/global-policy-layer.md`](docs/design/global-policy-layer.md) §8 (item 6). Depends on the PolicyService + router issues landing.

[#1307](https://github.com/wifihaven/wifihaven/issues/1307) ships the global infra allowlist as a near-term fix by **copying** it into every profile's `extraAllowed`. Once `global.extraAllowed` is emitted by PolicyService and consumed by the router, remove that per-profile copy so the set lives in exactly one place.

## Scope
- Remove the #1307 copy path in PolicyService.
- Confirm no profile/device still carries the infra hosts redundantly in `extraAllowed`.
- Regression: the always-reachable hosts still survive every block via `global.extraAllowed` (covered by the tests issue).

sameerparekh:
Closes #1321. Part of the global policy layer (#1308); depends on #1318 (global section incl. UI hosts) and #1319 (router `@global_allow`), both merged.

## What & why

The #1307 near-term fix copied the curated `PolicyService.infraAllowHosts` list (connectivity-check / OCSP / PKI / captive-portal / netcts / gvt2 — incl. the #1337/#1339/#1411 additions) into **every** profile's `extraAllowed` so it would survive a whole-MAC block. Now that the global section carries fleet-wide always-reachable hosts once (#1318) and the router enforces them flat across all MACs (#1319), this retires the per-profile copy.

- `PolicyServiceLive.snapshot` unions `infraAllowHosts` into `global.extraAllowed` (right next to the relocated UI hosts), so the curated set is emitted **once**. A change to it moves only the global section's ETag, not every profile entry.
- `computeBlockRules` no longer appends `infraAllowHosts` to per-profile `extraAllowed` — the per-`(MAC)` `ea_` duplication is gone.

No wire-shape change: this stays purely functional (#1311). `infraAllowHosts` remains a vetted code constant — it just rides the same global path the UI hosts already use, rather than being seeded into the mutable `global_allow` table.

## Behavior preserved at the router

Infra hosts still beat **every** block path for every MAC — now via the router's `@global_allow` carve-out (top of the precedence ladder, #1319) instead of per-`(MAC)` `ea_` copies. The #421 invariant (allow beats block) is intact.

**Hard pause (#1418):** still empties per-profile `extraAllowed` (the allowed-app host is cut, so the router drops the MAC unconditionally). The difference: infra hosts now survive a hard pause as part of the always-reachable global set — exactly like the deployment UI / block-page hosts already did. A hard pause remains a true off-switch for every *per-profile* carve-out; only the fleet-wide always-reachable set (UI + infra) is spared. (Confirmed as the intended behavior.)

## Tests (TDD red → green, real embedded Postgres)

- Per-profile `extraAllowed` no longer duplicates the infra set; the global section carries it (`PolicySnapshotGlobalAllowSpec`, `PolicySnapshotAppsSpec`, `PolicySnapshotGlobalSectionSpec`).
- Behavior preservation: an infra host (e.g. `connectivitycheck.gstatic.com`, `gvt2.com`) is still reachable for a MAC under a whole-MAC block (paused / time-limit) via `global.extraAllowed`.
- `#1337` / `#1411` hosts asserted in `global.extraAllowed` (not per-profile).
- Hard-pause / soft-pause specs updated: per-profile `extraAllowed` behavior unchanged; infra now verified via the global section.

`mill __.test` — 269 specs green. `scalafmt --check` clean.

## Docs

`docs/architecture.md` status note, §0.3 motivating example, and the hard-pause note updated to reflect the consolidated global path.

### DIFF
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -97,8 +97,8 @@ class PolicyServiceLive(
   // hosts, so they live in `global.extraAllowed` (carving out every block at the
   // router) rather than being copied into every profile's `extraAllowed`. The
   // DB-backed `global_allow` set is unioned with these per-deployment config
-  // hosts when assembling the global section. (The #1307 infra-allow copy is
-  // retired separately in #1321.)
+  // hosts when assembling the global section. (#1321 moved the curated
+  // `infraAllowHosts` onto the same global path, retiring its per-profile copy.)
   private val uiGlobalAllow: List[Hostname] = uiAllowedHosts
 
   // #1069: per-host /decision fallback must see the same schedule downtime as the snapshot —
@@ -218,11 +218,19 @@ class PolicyServiceLive(
         c -> Blocklist(version = version, url = BlocklistUrl.unsafe(s"/api/blocklists/${c.value}"))
       }.toMap
 
-      // #1318: union the per-deployment UI / block-page hosts into the DB-backed
-      // global allow set. These are the relocated `uiAllowedHosts` — fleet-wide
-      // always-reachable hosts that carve out every block at the router.
+      // #1318/#1321: union the fleet-wide always-reachable hosts into the
+      // DB-backed global allow set. Two relocated sources join `global_allow`:
+      //   - `uiGlobalAllow` — the per-deployment UI / block-page hosts (#1318).
+      //   - `infraAllowHosts` — the curated connectivity-check / OCSP / PKI /
+      //     captive-portal / gvt2 device-level deps (#1307), which #1321 stops
+      //     copying into every profile's `extraAllowed` and ships here ONCE.
+      // The router applies `global.extraAllowed` as the top of the precedence
+      // ladder (@global_allow, #1319), carving every drop out for every MAC — so
+      // these hosts beat every block path identically to the old per-(MAC) ea_
+      // copies, with no per-profile duplication and a single ETag-moving source.
       val globalRules = globalRaw.copy(
-        extraAllowed = (globalRaw.extraAllowed ++ uiGlobalAllow).distinct,
+        extraAllowed =
+          (globalRaw.extraAllowed ++ uiGlobalAllow ++ PolicyService.infraAllowHosts).distinct,
       )
 
       val core = SnapshotCore(globalRules, devicePolicies, profilePolicies, pBlocklists)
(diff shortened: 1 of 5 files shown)


sameerparekh:
🚧 **Holding this as a draft — do not merge yet.**

Gated on #1460 (e2e: router enforcement of the global section). This PR removes the #1307 per-profile infra-allow safety-net and relies entirely on the router's `@global_allow` path (#1319) to keep infra hosts reachable under a whole-MAC block. Before merging we want:

1. the e2e scenarios in #1460 to exist and pass, **and**
2. a router/agent version that has passed them **deployed to the fleet** — so the live `@global_allow` path is proven before the per-profile copy is retired.

Otherwise a gap in the deployed router's global enforcement would silently break infra reachability fleet-wide (DNS still resolves, so it wouldn't be obvious). Will un-draft once #1460 lands and the proven agent is rolled out.

sameerparekh:
✅ **Gate satisfied — un-drafting.**

The two preconditions from the hold are met:
1. **Global-policy e2e tests merged** — `scripts/e2e/scenarios_fake/test_global_policy.py` (suite H) is on `main`, proving a real OpenWRT agent enforces `@global_allow` / `@global_block`.
2. **A tested router release is published** — the green Master Router CD run [27063576535](https://github.com/wifihaven/wifihaven/actions/runs/27063576535) (commit `147368bf`, the #1489 dnsmasq-line-limit fix) passed **all gates including Gate 3a (apk+ipk)** and its *Publish OpenWRT release* job succeeded. `openwrt-latest` now carries **v0.3.3-1** (assets updated 2026-06-06; rolling-release `publishedAt` stays at the original date by design).

This branch has been **merged up to current `main`** (which since landed per-app / named schedules touching `PolicyService`); both core edits survived intact — the global section unions `infraAllowHosts` once, and `computeBlockRules` no longer copies it per-profile — and the **full suite is green (269/269)** on the merged tree.

⚠️ **Merge-timing note for the operator:** I could not SSH-verify the on-device agent version (prod-shell access wasn't authorized in this session), so please confirm the fleet has auto-updated to ≥0.3.3 (the build that enforces `@global_allow`, #1319) **before merging** this. Merging removes the per-profile infra safety-net; it's only safe once the running agents enforce the global carve-out. Un-drafting (review-ready) is safe regardless of rollout state.

Issue #891, opened 2026-05-22
vm: parameterize bridge/ports/MACs in VM bring-up to allow concurrent qemu pairs on one host


sameerparekh:
Related: #657 (capacity planning), #685 (.ipk/.apk matrix), #654 (Gate 2 parent).

## Problem

The "VM e2e must be serialized" invariant (only one qemu pair per host at a time) exists not because of host capacity — a dev box can easily run two router+client pairs (~2 GB RAM, 6 vCPUs total) — but because the VM bring-up scripts hardcode shared resources:

1. **Bridge name** — `wh-lan0` is hardcoded in `scripts/vm/router-up.sh` and `client-up.sh`. Two routers attached to the same bridge are L2-adjacent: ARP races, duplicate DHCP servers, UCI "I am the gateway" assumptions all break.
2. **Host port forwards** — `127.0.0.1:2222` (router SSH), `127.0.0.1:2223` (client SSH), `127.0.0.1:18081` (HTTP block-page probe) are baked into the qemu cmdlines. Second qemu fails to bind with EADDRINUSE.
3. **MAC addresses** — `52:54:00:fd:00:01/02` (router WAN/LAN) and the client MAC are hardcoded. Two routers with the same MAC on the same bridge → silent blackhole.
4. **`.run/` paths** — already per-workspace (workspace-relative), so this one is fine.

## Why it matters

- Today a local dev VM e2e run on a self-hosted-runner host blocks CI Gate 2 from making progress, and vice versa. The user noticed this during #685 validation: a CI Gate 2 run held the host for ~30m while local validation was queued.
- #657's "beefier single runner → matrix entries as parallel qemu instances" option depends on this work.
- Even on hosts with no CI runner, two devs sharing one box can't both iterate locally.

## Proposed approach

Introduce env-var inputs (with `WH_` prefix, per repo convention):

- `WH_LAN_BRIDGE` — defaults to `wh-lan0`. Router/client up scripts read this; bridge-creation helper in `scripts/vm/bridge.sh` (or wherever it lives) takes it as arg.
- `WH_PORT_BASE` — integer; router SSH = `$WH_PORT_BASE`, HTTP probe = `$WH_PORT_BASE + 1`, client1 SSH = `$WH_PORT_BASE + 2`. Defaults to `2222`. Two pairs = `2222` and (say) `2322`.
[...]

PR #1469, opened 2026-06-05
test(#1460): VM e2e for @global_allow / @global_block enforcement

### LINKED ISSUE #1460: e2e: router enforcement of the global section (@global_allow / @global_block) on a real OpenWRT agent
Part of the global policy layer (#1308). Gating dependency for #1321 / PR #1459.

## Why

#1318 emits `snapshot.global` (a fleet-wide `BlockRules`) and #1319 teaches the OpenWRT agent to compose it into `@global_allow` / `@global_block` nftables sets applied to **every** MAC. Both are covered by **feature/unit** tests (#1322, `api/test/**`, `openwrt/test/**`) but there is **no end-to-end proof** that a real deployed agent actually enforces the global section against live traffic.

This matters because #1321 (PR #1459) removes the #1307 per-profile infra-allow copy and relies **entirely** on `global.extraAllowed` (`@global_allow`) to keep connectivity-check / OCSP / PKI / captive-portal / gvt2 hosts reachable under a whole-MAC block. If the deployed router's `@global_allow` path has any gap, retiring the per-profile copy would silently break infra reachability **fleet-wide** — the exact failure mode the "DNS resolves ≠ reachable" model makes hard to notice. So we want an e2e gate before that change ships.

## Scope — new fake-api VM scenario(s)

Add to `scripts/e2e/scenarios_fake/` (alongside `test_pause.py` / `test_extra_blocked.py`), driving a real OpenWRT agent against the fake API:

1. **`global.extraAllowed` beats a whole-MAC block.** A MAC is `blocked = true` (e.g. paused) AND a host sits only in `global.extraAllowed` (not in the profile's `extraAllowed`). Assert the host is **reachable** (connection-layer, not DNS) for that MAC — i.e. `@global_allow` carves out the `@blocked_macs` drop. This is the direct #1321 regression guard.
2. **`global.extraAllowed` applies to *every* MAC.** Same host reachable from a second, differently-profiled MAC with no per-profile allow — proving the carve-out is flat across the fleet, not per-`(MAC)`.
3. **`global.extraBlocked` blocks fleet-wide** and is **not** un-blockable by a per-profile `extraAllowed` (only `global.extraAllowed` carves it out) — the precedence ladder from the design doc §5.
4. *(stretch)* **`global.blocked` lockdown** drops all forwarded traffic for every MAC while `global.extraAllowed` (UI/block-page hosts) still resolves the block page.

May require extending `scripts/e2e/scenarios_fake/snapshot_builder.py` / the fake API to emit a populated `global` section — do that here.

## The gate (blocks #1321 / PR #1459)

PR #1459 must **not** merge until:
- these e2e scenarios exist and pass, AND
- a router/agent version that has passed them is **deployed to the fleet** (so the live `@global_allow` path is proven before the per-profile infra safety-net is removed).

Until then #1459 stays a draft, blocked on this issue.

sameerparekh:
Closes #1460.

VM-level end-to-end coverage proving a **real OpenWRT agent** enforces the
snapshot's fleet-wide `global` section (#1318) via the `@global_allow` /
`@global_block` nftables sets (#1319). The prior #1322 work covered this at the
feature/unit level (`api/test/**`, `openwrt/test/**`); this adds the missing
proof that a deployed agent applies it against **live traffic**.

## This is the gate for #1321 / PR #1459

PR #1459 removes the #1307 per-profile infra-allow copy and relies **entirely**
on `global.extraAllowed` (`@global_allow`) to keep connectivity-check / OCSP /
PKI / captive-portal hosts reachable under a whole-MAC block. If the live
`@global_allow` path has any gap, retiring the per-profile safety net would
silently break infra reachability **fleet-wide** — the exact failure the "DNS
resolves ≠ reachable" model hides. **A router/agent version that passes this
suite must be deployed to the fleet before #1459 un-drafts.** (This PR does not
touch #1459 or its branch.)

## Scenarios — `scripts/e2e/scenarios_fake/test_global_policy.py`

All assertions are **connection-layer** (nftables forward-drop / block-page
DNAT), never "DNS resolved ⇒ allowed".

| | Asserts |
|---|---|
| **H1** `global.extraAllowed` beats a whole-MAC block | A paused (`blocked=true`) MAC still reaches a host that is **only** in `global.extraAllowed` (not the profile's own allow). Proves `@global_allow` carves the resolved IPs out of the `@blocked_macs` drop. Corroborated by the per-MAC `wh_drop:<mac>` rule being present (the block is real) and no per-MAC `ea_` set existing. **Direct #1321 regression guard.** |
| **H2** `global.extraAllowed` is flat across the fleet | Boots a **second, differently-profiled, also-blocked** client VM; the same global-allowed host is reachable from **both** MACs. Proves the carve-out is one fleet-wide set, not silently per-`(MAC)`. |
| **H3** `global.extraBlocked` blocks fleet-wide | A host in `global.extraBlocked` is blocked **even though** the MAC's profile lists it in `extraAllowed` — only `global.extraAllowed` carves out a global block (precedence ladder §5.2, step 2 beats step 3). Control host stays reachable + MAC not wholesale-blocked. |
| **H4** (stretch) `global.blocked` lockdown | `global.blocked=true` drops all forwarded traffic for every MAC while a `global.extraAllowed` host still reaches its upstream; a non-allowed host hits the block page. |

## Harness changes (in scope per the issue)

- `snapshot_builder.SnapshotBuilder.set_global()` — emits a populated `global`
  `BlockRules`. Omitted from `build()` unless set, so the default top-level
  shape is unchanged and snapshots without it render byte-identically to
  pre-#1319 (render.lua treats an absent `global` as `allowAll`).
- `_observers` — `wait_global_allow_populated` / `wait_global_block_populated`
  read the fleet-wide `@global_allow` / `@global_block` nft sets;
  `ea_set_exists_for_mac` proves the carve-out is the shared global set, not a
  per-`(MAC, host)` `ea_` set.
- `pytest.ini` — new `global_policy` marker (suite H). `README.md` scenario list.

No production code touched (`api/src` untouched, no Flyway migration) — this is
e2e-only.

## Checks run

- ✅ `pytest scenarios_fake/test_snapshot_builder.py` — 12 passed, including the
  4 new pure-dict `set_global` tests (run locally, no VM).
- ✅ `pytest scenarios_fake --collect-only` — full suite collects (43 tests);
  the 4 new H scenarios collect with the registered marker under
  `--strict-markers`.
- ✅ `py_compile` on all touched modules.
- ⏳ The H1–H4 VM scenarios require KVM and run on the **`e2e-vm-fake.yml`** CI
  gate (no local KVM on this machine). No repo Python linter is configured for
  `scripts/e2e`, so none was run.

### DIFF
--- a/scripts/e2e/README.md
+++ b/scripts/e2e/README.md
@@ -74,6 +74,7 @@ scripts/e2e/
     test_time_limit.py · test_reassignment.py
     test_blocked_mac_events.py · test_install_health.py
     test_port_alloc.py · test_snapshot_builder.py
+    test_global_policy.py             suite H (#1460) — @global_allow / @global_block
   gate3/                              Gate 3 smoke + conftest
     test_smoke.py
 ```
--- a/scripts/e2e/pytest.ini
+++ b/scripts/e2e/pytest.ini
@@ -18,6 +18,7 @@ markers =
     install_health: post-install router health asserts (#922 — cron, etc.)
     nflog: #1122 — forward-drop visibility via nflog (per-MAC drop comment + agent emission)
     cname_direct_requery: suite G (#1351) — ea_/eb_ population for directly-queried CNAME targets
+    global_policy: suite H (#1460) — @global_allow / @global_block fleet-wide enforcement (gate for #1321)
     smoke: fast subset across suites (one short case each) for PR gating
 log_cli = true
 log_cli_level = INFO
--- a/scripts/e2e/scenarios_fake/_observers.py
+++ b/scripts/e2e/scenarios_fake/_observers.py
@@ -170,6 +170,54 @@ def wait_bl_set_populated(bl_id: str, *, timeout_s: float = 90) -> list[str]:
     )
 
 
+# ── #1319 global policy sets (@global_allow / @global_block) ─────────────────
+#
+# Fleet-wide ipsets render.lua declares when the snapshot carries a populated
+# `global` section. Unlike the per-(MAC, host) `ea_<mac>_<host>` sets, these
+# are NOT keyed by MAC — one `global_allow` / `global_block` set applies to
+# every MAC. Populated at DNS resolve time by dnsmasq `nftset=` callbacks, same
+# as eb_/bl_. See render.lua GLOBAL_ALLOW4 / GLOBAL_BLOCK4.
+
+GLOBAL_ALLOW_SET = "global_allow"
+GLOBAL_BLOCK_SET = "global_block"
+
+
+def _wait_set_populated(name: str, *, timeout_s: float = 90) -> list[str]:
+    def probe():
+        elems = router_nft_set(name)
+        return elems if elems else None
+    return wait_until(
+        probe, timeout_s=timeout_s, interval_s=3,
+        description=f"nft set {name} to gain at least one element",
+    )
+
+
+def wait_global_allow_populated(*, timeout_s: float = 90) -> list[str]:
+    """Wait until the fleet-wide @global_allow v4 set has >=1 resolved IP."""
+    return _wait_set_populated(GLOBAL_ALLOW_SET, timeout_s=timeout_s)
+
+
+def wait_global_block_populated(*, timeout_s: float = 90) -> list[str]:
+    """Wait until the fleet-wide @global_block v4 set has >=1 resolved IP."""
+    return _wait_set_populated(GLOBAL_BLOCK_SET, timeout_s=timeout_s)
+
+
+def ea_set_exists_for_mac(mac: str) -> bool:
+    """True iff any per-(MAC, host) ea_<mac>_<host> allow-set has been rendered.
+
+    Used to prove the global allow carve-out is genuinely *global* (a single
+    @global_allow set), not silently degraded into per-MAC ea_ sets. render.lua
+    names ea_ sets `ea_<sanmac>_<sanhost>`; a `nft list table` scan for the
+    `ea_<sanmac>_` prefix tells us whether the MAC got a per-MAC allow path.
+    """
+    prefix = "ea_" + _san(mac) + "_"
+    res = router_ssh(
+        "nft list table inet wifihaven 2>/dev/null || true",
+        check=False, timeout=10,
+    )
+    return prefix.lower() in (res.stdout or "").lower()
+
+
 def wait_event_for_mac(fake_api, mac: str, *, allowed: bool, reason: str,
                        timeout_s: float = 60):
     """Block until fake captured a connection_attempt event matching predicates."""
--- a/scripts/e2e/scenarios_fake/snapshot_builder.py
+++ b/scripts/e2e/scenarios_fake/snapshot_builder.py
@@ -61,6 +61,11 @@ class SnapshotBuilder:
         self._profiles: dict[str, dict[str, Any]] = {}
         self._devices: dict[str, dict[str, Any]] = {}
         self._blocklists: dict[str, dict[str, Any]] = {}
+        # The fleet-wide `global` BlockRules (#1316/#1318). None until
+        # `set_global()` is called, so snapshots that don't exercise the global
+        # layer omit the key entirely and render byte-identically to pre-#1319
+        # (render.lua treats an absent `global` as BlockRules.allowAll).
+        self._global: dict[str, Any] | None = None
 
     def add_profile(
         self,
@@ -124,6 +129,45 @@ class SnapshotBuilder:
         self._devices[mac] = entry
         return self
 
+    def set_global(
+        self,
+        *,
+        blocked: bool = False,
+        block_reason: str | None = None,
+        extra_blocked: list[str] | None = None,
+        extra_allowed: list[str] | None = None,
+        blocklist_ids: list[str] | None = None,
+        block_ip_only: bool = False,
+    ) -> "SnapshotBuilder":
+        """Populate the snapshot's fleet-wide `global` BlockRules (#1318/#1319).
+
+        Applied to *every* MAC by the agent alongside that MAC's resolved
+        per-MAC rules, with the precedence ladder from
+        `docs/design/global-policy-layer.md` §5.2:
+
+          - `extra_allowed` → `@global_allow` — carves out **every** drop
+            (per-MAC or global), all MACs. Top of the ladder.
+          - `extra_blocked` / `blocklist_ids` → `@global_block` — blocks
+            fleet-wide; a per-profile `extraAllowed` may **not** un-block them,
+            only `global.extraAllowed` can.
+          - `blocked` → whole-network lockdown; only `global.extraAllowed`
+            survives it.
+
+        Unlike `add_profile(blocked=True)` this does **not** require a
+        `block_reason`: a global lockdown's reason is cosmetic (block-page copy
+        only) and has no `MacBlockReason` enum case. Pass one if a test asserts
+        on the rendered block-page text.
+        """
+        self._global = profile_rules(
+            blocked=blocked,
+            block_reason=block_reason,
+            extra_blocked=extra_blocked,
+            extra_allowed=extra_allowed,
+            blocklist_ids=blocklist_ids,
+            block_ip_only=block_ip_only,
+        )
+        return self
+
     def add_blocklist(
         self,
         *,
@@ -149,13 +193,19 @@ class SnapshotBuilder:
                     f"device {mac} references unknown profileId={pid}; "
                     f"declared profiles: {sorted(self._profiles)}"
                 )
-        return {
+        snap: dict[str, Any] = {
             "etag": etag,
             "generatedAt": self._generated_at,
             "devices": dict(self._devices),
             "profiles": dict(self._profiles),
             "blocklists": dict(self._blocklists),
         }
+        # Emit `global` only when explicitly set, so the default top-level key
+        # set stays {etag, generatedAt, devices, profiles, blocklists} for the
+        # scenarios (and test_snapshot_builder) that don't touch it.
+        if self._global is not None:
+            snap["global"] = dict(self._global)
+        return snap
 
 
 # ── Convenience helpers kept for the #683 canary (test_pause) ───────────────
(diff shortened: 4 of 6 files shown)


sameerparekh:
## VM e2e validation run (local KVM host)

Ran this branch end-to-end on a Linux KVM host. Pure-dict / collection passes; **H3 (`test_global_block_not_unblockable_by_profile_allow`) fails reproducibly** under the harness on this branch's HEAD.

### Environment
- HEAD: `131cbbc` (worktree off `origin/claude/impl-1460-global-e2e`, no edits)
- Host: Linux 6.8.0 (Ubuntu), `/dev/kvm` r/w, 16 cores
- QEMU: `qemu-system-x86_64 8.2.2 (Debian 1:8.2.2+ds-0ubuntu1.16)`
- Docker: `Server Version 29.1.3`
- Router image: built locally via `scripts/vm/build-router-image.sh` (PKG_FORMAT=ipk → OpenWRT 23.05.5, `openwrt-wifihaven-ipk-23.05.img`)
- Client base: built locally via `scripts/vm/build-client-base.sh` (Alpine NoCloud)
- LAN bridge pool `wh-lan0..wh-lan6` present, `qemu-bridge-helper` capable, NOPASSWD scoped to the specific `ip link` commands per `docs/ops/kvm-runner.md` (broader `sudo -n ip -V` doc check fails, but that's not what the harness invokes — runs fine).
- Branch diff vs `origin/main`: e2e-only (no `api/src`, no Flyway migration), matches PR description.

### Result matrix

| | Check | Result | Wall |
|---|---|---|---|
| U | `scenarios_fake/test_snapshot_builder.py` (incl. 4 new `set_global*` cases) | ✅ 12 passed | 0.02s |
| C | `scenarios_fake/test_global_policy.py --collect-only` | ✅ 4 collected (H1–H4 names match) | 0.02s |
| **H1** | `test_global_allow_beats_whole_mac_block` | ✅ PASS | — |
| **H2** | `test_global_allow_applies_to_every_mac` (2nd concurrent client VM) | ✅ PASS | — |
| **H3** | `test_global_block_not_unblockable_by_profile_allow` | ❌ **FAIL** | timed out after 120s `wait_block_page` |
| **H4** | `test_global_blocked_lockdown_with_allow_carveout` | ✅ PASS | — |

Full run: `scripts/e2e-vm.sh --mode=fake -- -m global_policy -v` → `1 failed, 3 passed, 39 deselected in 381.09s`.
Reproduced in a clean re-run isolating H3 only (`-k test_global_block_not_unblockable`) → same failure, same symptom, `221.19s`.

### H2 — concurrent 2nd client VM works fine

This is the first fake-mode scenario to spin up `client2` on its own SSH hostfwd while sharing the router LAN bridge. It passed on first run with no infra trouble (no DHCP lease retries, no hostfwd port conflict, both MACs reached the `@global_allow` host). The wh-lan pool + WH_RUN_ID parameterization from #891 handles it cleanly.

### H3 — root cause read: looks like a real **policy regression**, not harness/infra flake

The failure mode is *not* "host can't reach upstream":

- Both runs show conntrack-watcher posting events for the client MAC during the `wait_block_page` polling window, **all classified `allowed=True reason='allow'`** by the router agent. There are no `blocked`/`global_block` events for `example.org`.
- Re-run (H3 in isolation, client MAC `02:e2:fa:cc:e0:62`) shows `wait_block_page` probes hitting destIPs `216.144.228.179`, `67.84.224.198`, `23.186.168.128`, `15.204.246.57` (all consistent with Akamai-served `example.org`) — and each one was forwarded with `reason='allow'`. The router applied policy, DNS resolved, the connection completed, no DNAT-to-block-page fired.
- H4 (`global.blocked=true` lockdown with `@global_allow` carve-out) passes, so the global ipset / DNAT plumbing isn't dead in general. What H3 specifically exercises is the asymmetry "global.extraBlocked beats profile.extraAllowed for the **same host**", and that's exactly the assertion the test was designed to guard ([#1319 precedence ladder §5.2, step 2 beats step 3](https://github.com/wifihaven/wifihaven/issues/1319)).

By inspection in this branch's `openwrt/files/usr/lib/lua/wifihaven/render.lua`, the forward-drop rule for `@global_block` (≈L1012–1022) and the block-page DNAT (≈L1079–1088) intentionally use only `ga_suffix` (carve-out by `@global_allow`), **not** `ea_suffix` — so on paper the per-MAC `ea_<m>_example_org` allow should not save a `@global_block` drop. But the observed live behavior contradicts that. Possibilities the PR author should pin down:
1. dnsmasq `nftset=` callbacks for `@global_block` (v4/v6) aren't actually being emitted when the snapshot carries `global.extraBlocked` for a host that ALSO appears in a profile's `extraAllowed` — `@global_block` empty at resolve time → drop never matches → forward chain falls through to accept.
2. Rule ordering / chain composition emits an earlier `ea_<m>_example_org` accept-equivalent before the `@global_block` drop is evaluated (i.e. the `daddr != @ea_…` predicate is missing on the global-block drop but an earlier rule with a positive ea_ match is short-circuiting first).
3. `has_global_block` is computed `false` for this snapshot shape because `extraAllowed[host] ∩ extraBlocked[host] ≠ ∅` triggers a pre-filter — possible if `global_block_hosts()` subtracts global allows but doesn't intend to subtract per-profile allows.

I did not chase this further — that's PR author territory. The harness's failure surface is correct and reproducible; the assertion is the right one.

#### Diagnostic notes (limits of what I could capture)
- Captured artifacts: router console log under `scripts/vm/.run/<run_id>/router/console.log`, agent log via `logread -e wifihaven | tail -n 80` (the conftest hook). The agent-log tail rolls over boot noise before H3's snapshot-apply log lines, so we don't see the H3 `policy.apply` lines or any `global_block` populator counters. The hook does **not** snapshot `nft list table inet wifihaven` on failure — adding that would be a cheap follow-up (independent of this PR's scope).
- `dns-tail` periodic summaries show `eb_adds=0, ea_adds={32,72,104,...}, bl_adds=0` — but there is no `global_allow_adds` / `global_block_adds` counter in those summary lines, so the captured agent log can't confirm/deny @global_block population.

### Verdict

**Not safe to treat this branch as the e2e gate for #1321 / PR #1459 yet.** The scenarios are well-formed and H1/H2/H4 give confidence in `@global_allow`. But **H3 fails on a clean local KVM host, and the failure shape is consistent with a real `@global_block` enforcement gap when a per-profile `extraAllowed` names the same host** — which is exactly one of the regressions #1459 needs this gate to catch. Either:

- the agent has a bug that needs fixing in a separate PR (most likely, based on the evidence), and #1459 should not un-draft until the fix lands AND this suite goes 4/4 on a real CI / KVM run; or
- the test itself has a setup gap (e.g. needs to await `@global_block` population before polling, or extend the timeout for cold dnsmasq) — but even then it's flagging a behavior the implementation should make robust before fleet retirement of the per-profile infra-allow safety net.

Suggest waiting for the CI `e2e-vm-fake.yml` arms to run on this PR to confirm reproduction on the canonical self-hosted runner. If they also fail H3, that's the signal to root-cause the agent before merging this gate or un-drafting #1459. Either way, do **not** drop the per-profile infra-allow copy until H3 is green on real CI.

sameerparekh:
## Update — root cause confirmed, fix opened as #1483

Iterated on the H3 failure with live-router introspection. Posting the diagnosis here so #1469's verdict is no longer "validation blocked":

### Root cause

dnsmasq honours only ONE matching `nftset=/<host>/…` directive per domain — the rest are silently dropped. The H3 snapshot puts `example.org` in both `profile.extraAllowed` and `global.extraBlocked`, so `render.dnsmasq()` emitted two `nftset=/example.org/…` lines (one for `ea_…`, one for `@global_block`). dnsmasq fired only the first → `@global_block` stayed empty → the per-MAC `ip daddr @global_block drop` rule never matched → `example.org` was forwarded with `reason='allow'`.

Verified live on OpenWRT 23.05 / dnsmasq 2.91 by editing the kept-router's `wifihaven.conf`:
- Two separate `nftset=/example.org/…` directives → log shows `nftset add 4 inet wifihaven ea_…` only.
- One merged `nftset=/example.org/4#…ea_,6#…ea6_,4#…global_block,6#…global_block6` directive (after a real dnsmasq restart, not just HUP) → log shows `nftset add` for BOTH sets, both populate, the global block fires.

### Fix — #1483

[#1483](https://github.com/wifihaven/wifihaven/pull/1483) merges every per-host `ea_` / `eb_` / `bl_` / `@global_allow` / `@global_block` spec into a single `nftset=/<host>/…` directive in `render.dnsmasq()`, adds a `render_spec.lua` regression test pinning the H3 shape, and tightens `_observers.mac_drop_rule_present` to its docstring intent (whole-MAC reasons only) so H3's control assertion stops misfiring on the now-correctly-emitted per-MAC `wh_drop:<mac>:global_block`.

### Validation after fix (local KVM host)

| | | |
|---|---|---|
| H1 `test_global_allow_beats_whole_mac_block` | ✅ | |
| H2 `test_global_allow_applies_to_every_mac` | ✅ | concurrent 2nd client VM works |
| H3 `test_global_block_not_unblockable_by_profile_allow` | ✅ | was the failing one |
| H4 `test_global_blocked_lockdown_with_allow_carveout` | ✅ | |
| Full Gate 2 (`scripts/e2e-vm.sh --mode=fake`) | ✅ | 40 passed / 1 pre-existing skip / 2 pre-existing xfail, no regressions |
| `openwrt/test/render_spec.lua` (busted) | ✅ | 170 passed / 0 failed |

### Revised verdict

This branch (#1469) is the correct e2e gate for #1321 / PR #1459 — the suite caught a real `@global_block` enforcement gap that no other test covered. Merge #1483 first, then this PR's CI run on `e2e-vm-fake.yml` should go 4/4 on H1–H4 and unblock un-drafting #1459. (Do not drop the per-profile infra-allow copy in #1459 until #1483 ships and this suite is green on real CI.)

The validation findings in [the earlier comment](https://github.com/wifihaven/wifihaven/pull/1469#issuecomment-4634509247) are now fully resolved.

sameerparekh:
Closing in favour of #1483, which carries the same H1–H4 VM scenarios plus the render.lua fix that makes H3 actually pass (the per-host nftset= merge — see the diagnosis in the previous comment). Keeping the e2e and the agent fix in one PR avoids landing tests that need a not-yet-merged fix to go green on CI.

Issue #1380, opened 2026-06-04
per-app schedules: web UI — assignment schedule rules editor


sameerparekh:
Part of the per-app schedules epic #1376. **Blocked on the PolicyService eval issue** (and transitively #1069).

## Scope
Per `docs/design/per-app-schedules.md` §7, add a **Schedules** section to the app/profile assignment editor:

- An **add-rule** control producing one or more rule rows (one-to-many).
- Per rule: a **mode** toggle — *Allowed during* / *Blocked during* — plus a **named-schedule picker** (the #1069 picker; pick an existing 'Bedtime'/'School hours' or create-new inline). No bespoke time editor — the named-schedule primitive owns day/time editing.
- Surface the assignment's **`exemptFromDaily`** flag alongside the rules, with copy that makes the cap behaviour explicit (rows 9a–9c): exempt → reachable in-window even past the daily limit; non-exempt → the daily limit still blocks it. So the cap-bites-non-exempt behaviour is not a surprise.
- **Autosave** (#995/#423): debounced PATCH of the assignment, no Save button. Additive `scheduleRules: List[AppScheduleRule]` on the payload.

## Dependencies
- Blocked on the PolicyService eval issue.
- Uses the #1069 named-schedule picker.

sameerparekh:
Shipped in PR #1481 (assignment-editor schedule-rule editor); the API (#1379) landed in #1486 and the write-path gap was fixed in #1494. Closing.

Issue #1379, opened 2026-06-04
per-app schedules: PolicyService eval + models/repo + tests (TDD)


sameerparekh:
Part of the per-app schedules epic #1376. **Blocked on the DB-migration issue** (and transitively #1069).

## Scope
Implement the server-side collapse of per-app schedule rules into the existing per-MAC `BlockRules` — **no snapshot/wire change, no router change** (AGENTS.md invariant). Per `docs/design/per-app-schedules.md` §3.4–§5:

- **Models** (`shared/src/Models.scala`, API-internal — NOT a `PolicySnapshot` field): `enum AppScheduleMode { AllowedDuring, BlockedDuring }`, `case class AppScheduleRule(id, assignmentId, scheduleId, mode)`. Extend `UpsertAppAssignmentRequest` additively with `scheduleRules: List[AppScheduleRule] = Nil`.
- **Repo**: load `app_policy_schedule_rules` per assignment; resolve each rule's #1069 named schedule to its window list.
- **Eval** (in **both** `PolicyService.snapshot` and `PolicyService.decide`): the §4.1 effective-disposition resolution — a rule is active iff any of its named schedule's windows is active at `now` (reuse `scheduleActiveAt`, household-local TZ via injected `Clock`, never `java.time`). `allowed_during` active → AllowedDuring candidate; `blocked_during` active → Blocked; else base mode. Window overrides base mode.
- **Cap gate** (§5, rows 9a–9c): `carveIntoExtraAllowed = AllowedDuring AND NOT (dailyCapExhausted AND NOT exemptFromDaily)`. `dailyCapExhausted` computed directly from `ProfileDayState` (used/limit/extensions), independent of the collapsed `blockReason`. An `allowed_during` window beats Schedule/Paused/Manual always, but yields to the daily time limit unless the app is `exemptFromDaily`.
- Confirm **no** `PolicySnapshot` field is added; the only outputs are `extraAllowed`/`extraBlocked`.

## Tests (TDD — ship with the code)
- **Unit (schedule-boundary):** exact on/off edges, overnight wrap, DOW membership, DST transition, `allowed_during`-beats-`blocked_during` tiebreak, multi-window named-schedule resolution, and the cap gate (true for exempt-over-cap, false for non-exempt-over-cap, true for either under cap). `Clock.TestClock` fixtures.
[...]

Issue #995, opened 2026-05-24
app: autosave rollout — replace explicit Save buttons with debounced autosave across edit forms


sameerparekh:
Umbrella for rolling out debounced autosave across the SPA's edit forms. Operator preference (per #965): edit forms should autosave, not require explicit Save buttons. Profile is the first landing spot (#965 sub-2..6); this issue tracks the same shape rolled out to the other resources.

## Shape (same for every resource)
- Debounced (~600ms) autosave per field/section. No Save buttons.
- "Saved just now" (transient) + "Unsaved" (while debounce timer or in-flight request) indicators per subsection.
- Errors surface inline with a Retry; the dirty value stays in the form.
- Per-field PATCH under the hood. No PUT-on-keystroke.

## Hard prereqs per resource
Every per-resource autosave issue blocks on its corresponding API prereq:
- Profile: #423 (PATCH) + #581 (symmetric shapes) — existing.
- Devices: #996.
- Users: #997.
- Household: #998.
- Apps: #999.

## API prereqs
- [ ] #996 — PATCH /api/devices/:mac + symmetric Device shapes
- [ ] #997 — PATCH /api/users/:id + symmetric User shapes
- [ ] #998 — PATCH /api/household + symmetric Household shapes
- [ ] #999 — PATCH /api/apps/:id + symmetric App shapes
- [ ] #423 — PATCH /api/profiles/:id (existing)
- [ ] #581 — symmetric Profile shapes (existing)

## Web rollout (each depends on its API prereq above)
- [x] #1000 — autosave Devices edit form (blocks on #996)
- [x] #1001 — autosave Users edit form (blocks on #997)
- [x] #1002 — autosave Admin / Household settings (blocks on #998)
- [ ] #1003 — autosave Apps edit form (blocks on #999)
- [ ] #973-#977 — Profile inline subsections under #965 (block on #423 + #581)

## Explicitly out of scope
- **Account password change** ([`AccountPage.tsx`](web/src/pages/AccountPage.tsx)) — security-sensitive submit, stays explicit.
- **Routers** ([`RoutersPage.tsx`](web/src/pages/RoutersPage.tsx)) — create-only flow with token generation; no edit form to autosave.
- **Create flows** in general — autosave applies to *edit*, not *create*. Create modals stay explicit so the operator can abandon.

## Cross-cutting design opportunity
[...]

PR #1481, opened 2026-06-05
feat(#1380): per-app schedules — assignment-editor schedule-rule editor

sameerparekh:
Implements #1380 (per-app schedules epic #1376) — the **web UI sub-issue (§7 / sub-issue 3)** of `docs/design/per-app-schedules.md`.

Adds a **Schedules** section to each app's per-profile assignment row (the `AppRow` editor on the Profiles page → app subsection).

## What's here
- **Per-rule editor.** Attach one or more rules to an assignment, each a #1069 **named schedule** + a **mode** toggle — *Allowed during* / *Blocked during*. Rules are deduped on `(scheduleId, mode)`.
- **Reuses the #1069 `SchedulePicker`** (landed in #1487) for schedule selection — so the household named-schedule catalog **and** the inline "Custom (define inline)…" create-a-reusable-schedule flow come for free, exactly as the §7 ask intended ("create-new inline from the same control").
- **Clear semantics in copy.** Allowed-during reads "reachable while this window is active — even during profile downtime"; blocked-during reads "blocked while this window is active — even when the profile is otherwise allowed."
- **`exemptFromDaily` surfaced alongside allowed-during rules** with the cap copy from rows 9a–9c (exempt → reachable in-window past the cap; non-exempt → the daily limit still blocks it), so the cap-bites-non-exempt behaviour is not a surprise. Hidden for `time_limited` apps, whose existing "Counts toward daily limit" row already governs the same flag.
- **Autosave (#995/#423).** Add/remove persists the whole assignment immediately — no Save button — consistent with the existing block/allow/time-limit toggles in the same row.

## Wire shape (depends on #1379)
The assignment payload extends `UpsertAppAssignmentRequest` **additively** with `scheduleRules`. The field is **omitted when empty**, so existing no-rule payloads (and their assertions) are unchanged. The router/snapshot wire is untouched — per the design, per-app schedules collapse server-side into the existing `extraAllowed`/`extraBlocked` fields.

### Assumptions to reconcile once #1379 lands
#1379 (PolicyService eval + `shared/Models.scala` additions) is being built in parallel and is **not merged yet**, so the TS types here are written against the design contract:
- **Rule shape on the wire:** I send `scheduleRules: [{ scheduleId, mode }]`, where `mode` is the string `allowed_during` / `blocked_during` (matching the existing `ScheduleMode` JSON codec in `shared`). The design's `AppScheduleRule` case class also carries `id` / `assignmentId`; for an *upsert authoring* request those aren't known, so I followed the `SetProfileSchedulesRequest` precedent (ids/refs only). **If #1379's request decoder expects a different rule field set, this is the one spot to reconcile** — `web/src/types/api.ts` (`AppScheduleRule`) and the payload built in `AppRow.apply`.
- **Read-back:** `AppPolicyAssignment.scheduleRules?` is optional (back-compat) and rendered when present.

## Rebased onto current main (#1069 landed)
This branch was rebased after #1069's named-schedule web UI (#1487) merged. That PR shipped the named-schedule client (`api.schedules.*`), the `useNamedSchedules` query, the `NamedSchedule`/`ScheduleWindow` types, **and** the reusable `SchedulePicker`. This branch now **reuses all of that** instead of the duplicate plumbing it originally carried — the diff is down to three files (`ProfilesPage.tsx`, `ProfilesPage.test.tsx`, `types/api.ts`) and the only net-new types are the per-app-rule ones (`AppScheduleMode` / `AppScheduleRule` / `scheduleRules`).

## Tests (TDD)
`web/src/pages/ProfilesPage.test.tsx` — new `#1380` describe:
- renders attached rules with schedule name + mode
- attaching allowed-during / blocked-during calls `setPolicy` with the right additive payload
- removing a rule persists the assignment without it (empty → field omitted)
- an allowed-during rule surfaces the exempt-from-daily cap copy

Full web suite green (381 tests), `tsc --noEmit` clean, eslint clean. No Scala touched.

Depends on #1379.

### DIFF
--- a/web/src/pages/ProfilesPage.test.tsx
+++ b/web/src/pages/ProfilesPage.test.tsx
@@ -36,6 +36,12 @@ vi.mock('@/api/client', () => ({
       setPolicy: vi.fn(),
       deletePolicy: vi.fn(),
     },
+    // #1380 — the per-app schedule-rule editor embeds the #1069 SchedulePicker,
+    // which reads/writes the household named-schedule catalog.
+    schedules: {
+      list: vi.fn(),
+      create: vi.fn(),
+    },
     time: {
       summaryAll: vi.fn(),
       grantExtension: vi.fn(),
@@ -162,6 +168,8 @@ beforeEach(() => {
   ;(api.apps.list as unknown as ReturnType<typeof vi.fn>).mockResolvedValue([])
   ;(api.apps.setPolicy as unknown as ReturnType<typeof vi.fn>).mockResolvedValue(undefined)
   ;(api.apps.deletePolicy as unknown as ReturnType<typeof vi.fn>).mockResolvedValue(undefined)
+  ;(api.schedules.list as unknown as ReturnType<typeof vi.fn>).mockResolvedValue([])
+  ;(api.schedules.create as unknown as ReturnType<typeof vi.fn>).mockResolvedValue(undefined)
   ;(api.profiles.usageByApp as unknown as ReturnType<typeof vi.fn>).mockResolvedValue({
     profileId: 1, profileName: 'Kids', from: '2026-05-26', to: '2026-05-26', apps: [],
   })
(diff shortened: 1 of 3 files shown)


sameerparekh:
Coordination from the #1069 SPA work (#1487, now open): that PR adds the named-schedule management page **and a reusable `SchedulePicker`** (`web/src/components/SchedulePicker.tsx`) — exactly the "inline create + window authoring" surface this PR's scope note deferred. It offers household schedules + "Always" + "Custom (define inline)", where Custom transparently creates a household-scoped named schedule.

When both land, suggest rebasing this onto #1487 and:
- replacing the inline assignment-row picker here with `<SchedulePicker value={…} onChange={…} />`, and
- dropping the duplicate `api.schedules.create` / `useNamedSchedules` additions (#1487 adds the same `api.schedules.{list,create,update,delete}` + `useNamedSchedules` to `client.ts` / `queries.ts`, so there will be a small conflict to resolve there).

No rush — flagging so the two don't drift.

Issue #1494, opened 2026-06-05
bug: profile-editor schedule WRITE path still targets legacy schedules table — edits silently ineffective after #1482 read-switch


sameerparekh:
## Bug — schedule write path not migrated; profile-editor edits are silently ineffective
After #1482 (PR #1490), schedule **enforcement reads from `named_schedules`/`profile_schedule_rules`**. But the **profile editor still WRITES old-style per-profile schedules** (the legacy V1 `schedules` table) — observed in the SPA: the profile editor's schedule section edits old-style schedules, not named schedules. Since nothing reads the legacy table anymore, **editing a profile's schedule in the profile editor now does nothing functionally** (it writes a dead table). Split-brain: new read path, old write path.

This is the write-path half that #1482 (read-only switch) and #1069 (named-schedule management UI + picker, but not wired INTO the profile editor) left incomplete. Filing fresh (both merged/closed).

## What exists vs what's missing
- ✅ Named-schedule CRUD API + SPA (`/api/schedules`, the management page + reusable picker — #1438/#1487).
- ✅ Enforcement reads `named_schedules` via `profile_schedule_rules` (#1490).
- ❌ The **profile editor** still edits inline old-style schedules (likely an inline `schedules` array on the profile upsert → writes the legacy `schedules` table).
- ❌ Likely missing/unwired: a profile↔named-schedule **attach** write path (`profile_schedule_rules` CRUD) consumed by the profile editor.

## Fix
1. **API write path:** ensure profiles attach/detach named schedules via `profile_schedule_rules` (add the endpoint if it doesn't exist, or confirm the #1438 API covers it). Stop accepting/writing the legacy inline `schedules` array on the profile upsert (or redirect it to create a profile-scoped named schedule + rule). After this, NOTHING writes the legacy `schedules` table.
2. **SPA profile editor:** replace the inline old-schedule editor in the profile editor with the **#1069 reusable schedule picker** (dropdown of household schedules + Always + Custom inline), calling the named-schedule attach API. Editing a profile's schedule must go through named schedules end to end.
3. Verify round-trip: edit a profile's schedule in the editor → it persists as a `profile_schedule_rules` row → enforcement (the #1490 read path) reflects it on the next snapshot.

## Relationship to the migration
[...]

Issue #1485, opened 2026-06-05
schedules: DROP legacy per-profile schedules table (phase 3, destructive — land ~1 week after #1482 deploys)


sameerparekh:
## Ask (deferred destructive — target ~2026-06-12, about a week out)
Phase 3 of the schedule unification: **drop the old per-profile `schedules` table** (V1/V16) once the new named-schedule read-path (#1482) has rolled out to prod and soaked, so rollback to a pre-#1482 image is no longer a goal.

This is the destructive tail of the two-phase pattern — deliberately separated from #1482 so it lands only after the new enforcement path is proven in prod.

## Preconditions (do NOT land until ALL hold)
1. #1482 PR 2 (PolicyService reads `named_schedules`/`profile_schedule_rules`) is **merged and deployed to prod**.
2. It has run in prod for **~1 week** with schedule enforcement confirmed correct (no regression vs the migrated bedtime/etc. blocks).
3. Nothing in the codebase still reads the old `schedules` table — grep confirms zero references (the api test suite passing with the table gone is the proof).
4. We're past the point where rolling back to a pre-#1482 image is a recovery goal (after a week + a deploy or two, per the #696 two-phase-delete guidance).

## Work
- A destructive migration-only PR (`V{next}__drop_legacy_schedules.sql`) dropping `schedules` (and any now-dead per-profile schedule columns/indexes).
- Migration-isolation rule applies (migration-only PR).
- Take/confirm a DB snapshot exists before applying in prod (drop is irreversible without restore) — note this in the PR.

## Acceptance
- Old `schedules` table dropped; no code references it; api test suite green; prod schedule enforcement unaffected (still driven by named_schedules).

Refs: #1482 (the migration + read-switch this depends on), #1069, `docs/design/per-app-schedules.md` §3.3, #696 (migration-isolation / two-phase-delete pattern).

sameerparekh:
Audited preconditions today (2026-06-14) — **abort: precondition #3 fails**.

#1482 is merged + soaked (a25ae618, 2026-06-05, 9 days), so timing is fine. But the codebase still has live readers and writers of the `schedules` table:

- `api/src/db/Repos.scala` — `ScheduleRepoLive` is still wired (SELECT/INSERT/DELETE on `schedules` at lines 890/897/904/906) and the `scheduleRepo` ZLayer (3308) is constructed at startup via `api/src/Main.scala:261`.
[...]

Issue #1482, opened 2026-06-05
schedules: migrate existing per-profile schedules into named-schedule model + switch enforcement read-path


sameerparekh:
## Ask
Migrate existing **per-profile schedules** (the V1 `schedules` table) into the new **household-scoped named-schedule** model (#1069: `named_schedules` + `schedule_windows` + `profile_schedule_rules`) and switch PolicyService to enforce from the new model — so there is ONE schedule system, not two. Prerequisite for retiring the old table and for per-app/blocklist schedules to share one concept.

## Current state (verified)
- **Old (live):** `schedules` (V1/V16) — profile-scoped: `profile_id, name, days TEXT[], start_local TIME, end_local TIME, tz TEXT`. **Deployed enforcement still reads this** — it backs all schedule blocking today.
- **New (additive, #1069):** `named_schedules` + `schedule_windows` + `profile_schedule_rules` (V50). API model/CRUD/snapshot-expansion merged (#1438) but the old table is still the enforcement source. The V50 comment + `docs/design/per-app-schedules.md` §3.3 describe this migration.

## Work (two PRs, per the migration-isolation rule)
1. **Data migration (migration-only PR).** For each existing `schedules` row: create a `named_schedule` + its `schedule_windows` (from `days`/`start_local`/`end_local`/`tz`), and a `profile_schedule_rules` row linking the profile → that named schedule with the correct mode (old schedules = internet **blocked during** the window). Faithful 1:1 by default (don't aggressively dedup; users can merge in the UI later) — but give identical schedules a sensible shared/derived name. Append-only; safe on prod volume.
2. **PolicyService read-path switch (code PR).** Make schedule enforcement read from `named_schedules`/`profile_schedule_rules` instead of the old `schedules` table, so there's a single source of truth. ZERO router-side change — snapshot expansion stays API-side; don't touch `shared/contract/api-to-router/policy_snapshot.json`.

## Critical invariant
Enforcement must be **byte-for-byte identical** before vs after for existing profiles — a profile with a bedtime block keeps blocking at the exact same times. The migration is a representation change, not a behavior change.

## NOT in scope (future, two-phase destructive)
Dropping the old `schedules` table — a later PR once the new path has rolled out and the old rows are dead (per §3.3).

## TDD
- Migration test: an existing `schedules` row produces the equivalent named_schedule + window + profile rule.
[...]

PR #1500, opened 2026-06-06
fix(#1494): migrate the schedule e2e write path to named schedules (unblock Master API/UI CD)

sameerparekh:
## Why

Master API/UI CD is **RED** on `scripts/e2e-router.sh` §7 ("Live e2e against staging stack"), blocking all prod deploys:

```
✗ expected blockReason=Schedule, got 'TimeLimit'
```

This is the e2e half of the #1494 split-brain. Since #1490 switched schedule **enforcement** to read `named_schedules` / `profile_schedule_rules`, and #1495 stopped the profile upsert from **writing** the legacy V1 `schedules` table (an inline `schedules` array is now an ignored unknown field), the e2e was the last client still authoring its schedule via that dead inline array on the profile PUT. The schedule was never created → the snapshot fell through to the §3 TimeLimit block → `got TimeLimit`.

This is the same shape as the live prod regression #1494 describes: any schedule authored via the legacy inline path is silently unenforced after the #1490 read-switch. The API and SPA write paths were already migrated in #1495; this PR closes the remaining gap (the e2e) so CD goes green and the precedence is pinned by a unit-level test.

## What changed

- **e2e (`scripts/e2e-router.sh` §7):** author the always-on schedule the way every client now must — `POST /api/schedules` to create an all-days/all-hours **named** schedule, then `PUT /api/profiles/{id}/schedules` to attach it (`profile_schedule_rules`). The 1-min daily limit stays hot so §7 still proves **Schedule > TimeLimit** precedence end to end. Assertion stays strict (must be `Schedule`, never just "some block").
- **API feature test (`PolicySnapshotNamedScheduleSpec`):** add the snapshot-level counterpart — an attached named schedule + an exceeded daily limit yields `blockReason=Schedule` (Schedule #2 > TimeLimit #3), plus a control (schedule inactive → falls through to `TimeLimit`) proving the limit is genuinely hot. Both run against real embedded Postgres via the same attach path the e2e exercises.

No production source changed — the API/SPA write-path migration shipped in #1495; nothing writes the legacy `schedules` table.

## Verify

- `mill api.test.testOnly 'wifihaven.api.feature.PolicySnapshotNamedScheduleSpec'` → 6 passed (incl. the two new precedence tests).
- `scalafmt --check` clean; `bash -n scripts/e2e-router.sh` clean.

## Notes

- This **unblocks Master API/UI CD** and restores confidence that profile schedules are enforced end to end through the #1490 read path.
- #1485 (drop the legacy `schedules` table) stays blocked until this + #1482 soak.

Refs #1494, #1490, #1495, #1485, #1482.

### DIFF
--- a/api/test/src/feature/PolicySnapshotNamedScheduleSpec.scala
+++ b/api/test/src/feature/PolicySnapshotNamedScheduleSpec.scala
@@ -10,7 +10,7 @@ import io.zonky.test.db.postgres.embedded.EmbeddedPostgres
 import zio.{Clock as _, *}
 import zio.test.*
 
-import java.time.{LocalDateTime, LocalTime, ZoneId}
+import java.time.{LocalDate, LocalDateTime, LocalTime, ZoneId, ZoneOffset}
 
 /**
  * #1069: the policy snapshot must fold a profile's *named* schedule (referenced via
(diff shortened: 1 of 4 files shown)


sameerparekh:
Swept the rest of the e2e surface for the legacy inline `schedules` array (follow-up commit 62836bc):

- **`scripts/e2e/lib/api_admin.py`** — `add_schedule`/`remove_schedule` folded a schedule into the inline array and PUT it on the profile, a silent no-op since #1495. No callers today (no live-mode `scenarios/` suite uses them), so nothing in CD was broken by them — but they were a landmine. Re-pointed at the named-schedule write path (`POST /api/schedules` → `PUT /api/profiles/{id}/schedules`), added `attached_schedule_ids`/`set_profile_schedules`, and stopped folding the dead `schedules` field through `create_profile`/`apply_profile_update`.
- **`scripts/e2e-tests.sh`** — dropped the now-meaningless `"schedules":[]` from the profile-create body.

The fake-mode suite (`scenarios_fake/test_schedule.py`) POSTs `blocked`/`blockReason` straight to a fake API and never touches the real PolicyService write path, so it needed no change. After this, **no e2e script authors a schedule the API will ignore.**

Issue #1513, opened 2026-06-06
bug(prod): exempt-from-daily time-limited app (Math Academy) blocked when profile hits DAILY limit — apex-only host-set escapes the exempt carve-out


sameerparekh:
## Bug (prod, confirmed)
An **exempt-from-daily, time-limited app** (Math Academy) is **blocked when the controlling profile hits its DAILY limit**, even though it's under its own app limit and is explicitly exempt from the daily total. Operator repro: daily was 35m over the 30m limit; Math Academy showed ~14m (under its own limit); Math Academy was blocked; bumping the profile daily limit to 60m made it work again. No blocked traffic appeared in Connection Events for the device.

## Prod config (verified via /api/apps, profile 1 "Kids")
- Math Academy (app 13) assignment: `mode: "time_limited", dailyMinutes: 30, exemptFromDaily: true`.
- **HOSTS: `["mathacademy.com"]` — apex only.** No CDN/API/asset hosts.
- `/api/time/status`: daily used **40 / limit 30** (over → whole-MAC TimeLimit block). Math Academy site usage **usedMins 0** (its real traffic isn't attributed to the app — lands in "Other").

## Root cause
1. **Apex-only host-set (depends on #1505).** When the profile is over its daily limit, enforcement applies a whole-MAC block (reason TimeLimit). The exempt carve-out (#1105) can only protect Math Academy's listed hosts = `mathacademy.com`. The app's actual traffic (CDN/assets/API — where essentially all the bytes go; that's why site usage reads 0) is NOT in the carve-out → dropped → app blocked. Raising the daily limit removed the whole-MAC block, so everything became reachable → masked the bug.
2. **Verify the exempt carve-out itself.** Confirm `computeBlockRules` adds an exempt-from-daily time_limited app's FULL host-set to `extraAllowed` whenever the MAC is blocked for daily TimeLimit AND the app is under its own limit (the #1105 `appExemptAllowedHosts` path). With apex-only hosts it's untestable today; once #1505 lands the full host-set, this carve-out must use it.
3. **Site-usage attribution (#1505 again):** because only the apex is attributed, Math Academy's own usage reads 0 — so its own limit isn't actually enforced either, and the "under own limit" budget check that gates the carve-out is computed on bad data.

## Secondary finding — observability gap
[...]

Issue #1505, opened 2026-06-06
app attribution: map app host-sets beyond the apex so off-domain assets count toward the app + its site limit


sameerparekh:
Part of the **Tuning** epic #1445. Surfaced by the #1499 over-count analysis (see `docs/design/presence-tuning-overcount.md`, PR #1501).

## Problem
App→domain attribution matches a traffic host to an app only when the host's apex is a stored `app_hosts` row (`HostMatch.lookupApex`, `shared/src/types/HostMatch.scala`). Apps are seeded with **apex-only** hosts — Math Academy = `["mathacademy.com"]` (verified on prod). Suffix matching covers subdomains of the branded apex (`www.mathacademy.com` ✓), but **off-domain** assets/CDN (anything not under `mathacademy.com`) miss both:
- the app attribution → land in the traffic page's "Other" bucket, and
- the `mathacademy.com` **site-limit pattern** → never tick the 30m limit.

Note this is **distinct from the closed infra-allow work** (#1337/#1411), which was about *enforcement* (carving CDN out of the block); this is about *usage attribution + site-limit counting*.

## Ask
Give each app a curated host-set beyond the apex so its real asset/CDN hosts attribute to the app and tick its site limit. Consider:
- A mechanism to surface "Other" hosts that co-occur with an app's sessions, so the operator can promote them into the app's host-set.
- Avoid pinning *rotating* CDN mapping artifacts (`*.akamai.net`/`*.fastly.net`) directly where they rot; prefer the app's stable branded asset domains.

## Composition
Pairs with the per-site stitch migration (#1504, the bigger reason Math Academy reads 0) and feeds the suppression-gate guardrail (app-attributed hosts are never suppressed as infra).

Issue #1413, opened 2026-06-05
bug(prod): allowed app (mathacademy.com) blocked while profile is Paused — extraAllowed not carving around pause


sameerparekh:
## Report
On **prod**, the operator paused a kid's profile. The kid tried to use **mathacademy.com - an allowed (extraAllowed) app - and was blocked.**

Per the architectural invariant (#421, `memory/feedback_extraallowed_beats_blocked.md`), `extraAllowed` must beat EVERY whole-MAC block path - including the Paused block - via the router's `ip daddr != @ea_<m>_<a>` carve-out. So an allowlisted app should remain reachable during pause, exactly as #1307/#1309 made it survive the daily-time-limit block.

This is the **Paused-reason sibling of #1307** (which fixed the TimeLimit reason). Filing fresh.

## Product intent to confirm (flag in the PR)
Confirm intended semantics: an explicitly-allowed app SHOULD stay reachable when a profile is paused (treat pause like the other whole-MAC blocks, honoring extraAllowed). If pause is instead meant to be a HARD stop overriding even allowlists, that's a different design and this isn't a bug - but the documented invariant (#421) and the #1307 precedent both say extraAllowed wins, so default to: **allowed apps survive pause.**

## Candidate root causes (check in order)
1. **Pause block applied via a path that doesn't carry the ea_ carve-out.** Verify the Paused block uses the SAME `@blocked_macs` mechanism WITH the `ea_` carve-out - not a separate/earlier drop rule that skips it. Inspect render.lua for whether the pause path emits `ip daddr != @ea_...`.
2. **Snapshot:** for a paused profile, does the MAC's BlockRules still include the app's hosts in extraAllowed? Check `computeBlockRules` - when blocked=true reason Paused, extraAllowed should still be populated unconditionally (same as the #1307 fix).
3. **ea_ ipset not populated** for the app's hosts under pause.
4. **Stale agent** - the #1309 carve-out may not be deployed to the operator's router.
5. **blockIpOnly** with no carve-out for IP-only flows.

## Division of labor
Claude drives API/prod (fetch the paused profile's snapshot for the kid MAC, confirm extraAllowed contents). Operator runs router-side: `nft list ruleset` to see whether the pause drop carries the ea_ exception, agent version.

## TDD / acceptance
- API snapshot test: a Paused profile still emits the allowed app's hosts in extraAllowed.
[...]

Issue #1325, opened 2026-06-02
metrics: agent emits nftables forward-drop / block counters


sameerparekh:
Prerequisite for a real enforcement dashboard (see #1280, reframed).

## Why
There is currently **no metric for the actual enforcement plane**. Per the architectural model, enforcement is nftables forward-drop on the gateway — dnsmasq is only hostname attribution, not enforcement. The agent metrics shipped in #1206 cover apply/poll/restart telemetry (the *control* path), not *how much traffic is actually being dropped*. An operator cannot today answer "are blocks taking effect, and for what reason" from metrics.

## Scope
Add a cumulative counter to the agent's in-process registry (the same one #1206's 60s push to `POST /api/router/metrics` drains) for nftables forward-drops, sliced only by a **bounded reason enum**:

- `enforcement_drops_total{reason}` where `reason ∈ {whole_mac, host_block, category_block, ip_only}` — mirrors the per-flow `BlockReason` cases the router emits at drop time (whole-MAC `@blocked_macs` drop, per-host `eb_`/`bl_` drop, category blocklist drop, `blockIpOnly` drop).

Source the deltas from the nftables rule `counter` statements (each drop rule carries a named counter); read them on the metrics tick and fold the delta into the registry. **No per-mac / per-host / per-ip label** — those are forbidden by the §4 cardinality firewall.

Server side:
- Add `enforcement_drops_total -> Set("reason","router_id","installation_id")` to `MetricGuard.Allowed` in `api/src/metrics/Metrics.scala`.

Contract:
- Extend the `router_metrics_batch.json` golden fixture (#634) with a representative `enforcement_drops_total` series.

## Out of scope
The dashboard itself — that's #1280 (reframed), which depends on this.

## TDD
- busted unit test for the nft-counter delta read/fold in the agent.
- Contract round-trip test stays green after the fixture extension.

sameerparekh:
Blocks the reframed enforcement dashboard #1280.

PR #1522, opened 2026-06-06
fix(#1513): carve & limit exempt time_limited apps at the app level (full host-set), not per host

sameerparekh:
## Problem (prod, confirmed)

An **exempt-from-daily time_limited app** (Kids profile / Math Academy) was **blocked when the profile hit its DAILY limit**, even though it was under its own app limit and explicitly exempt. Operator repro: daily 40/30 (over) → whole-MAC `TimeLimit` block; Math Academy showed ~14m (under its own 30m); bumping the daily limit to 60m unblocked it.

## Root cause

The exempt carve-out (`appExemptAllowedHosts`) **and** its "under its own limit" budget check were evaluated **per host**:

- Math Academy's host-set is **apex-only** (`mathacademy.com`), so the carve only protected the apex. The app's real traffic (CDN/asset/API hosts) fell outside `extraAllowed` and was dropped by the whole-MAC `TimeLimit` block. (Same apex-only attribution is why its site usage read 0m.)
- Per-host evaluation also **splits an app's usage across its hosts**, so a multi-host app's own limit never bites — and when over budget, a per-host check would still carve the hosts that are individually under, leaving the app reachable past its limit.

## Fix

`PolicyService.computeBlockRules` now groups the per-host `SiteDayState` rows by their shared `label` (= `app:<slug>` — one time_limited app emits one row **per host**, see `SiteTimeLimitRepoLive.listForProfile`), **sums the app's usage across its hosts**, and carves the app's **ENTIRE host-set** into `extraAllowed` iff it is exempt **and under its single daily limit** — otherwise none, so the whole-MAC block drops it.

- With today's apex-only host-sets this is **identical** to the prior per-host result; it becomes correct automatically once **#1505** maps each app's full asset/CDN host-set. This fix **consumes** that set rather than hardcoding the apex.
- `extraAllowed` beats the whole-MAC block at the router (#421). **The router is unchanged** — it's a dumb applier and already carves every `extraAllowed` host (`render_spec` locks in the multi-host carve under a `TimeLimit` block).
- Non-exempt apps are unaffected (regression guard retained).
- The `decide` per-host fallback is intentionally left as-is: it's not the enforcement plane (the router uses the snapshot) and stays correct for today's apex-only data.

## Secondary finding (observability)

The prod block surfaced **no blocked Connection Events**. `render_spec` confirms the whole-MAC `TimeLimit` drop carries `log group 1 counter drop comment "wh_drop:<mac>:TimeLimit"`, so the nft drop **is** instrumented — the gap is in how the agent surfaces whole-MAC / IP-only drops as Connection Events. Tracked by the enforcement drop-counter work in **#1325**; not expanded here.

## Tests (TDD: red → green)

- `PolicySnapshotAppsSpec` (real embedded Postgres):
  - exempt multi-host app **UNDER** its own limit → **ENTIRE host-set** in `extraAllowed` under the daily block (reachable).
  - exempt multi-host app **OVER** its own limit (each host individually under) → **NOT carved** → blocked. *(This is the red commit — fails on the old per-host logic.)*
  - existing non-exempt / single-host #1105 / #1307 / #1413 guards still pass (20/20).
- `openwrt/test/render_spec.lua`: the whole-MAC `TimeLimit` drop carries one `ip daddr != @ea_<m>_<host>` clause **per host** (v4 + v6) — 176/0.

The red commit (`af807bc3`) precedes the fix commit (`a21c01b7`) so the progression is visible.

Refs #1513. Depends on #1505 (full app host-set beyond the apex).

### DIFF
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -681,18 +681,42 @@ object PolicyService {
         Hostname.unsafe(sl.domainPattern)
     }
 
-    // #1105: time_limited app hosts with exemptFromDaily=true carve around the
-    // MAC-level @blocked_macs drop while they still have per-host budget. The
-    // exempt flag's original role was just to exclude the host from the daily
-    // tally; without this carve-out, hitting the profile cap silently dropped
-    // the exempt app too, violating the "Khan doesn't count" contract.
-    // Naturally transitions allow → block as the per-host budget exhausts
-    // (siteLimitExtraBlocked above takes over and extraAllowed-beats-extraBlocked
-    // at the router; see feedback_extraallowed_beats_blocked).
-    val appExemptAllowedHosts: List[Hostname] = state.perSite.collect {
-      case sd if sd.exemptFromDaily && sd.usedMinutes < sd.dailyLimitMinutes =>
-        Hostname.unsafe(sd.domainPattern)
-    }
+    // #1105/#1513: time_limited app hosts with exemptFromDaily=true carve around
+    // the MAC-level @blocked_macs drop while the app still has budget. The exempt
+    // flag's original role was just to exclude the host from the daily tally;
+    // without this carve-out, hitting the profile cap silently dropped the exempt
+    // app too, violating the "Khan doesn't count" contract.
+    //
+    // #1513: the carve gate AND the "under its own limit" budget check are
+    // evaluated at the APP level over the app's FULL host-set — NOT per host. One
+    // time_limited app emits one `SiteTimeLimit`/`SiteDayState` row PER host, all
+    // sharing the same `label` (= `app:<slug>`), daily limit, and exempt flag
+    // (see SiteTimeLimitRepoLive.listForProfile), so we group by `label` to
+    // recover the owning app. A per-host check was the prod bug: with an app's
+    // real asset/CDN hosts (#1505) it (a) splits the app's usage across hosts so
+    // its own limit never bites, and (b) — when over budget — would still carve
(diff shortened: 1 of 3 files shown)


sameerparekh:
Closing without merging — folding this into the existing app-aggregation thread instead of shipping a standalone interim fix.

The root cause is confirmed and stands: the exempt carve-out + the "under its own limit" budget check were evaluated **per host**, and with Math Academy's apex-only host-set the carve only protected the apex while the app's real CDN/asset traffic was dropped by the whole-MAC `TimeLimit` block.

But the correct fix — per-app cap evaluation over the app's full, gap-bridged host-set — is already scoped as a coordinated thread under #1510:

- #1514 — app-aware presence primitive (`appSecondsForProfile`, cross-host gap-bridged) — the correct per-app usage this PR only approximated by summing per-host minutes.
- #1515 — enforcement: per-app cap evaluation; block all app hosts together; `SiteTimeLimit`→per-app shape.
- #1516 — per-(profile, app, date) rollup (schema-only migration first).
- #1518 / #1519 — App-vs-Site naming pass + drop the synthetic "Other" bucket.

Shipping the interim grouping-by-`label` fix here would pre-empt #1514's design (cross-host gap-bridging is an open question in #1510) and overlap #1515/#1518. Per #1515, the enforcement fix lands there, sequenced after #1514.

The red→green TDD tests written here (multi-host exempt app: under-limit → full host-set carved; over-limit → blocked; render_spec multi-host `ip daddr != @ea_` carve) are on branch `claude/fix-1513-exempt-daily` for reuse in #1515's adoption PR.

Issue #1507, opened 2026-06-06
dashboard: surface 'background/infra excluded' on the per-device usage drill-in


sameerparekh:
Part of the **Tuning** epic #1445. Surfaced by the #1499 over-count analysis (see `docs/design/presence-tuning-overcount.md`, PR #1501).

## Ask
Once presence suppression is expanded (#1503), show the operator what was excluded and why on the per-device usage drill-in — e.g. a 'background/infra (not counted)' grouping alongside counted app/host time. Observability for the new suppression so an operator can sanity-check that the daily total reflects engaged use.

Follow-up; depends on #1503 landing. A new metric/panel should follow the repo's 'new functionality ships with metrics' + 'a new metric ships with its dashboard' rules.

Issue #1506, opened 2026-06-06
presence: make background-suppression app-aware — a host attributed to an active app is never suppressed as infra


sameerparekh:
Part of the **Tuning** epic #1445. Surfaced by the #1499 over-count analysis (see `docs/design/presence-tuning-overcount.md`, PR #1501).

## Problem / guardrail
When background/infra suppression is expanded (#1503) and a real foreground/engagement model lands (#842), there is a risk that a host an app genuinely depends on gets suppressed as "infra," re-opening the #1446 undercount.

## Ask
Feed app host-set membership into the suppression decision so a host attributed to a known/active app can **never** be dropped as background infra. This is the clean seam that lets the suppression work (#1503), the foreground model (#842), and the #1464 session-stitch coexist.

## Dependency
Blocked on the app host-set attribution work (#1505). Lower priority than #1503/#1504.

sameerparekh:
PR #1557

Issue #1504, opened 2026-06-06
per-site time limit undercounts: migrate display + enforcement off legacy bucket-max to the #1464 session-stitch model


sameerparekh:
Part of the **Tuning** epic #1445. Surfaced by the #1499 over-count analysis (see `docs/design/presence-tuning-overcount.md`, PR #1501).

## Problem
A site-limited app's own limit reads **0** despite real use. Prod: Math Academy shows `usedMins 0` against its 30m limit while `www.mathacademy.com` has ~16 min of stitched presence; the profile total is meanwhile 31/30.

Root cause is a **presence-model inconsistency**:
- Daily total → `Presence.totalMinutesByMac` = the #1464 **session-stitch** model (`api/src/routes/Routes.scala:1185`).
- Per-site limit → **both** display (`Routes.scala:1194`) **and** enforcement (`api/src/policy/PolicyService.scala:333`) → `Presence.patternMinutesByMac` (`api/src/presence/Presence.scala:354`) = the **legacy bucket-max** model (`Σ max(activeSeconds) per bucket`).

`patternMinutesByMac` was never migrated to the session-stitch primitive, so it still has the ~3.3× undercount that #1464 fixed for the daily total (`activeSeconds` bottoms at the 10s sample floor). This hits both directions: the operator sees 0, and enforcement lets the kid run well past 30 real minutes before the per-site block triggers.

## Ask
Migrate the per-site count — **display and enforcement** — to a pattern-scoped session-stitch (a pattern-filtered variant of `proportionalHostSeconds`): stitch the spans of rows whose host matches the limit's pattern, per device, combined per the profile's `crossDeviceOverlapMode`. This is the per-site counterpart of the #1464 daily-total fix.

## Guardrail
This is the **undercount** direction — reuse the #1464 primitive; do NOT apply background-suppression here (don't drop low-byte app requests). Composes with #1446.

## Tests (TDD)
Failing-first: a fixture where ~16 min of `www.mathacademy.com` presence reads ~16 (not 0) for the `mathacademy.com` site limit, and the enforcement decision blocks at the true 30-minute mark.

sameerparekh:
[...]

Issue #1503, opened 2026-06-06
presence/enforcement: unify infra-allow + heartbeat host lists into one canonical infra set, and expand it


sameerparekh:
Part of the **Tuning** epic #1445. Surfaced by the #1499 over-count analysis (see `docs/design/presence-tuning-overcount.md`, PR #1501).

## Problem
`PolicyService.infraAllowHosts` (`api/src/policy/PolicyService.scala:564`, from #1307/#1337/#1411) and the `heartbeatHostPatterns` in `household_settings` (seeded V24) are the **same conceptual set** — device-level infrastructure the user did not initiate — maintained as two hand-curated lists:
- infra-allow = carve these *out of the block* so apps work.
- heartbeat = carve these *out of presence counting* so they don't inflate `usedMins`.

They have **drifted**: the heartbeat list is missing most of what infra-allow already lists — notably `gvt2.com` (36% of the counted background rows in the #1499 prod sample) and the OCSP responders. That drift is the leak that makes the Kids profile over-count (~93% of counted daily-cap rows are background OS/telemetry/infra).

## Ask
1. **Unify**: one canonical infra-host list as the source of truth, consumed by both `PolicyService` (allow carve-out) and `Presence` (suppression in `countedRows`/`isHeartbeat`). Prefer doing this at the **code level** (have `Presence` consume the same canonical host set) to avoid a Flyway migration; if a DB seed is unavoidable, follow the migration-isolation two-PR rule.
2. **Expand** the unified list with the observed leaking infra classes: `*.gvt2.com`, `*.gvt3.com`, `*.ls.apple.com`, background `*.googleapis.com`, `*.nel.goog`, analytics SaaS (`launchdarkly`, `adobe` telemetry, `app-analytics-services.com`), OCSP/cert and safe-browsing hosts.

## Boundary (must preserve)
Device-level infra **only** — do NOT absorb per-app CDN/asset hosts (`*.akamai.net`, `*.fastly.net`). Those must **attribute to the app and count** (the app host-set issue), not be suppressed. This is the seam that keeps suppression from re-opening the #1446 undercount.

## Tests
Add a presence regression fixture: the #1499 background-heavy Kid-device day must drop toward observed engaged use, while the #1446 sparse-app fixtures still read correctly (suppression keys on host *identity*, never on low-byte/short-activity).

PR #1523, opened 2026-06-06
feat(#1505): aggregate the per-site time limit across an app's whole host-set

### LINKED ISSUE #1505: app attribution: map app host-sets beyond the apex so off-domain assets count toward the app + its site limit

sameerparekh:
Closes #1505. Surfaced by the #1499 over-count analysis (`docs/design/presence-tuning-overcount.md`, Fix C).

## Problem
App→host attribution already buckets off-domain asset/CDN hosts under their app on the traffic page (`app_hosts` + `HostMatch.lookupApex`). But the **per-site time limit** was synthesized as one `SiteTimeLimit` row per `(assignment × host)`, so a multi-host app got **N independent budgets**:

- Traffic to an off-domain asset host ticked **its own** 30 m budget (and rendered its own bar), not the app's single limit.
- The whole set was never blocked together — only the host whose own bucket-count crossed the cap landed in `extraBlocked`.

So the gap #1505 names ("off-domain assets count toward the app *and its site limit*") was on the **accounting** side, not the seed data — the templates already carry host-sets (e.g. `roblox.com` + `rbxcdn.com`). (Math Academy itself is genuinely apex-only: its assets are served on `mathacademy.com`; the #1499 day was all `www.mathacademy.com`. The deliverable is the mechanism that unblocks any app that *does* serve off-domain assets.)

## Change — per-site limit is now per-app, aggregated across the host-set
- **`Presence.patternGroupMinutesByMac`** — bucket-max minutes per `(mac, group)`, where a group is the app's host-set. A bucket counts **once** toward the app when **any** of its hosts match (no double-count within an app; independent across apps).
- **`TimeStatusService.{groupSiteLimits, siteDayStates}`** — collapse the per-host rows into one `SiteDayState` per app carrying the full host-set + aggregated usage; `fold`/`assemble` emit one bar per app.
- **`PolicyService.computeBlockRules`** — when an app's aggregate hits its limit, **all** its hosts go to `extraBlocked`; while under cap an `exemptFromDaily` app carves out its **whole** set. `decide()` mirrors this (`byApp` + a label-keyed `timeLimitBlockFromDb`).
- **Routes** — per-profile and per-device time status exempt **and** count each app's whole host-set from the daily cap, one `SiteUsage` row per app.

Still **bucket-max** counting — migrating this surface to the #1464 session-stitch model is **Fix D (#1504)**, which composes on top of this.

## Coordination
- App host-sets are the app's **own dependency domains** and stay **distinct** from the device-level infra-allow list (#1337/#1411) / the #1503 infra+heartbeat unification: app assets *attribute and count*, device infra is *allowed and suppressed*. No host-set is defined twice.
- **Unblocks #1506** — the attribution result (a host belonging to an active app) is available on the presence/policy path, so #1506 can ensure an app-attributed host is never suppressed as infra.
- **Unblocks #1507** — the dashboard can surface the per-app aggregated limit.
- Does **not** regress #1446/#1464/#1465 (session-stitch presence is untouched; this is the under-count-safe per-site direction) or the #1499 undercount fixes.

## Wire contract
No wire change. The router still receives `extraBlocked` / `extraAllowed` host strings; the snapshot/usage shapes are unchanged.

## Tests (TDD red→green, real embedded Postgres)
- `AppHostSetAttributionSpec` — off-domain asset host counts toward the app's **single** limit (one bar, 20/30 not two 10/30 bars) and stays exempt from the daily cap; plus a group-by-app attribution guardrail (asset host → app, not "Other"). The first test is committed red before the implementation.
- `PolicySnapshotAppsSpec` — exhausting an app's budget on **one** host now blocks the **whole** host-set; under-budget keeps the set reachable.
- `PresenceSpec` — unit coverage for `patternGroupMinutesByMac` (aggregate across hosts, dedup within a group, independent across groups, no-match).

Affected specs run green individually: `AppHostSetAttributionSpec`, `PolicySnapshotAppsSpec`, `PresenceSpec`, `TimeApiSpec`, `TimeStatusServiceSpec`, `TimeStatusCacheSpec`, `TimeUsedRollupSpec`, `PolicySnapshotConsolidationSpec`, `RouterDecisionSpec`, `RouterDecisionPerAppScheduleSpec`, `PolicySnapshotPerAppScheduleSpec`, `UsageApiSpec`. (The local full-suite run hit unrelated embedded-Postgres `initdb` flakiness from parallel forks + a pre-existing timing flake in `MetricsExportSpec` that also fails on `origin/main`.)


sameerparekh:
Added a docs-only commit reframing the model as **app-focused, not site-focused** (`AGENTS.md`, `docs/architecture.md`):

- Usage and time limits are framed around **apps**; an app's limit is its whole host-set aggregated as one budget.
- A host in **no** configured app is its **own single-host app** — not a member of a catch-all "Other" app. "Other" exists only as a **display rollup** of the long tail (top-N + remainder); a host there is *low on the list*, not *part of an Other app*.

The broad code/identifier rename (`SiteTimeLimit`/`SiteUsage`/`perSite` → app, SPA `siteUsage` field) **and** the single-host-app rework of the `__other__` membership are tracked in **#1526** (Tuning epic) — kept out of this PR to stay focused and non-breaking. The wire/persisted `site_time_limit:` reason token stays frozen until wire versioning (#376).

Issue #1572, opened 2026-06-07
router: traffic/usage records still attributed to resolved CNAME target, not queried brand (incomplete #1344 — best-effort alias recovery)


sameerparekh:
## Summary
Traffic / usage records (per-(mac,host) `traffic_reports` + `time_usage`) are **still being attributed to the resolved CNAME target** (opaque CDN edge names) instead of the brand the client queried — the exact problem #1344 set out to fix. #1344's alias-based head-recovery landed and helps, but it is **best-effort and incomplete in practice**, so the pollution persists. Surfaced while root-causing the #1569 usage-ingest incident.

## Live evidence (from `root@router.lan`, 2026-06-07)
The agent's in-memory IP→host cache (`/tmp/wifihaven-dns-cache.txt`, the map that drives usage/connection attribution) currently holds:

- **1319** total IP→host entries
- **~33** whose host is an unmistakable **CDN-edge CNAME target**, not a brand a user would type, e.g.:
  - `dn0qt3r0xannq.cloudfront.net`, `difbymfufwhaj.cloudfront.net`, `a4de38…profile.fco50-p4.cloudfront.net`
  - `media.amazon.map.fastly.net`, `paypal-dynamic.map.fastly.net`
  - `eyaqfachaahtejqacqcxiaaacvvclw2f-p7o7rm-ee23fcb59-clienttons-s.akamaihd.net`

dnsmasq confirms these arrive as **direct queries to the CDN target** (the client re-queries the final CNAME target as its own lookup — the Apple/CDN behaviour #1346 describes):

```
5371 192.168.10.133/59331 query[A] media.amazon.map.fastly.net from 192.168.10.133
5371 192.168.10.133/59331 reply media.amazon.map.fastly.net is 151.101.129.16
711  192.168.10.160/61558 reply dn0qt3r0xannq.cloudfront.net is 3.169.202.101
```

## Why #1344's fix doesn't fully hold
`dns_log.lua` does the right thing *by design*: on a `reply`, `head = resolve_head(pending[qid].qname or r.name)` and it stores the IP against the recovered branded **head**, learning CNAME→brand alias edges as chains are observed (#1344). But recovery is only as good as the **alias map**, which is:

- **in-memory only** — lost on every `wifihaven-dns-tail` / agent restart;
- **TTL-bounded** (`ttl_seconds`, 3600s) and **LRU-capped** (`max_aliases`, 10k) — edges expire/evict;
[...]

Issue #1126, opened 2026-05-27
wire nflog production reader in wifihaven-agent (#1122 follow-up)


sameerparekh:
Follow-up to #1122. That PR adds:
- `log group 1 counter drop comment \"wh_drop:<mac>:<reason>\"` on every per-MAC drop rule in render.lua.
- `openwrt/files/usr/lib/lua/wifihaven/nflog.lua` — parser + dedup + event builder with an injectable `read_fn`.

What's still missing: a non-blocking production reader that drives nflog.run() inside wifihaven-agent's cooperative on_tick loop alongside conntrack.watch.

Options:
1. **ulogd2 sidecar** — install on the OpenWRT image, configure NFLOG group 1 → logfile, tail the logfile in the agent.
2. **nflog2syslog tiny C helper** — bind libnetfilter_log, emit one syslog line per packet, agent tails logread.
3. **log prefix instead of log group** — switch render.lua to `log prefix \"wh_drop:...\" level info counter drop`. Loses the netlink-group multiplexing but works with stock OpenWRT and \`logread -f\`.

Recommendation: (3) for the initial rollout — minimum new dependencies, the comment+prefix encoding survives, and the agent already understands logread.

Acceptance:
- [ ] Blocked-MAC traffic appears in /api/logs?blocked=true within one flush window of the drop on a real router.
- [ ] e2e scenario scripts/e2e/scenarios/test_blocked_mac_events.py passes against a live VM stack.
- [ ] No new daemons except those installed by the package itself.

sameerparekh:
Confirmed still-needed in prod (see #1524): the Kid device's connection_events on 2026-06-06 captured only 3 background domains; an exempt app (Math Academy) blocked under a whole-MAC daily-limit drop produced NO blocked connection events — its blocked traffic was invisible. Wiring the nflog production reader is the fix so enforcement drops surface in Connection Events (and feed the #1338 most-recently-blocked panel).

Issue #864, opened 2026-05-22
ops: metric for traffic_reports rows filtered as zero-bytes/zero-seconds


sameerparekh:
## Goal

When [#858](https://github.com/wifihaven/wifihaven/issues/858) is rooted out (or even before), expose a counter / log signal for the rate at which the server is dropping \`traffic_reports\` rows with bytesIn = bytesOut = activeSeconds = 0. That way if the agent-side regression filed in #858 ever returns silently, the operator can see it.

## Context

PR for #846 adds a server-side filter that drops these rows from the Traffic Usage page (\`UsageTraffic.cleanRows\`). Each request logs a warning with the drop count. There's a TODO marker pointing here for the proper metric once the observability stack lands.

## What to build

- Pick a metrics framework once one exists (#TBD — observability has no umbrella yet).
- Add a \`traffic_reports_filtered_zero_bytes_total\` counter incremented per filtered row.
- Alert / dashboard panel if the rate exceeds some threshold (operator-tunable).

## Acceptance

- Filter site in [api/src/usage/UsageTraffic.scala](api/src/usage/UsageTraffic.scala) reports the counter and the TODO comment is removed.
- Dashboard shows the metric.

## Related

- #858 (agent-side root cause investigation)
- #846 (Traffic Usage page that surfaces the rows)

PR #1574, opened 2026-06-07
fix(#1569): log usage decode failures + skip/meter malformed records instead of dropping the batch

### LINKED ISSUE #1569: INCIDENT(prod): /api/router/usage failing — investigate (DB contention from recent rollup/decide merges? #809-class) + restore ingest
## PROD INCIDENT — `/api/router/usage` failing
Grafana alert (router usage endpoint failing): https://wifihaven.grafana.net/goto/sbd7br?orgId=stacks-1674139

Impact: router usage ingest failing ⇒ **per-(mac,host) traffic + time-usage data loss** while it persists, and time-limit/cap accounting drifts. Treat as a deploy-blocking-class incident.

## What I ruled in/out from the code (origin/main)
- The ingest handler `RouterIngestRoutes.handleUsage` maps decode/validation errors to **400**, so a *failing* (5xx / timeout) endpoint is an unhandled defect OR downstream DB failure, not a bad-request. The ingest write path is just `trafficRepo.insertBatch` + `applyDelta` (time_usage + device last_seen) — unchanged recently.
- **#1510 (PR #1558, just merged) does NOT touch the ingest route.** It added a periodic rollup JOB (`TimeUsedRollupJob`/`AppUsedRollupService`/`AppUsedRollupRepo`, table `app_used_daily` V53) and a read accessor. Its read reuses the EXISTING `trafficRepo.listPresenceRowsSince/listPresenceRows` (same query the profile rollup uses), so it's probably not a brand-new unindexed query — but it DOES add DB work per job tick.
- **#1544 (decide() collapse)** made `PolicyService.decide()` read `TimeStatusService.todaysState` — extra DB reads on the per-host decision path.

## Ranked hypotheses
1. **DB CPU / connection-pool exhaustion (the #809/#1254 pattern).** The cluster of time-accounting merges (#1510 per-app rollup job + #1544 decide() DB reads) raised DB load at prod scale; Postgres CPU pegs / HikariCP pool saturates, and the **highest-frequency** endpoint (`/api/router/usage`, ~60s × every router) is the first to fail/timeout. CLAUDE.md "Validate query performance before merge" exists because this exact pattern took prod down on 2026-05-31. **Most likely.**
2. **Slow ingest tipping into timeouts** — #1511 already noted usage ingest is ~1s; added DB contention pushes it past a client/proxy timeout → "failing".
3. **Genuine 5xx defect in the ingest write path** — less likely (unchanged), but possible if a recent migration/constraint made an insert fail (verify no recent ALTER on traffic_reports/time_usage).
4. **Rollup job error** — forkDaemon, shouldn't 5xx the request, but if it holds connections/locks it feeds (1).

## Diagnostics needed (require prod/Render/Grafana access — operator or a chip with creds)
- The exact Grafana metric: is it **5xx rate** (defect) or **latency/timeout / request-duration** (contention)? Status-code breakdown for route `/api/router/usage`.
- **Onset time** vs deploy timeline: did failures start right after #1510 (#1558) or #1544 deployed? (This alone narrows it.)
- Render logs for `/api/router/usage`: stack trace / error text.
- Postgres: CPU %, active connection count, HikariCP pool saturation/wait, slow-query log; check for a seq-scan/expensive plan from the rollup job at prod row counts.

## Fast mitigation (if timeline correlates with a deploy)
Revert the correlated PR (#1558/#1510 first suspect, else #1549/#1544) — CD redeploys image-(N-1); usage ingest is more important than the new per-app rollup. Confirm onset-vs-deploy timing before reverting.

## Fix paths (per root cause)
- Contention/CPU: EXPLAIN the rollup-job + decide() queries at prod scale; add the missing index (separate schema-only PR per migration-isolation) or throttle/space the rollup job; consider connection-pool sizing. Capture before/after EXPLAIN.
- Defect: reproduce with a feature test against the ingest route (red→green) and fix.

## Note
This is also a gap in our process: #1510/#1544 should have had EXPLAIN-at-prod-scale validation per CLAUDE.md before merge (and the PR-review gate #1563 just landed to enforce it). If root cause is (1), fold the learning into the review checklist.

sameerparekh:
## Incident fix — restore `/api/router/usage` ingest (awaiting expedited review/merge)

Fixes the live ingest loss in #1569. **This is a prod incident fix — please prioritize review/merge.**

### Root cause (confirmed from live router logs, `root@router.lan`)
`POST /api/router/usage` decoded the body as a whole `UsageReport` and mapped the decode error **straight into the 400 response body without logging it**. When a single record carried a host value that fails `Hostname` validation, the **entire batch 400'd and every valid record was dropped** — and the offending field was invisible server-side (the sibling `/events` handler has logged decode failures since #1126; usage never got it).

The offending value, pulled from the router's own `usage.post` log:

```
.records[132].host(invalid hostname label
  '73-169-39-14_s-23-196-4-147_ts-1780860759-clienttons-s' in
   73-169-39-14_s-23-196-4-147_ts-1780860759-clienttons-s.akamaihd.net)
```

An Akamai CDN CNAME-target hostname with **underscores** in a label. `Hostname.LabelPattern` (`^[a-z0-9]([a-z0-9-]{0,61}[a-z0-9])?$`) rejects it. On prod this stuck whole usage buckets in the agent's in-memory retry queue (observed at retry attempt 6→10, never succeeding), losing per-(mac,host) traffic + time-usage data for those windows.

### What this PR changes
1. **Logs the usage decode failure.** A genuinely unparseable envelope still returns 400, but now emits a warning with router id, body length, a bounded 200-char snippet (no full body / no extra PII), and the zio-json error — mirroring the #1126 events pattern. Timestamp and router-id-mismatch 400s now log too.
2. **Tolerates a single bad record instead of failing the whole batch.** The envelope (`routerId`, `periodStart`, `periodEnd`) + `records` is decoded as a raw JSON array (`RawUsageReport`), then each record is decoded individually. Malformed records are **skipped + logged + metered**; valid records ingest normally. Idempotency / `ON CONFLICT DO NOTHING` and the #864 zero-byte filter are unchanged.
3. **New bounded metric** `usage_records_rejected_total{reason}` (enum: `decode_error` today) via `AppMetrics`/`MetricGuard` (entry added to `MetricGuard.Allowed`), plus two panels on the `data-quality-ingest` Grafana dashboard (timeseries by reason + 24h stat).

### Wire compatibility
Decode is made **strictly more lenient** — additive per the back-compat "tolerate unknown/partial input" rule. No snapshot/request/response shape change; older agents are unaffected, and a well-formed envelope with all-valid records behaves exactly as before.

### Tests (TDD red→green in history)
`api/test/src/feature/RouterIngestSpec.scala`:
- batch with one malformed record (the real prod akamaihd.net host) → **200**, valid record ingests into `traffic_reports` + `time_usage`, `usage_records_rejected_total` += 1.
- all-malformed batch (well-formed envelope) → **200** no-op-ingest, meters each rejected record.
- malformed envelope (bad timestamp) → **400** + a warning is logged + no record-reject metric charged.
- non-JSON body → **400** + envelope-deserialization warning logged (incl. `router=<id>`).
- all existing `RouterIngestSpec` / `UsageApiSpec` / `UsageRoutingSpec` / `RouterMetricsIngestSpec` stay green.

Grafana gate: `terraform fmt -check` + `terraform validate` + `python3 -m json.tool` on every dashboard all pass.

### Follow-up
This restores ingest and surfaces the offending field; it does **not** fix the mis-attribution that produced the bad host. That's #1572 — traffic/usage records still attributed to the resolved CNAME target instead of the queried brand (incomplete #1344).

Closes #1569.

### DIFF
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -76,6 +76,11 @@ object MetricGuard {
     "auth_failures_total"                       -> Set("reason"),
     "agent_connected_routers"                   -> Set.empty[String],
     "traffic_reports_filtered_zero_bytes_total" -> Set.empty[String],
+    // #1569 — usage-ingest records dropped at decode. A single malformed record
+    // (e.g. a host that fails Hostname validation — a CDN CNAME target with
+    // underscores, see #1572) is skipped + metered here instead of 400-ing the
+    // whole batch. `reason` is a small fixed enum (currently just `decode_error`).
+    "usage_records_rejected_total"              -> Set("reason"),
     // #1318 — global-policy-layer visibility. `global_allow_hosts` is the size of the
     // security-sensitive fleet-wide always-reachable set (a host here bypasses every block);
     // `default_deny_profiles` counts profiles running the block-all baseline. Both are unlabelled
@@ -236,6 +241,19 @@ object AppMetrics {
           Map.empty,
           rows.toLong,
         ),
+      )
+      .unit
+
+  // ── #1569: usage records dropped at decode ───────────────────────────────────
+  // A malformed record (host failing Hostname validation, etc.) is skipped at
+  // ingest instead of failing the whole batch. A rising rate flags an agent/DNS
+  // source emitting host values the API rejects (the #1572 CNAME-target case).
+  // `reason` is a small fixed enum — only `decode_error` today.
+
+  def recordUsageRecordsRejected(count: Int, reason: String = "decode_error"): UIO[Unit] =
+    ZIO
+      .when(count > 0)(
+        MetricGuard.counter("usage_records_rejected_total", Map("reason" -> reason), count.toLong),
       )
       .unit
 
--- a/api/src/routes/RouterIngestRoutes.scala
+++ b/api/src/routes/RouterIngestRoutes.scala
@@ -8,6 +8,7 @@ import wifihaven.shared.types.*
 import zio.{Clock as _, *}
 import zio.http.*
 import zio.json.*
+import zio.json.ast.Json
 
 import java.time.{Duration, Instant}
 
@@ -34,20 +35,48 @@ object RouterIngestRoutes {
       Method.POST / "api" / "router" / "usage"  ->
         handler { (req: Request) =>
           for {
-            router   <- auth.authenticate(req)
-            body     <- req.body.asString.orElseFail(Response.badRequest(""))
-            rep      <- ZIO
-              .fromEither(body.fromJson[UsageReport])
+            router <- auth.authenticate(req)
+            body   <- req.body.asString.orElseFail(Response.badRequest(""))
+            // #1569: decode the *envelope* (routerId, periodStart, periodEnd) +
+            // `records` as a raw JSON array, deferring per-record validation. A
+            // genuinely unparseable envelope (bad routerId/timestamps, not an
+            // object, `records` not an array, truncated body) still 400s — but
+            // now LOGGED (the prior code dropped the zio-json error straight into
+            // the 400 body and never logged it, so the offending field was
+            // invisible server-side; mirrors the #1126 events-handler pattern).
+            raw    <- ZIO
+              .fromEither(body.fromJson[RawUsageReport])
+              .tapError(e =>
+                ZIO.logWarning(
+                  s"router usage: envelope deserialization failed for router=${router.id} " +
+                    s"bodyLen=${body.length} bodySnippet=${snippet(body)} err=$e",
+                ),
+              )
               .mapError(e => Response.badRequest(e))
-            _        <- ZIO
+            _      <- ZIO
               .fail(Response.badRequest("router_id mismatch"))
-              .when(rep.routerId != router.id)
-            ps       <- parseInstant(rep.periodStart)
-            pe       <- parseInstant(rep.periodEnd)
+              .when(raw.routerId != router.id)
+            ps     <- parseInstant(raw.periodStart, router.id)
+            pe     <- parseInstant(raw.periodEnd, router.id)
+            // #1569: decode each record individually. A single malformed record
+            // (e.g. a host value that fails Hostname validation — an Akamai/CDN
+            // CNAME target with underscores, see #1572) used to fail the WHOLE
+            // batch, dropping every valid record with it. Now the bad records are
+            // skipped, logged (bounded), and metered, and the valid ones ingest.
+            decoded  = raw.records.zipWithIndex.map((j, i) => (i, j.as[UsageRecord]))
+            rejected = decoded.collect { case (i, Left(err)) => (i, err) }
+            records  = decoded.collect { case (_, Right(r)) => r }
+            _        <- ZIO.foreachDiscard(rejected) { (i, err) =>
+              ZIO.logWarning(
+                s"router usage: skipping malformed record[$i] for router=${router.id}: $err",
+              )
+            }
+            _        <- AppMetrics.recordUsageRecordsRejected(rejected.size)
             _        <- ZIO.logDebug(
-              s"router usage: router=${router.id} period=$ps..$pe records=${rep.records.size}",
+              s"router usage: router=${router.id} period=$ps..$pe records=${records.size} " +
+                s"rejected=${rejected.size}",
             )
-            _        <- ZIO.foreachDiscard(rep.records)(r =>
+            _        <- ZIO.foreachDiscard(records)(r =>
               ZIO.logDebug(
                 s"  usage record: mac=${r.mac} ip=${r.ip.getOrElse("-")} " +
                   s"host=${r.host.value} secs=${r.activeSeconds} bIn=${r.bytesIn} bOut=${r.bytesOut}",
@@ -58,7 +87,7 @@ object RouterIngestRoutes {
               router.id,
               ps,
               pe,
-              rep.records,
+              records,
               settings,
               trafficRepo,
               timeUsageRepo,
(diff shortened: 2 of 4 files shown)


sameerparekh:
## Independent review — PR #1574 (#1569 usage-ingest decode resilience)

Read-only adversarial pass against the STANDARD PR REVIEW checklist. Ran `RouterIngestSpec` (41 passed), validated the dashboard JSON, and verified the TDD red→green history.

### BLOCKERS
None.

### SHOULD-FIX
None.

### NITS
1. **`data-quality-ingest.json` — whole-file reformat inflates the diff.** The dashboard was re-emitted in `python3 -m json.tool` expanded style, so ~530 of the 560 changed lines are pure whitespace/`\uXXXX`-escape churn around the two real new panels (ids 9, 10) and the description string. Semantically identical (verified valid JSON; em-dash/arrow `—`/`→` are escape-only), but it buries the substantive change and complicates future blame. Not worth re-rolling now; just flagging the pattern.
2. **`RouterIngestRoutes.scala:72` warn-log includes the offending host value** (`...skipping malformed record[$i]...: $err`, where `$err` is the zio-json message containing the rejected hostname, e.g. the Akamai CNAME target). This is intentional (the whole point is surfacing the offending field per #1572) and bounded — it's the single bad field, not the body — so acceptable. The envelope-failure path correctly uses the truncated `snippet(body)` (200-char cap) rather than the full body. Calling it out only so the PII tradeoff is on the record.

### Checklist findings
1. **Single source of truth — PASS.** `RawUsageReport` (RouterIngestRoutes.scala:158-163) carries the identical envelope fields (routerId/periodStart/periodEnd) as `UsageReport` (Models.scala:1276) with `records: List[Json]`. Per-record decode is `j.as[UsageRecord]` (line 65) — the *same* derived `JsonDecoder[UsageRecord]`, so unknown-field tolerance is preserved (confirmed: existing "clockSkewSeconds is discarded" test still passes through the new path). No re-derivation of any decision/quantity.
2. **Testing integrity — PASS.** Red commit `9de9a97` touches **only** RouterIngestSpec.scala (123 lines, no impl); green commit `9e96f3e` adds the impl; `1b2598d` is scalafmt-only. No existing assertions weakened. New behavior is asserted concretely: 200 + valid record ingested into both traffic_reports and time_usage + counter +1 (one-bad-record test); all-bad → 200 no-op + counter +2; bad-timestamp envelope → 400 + warn logged + counter **unchanged** (+0); non-JSON → 400 + envelope-decode warn logged. Counter delta is read via before/after off the live registry — correct given the shared cumulative counter.
3. **Architectural invariants — PASS.** No router-side decision logic; wire vocabulary unchanged (`RawUsageReport` is a private server-side parse aid, explicitly documented as not a wire type); ZIO typed errors throughout; no `sys.env`/hardcoded config.
4. **Wire/back-compat — PASS, strictly more lenient.** `RawUsageReport` has **no** `@jsonNoExtraFields` (neither does `UsageReport`/`UsageRecord`), so unknown input still tolerated. A previously-all-valid batch decodes identically and ingests the same records (rejected list empty, counter +0). Change is additive/more-permissive: no rename/remove/retype. Empty `records:[]` → 200 no-op (envelope parses). `records` not an array → envelope decode fails → 400 (correct).
5. **Migrations — PASS.** None present.
6. **Metrics + dashboard — PASS.** `usage_records_rejected_total` routed via `AppMetrics.recordUsageRecordsRejected` → `MetricGuard.counter`, label bounded to `reason` only (no per-mac/host/ip); `MetricGuard.Allowed` updated (Metrics.scala). Two Grafana panels added targeting the exact emitted series — PromQL `sum by (reason) (rate(usage_records_rejected_total{env="$env"}[5m]))` and `sum(increase(...[24h]))` match code (`reason="decode_error"`). Counter name is not `_total`-suffixed by the connector → name in code == name in query. JSON valid, panel ids unique.
7. **Scope/hygiene — PASS.** Diff limited to the 4 intended files. No drive-by logic edits, no leftover debug logging (the added logs are intentional and leveled Warning/Debug). `snippet` truncation actually bounds to 200 chars. No Claude/Anthropic/AI mentions in commits or PR body.
8. **Security — PASS.** No secrets; ingest write path (`handleUsage`) unchanged, so SQL still parameterized and `ON CONFLICT`/idempotency + zero-byte filter intact for the records that do ingest. `records.zipWithIndex` index is used **only** in the warn-log string — dropping a rejected record does not shift attribution of the surviving records (they're collected independently via `collect { case Right }`). bodySnippet bounds length.

A regression test would fail without the fix: pre-fix the whole-`UsageReport` decode 400s on the bad record (no ingest, no counter, no envelope log) — exactly what the red commit demonstrated.

### VERDICT: APPROVE
No blockers. Two nits (dashboard reformat churn; intentional offending-host value in the warn log) are non-gating.

3-line summary:
- Correctly converts a single-bad-record whole-batch 400 into skip+log+meter while ingesting the valid records, with the envelope still 400-ing (now logged) — strictly more-lenient, fully back-compatible, no wire-shape or router-policy change.
- Clean TDD red→green history, concrete assertions (ingest + counter delta + log presence/absence), metric+dashboard shipped together with matching PromQL, no migrations, no scope creep.
- Only nits: the dashboard JSON was reformatted whole-file (diff noise), and the per-record warn log intentionally surfaces the offending host value (bounded, by design for #1572).

sameerparekh:
## Independent review (re-run) — PR #1574 (#1569 usage-ingest decode resilience)

Re-reviewed after the new push. Only commit since my prior review (`1b2598d`) is `41ef0a1` — `style(#1569): minimize data-quality dashboard diff to the two new panels`. Confirmed PR scope via three-dot merge-base (`git diff origin/main...HEAD`): touches exactly the **4 intended files** (`Metrics.scala`, `RouterIngestRoutes.scala`, `RouterIngestSpec.scala`, `data-quality-ingest.json`) — no unrelated skill-file deletion (that was a two-dot artifact from `main` advancing).

### Prior-findings status

| # | Prior finding | Severity | Status |
|---|---------------|----------|--------|
| 1 | `data-quality-ingest.json` re-emitted in whole-file expanded style — ~530 lines of whitespace/escape churn burying the 2 real panels | NIT | **ADDRESSED** — `41ef0a1` re-rolls the edit as a surgical insertion of panels 9/10 in the file's original hand-formatting. Incremental delta is now `+98/-422`, the reformat is gone. |
| 2 | warn-log at `RouterIngestRoutes.scala:72` includes the offending host value (intentional per #1572, bounded to the single bad field) | NIT (informational) | **UNCHANGED** — impl untouched by this push; tradeoff stands as previously noted. |

### Verification of the dashboard fix
- **JSON valid** — `python3 -m json.tool` passes.
- **Panels still target the emitted series** — panel 9 `sum by (reason) (rate(usage_records_rejected_total{env="$env"}[5m]))`, panel 10 `sum(increase(usage_records_rejected_total{env="$env"}[24h]))`. Byte-identical expr/title/id to the prior-reviewed version (verified via canonical-JSON compare of ids 9/10).
- **Metric name + label match code** — `Metrics.scala:256` emits `usage_records_rejected_total` with `reason` (allowlisted at `Metrics.scala:83`); `reason` is a bounded enum (`decode_error`). Group-by stays low-cardinality. ✓
- **Style commit is semantically a no-op** — canonical-JSON diff of `1b2598d...HEAD` shows **only** two cosmetic prose tweaks in the panel `description` strings ("usage ingest"→"usage-ingest", "pair with"→"pair it with", backticks around `reason`, and the 24h description rewritten to lead with the literal `sum(increase(...[24h]))` query — matching the dashboard's own convention). No change to any `expr`, `title`, `id`, `gridPos`, or panel structure.

### New findings (incremental delta `1b2598d...HEAD`)
None. The delta is the single dashboard file and is a pure reformat-plus-doc-copy improvement.

### Checklist (re-confirmed against current HEAD)
- **Single source of truth — PASS.** No code change; per-record decode still reuses the derived `JsonDecoder[UsageRecord]`.
- **Testing integrity — PASS.** No test edits in this push; red→green history intact.
- **Architectural invariants / wire compat — PASS.** `RawUsageReport` remains a private server-side parse aid, strictly-more-lenient, no `@jsonNoExtraFields`.
- **Metrics ship with a dashboard — PASS, now cleanly.** `grafana-terraform` gate (`json.tool`) passes; panel JSON is hand-formatted and minimal.

### VERDICT: **APPROVE**

The one outstanding nit from the prior pass is resolved; no new issues. Ready to merge.

Issue #1570, opened 2026-06-07
refactor: standard centralized error handler across all endpoints (consistent logging/status/body/metrics) — replace per-route bespoke handling


sameerparekh:
## Ask
Introduce a **standard, centralized error handler applied across all endpoints** instead of per-endpoint bespoke error handling, so logging, status mapping, response-body shape, and metrics are consistent for every route.

## Why (motivated by the #1569 incident)
Error handling is hand-rolled per route: `ErrorMapper.dbErrorToResponse` exists but is called manually, and there are ~hundreds of ad-hoc `.mapError(...)` / `Response.badRequest(e)` / `orElseFail(...)` sites across all 14 files in `api/src/routes/` (Routes.scala alone has ~172). There is no error middleware at the server boundary. Consequences:
- **Inconsistent logging** — POST `/api/router/events` logs decode failures (`tapError` → `logWarning`, #1126); POST `/api/router/usage` does NOT — it returned the decode error only in the 400 body, so the **#1569 incident's failing field was invisible in our logs**. This is the same drift class as #1532 (the same job done differently in two places).
- **Inconsistent status mapping & body shape** — each route decides its own status codes and error payloads; clients (router agent, SPA) see non-uniform error bodies.
- **Inconsistent metrics** — error responses aren't uniformly metered by route/status.

## Design
A single error-handling layer applied to ALL routes (ZIO HTTP middleware / `HandlerAspect`, or a shared `Routes` transformer applied where routes are combined in Main), that:
1. **Maps errors → Response in ONE place.** Centralize/extend `ErrorMapper` into a single `errorToResponse` (covering typed domain errors, decode/validation failures, DB throwables, auth failures). Routes should `ZIO.fail` TYPED sealed errors and let the handler map them, instead of constructing `Response.badRequest`/`notFound` inline. (Aligns with the AGENTS.md "domain errors as sealed traits" + the single-source-of-truth convention #1561.)
2. **Logs consistently:** 4xx (incl. decode/validation) at WARN with context (route template, method, request id, body length + short snippet for decode failures — bounded, no PII dumps); 5xx at ERROR with the cause/stack. Uniformly closes the #1569-class gap.
3. **Meters consistently:** error responses counted by templated route + status via AppMetrics/MetricGuard (bounded labels only — route/status; never per-mac/host/ip). Ships/extends a Grafana panel per the metrics rule.
4. **Standard error body shape** for consistency.

## CRITICAL — wire/back-compat
[...]

Issue #1561, opened 2026-06-07
docs(AGENTS.md): codify 'single source of truth — never duplicate a decision/computation path' (anti-divergence convention)


sameerparekh:
## Ask
Add a durable **"Single source of truth — do not duplicate decision/computation paths"** convention to `AGENTS.md` / `CLAUDE.md` (the agent + contributor instructions) so future work doesn't reintroduce the divergence class that has repeatedly shipped prod bugs.

## Why
Duplicated logic drifts: one copy gets updated, the others don't, and the surfaces silently disagree. Recent prod incidents were all this class:
- **#1531** — displayed daily total didn't subtract exempt-from-daily app time while the cap-evaluation path did (display vs enforcement).
- **#1539** — SPA schedule chip read the dead legacy `schedules` table while enforcement read named schedules (display vs enforcement).
- **#1532 audit** found more, now being collapsed: `decide()` hand-mirrored the whole snapshot fold (#1544, fixed), block-reason strings in 3 hand-written copies (#1545, fixed a live mislabel), per-device totals recomputed off the canonical total (#1546, fixed a live Dedup over-count).

The codebase already self-documents the smell with comments like "must mirror X", "same branch as snapshot", "keep in sync."

## What to add (new subsection under "Coding conventions")
Proposed heading: **"Single source of truth — never duplicate a decision or computation."** Content:
- The same logical quantity/decision (e.g. "minutes used today", "is this MAC blocked", "the block reason", "engaged seconds for an app", a wire reason string) must be computed in exactly ONE place; all other consumers call it.
- Before adding a new path that computes such a value, search for the existing one and reuse it.
- Resolve unavoidable proximity by **COLLAPSE** (extract a shared function / have one path call the other) or **TYPE-ENFORCE** (a shared sealed type / wire round-trip the compiler keeps exhaustive — e.g. `BlockReason.asWire`/`fromWire`). Treat any "must mirror" / "keep in sync by hand" comment as a defect to remove, not a pattern to follow.
- If two paths genuinely must stay separate, **ACCEPT + TEST-PIN**: add a test that fails when they diverge.
- Carve-out: intentional **wire-shape** redundancy the architecture mandates is NOT a violation (the infra allowlist copied into every profile's `extraAllowed`, #1311; the profiles map as a wire dedup aid). Reference the existing AGENTS.md callout so the two ideas don't conflict.
- Cross-link the audit umbrella (#1532) as the worked example / backlog of outstanding collapses.

[...]

Issue #1532, opened 2026-06-06
audit: find duplicated decision/computation paths that must be manually kept in sync (divergence bug class — #1531, #1513)


sameerparekh:
## Why
A recurring bug class: the same logical computation is implemented in **two or more places** that must be kept in sync by hand, a change lands in one but not the others, and the surfaces silently disagree. Most recent instances:

- **#1531** — the displayed per-profile \"used / daily\" total did NOT subtract `exemptFromDaily` app time, while the cap-evaluation path (`TimeStatusService.usedSecondsForProfile`) did. Display said 45/30; enforcement correctly didn't block. Two totals, one contract, drifted apart.
- **#1513** — the exempt carve-out was applied apex-only on the enforcement side while attribution used the whole host-set. Same root: parallel paths, one updated, one not.

These aren't isolated; the codebase already self-documents the fragility. A grep for the manual-sync seams turns up:
- `PolicyService.scala:292` — \"`#751: same branch as snapshot — keeps decide() consistent with the snapshot's cap evaluation`\" (the snapshot path and the live `decide()` path each re-derive the cap independently).
- `TimeStatusService.scala:360` — \"`Block precedence (mirrors PolicyService.computeBlockRules): Paused > Schedule > TimeLimit`\" (precedence ordering duplicated in two functions).
- `TimeStatusService.scala:411,435` — the `#1160` \"source-of-truth invariant\": cached-read path and live path \"stay structurally identical\" — by hand.
- `routes/Routes.scala:1223` — \"`must mirror the SQL's group_key concatenation order exactly`\".
- `routes/Routes.scala:863` and `:1112` — `exemptPats = ...filter(_.exemptFromDaily)` open-coded at multiple call sites (the seam #1531 fell through).
- `routes/BlockedRoutes.scala:93` — block-reason literals \"mirror the literals emitted by `PolicyService.decide`\".

Every \"mirrors X\" / \"must match X exactly\" / \"same branch as Y\" comment is a divergence waiting to happen. The comment is doing the job a shared function or a type should be doing.

## Scope of the audit
Survey the codebase for logic that is **duplicated where it shouldn't be** — specifically the cases where a future change *should* propagate to all copies but nothing forces it to. Prioritize the highest-blast-radius surfaces:

[...]

PR #1577, opened 2026-06-07
refactor(#1570): centralized error handler (Stage 1 — boundary handler + ingest routes)

sameerparekh:
## What

Stage 1 of #1570: introduce a **single error-handling layer at the server boundary** so every error response is **mapped, logged, and metered consistently**, replacing per-route bespoke handling. Migrate the **router ingest routes** (`/api/router/usage` + `/api/router/events`) as the first consumers — they are the #1569 incident surface.

Closes the structural gap behind #1569 and is anti-divergence infrastructure alongside #1532 / #1561.

## Why (#1569)

Error handling was hand-rolled per route. `POST /api/router/events` logged decode failures (`tapError → logWarning`, #1126); `POST /api/router/usage` did **not** — it returned the decode error only in the 400 body, so **the #1569 incident's failing field was invisible in our logs**. Same job, two implementations → drift (the #1532 class).

## The central map / log / meter

- **Map — one place.** `ErrorMapper.errorToResponse` covers the typed `ApiError` vocabulary (`BadRequest`, `DecodeFailure`, `Unauthorized`, `Forbidden`, `NotFound`, `Db`, `Internal`, `Wrapped`). Routes `ZIO.fail` a typed error and let the mapper produce the `Response`, instead of constructing `Response.badRequest`/`notFound` inline.
- **Log — one place.** `ErrorBoundary.observe` wraps the real API routes (mirrors `HttpMetrics.instrument`: per-route transform → reads the templated path), inspects the final response, and logs **4xx at WARN / 5xx at ERROR** with the templated route, method, status, and a **bounded one-line body snippet**. The snippet carries the decode error's failing field — this is what closes the #1569 visibility gap. It sits **inside** the readiness gate (no startup-503 log spam) and excludes health / metrics / SPA static (their own concerns; asset 404s aren't API errors).
- **Meter — one place.** `api_errors_total{route,status}` via `AppMetrics`/`MetricGuard` (bounded labels only — added to `MetricGuard.Allowed`), plus a Grafana panel ("API error rate by route + status") on the `api-self-metrics` dashboard.

## Wire back-compat (CRITICAL — preserved)

Error responses are part of the cross-process wire contract. Audited consumers first:
- **OpenWRT agent** branches on **status only** — 4xx = drop (no retry), 5xx / conn-error = retry/back-off. Does **not** parse error bodies.
- **SPA** reads the error body as **plain text** for display and sniffs the 403 body for `password_change_required`.

This PR changes **no status codes and no body shapes**. Every `ApiError` case reproduces the exact `Response` the hand-rolled code produced (same status, same constructor). The DB case stays **503 + `{"status":"error","db":"<class>"}` + `Retry-After: 30`**. A dedicated test pins these shapes.

## Built on top of #1574 (the #1569 hotfix) + logging dedupe

This branch is **rebased onto #1574**, the per-record-skip hotfix that merged after it was cut. The usage handler **keeps** #1574's behavior — `RawUsageReport` envelope decode, per-record `as[UsageRecord]` skip, the `skipping malformed record[i]` WARN log, and `usage_records_rejected_total` metering — and only routes the **error channel** through `ApiError`/`ErrorMapper`. The per-record skip stays inline (a bad record returns 200, so the boundary never sees it).

With the boundary as the single emitter, the now-redundant inline error logging is removed: #1574's **envelope** + **timestamp** decode `logWarning`s (and its `snippet` helper), plus the events route's decode `logWarning`, routerId-mismatch `logWarning`, and "returning status=" `logInfo`. Those 400s are now logged once, at WARN, by the boundary — with the response-body snippet (the failing field). `RouterIngestSpec.buildRoutes` wraps the routes in `ErrorBoundary.observe` (production-faithful), and #1574's two envelope/timestamp "logs a warning" tests were updated to assert the boundary's WARN log instead of the removed inline strings — behavior preserved, emitter relocated.

## Tests (TDD — red commit first)

`ErrorBoundarySpec`:
- `errorToResponse` maps each `ApiError` class to its preserved status + body; the DB case preserves the 503 + `{status,db}` body + `Retry-After`.
- Boundary logs 4xx at WARN (with body snippet) / 5xx at ERROR, and meters `api_errors_total`; observes errors on both the success and error channel; leaves 2xx alone.
- **Regression pin for #1569**: a `/api/router/usage` decode failure is **logged at WARN** (not just 400-bodied) and metered, while the 400 body is unchanged for the client.
- An invalid router bearer → 401, logged at WARN.

Full `api.test` suite green. `python3 -m json.tool` passes on all dashboards (no `.tf` changed, so `terraform fmt/validate` is unaffected). `scalafmt --check` clean.

## Stage 2+ — migrate remaining route files off bespoke handling

Each migration must NOT change observable status/body for existing consumers; one PR per file (or small group), each with tests.

- [ ] `Routes.scala` (~172 inline sites — the bulk; likely split further)
- [ ] `UsageRoutes.scala`
- [ ] `ProfileRoutes.scala` / `DeviceRoutes.scala` / `ScheduleRoutes.scala`
- [ ] `AuthRoutes.scala` (note SPA 401 redirect + 403 `password_change_required` sniff)
- [ ] `AppRoutes.scala` / `AlertRoutes.scala` / `GlobalPolicyRoutes.scala` / `HouseholdSettingsRoutes.scala`
- [ ] `LogRoutes.scala` / `DashboardNowRoutes.scala` / `BlocklistRoutes.scala`
- [ ] `RouterRoutes.scala` / `AdminRouterRoutes.scala` / `RollupAdminRoutes.scala` / `RouterMetricsRoutes.scala`
- [ ] `BlockedRoutes.scala` / `VersionRoutes.scala` / `DebugRoutes.scala`
- [ ] Migrate `RouterAuth.authenticate` to a typed error (currently bridged via `ApiError.Wrapped`)
- [ ] Optionally enrich 5xx logs with the cause/stack once typed errors carry the throwable to a logging mapper

Refs #1570, #1569, #1532.

### DIFF
--- a/api/src/ErrorBoundary.scala
+++ b/api/src/ErrorBoundary.scala
@@ -1,0 +1,101 @@
+package wifihaven.api
+
+import wifihaven.api.metrics.AppMetrics
+import zio.*
+import zio.http.*
+
+/**
+ * #1570: the single error-handling layer at the server boundary. Mirrors [[metrics.HttpMetrics]] —
+ * it transforms each route individually so it can read the route's own templated path
+ * (`routePattern.pathCodec`), then inspects the response the handler ultimately produced (success
+ * OR error channel) and, when that response is an error (status >= 400), LOGS and METERS it
+ * uniformly.
+ *
+ * This is what makes error handling consistent regardless of how a route produced the response:
+ *   - migrated routes fail with a typed [[routes.ApiError]] that
+ *     [[routes.ErrorMapper.errorToResponse]] maps; un-migrated routes still build a `Response`
+ *     inline. Either way the boundary sees the final response and treats it the same.
+ *
+ * It closes the #1569 gap structurally: POST `/api/router/events` hand-logged decode failures but
+ * POST `/api/router/usage` did not, so the incident's failing field was invisible in our logs. Now
+ * every 4xx — including a decode 400 whose body is the zio-json error naming the failing field — is
+ * logged once, here, at WARN; every 5xx at ERROR. The per-route bespoke logging is removed so there
+ * is exactly one emitter (no double-logging, no drift).
+ *
+ * Logging policy:
+ *   - 4xx (incl. decode/validation/auth) → WARN
+ *   - 5xx (incl. the DB 503) → ERROR
+ *   - context: templated `route`, `method`, `status`, and a bounded one-line snippet of the
+ *     response body. The body of an error response is our own message (the decode error, the
+ *     "router_id mismatch" text, the `{"status":"error","db":...}` JSON) — never request PII — and
+ *     it carries the diagnostic detail, so a small truncated snippet is logged. Reading is guarded
+ *     by `knownContentLength` (our error bodies are small, complete, in-memory strings) and any
+ *     read failure degrades to no-snippet, so it can never disturb the response sent to the client.
+ *
+ * Metering: `api_errors_total{route,status}` via [[AppMetrics.recordHttpError]] (bounded labels
+ * only — the templated route + the status code; never a per-mac/host/ip value).
+ *
+ * Like HttpMetrics, an unmatched request never reaches a transformed handler, so a raw
+ * attacker-controlled path can never become a label.
+ */
+object ErrorBoundary {
+
+  /** Max characters of an error-response body logged as context. Bounded — no PII dumps. */
+  private val MaxSnippet = 256
+
+  /** Don't even attempt to read a body larger than this as a snippet. */
+  private val MaxReadBytes = 64L * 1024L
+
+  private def renderTemplate(rp: RoutePattern[?]): String =
+    rp.pathCodec.render(":", "")
+
+  def observe(routes: Routes[Any, Response]): Routes[Any, Response] =
+    Routes.fromIterable(routes.routes.map(observeOne))
+
+  private def observeOne(route: Route[Any, Response]): Route[Any, Response] = {
+    val routeLabel = renderTemplate(route.routePattern)
+    val method     = route.routePattern.method.render
+    route.transform[Any] { handler =>
+      Handler.fromFunctionZIO[Request] { req =>
+        handler(req).either.flatMap { result =>
+          val response = result.merge
+          val observe  =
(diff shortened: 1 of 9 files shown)


sameerparekh:
Rebased onto latest `origin/main` to build **on top of** #1574 (the #1569 per-record-skip hotfix), which merged after this branch was cut. An independent review caught that the original branch's whole-batch usage decode would have reverted #1574's per-record-skip-and-meter fix on merge.

The reconciled usage handler now **keeps** #1574's behavior — `RawUsageReport` envelope decode, per-record `as[UsageRecord]` skip, the `skipping malformed record[i]` WARN log, and `usage_records_rejected_total` metering — and only routes the **error channel** through `ApiError`/`ErrorMapper.errorToResponse`. The per-record skip stays inline (a bad record still returns 200, so the boundary never sees it). The now-redundant inline **envelope** + **timestamp** decode `logWarning`s (and the `snippet` helper) are removed — the boundary logs those 400s once, with the response-body snippet (the #1569 dedupe: single emitter).

`RouterIngestSpec.buildRoutes` now wraps the ingest routes in `ErrorBoundary.observe` (production-faithful); #1574's two envelope/timestamp "logs a warning" tests were updated to assert the boundary's WARN log instead of the removed inline strings — behavior preserved, emitter relocated. Full `api.test` green (incl. #1574's per-record tests, now 41 in RouterIngestSpec).

sameerparekh:
Extended the boundary to cover the **SPA catch-all** (`StaticRoutes`), not just the API route families. Rationale: an unmatched `/api/*` path 404s there (`no such API route`) and a malformed URI 400s — both are real API errors, and excluding them also left `api_errors_total` asymmetric with `http_requests_total` (which already counts SPA responses via `HttpMetrics.instrument`). The SPA fallback serves `index.html` (200) for client routes, so there's no asset-404 flood, and the `route` label is the bounded trailing template (`/...`) so no per-path leak. Only `/api/health` (constant-poll 503-while-starting) and `/metrics` (scrape) remain excluded. Added a test pinning that an unmatched `/api/*` 404 is logged at WARN + metered through the boundary. Full `api.test` green.

Issue #1519, opened 2026-06-06
usage-by-app presentation: per-app rows + per-orphan-site rows, drop 'Other' bucket


sameerparekh:
Sub-issue of #1510 (App-level usage aggregation). Part of **Usage & Screen-Time**.

In the app table/graphs, list non-app hosts **individually, one row per site** — drop the synthetic "Other" bucket.

## Why
`buildUsageByApp` (UsageRoutes.scala) maps each host to its owning app and dumps every host **not in any app** into a single `appId = null` bucket named **"Other"** (`key.flatMap(appById.get).map(_.name).getOrElse("Other")`). That hides which individual sites a kid actually used outside defined apps.

## Ask
- App view = one row per app (aggregated + gap-bridged) **plus** one row per **orphan site** (named by its host/apex), with **no** "Other" catch-all.
- Orphan-site rows use the single-host engaged-minutes path (`proportionalHostSeconds`/per-host), so they're the "Site" case from the naming sub-issue.
- Stable ordering/tiebreak preserved; deterministic host→app attribution (lowest appId wins) unchanged for hosts that *do* belong to an app.

## Tests (TDD)
- Two apps + two orphan hosts → 2 app rows + 2 per-site rows, **no** "Other" row.
- A host matching an app's apex still attributes to the app, not a per-site row.

Issue #1514, opened 2026-06-06
App-aware presence primitive: cross-host gap-bridged appSecondsForProfile


sameerparekh:
Sub-issue of #1510 (App-level usage aggregation). Part of **Usage & Screen-Time**. Depends on the app-aware presence primitive sub-issue.

Make per-app time-limit **enforcement** evaluate the cap once per app against the gap-bridged app aggregate, and block all the app's hosts together.

## Why
`SiteTimeLimitRepoLive.listForProfile` (api/src/db/Repos.scala) fans each `time_limited` app into **one `SiteTimeLimit` per `app_hosts` row**, each carrying the app's `daily_minutes` and `domainPattern = ah.host`. `TimeStatusService.fold` then computes `usedMinutes` per `domainPattern` independently, so a multi-host app's cap is checked **per host** — host A can be under cap while host B is over, partially blocking one app incoherently.

## Ask
- Reshape `SiteTimeLimit` / `SiteDayState` to be **per-app**: carry the app's full host set (or app id + resolved hosts) instead of one row per host.
- `TimeStatusService.fold` computes per-app used-minutes via `appSecondsForProfile` (the primitive sub-issue), combined per `crossDeviceOverlapMode`.
- Cap evaluation is per app; when exhausted, **all** the app's hosts go into the block derivation together (`PolicyService` `extraBlocked` / snapshot + the `decide()` per-host fallback agree).
- Preserve `exempt_from_daily` semantics (carve from the daily total) at the app level.

## Compat
`SiteTimeLimit`/`SiteDayState` are **server-side only** (not in the router snapshot — the snapshot carries collapsed `BlockRules`), so this refactor is internal. Keep `DeviceTimeStatus`/`SiteUsage` API responses backwards-compatible or coordinate with the SPA (the naming sub-issue tracks renames).

## Tests (TDD)
- Multi-host app, usage split across two hosts summing past the cap → app blocked, **all** its hosts blocked; under cap → none blocked.
- `decide()` per-host fallback agrees with the snapshot for both hosts of the app.
- Exempt-from-daily app still carves from the profile daily total.

sameerparekh:
PR #1553

sameerparekh:
Landed via PR #1553 (merged). appSecondsForProfile is the single app-aware gap-bridged presence primitive; #1510 (rollup builder) and #1517 (graph) consume it.

Issue #1510, opened 2026-06-06
App-level usage aggregation & rollups: stitch across an app's hosts, per-app daily rollup, App-vs-Site naming, drop 'Other' bucket


sameerparekh:
Follow-up to #1504 (PR #1509). Part of the **Tuning** epic #1445.

#1504 fixed the per-site time-limit *counting model* (bucket-max → #1464 session-stitch) but **left the aggregation granularity at the host level**. This issue closes the gap: make usage **app-aware** end-to-end — counting, enforcement, rollups, and graphs — with correct naming and presentation.

## Background — where we are after #1504

A `time_limited` app is authored **once per app** (one `daily_minutes` on `app_policy_assignments`), but the read path fans it out **per host**:

```sql
-- SiteTimeLimitRepoLive.listForProfile (api/src/db/Repos.scala)
SELECT apa.profile_id, ah.host, COALESCE(apa.daily_minutes,0), a.slug, apa.exempt_from_daily
FROM app_policy_assignments apa
JOIN apps a       ON a.id = apa.app_id
JOIN app_hosts ah ON ah.app_id = apa.app_id
WHERE apa.mode = 'time_limited'
```

→ one `SiteTimeLimit` per `(app, host)`, each carrying the app's cap and `domainPattern = ah.host`.

`Presence.patternSecondsForProfile` then counts **each `domainPattern` independently**, and the daily total's `sessionSpans` stitches per `(mac, host)` then unions. So for a multi-host app:

1. The cap is checked **per host**, not against the union of the app's hosts (host A can be under, host B over → incoherent partial blocking of one app).
2. The 120s idle gap (`presenceContinuationSeconds`) **only bridges within a single host's stream**. Traffic to `hostA` then `hostB` of the *same app* within the gap is recorded as two separate sessions, not continuous app usage.

### 1. App is the stitch unit (cross-host gap-bridging)
For an app, stitch the **union of all the app's host rows on a device as one activity stream**, on the idle gap, per device — so traffic to `hostA` and `hostB` of the same app within `presenceContinuationSeconds` (currently 120s) counts as **continuous** usage. Then combine across the profile's devices per `crossDeviceOverlapMode`. The cap is evaluated **once per app** against this aggregate.

[...]

Issue #1492, opened 2026-06-05
presence/tuning: make per-profile usage graph presence-based (total doesn't match graph; show per-app contribution to time-used)


sameerparekh:
## Problem
In prod, a kid's profile shows **56m used** (the time-used total that drives time limits), but the **per-profile usage graph does not reflect that 56m** — the graph and the headline total disagree. Strong suspicion: **the usage graph is NOT presence-based** (it's plotting raw traffic bytes / a different aggregation), while the 56m total comes from the presence/time-used computation. So the operator can't see *which app usage added up to the total time used*.

## Ask
Make the **per-profile usage graph presence-based**, and break it down by app, so:
1. The graph's total visibly **reconciles with the headline time-used** (the 56m), and
2. You can see **which apps' usage contributed** the minutes that sum to the total — the app-by-app presence contribution over the day.

## Investigate
- Confirm what the per-profile usage graph currently plots (`web/src/...` profile/dashboard usage chart + the API endpoint it calls — likely `/api/usage/*` or `/api/usage/series` plotting bytes or raw-bucket minutes) vs what the 56m time-used number is computed from (the presence path, `api/src/presence/Presence.scala` / `time_used_daily`). Identify exactly why they diverge.
- Decide the right series: the graph should source from the **same presence/session model** as the time-used total (after #1464/#1465 land, that's the session-stitch model) so per-app minutes sum to the cap.

## Dependency / coordination
This is the **Tuning epic** and sits directly on top of the presence rework:
- #1464 (profile presence = session-stitch + interval-union) defines the total.
- #1465 (per-app presence = session span) defines the per-app breakdown the graph needs.
Build the graph against that model so total = sum-of-app-contributions (modulo the union/overlap semantics — surface concurrency honestly). If it lands before #1464/#1465, at minimum point the graph at the same source the time-used total uses so they stop disagreeing.

## Acceptance
- The per-profile usage graph is presence-based; its total reconciles with the headline time-used (the 56m case); per-app contributions are visible and explain the total. A feature/UI test pins that the graph total matches the time-used total for a known fixture.

Part of the Tuning epic #1445. Refs: #1446 (presence design), #1464/#1465 (presence impl), #1433 (per-app time-used on rows).

PR #1578, opened 2026-06-07
feat(#1517): per-app graph series from app-aggregated engaged minutes (reconcile with rollup + cap)

### LINKED ISSUE #1517: Graph-serving rollups/series reflect app-aggregated engaged minutes (reconcile with cap + #1492)
Sub-issue of #1510 (App-level usage aggregation). Part of **Usage & Screen-Time**.

Naming pass: use **App** wherever it's an app (a defined bundle, incl. multi-host); reserve **Site** for a **single-host** limit/series.

## Why
Current naming is muddled: `site_time_limits` table and `SiteTimeLimit`/`SiteDayState`/`SiteUsage` types actually describe **app** assignments (`app_policy_assignments`, mode `time_limited`), and the per-app read fabricates a `label = "app:<slug>"`. "Site" should mean single-host only.

## Ask
- Rename server-side model/repo/rollup identifiers to the App-vs-Site distinction (`SiteTimeLimit`→app-limit shape where it represents an app; keep a Site form only for genuine single-host limits).
- Align rollup table / series / metric names introduced by the sibling sub-issues.
- SPA-facing response fields (`ProfileAppUsage`, `SiteUsage`, `HostUsage`, `DeviceTimeStatus`) renamed only with SPA coordination — **additive/backwards-compatible** on the API↔SPA boundary (self-hosted bundles the SPA atomically; cloud rolls SPA/API independently). Stage renames behind compat where needed.

## Scope
Naming/shape only — behavior is owned by the aggregation/enforcement sub-issues. Land after those so the shapes are settled.

sameerparekh:
Closes #1517 (App-Centric Model epic; sub-issue of #1510).

## Problem

The per-app graph series (`GET /api/usage/series?groupBy=app`) was built in
`UsageSeries.buildEntries` by grouping each app's **per-host** stitched spans
and concatenating them (`_ ++ _`). That concatenation:

- never **bridges the cross-host idle gap** a multi-host app spans (the apex
  going quiet while an off-domain asset / CDN host carries the next request,
  #1505), and
- never **dedups within-app overlap** (two hosts of the same app active in the
  same window double-count).

So the per-app series disagreed with the **gap-bridged per-app cap aggregate**
(`TimeStatusService.siteDayStates` / `appSecondsByApp`) and the **#1510 per-app
daily rollup** — both of which read the single per-app primitive
`Presence.appSpansForProfile` (#1514/#1532).

## Fix

Derive each app entity's spans from that same single primitive
`Presence.appSpansForProfile` over the app's **whole host-set** — the
gap-bridged, overlap-deduped, cross-host union the cap and the rollup already
read. Non-app hosts still surface standalone with their per-host spans. The
per-hour **total** bar is unchanged (the device session union), so the
host/device axes and the #1492 headline reconciliation are untouched.

`buildEntries` now takes an `AppAxis` (per-host attribution + per-app host-set
patterns), built once from one `app_hosts` snapshot in
`UsageRoutes.loadAppLookup`, so attribution and the span host-set can't drift
(single source of truth, #1532). The app-time math is **consumed, not
re-implemented** (#1514/#1510).

## Reconciliation invariant (#1492 extended to the app dimension)

For a profile + date, per app:

> **per-app series ⇄ #1510 per-app rollup ⇄ per-app cap aggregate** — all read
> the same gap-bridged engaged seconds, floor-of-sum.

and, in the non-bridged / non-overlapping / non-exempt case, the chain also
reconciles with the **profile headline total** (the restricted equality
documented on `AppUsedRollupService` — under cross-host bridging a per-app
number can exceed the per-`(mac,host)` headline, which is expected).

## Tests (TDD — failing test committed first)

Two embedded-Postgres feature tests in `UsageApiSpec`, both reading the live
chain (the #1510 rollup accessor, the per-app cap aggregate, and the series
endpoint):

- **cross-host gap-bridge**: app `Social` = `social.com` + `cdn.social.net`,
  active 08:50–08:55 then 09:00–09:05 (300 s cross-host gap < effective gap).
  Series day total = **15** (bridged) = rollup = cap, **not** the per-host
  proportional sum (10); hour-clipping sums back (10 in h8 + 5 in h9).
- **within-app overlap**: both hosts active in the same three windows. Series =
  **15** (overlap-deduped), and the **full chain holds exactly**: profile total
  ⇄ rollup ⇄ cap ⇄ series, all 15, while the per-host sum (30) is what the old
  un-deduped series plotted.

The red commit shows the series returning `Some(10)` while the rollup/cap
already read 15 — pinpointing the series as the broken link.

## Wire compatibility

Response shape is **unchanged** (`topEntries` / `bucketsByEntry`) — additive,
the SPA is unaffected. No migration (the series is presence-based off raw
rows; it does not read rollup tables). No new metric series, so no dashboard
change.

## Deferred to siblings (out of scope here)

- App-vs-Site naming pass → #1518
- `Other` bucket handling → #1519

### DIFF
--- a/api/src/routes/UsageRoutes.scala
+++ b/api/src/routes/UsageRoutes.scala
@@ -177,7 +177,7 @@ object UsageRoutes {
             groupByApp = req.url.queryParam("groupBy").exists(_.split(',').contains("app"))
             appLookup <-
               if (groupByApp) loadAppLookup(appRepo)
-              else ZIO.succeed((_: HostId) => Option.empty[UsageSeries.AppInfo])
+              else ZIO.succeed(UsageSeries.AppAxis.empty)
             settings  <- hsRepo.get.mapError(ErrorMapper.dbErrorToResponse)
             resp      <- (macOpt, profileIdOpt) match {
               case (Some(mac), _) =>
@@ -240,7 +240,7 @@ object UsageRoutes {
             groupByApp = req.url.queryParam("groupBy").exists(_.split(',').contains("app"))
             appLookup <-
               if (groupByApp) loadAppLookup(appRepo)
-              else ZIO.succeed((_: HostId) => Option.empty[UsageSeries.AppInfo])
+              else ZIO.succeed(UsageSeries.AppAxis.empty)
             settings  <- hsRepo.get.mapError(ErrorMapper.dbErrorToResponse)
             resp      <- buildBatch(
               pids,
@@ -278,7 +278,7 @@ object UsageRoutes {
       userProfileRepo: UserProfileRepo,
       siteTimeLimitRepo: SiteTimeLimitRepo,
       groupByApp: Boolean,
-      appLookup: HostId => Option[UsageSeries.AppInfo],
+      appLookup: UsageSeries.AppAxis,
       settings: HouseholdSettings,
   ): IO[Response, UsageSeriesBatchResponse] =
     for {
@@ -464,28 +464,39 @@ object UsageRoutes {
       )
     }
 
-  // #1079 — build the host → owning-app lookup used by the unified-axis
-  // builder. Mirrors the deterministic "lowest appId wins" tiebreak from
-  // #1061 and the apex-aware match from #1085 so subdomain traffic attributes
-  // to the apex-form app entry.
+  // #1079 — build the by-app axis used by the unified-axis builder. `appOf` mirrors the
+  // deterministic "lowest appId wins" tiebreak from #1061 and the apex-aware match from #1085 so
+  // subdomain traffic attributes to the apex-form app entry. #1517: also carry each app's full
+  // host-set (`patternsBySlug`) so the per-app series spans are gap-bridged over the SAME host-set
+  // the per-app cap aggregate and the #1510 rollup use (via Presence.appSpansForProfile), keyed by
+  // slug. Both are built from one `app_hosts` snapshot so the attribution and the span host-set
+  // can't drift.
   private def loadAppLookup(
       appRepo: AppRepo,
-  ): IO[Response, HostId => Option[UsageSeries.AppInfo]] =
+  ): IO[Response, UsageSeries.AppAxis] =
     for {
       apps     <- appRepo.listAll.mapError(ErrorMapper.dbErrorToResponse)
       mappings <- appRepo.listAllHostMappings.mapError(ErrorMapper.dbErrorToResponse)
     } yield {
-      val appById = apps.iterator.map(a => a.id -> a).toMap
-      val byApex  = mappings
+      val appById        = apps.iterator.map(a => a.id -> a).toMap
+      val byApex         = mappings
         .groupBy(_.host.value)
         .view
         .mapValues(ms => ms.iterator.map(_.appId).minBy(_.value))
         .toMap
-      (h: HostId) =>
+      val appOf          = (h: HostId) =>
         h.asFqdn
           .flatMap(fqdn => HostMatch.lookupApex(fqdn.value, byApex))
           .flatMap(appById.get)
           .map(a => UsageSeries.AppInfo(a.id, a.slug, a.name, a.icon))
+      val patternsBySlug = mappings
+        .groupBy(_.appId)
+        .iterator
+        .flatMap { case (aid, ms) =>
+          appById.get(aid).map(a => a.slug -> ms.map(_.host.value).distinct)
+        }
+        .toMap
+      UsageSeries.AppAxis(appOf, patternsBySlug)
     }
 
   private def buildForDevice(
@@ -498,7 +509,7 @@ object UsageRoutes {
       trafficRepo: TrafficReportRepo,
       userProfileRepo: UserProfileRepo,
       groupByApp: Boolean,
-      appLookup: HostId => Option[UsageSeries.AppInfo],
+      appLookup: UsageSeries.AppAxis,
       settings: HouseholdSettings,
   ): IO[Response, UsageSeriesResponse] =
     for {
@@ -556,7 +567,7 @@ object UsageRoutes {
       userProfileRepo: UserProfileRepo,
       siteTimeLimitRepo: SiteTimeLimitRepo,
       groupByApp: Boolean,
-      appLookup: HostId => Option[UsageSeries.AppInfo],
+      appLookup: UsageSeries.AppAxis,
       settings: HouseholdSettings,
   ): IO[Response, UsageSeriesResponse] =
     for {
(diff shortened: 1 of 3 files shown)


sameerparekh:
## Review — PR #1578 (per-app graph series from app-aggregated engaged minutes)

Read-only review of `git diff origin/main...HEAD` (merge-base). Scope: 3 files, no migration, no router/agent code, no wire-shape change.

### Single source of truth (highest priority) — PASS
The per-app series now genuinely reads the single per-app primitive and does **not** re-implement gap-bridging/stitching. `UsageSeries.buildEntries` (`api/src/usage/UsageSeries.scala:302-311`) calls `Presence.appSpansForProfile(rows, <slug -> patterns>, overlap, filter, continuationSeconds)` — the same gap-bridged + overlap-deduped union the cap (`TimeStatusService.siteDayStates`→`patternGroupMinutesForProfile`→`appSpansForProfile`) and the #1510 rollup read. The old `groupMapReduce(_ ++ _)` per-host concat is gone.

Host-set provenance reconciles: the cap's host-set comes from `groupSiteLimits` over `SiteTimeLimitRepo.listForProfile`, which is synthesized from `app_hosts` (`api/src/db/Repos.scala:1097-1108`); the series' `patternsBySlug` is built from the same `app_hosts` via `appRepo.listAllHostMappings` grouped by `appId` (`UsageRoutes.scala:490-496`). Both attribution (`appOf`) and span host-set (`patternsBySlug`) come from one `app_hosts` snapshot in `loadAppLookup`, so they can't drift within a request. `apps.slug` is `UNIQUE` (V28), so the slug join key across `appsPresent`/`appSpans`/`app:<slug>` cap labels is collision-free. `overlap` passed to `buildEntries` is `profile.crossDeviceOverlapMode` — same mode the cap uses.

One benign asymmetry worth noting (not a defect): the cap only forms `app:<slug>` groups for apps with a `time_limited` assignment **to this profile**, whereas the series surfaces **all** apps (`listAllHostMappings`). So the reconciliation invariant only binds for time-limited apps; non-time-limited apps appear on the graph with bridged spans but have no cap to reconcile against. That is the intended behavior (the graph shows every app), and the tests exercise a time-limited app, so the asserted chain is the meaningful one.

### Double-surface / vanishing-app — PASS (examined `buildEntries`)
- **No double-surface.** An app-attributed host is excluded from standalone host entries by `case (h, spans) if axis.appOf(h).isEmpty` (`UsageSeries.scala:319-322`). App entries come solely from `appsPresent` (keyed by slug); host entries only when `appOf(h)` is empty. Clean partition — a host never surfaces both as an app member and standalone.
- **No silent vanish.** A present app with empty `patternsBySlug` would yield `Nil` spans → `daySecs=0` → filtered out. But this is impossible by construction: `appsPresent` is populated only via `axis.appOf(h)`, which matches through `byApex` built from the same `mappings` that build `patternsBySlug`. Any app that `appOf` returns therefore has ≥1 mapping ⇒ a non-empty `patternsBySlug` entry. `appOf`/`patternsBySlug`/`appSpansForProfile` are all apex-aware on the same apex-form hosts, so subdomain traffic matches consistently. The `AppAxis.empty` path is only hit when `groupByApp=false`, where `appsPresent` is empty too.

### Correctness of the bridging claim — PASS
Gap fixture: social.com 08:50–08:55 + cdn.social.net 09:00–09:05, 300s cross-host gap < effectiveGap(600s) ⇒ bridges to one [08:50,09:05]=900s=15min; hour-clip 10(h8)+5(h9). The per-app entity spans come from `appSpansForProfile` (bridged) while non-app hosts keep `hostSessionSpans` (per-host). Internal consistency with cap/rollup is guaranteed because all three read the same primitive, independent of the absolute bucket→span convention.

### Testing integrity — PASS (with one hygiene note)
- Red→green visible: `06e0209` (test only, +109/-1, the -1 is an `import` line) precedes impl `f06dfb9`.
- No **pre-existing** test assertion was weakened/deleted: the net `origin/main...HEAD` test diff has no removed assertions on existing tests (only import additions).
- New behavior is genuinely asserted: gap-bridged 15 vs per-host-prop 10, plus rollup==15, cap==15, hour-clip 10/5; and a second overlap fixture where the full chain (profile total ⇄ rollup ⇄ cap ⇄ series) holds exactly at 15 vs per-host concat 30.
- No forbidden mocks: real embedded Postgres, `TestClock`, no repo/AuthService/Clock mocks.

**SHOULD-FIX (TDD hygiene, non-blocking):** The red commit `06e0209`'s single test used the gap-bridge fixture yet asserted `presenceTotalMins == 15` (`UsageApiSpec.scala` red rev, line ~102). For the gap fixture the headline stitches per-`(mac,host)` and does NOT bridge the cross-host gap, so the headline is 10, not 15 — that assertion was factually wrong and the red test could not have gone green as authored. The green commit corrected it by dropping that assertion from the gap test and adding a separate **overlap** fixture where the headline genuinely is 15. This is a legitimate red-test correction (the gating assertion `seriesDayMins.contains(15)` is preserved and still fails against old code), not regression-hiding — but the red commit as committed was not a clean red. Ideally the red commit should model the headline correctly from the start.

### Wire compat — PASS
`UsageSeriesResponse` shape unchanged (`topEntries`/`bucketsByEntry`/`dayMins`/`presenceTotalMins` all intact). The `buildEntries` signature change (`appOfHost: HostId => Option[AppInfo]` → `axis: AppAxis`) is internal API, not on the wire. `AppAxis` is a new internal type. Additive-only; SPA unaffected.

### Migrations / Metrics / Security — PASS
No migration (correct — series is presence-based off raw rows). Pure refactor of an existing series-builder path; no new meaningful code path warranting a metric, so no dashboard change needed. SQL unchanged and parameterized; the series route's existing auth is untouched (tests still bearer-authenticate). No secrets.

### Scope / hygiene — PASS
Diff limited to the 3 intended files. No unrelated drive-by edits, no dead code, no debug logging, no TODO/FIXME. No Claude/Anthropic/AI mentions in commits or PR body.

---
### BLOCKERS
None.

### SHOULD-FIX
1. `api/test/src/feature/UsageApiSpec.scala` (red commit `06e0209`): the red test asserted `presenceTotalMins == 15` against the gap-bridge fixture where the headline is 10 — an incorrect assertion the green commit had to remove. Functionally fine post-fix (gating assertion preserved), but the initial red commit wasn't a clean red. Hygiene only.

### NITS
None.

---
## VERDICT: APPROVE

Per-app series now reads the single `Presence.appSpansForProfile` primitive — no re-implemented bridging, host-set provenance reconciles with cap/rollup, and the app/host partition is double-surface-free and vanish-safe by construction.
TDD red→green is visible and the new gap-bridge + overlap reconciliation is genuinely asserted with real embedded Postgres; one non-blocking hygiene note on an incorrect assertion in the red commit that green corrected.
Wire shape unchanged, no migration, scope tight, no AI mentions — clean.

Issue #1539, opened 2026-06-07
bug(prod): inconsistent schedule enforcement across profiles sharing the same named schedules — some 'paused for Schedule', some not, with 'Bedtime'+'Breakfast' both enabled


sameerparekh:
## Symptom (prod)
Several profiles each had the **same** named schedules attached as block/downtime rules — **'Bedtime'** and **'Breakfast'** — and during an active window **some profiles were paused for Schedule while others were not**. The window has since ended, so it is not currently observable.

## What I ruled out by reading `origin/main` (do NOT re-tread)
The schedule **enforcement logic is profile-uniform** — there is no per-profile branch that could legitimately produce different results for identical attachments:
- **Route** `PUT /api/profiles/{id}/schedules` (`Routes.scala` ~234) validates each id then `namedScheduleRepo.setProfileBlockSchedules(pid, scheduleIds)` — a full-set **replace** (delete all `blocked_during` for the profile, insert the given set). `GET /api/profiles[/{id}]` returns `blockScheduleIdsForProfile(pid)`.
- **Snapshot** (`PolicyService` snapshot build) takes each profile's `blocked` straight from `timeStatusService.dayStateAll(...)`. Both batch paths (`dayStateAllFromRollupHits` and `dayStateAllLive`) build `schedMap` identically: `syntheticWindows(p.id, windowsForAllProfiles.getOrElse(p.id, Nil))`, and `assemble` decides downtime via `schedules.exists(PolicyService.scheduleActiveAt(_, now))`. Schedule blocking is evaluated **live** in `assemble(now)` regardless of the rollup, so the rollup path can't desync it.
- **`windowsForAllProfiles` SQL** correctly joins `schedule_windows → profile_schedule_rules ON schedule_id` filtered to `mode='blocked_during'`, grouped by `profile_id`.
- **SPA write** (`ProfilesPage.tsx` `ScheduleSubsection`) attach/detach build the full next id-set from `pd.scheduleIds` with replace semantics and autosave via `api.profiles.setSchedules` — correct.
- `scheduleActiveAt` uses each window's own `tz`; a shared schedule has the same windows, so timezone can't make a shared schedule evaluate differently per profile.

**Conclusion: identical attachments → identical enforcement.** So the divergence is NOT in the evaluation. It is one of:

## Candidate root causes (for the investigation to discriminate)
[...]

PR #1598, opened 2026-06-09
test(#1547): pin no SPA surface renders pd.schedules

### LINKED ISSUE #1547: cleanup(divergence): stop shipping the dead legacy ProfileDetail.schedules field (test-pin now; emit Nil + remove with #1485)
Sub-issue of #1532 (divergence audit), **Finding 4 — MEDIUM, the #1539 trap half-closed.**

## Divergence
The server still computes and ships the dead legacy `ProfileDetail.schedules` field alongside the correct `scheduleIds`. The SPA was fixed (#1542) to read `scheduleIds` + named-schedule windows and ignore `pd.schedules`, so there is **no current user-visible bug** — but the stale field remains a loaded gun.

## Locations
- Still populated: `ProfileRoutes` fills `ProfileDetail.schedules` from `scheduleRepo.listForProfile` at Routes.scala ~206 (`GET /api/profiles`) and ~224 (`GET /api/profiles/:id`). These are the only remaining `ScheduleRepo.listForProfile` readers in api/src.
- Correct field: `ProfileDetail.scheduleIds` from `blockScheduleIdsForProfile` (Routes.scala ~208,226), enforced via `namedScheduleRepo.windowsForProfile`.

## Risk
The legacy `schedules` rows are stale and vary per profile. Any new SPA surface (or a careless revert) reading `detail.schedules` re-introduces the exact #1539 divergent-chip symptom.

## Fix — ACCEPT + TEST-PIN now, stop populating when the window allows
- The wire field can't be removed until the deprecation window (back-compat rules); the table drop is the planned destructive phase (#1485, gated ~Jun 12).
- **Now:** keep/extend the `ProfilesPage.test.tsx` "ignores the dead legacy schedules field" guard so no SPA surface renders `ProfileDetail.schedules`.
- **When the deprecation window allows (coordinate with #1485):** stop populating it server-side (emit `Nil`) so the trap can't fire even on a SPA regression, then remove the field in a later additive-safe step.

No behavior change now; this is a guard + a scheduled cleanup tied to #1485.

sameerparekh:
Closes #1547.

Sub-issue of #1532 (divergence audit) — Finding 4. Adds a cross-cutting guard to `ProfilesPage.test.tsx` that stuffs sentinel values into the dead legacy `ProfileDetail.schedules` field, expands the profile card, and asserts none of the sentinels (name, day token, start/end times, tz, id) reach the DOM.

The existing #1539 test pins one specific surface (the schedule chip); this broader trap catches a *new* surface that reads `pd.schedules` before its author notices. It is the guard the issue's "Now" phase asked for — the wire field stays until the #1485 deprecation window allows stop-populating + removal.

Verified no current SPA surface reads `pd.schedules`:
- `grep -rn "pd\.schedules\|detail\.schedules\|profile\.schedules" web/src` → only comments
- Router agent does not consume this field (router reads the policy snapshot, not `/api/profiles`).

No behavior change. No wire change.

## Test plan
- [x] `npx vitest run src/pages/ProfilesPage.test.tsx` → 90/90 green (89 prior + 1 new)
- [x] new `#1547` test fails if a SPA surface starts rendering any value from `pd.schedules`

### DIFF
--- a/web/src/pages/ProfilesPage.test.tsx
+++ b/web/src/pages/ProfilesPage.test.tsx
@@ -1892,4 +1892,53 @@ describe('ProfilesPage — schedule chip reads named schedules, not legacy (#153
     const chip = await screen.findByTestId('profile-pause-chip-1')
     expect(chip).toHaveAttribute('data-chip', 'active')
   })
+
+  // #1547 — broader guard: NO SPA surface may render any value derived from
+  // the dead legacy `ProfileDetail.schedules` field. The chip test above pins
+  // one specific surface (#1539); this one is the cross-cutting trap that
+  // catches a *new* surface that reads `pd.schedules` before its author
+  // notices. The legacy field stays on the wire only until #1485 drops it;
+  // until then this guard is what keeps a SPA regression from re-introducing
+  // the #1539 divergent-chip class of bug.
+  it('no SPA surface renders any value from `pd.schedules` (#1547)', async () => {
+    // Sentinels chosen so they cannot collide with any real label, day, time,
+    // tz, or id rendered elsewhere on the page.
+    const LEGACY_NAME  = 'ZZZ-legacy-schedule-name-1547'
+    const LEGACY_DAY   = 'qqq'        // not a valid weekday token
+    const LEGACY_START = '03:17'      // distinct minute, unlikely in fixtures
+    const LEGACY_END   = '04:23'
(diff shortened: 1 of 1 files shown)


sameerparekh:
Superseded by #1600 — that PR removes the dead `ProfileDetail.schedules` field outright from the wire model on both sides (Scala case class + TS interface + all fixtures). The test-pin from this PR becomes redundant: there is no field left to read, so no SPA surface can render its contents.

Issue #715, opened 2026-05-21
usage: per-FQDN time-on-site is bucket-max, not real time-on-site


sameerparekh:
## What

[`RouterIngestRoutes.applyDelta`](api/src/routes/RouterIngestRoutes.scala:141) groups by `(mac, host)` per batch and increments `time_usage.seconds_used` by **the max activeSeconds** in the bucket. That value is the same bucket-window flag (0 or 300, or 10s-multiples post #295) — not actual time spent on the host. Every host the device touched in a 5-min bucket gets credited with the whole bucket's duration.

This is correct for the per-device daily total (Presence collapses across hosts), but **wrong** for per-app/per-site graphs: if the kid had youtube.com open for 30 min and his phone touched icloud.com once in each of those 6 buckets, both rows show `seconds_used ≈ 30 min`.

## Evidence (prod, 2026-05-21, Sameer profile, `/api/time/status`)

Sameer Mac sat at 61 used minutes for the day. The per-host breakdown shows **10 different hosts all at 50–80 minutes each**:

```
inquisition.goguardian.com    81m
api.github.com                79m
prod-zephyr.goguardian.com    79m
bolt.dropbox.com              74m
nexus-websocket-a.intercom.io 59m
...
```

These are all polling APIs / websockets / dropbox sync. They are not 79 minutes of attention each. Per-app screen-time policy ("kid spent N minutes on YouTube") cannot be implemented on top of this.

## Why this matters

[#127](https://github.com/wifihaven/wifihaven/issues/127), [#301](https://github.com/wifihaven/wifihaven/issues/301), [#64](https://github.com/wifihaven/wifihaven/issues/64) all want per-app graphs. The current per-host numbers are unfit to drive a per-app limit decision.

## Proposal sketch

Two paths, not mutually exclusive:

1. **Foreground-host heuristic** in the agent: within a 10s sample, attribute the sample to the *single* (mac, dst_ip) pair with the most bytes/packets — not all of them. This is roughly what mobile OSes call "foreground app" detection. Implementable in [usage.lua:166](openwrt/files/usr/lib/lua/wifihaven/usage.lua:166) without wire changes.
2. **Server-side rule**: in the daily rollup expose two numbers per (mac, host): bucket-presence-minutes (current `seconds_used`) and proportional-time-minutes (bucket duration × this host's share of bucket bytes). Lets the UI choose.

[...]

PR #1611, opened 2026-06-10
feat(#842): router-side foreground-host heuristic for per-FQDN time-on-site

### LINKED ISSUE #842: router-side foreground-host heuristic for per-FQDN time-on-site (#715 Path 1)
Follow-up to #715. The server-side Path 2 has shipped — `/api/time/status` now exposes `proportionalMins` alongside `usedMins`, weighting each bucket's wall-clock duration by the per-host byte share. That fixes the screen-time UI for the prod case where 10 polling APIs all read as 60-minute hosts.

Path 1 from #715 is still worth doing eventually: in [usage.lua:166](openwrt/files/usr/lib/lua/wifihaven/usage.lua:166), within a 10s sample, attribute the sample to the *single* (mac, dst_ip) with the most bytes/packets — i.e. router-side "foreground app" detection. The bytes-share weight gets a sharper signal at agent time and the API-side proportional math becomes a more faithful surface of what actually happened on the wire (vs. a heuristic over after-the-fact buckets).

**Blocked on #654** — router agent is frozen until Gate 2 lands. Re-open once the freeze is lifted.

Tracked separately so #715 can close on the server-side fix.

sameerparekh:
Closes #842.

## What changes

Within each 10s sample tick in `usage.tracker_sample`, only the foreground (mac, dst_ip) — the flow with the largest byte delta for that MAC — earns the active-sample credit. Background flows that previously got a slice of every sample no longer inflate per-host time-on-site.

- Per-MAC: each device picks its own foreground dst_ip in the same tick.
- Tie-break: lexically smallest dst_ip wins (deterministic).
- Bytes accounting (`bytesIn` / `bytesOut`) is unchanged — only the seconds-attribution shifts.

The heuristic lives in one place inside `usage.lua` (per AGENTS.md §single-source-of-truth).

## Behavior change callout

This is a usage-attribution behavior change in the router agent. Already-stored historical `traffic_reports` rows are unaffected; only NEW samples shift. The change pairs naturally with the server-side `proportionalMins` shipped by #715 Path 2 — after a soak window in prod, operator may want to compare both metrics to decide which one becomes the canonical display surface.

## Validation

Ran the Lua busted suite on Mac:

```
$ LUA_PATH=… busted test/conntrack_spec.lua test/render_spec.lua test/policy_spec.lua test/usage_spec.lua …
636 successes / 0 failures / 0 errors / 1 pending
```

The pending case is `nft -c -f -` which needs the `nft` binary (not present on macOS).

**End-to-end validation needs the KVM runner.** Operator may want to run the Linux Gate-1/Gate-2 e2e suite manually before promoting to prod, or wait for the master-router CD to do it on merge — the unit-test coverage of the heuristic is solid but the wire shape was not exercised against a live router image here.

## Test plan
- [x] RED commit: three new busted cases describing the heuristic (largest-delta-per-MAC, per-MAC independence, deterministic tie-break)
- [x] GREEN commit: `tracker_sample` rewritten as per-MAC argmax-delta selection
- [x] Full Lua suite passes (636/0/0/1-pending)
- [ ] Linux Gate-1/Gate-2 e2e (operator / CD)

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/usage.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/usage.lua
@@ -281,26 +281,50 @@ function M.new_tracker()
   return { last = {}, active_samples = {} }
 end
 
--- Compare each counter to its previous byte total; an increase counts as one
--- active sample.  A counter that went DOWN is treated as a fresh start
--- (element expired and reappeared, or counters reset out-of-band); we still
--- count it as an active sample if the new value is > 0.
+-- Per sample tick, only the foreground (mac, dst_ip) — the one with the
+-- largest byte delta since the previous tick — earns an active sample for
+-- each MAC (#842 / #715 Path 1). Background flows that get a small slice of
+-- every sample no longer inflate per-host time-on-site; bytes accounting
+-- (build_report.bytesIn/bytesOut) is unchanged.
+--
+-- A counter that went DOWN is treated as a fresh start (element expired and
+-- reappeared, or counters reset out-of-band); its effective delta for the
+-- foreground contest is the new total.
+--
+-- Ties break on lexically smallest dst_ip so the choice is deterministic.
 function M.tracker_sample(tracker, counters)
+  local winner_key   = {}  -- mac → tracker_key of largest-delta flow
+  local winner_delta = {}  -- mac → delta of current winner
+  local winner_dst   = {}  -- mac → dst_ip of current winner (for tie-break)
+
   for _, c in ipairs(counters or {}) do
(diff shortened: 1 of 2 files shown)


sameerparekh:
Closing — operator feedback: byte-share is a bad foreground proxy. A background download can dominate bytes while the user is actively engaged with a low-bandwidth foreground app (reading, chat), and multi-tasking (music + browsing) loses the background flow's time entirely. The server-side proportionalMins from #715 Path 2 is the better surface; this heuristic would trade one bias for another. Closes #842 as won't-do.

Issue #1630, opened 2026-06-10
Collapse per-app disposition readers into one fold (#1629 root cause; #1532 audit miss)


sameerparekh:
**Rescoped 2026-06-10 (third time)** — earlier framings were too narrow. The fix isn't to unblock the SQL filter; it's to **collapse the two readers so the divergence can't recur**. This is the §single-source-of-truth treatment the #1532 audit was supposed to enforce, and didn't catch.

## The divergence

The same underlying object — "this profile's per-app dispositions" — is read by two paths that derive different facts from it:

- **`expandAppDispositions`** ([PolicyService.scala:821](api/src/policy/PolicyService.scala:821)) reads `List[AppPolicyAssignment]` (all rows, all modes) → folds `(mode, exemptFromDaily, dailyMinutes, schedules)` into `(extraAllowed, extraBlocked)`.
- **`exemptPatterns`** ([TimeStatusService.scala:559](api/src/policy/TimeStatusService.scala:559)) reads `List[AppTimeLimit]` (via [`AppTimeLimitRepoLive.listForProfile`](api/src/db/Repos.scala:1115), `WHERE mode='time_limited'`) → exempt host-set for daily total.

Both are decompositions of the same fact. They drift because each path has its own SQL filter and its own per-mode case-analysis. Today's bug — `mode=Allowed` apps reach `extraAllowed` but their `exemptFromDaily=true` doesn't reach `exemptPatterns` — is the exact "two readers, one filter changes, surfaces silently disagree" shape AGENTS.md §single-source-of-truth names. It is also what the #1532 audit umbrella exists to prevent.

Other readers in the same family that re-derive from various per-mode subsets and should fold into the same fix:

- `appHostPatterns` ([TimeStatusService.scala:608](api/src/policy/TimeStatusService.scala:608)) — union of every active app's host-set, also derived from `groupAppLimits(appLimits)`.
- `groupAppLimits` ([TimeStatusService.scala:417-ish](api/src/policy/TimeStatusService.scala)) — per-app cap groups.
- `appDayStates` / per-app cap evaluation — reads the same subset for the cap surface.
- `appSecondsByApp` ([:577](api/src/policy/TimeStatusService.scala:577)) — per-app engagement, derives groups from `groupAppLimits` too.

Every one of these is a projection of `List[AppPolicyAssignment]` plus the same per-profile per-app schedule windows.

## Fix shape — COLLAPSE, don't add

[...]

PR #1633, opened 2026-06-10
feat(#1628): blocked MAC produces zero usage records

### LINKED ISSUE #1628: bug: ProfileDayState.usedMinutes is not clamped at the cap when blocked=true, blockReason=TimeLimit (#1626 symptom 2)
Split from #1626 — root-caused via code read.

## Summary

After the daily-cap fires for a profile (\`blocked=true, blockReason=TimeLimit\`), the \"time used today\" counter shown on the dashboard keeps growing. The operator observed it growing by enough that >1h of extension was needed to compensate before the kids could use anything again — even though enforcement was correctly blocking forwarded traffic at the router. This is symptom 2 of #1626.

## Root cause

\`TimeStatusService.assemble\` (\`api/src/policy/TimeStatusService.scala:716\`) returns a \`ProfileDayState\` with two **independent** fields:

- \`blocked: Boolean\` / \`blockReason: Option[MacBlockReason]\` — computed from \`usedMinutes >= limit + extensionMinutes\` at \`:731\`.
- \`usedMinutes: Int\` — passed through verbatim at \`:752\` (= the raw aggregated total from presence rows / rollup).

There is no clamping. Once the cap fires, both the snapshot's \`blockReason\` and the dashboard's \`usedMinutes\` come from the same source (good — that was #1531's collapse), but the displayed counter still increments freely with any new traffic that arrives.

In practice new traffic DOES arrive: the carve-outs in \`extraAllowed\` (1password, kasandbox, kastatic, khanacademy, mathacademy in Prima's case) keep working past the cap; their byte / engagement contribution feeds back into \`usedMinutes\` via \`Presence.appSecondsForProfile\` / \`usedSecondsForProfile\`. The counter advertised to the SPA exceeds \`dailyLimit + extensionMinutes\` arbitrarily.

A reader of \`assemble\` would think \"the headline is the truth\" and that \`blocked\` is just a label. But once \`blocked=true, blockReason=TimeLimit\` is set, the displayed \`usedMinutes\` should freeze at \`limit + extensionMinutes\` — that's the contract the UI and the operator expect (\"how much extension do I need to grant?\").

## Why this matters (operator-facing impact)

The displayed counter is what the operator uses to decide how much extension to grant. If the counter shows \"180 min used / 120 min limit\" the operator thinks they need 60 min of extension just to break even — but the actual enforcement only ever blocked at the 120 min boundary, so the 60 min gap is illusory inflation. The operator ends up over-granting extension by the amount of \"phantom\" overshoot.

## Fix direction (do NOT implement here)

\`assemble\` (\`TimeStatusService.scala:725-758\`) clamps:

\`\`\`scala
val displayedUsed = (blocked, reason, dailyLimit) match {
  case (true, Some(MacBlockReason.TimeLimit), Some(limit)) =>
    limit + extensionMinutes  // freeze at the cap
  case _ =>
    usedMinutes
}
\`\`\`

and returns \`ProfileDayState(..., usedMinutes = displayedUsed, ...)\`. Keep the \"raw\" total internally only if a downstream consumer truly needs it (search the codebase — likely no consumer does; the cap-evaluation already uses the raw value before clamp).

## Acceptance

- Feature test: profile hits the cap, additional traffic from an \`extraAllowed\` host pushes raw aggregation to \`limit + 30\`; \`ProfileDayState.usedMinutes\` returned to the SPA is \`limit + extensionMinutes\` (clamped), \`blocked=true\`, \`blockReason=TimeLimit\`, \`remainingMinutes=0\`.
- Symmetric test for \`blockReason=Schedule\` and \`Paused\`: per \`AGENTS.md\`, those don't depend on \`usedMinutes\`, so they should NOT clamp (the raw used-today is still useful for the operator's history view).

## Related
- Parent: #1626
- #1531 — earlier display-vs-enforcement divergence in the same family
- #1532 — single-source audit umbrella; this is another instance.

Epic: Tuning.

sameerparekh:
Closes #1628.

## What

`usage.build_report` takes an optional `blocked_macs` table and drops any counter whose `mac` is in it; `wifihaven-agent` passes the shared `blocked_macs` ref (already maintained by `render.update_shared`) at every usage-timer tick.

## Why

The architectural contract from the issue is: **a `blocked=true` MAC produces zero new `traffic_reports` row growth and zero new `usedMinutes` accumulation**, regardless of `blockReason` (TimeLimit / Schedule / Paused / Manual).

Router-side investigation against live `openwrt.lan` (Prima `04:72:ef:d6:e4:5a`, `blocked=true, blockReason=TimeLimit`):

- `wifihaven_block` hooks `forward` at `priority filter` (0); `wifihaven_account_tx` / `wifihaven_account_rx` hook at `priority filter + 1`. `drop` at priority 0 terminates packet processing before the priority-1 accounting chains, so **dropped flows do NOT increment `mac_ip_tracking`** — confirmed by `nft list set inet wifihaven mac_ip_tracking` showing zero rows for the blocked MAC despite drop counters showing 3411 v4 / 2878 v6 packets dropped. Hypothesis 1's mechanism (drops inflating tx counter) does not occur.
- But the per-MAC drop rules carry `ip daddr != @ea_<mac>_<host>` carve-outs. **`ea_` carve-out traffic** (`1password.com`, `kasandbox.org`, `kastatic.org`, `khanacademy.org`, `mathacademy.com`) is legitimately allowed past the drop and DOES update the priority-1 counters. The agent ships those as `traffic_reports` rows for the blocked MAC; Presence stitches them (`continuationSeconds=120s`) into engaged-seconds against the daily cap — which is what the operator sees growing past the cap.

The `extraAllowed` carve-out exists for **functional access** (the kid can still hand in Khan / unlock 1Password after they've hit the daily limit), not to credit minutes against a daily limit they've already exhausted. Filtering at agent report time enforces the contract end-to-end: zero rows shipped, nothing for Presence to stitch across the block window, no need for any API-side display clamp or headline exclusion.

This also collapses the Hypothesis 2 concern (Presence stitching across the block window): with no agent-side rows during the block, there is nothing to bridge from.

## TDD

- **Red** (586285ff): `test/usage_spec.lua` adds a `build_report` case asserting that a counter for a MAC in `blocked_macs` is dropped while a counter for an unblocked MAC is still emitted, plus a back-compat check that `nil` blocked_macs is a no-op.
- **Green** (58e6b11c): `usage.build_report` and `wifihaven-agent` carry `blocked_macs` through.

## Tests

`sh openwrt/test/run_tests.sh` → 637 successes / 0 failures / 0 errors. No other module touched.

## Wire compatibility

Additive: `build_report` adds an optional 11th positional arg; `nil` is the old behavior. The API surface (`POST /api/router/usage`) is unchanged — this PR just emits fewer rows for blocked MACs.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/usage.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/usage.lua
@@ -54,7 +54,7 @@
 --
 --   usage.build_report(counters, nft_sets, period_start, period_end, router_id,
 --                      leases, lookup_hostname, tracker,
---                      sample_seconds, bucket_seconds)
+--                      sample_seconds, bucket_seconds, blocked_macs)
 --     → report table ready for JSON encoding and POST /api/router/usage
 --     nft_sets:        { hostname → { ip → true } }  (maintained by render.update_shared + dnsmasq)
 --     leases:          { mac → ip }  (optional; populates the ip field on each record)
@@ -318,12 +318,22 @@ end
 -- bucket_seconds.  A counter present in `counters` but missing from the
 -- tracker (e.g. set element first appeared after the last sample) falls back
 -- to one active sample when bytes > 0.
+-- #1628: blocked_macs (optional) is the shared `{ mac -> true }` table
+-- maintained by render.update_shared. When supplied, any counter whose mac is
+-- in the set is dropped before record emission so a blocked device produces
+-- zero traffic_reports rows — even for ea_ carve-out traffic (Khan / Math
+-- Academy / 1Password) that the kernel legitimately allows past the per-MAC
+-- forward-drop. The architectural contract is "blocked = no usage growth";
+-- the carve-out exists for functional access, not to credit minutes against
+-- the daily cap. With agent-side rows suppressed during the block window,
+-- Presence has nothing to stitch across the block boundary either.
 function M.build_report(counters, nft_sets, period_start, period_end, router_id,
                         leases, lookup_hostname, tracker,
-                        sample_seconds, bucket_seconds)
+                        sample_seconds, bucket_seconds, blocked_macs)
   local records = {}
 
   for _, c in ipairs(counters or {}) do
+    if not (blocked_macs and c.mac and blocked_macs[c.mac]) then
     local host    = host_for_ip(c.dst_ip, nft_sets, lookup_hostname)
     local samples = tracker.active_samples[tracker_key(c.mac, c.dst_ip)] or 0
     local active_seconds
@@ -362,6 +372,7 @@ function M.build_report(counters, nft_sets, period_start, period_end, router_id,
       end
       records[#records + 1] = rec
     end
+    end -- #1628 blocked_macs filter
   end
 
   return {
--- a/openwrt/files/usr/sbin/wifihaven-agent
+++ b/openwrt/files/usr/sbin/wifihaven-agent
@@ -820,10 +820,15 @@ conntrack.watch({
         -- last activity_sample_int tick still get an active sample counted.
         usage.tracker_sample(activity_tracker, counters)
         last_activity_sample = mono
+        -- #1628: skip records for MACs currently blocked (TimeLimit / Schedule
+        -- / Paused / Manual) so a blocked device produces zero traffic_reports
+        -- row growth — even when ea_ carve-out traffic legitimately reaches
+        -- the priority-1 accounting chain. blocked_macs is the shared ref
+        -- render.update_shared rebuilds from each device's effective rules.
         local report = usage.build_report(
           counters, nft_sets, period_start, period_end, router_id,
           nil, lookup_hostname, activity_tracker,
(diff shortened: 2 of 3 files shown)


sameerparekh:
Closing — the operator's latest investigation comment on #1628 (post-dating this work) reframed the root cause as #1630: `AppTimeLimitRepoLive.listForProfile`'s `WHERE mode='time_limited'` filter hides `exemptFromDaily=true` for `mode=Allowed` apps (Khan/Math/1Password) from `Presence.exemptPatterns`, so their legitimately-allowed ea_ carve-out traffic counts against the daily cap when it shouldn't.

This PR's agent-side suppression is the wrong fix: traffic to ea_ allowed hosts IS real and operators want it visible in `traffic_reports` / usage views — what shouldn't happen is those minutes counting against the daily cap. That belongs in #1630's fold over `AppPolicyAssignment`, not at the agent.

Closing #1628 as duplicate of #1630.

Issue #1626, opened 2026-06-10
bug(prod): Kids profile usage inflates overnight, keeps growing after time-limit cap, and exempt-from-daily Math Academy blocked when over-cap


sameerparekh:
## Symptoms (prod, observed 2026-06-09 → 2026-06-10 morning)

Three connected symptoms on the **Kids** profile, observed over a single overnight window:

### Symptom 1 — usage accumulated overnight despite scheduled block + sleeping kids
- The Kids profile is under a scheduled-downtime window overnight (`@blocked_macs` should drop all forwarded traffic).
- The kids were asleep (no actual app interaction).
- Yet the usage counter on the dashboard showed engaged minutes accumulating through the night.

### Symptom 2 — time used kept growing AFTER the daily time-limit cap fired
- Profile hit the daily limit during the day → `blocked=true` per the snapshot's per-MAC rules.
- Per the architectural model, nftables drop all forwarded traffic for that MAC.
- The displayed "time used today" continued to climb past the cap — by enough that **>1 hour of extra time was needed** to compensate before the kids could use anything.
- Suggests: either real traffic is reaching the device through the block (enforcement gap), or the display path counts things the enforcement path correctly ignores (display-vs-enforcement drift in the same family as #1531/#1539/#1513).

### Symptom 3 — Math Academy (exempt-from-daily) blocked while profile was over-cap
- Math Academy is configured `exempt_from_daily` — per #1505/#1513/#1515 design, its host-set should be carved out of the whole-MAC daily-cap block.
- When the profile was blocked for the daily cap, Math Academy did NOT work.
- Operator had to grant time extensions sufficient to unblock the WHOLE profile before Math Academy worked.
- Looks like a regression or incomplete carve-out of #1513 (PR #1523's exempt-from-daily fix for apex-only host-sets). Or the carve-out logic doesn't honor exempt-from-daily when the block reason is `TimeLimit` (vs `Schedule`/`Paused`).

### For symptom 1
- **Idle-device pings counting as engaged minutes.** Phones/tablets/laptops keepalive throughout the night. If the InfraHosts list (#1503) doesn't catch all of them, the session-stitch (#1464/#1466) bridges gaps up to `presence_continuation_seconds` (default 120s) → one long synthetic session per device. Real usage = 0; counted = several hours.
[...]

PR #1634, opened 2026-06-10
fix(#1627): exempt-from-daily app w/ no per-app cap carves around whole-MAC block

### LINKED ISSUE #1627: bug: exempt-from-daily app with no per-app limit never carves around whole-MAC block (#1626 symptom 3)
Split from #1626 — root-caused via router-side evidence + code read.

## Summary

When a profile assigns an app with `mode=time_limited`, `exempt_from_daily=true`, and **no per-app `daily_minutes` set** (i.e. `apa.daily_minutes IS NULL`), the exempt-from-daily carve-out **never** triggers, so the app is blocked at the router whenever the profile is whole-MAC blocked (`TimeLimit`, `Schedule`, `Paused`, `Manual`). This was Symptom 3 of #1626: Math Academy was unreachable when Prima's profile was over the daily cap, even though Math Academy is configured exempt-from-daily.

## Root cause (single source-of-truth violation: \`COALESCE(..., 0)\` collides with the carve-out gate)

1. \`AppTimeLimitRepoLive.listForProfile\` projects \`COALESCE(apa.daily_minutes, 0)\` into \`AppTimeLimit.dailyMinutes: Int\` — \`api/src/db/Repos.scala:1113\` (and the matching \`listAll\` at \`:1125\`). \`AppPolicyAssignment.dailyMinutes\` is \`Option[Int]\` on the wire (\`shared/src/Models.scala:341\`); the repo erases the \`None\` → \`0\` distinction.
2. \`TimeStatusService.groupAppLimits\` takes \`lims.map(_.dailyMinutes).max\`, so the assignment surfaces as an \`AppDayState\` with \`dailyLimitMinutes = 0, exemptFromDaily = true\` — \`api/src/policy/TimeStatusService.scala:432\`.
3. \`PolicyService.exemptUnderCapHosts\` gates the carve-out on \`sd.exemptFromDaily && sd.usedMinutes < sd.dailyLimitMinutes\` — \`api/src/policy/PolicyService.scala:786\`. With \`dailyLimitMinutes = 0\`, the predicate \`usedMinutes < 0\` is **never true** for any \`usedMinutes >= 0\`. The exempt app is **never** carved into \`extraAllowed\`.

Per the architectural model, \`extraAllowed\` beats the whole-MAC \`TimeLimit\` drop at the router (memory: \`extraAllowed beats blocked\`; #421). Because the carve-out never reaches \`extraAllowed\`, the per-MAC \`wh_drop:\$mac:TimeLimit\` rule drops Math Academy traffic.

This is a single-source-of-truth violation in the spirit of \`AGENTS.md\` §single-source-of-truth: the \`dailyLimitMinutes\` field is overloaded with two semantics ("the cap" vs "no cap set"), with no shared accessor, and two consumers (\`exemptUnderCapHosts\` and the per-app cap-enforcement path \`PolicyService.scala:841\`) read it differently.

## Evidence from prod router (10 Jun 2026, ~21:08 UTC, Prima MAC 04:72:ef:d6:e4:5a)

- Snapshot (\`/etc/wifihaven/policy.json\`): Prima profile has \`blocked=true, blockReason=TimeLimit, extraAllowed=[\"1password.com\",\"kasandbox.org\",\"kastatic.org\",\"khanacademy.org\",\"mathacademy.com\"]\`. (mathacademy.com IS in the list — but only because some OTHER mechanism is also putting it there; the empty-limit carve never fires.)
- nftables: \`wh_drop:04:72:ef:d6:e4:5a:TimeLimit\` counter shows **2938 pkts / 730678 bytes DROPPED** — Prima's iPad is actively trying and being blocked.
- \`ea_04_72_ef_d6_e4_5a_mathacademy_com\` was populated by dnsmasq nftset callbacks at 14:57 / 15:10 / 15:22 / 15:33 UTC (logged) — DNS resolution + ipset population is fine.
- The dnsmasq log shows mathacademy.com queries during the symptom window resolving to \`52.40.111.135\`.

So at the time the operator observed Math Academy blocked overnight, the apex \`mathacademy.com\` was the only host. The carve-out gate \`usedMinutes < 0\` is the failure: the snapshot Prima has \`extraAllowed=[..., \"mathacademy.com\"]\` **now** likely because the operator since added a non-zero per-app limit to compensate (or because the assignment path differs from \`exemptUnderCapHosts\`). Either way, the broken gate is the proximate cause.

## Fix direction (do NOT implement here)

The cleanest collapse — change \`dailyLimitMinutes\` to \`Option[Int]\` end-to-end, so \`None\` means "no per-app limit" and the carve-out gate becomes \`exemptFromDaily && dailyLimitMinutes.forall(usedMinutes < _)\` (no limit → always carve while exempt). The COALESCE in the repo goes away. This collapses the two semantics into the type system.

Cheaper alternative — keep the Int but change the gate to \`exemptFromDaily && (dailyLimitMinutes == 0 || usedMinutes < dailyLimitMinutes)\`. Worse because it codifies the \`0 == None\` overload instead of removing it; pins the same overload into the per-app cap-enforcement path (\`PolicyService.scala:841\`) too.

## Acceptance

- Feature test: a profile with an app assigned \`mode=time_limited, exempt_from_daily=true, daily_minutes=NULL\` produces an \`extraAllowed\` carve-out for that app's hosts even when the profile is \`blocked=true, blockReason=TimeLimit\`.
- Same test variant for \`blockReason=Schedule\` and \`blockReason=Paused\` — the carve-out is reason-agnostic per \`PolicyService.exemptUnderCapHosts\`.

## Related
- Parent: #1626
- #1513 (apex-only escape) — fix from #1523. This issue is the same family but a different failure mode (NULL limit, not apex-only).
- #1532 (single-source audit umbrella)
- #1505 / #1515 / #1518 (app-centric refactor)

Epic: Tuning.

sameerparekh:
## Summary

- Make `AppTimeLimit.dailyMinutes` and `AppDayState.dailyLimitMinutes` `Option[Int]`. `None` now means "no per-app cap configured", distinct from `Some(0)`.
- Drop the `COALESCE(apa.daily_minutes, 0)` in `AppTimeLimitRepoLive.{listForProfile,listAll}` — that was the proximate bug; it erased `NULL → 0` before the carve-out gate ever ran.
- Gate `PolicyService.exemptUnderCapHosts` on `exemptFromDaily && dailyLimitMinutes.forall(usedMinutes < _)` — no cap ⇒ never exhausted ⇒ always carve while exempt.
- `PolicyService.appCapExhaustedHosts` and `timeLimitBlockFromState` use the same Option-aware exhaustion check so the snapshot and `/decision` cannot drift on what "exhausted" means (AGENTS.md §single-source-of-truth).
- `AppUsage.{limitMins,remainingMins}` become `Option[Int]` on the SPA wire (SPA deploys atomically with the API). Web TS types updated.

## Why

Prod symptom (#1626 symptom 3): Math Academy was assigned `mode=time_limited`, `exempt_from_daily=true`, `daily_minutes=NULL`. Whenever Prima's profile was whole-MAC blocked (TimeLimit / Schedule / Paused), Math Academy was dropped at nftables along with everything else. Root cause: the carve gate's `usedMinutes < dailyLimitMinutes` test became `usedMinutes < 0` after the COALESCE collapse, which is always false for any non-negative `usedMinutes`.

## Test plan

- [x] Four new feature tests in `PolicySnapshotAppsSpec` covering exempt+`NULL` + cap exhausted (TimeLimit), paused, schedule, and own-usage > 0 — committed first as a failing red test (commit b8a31a1f), then green after the fix (commit 86605531).
- [x] `mill __.test` — all suites green.
- [x] `scalafmt --check` clean across the repo.
- [x] `npm run type-check` clean in `web/`.

Closes #1627

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -791,6 +791,44 @@ In the same PR:
    `terraform fmt -check`, `terraform validate`, and `python3 -m json.tool`
    on every dashboard. Run all three locally before pushing.
 
+## Every router-agent write to `/tmp` must be bounded {#bounded-tmp-writes}
+
+**On the OpenWRT target `/tmp` is `tmpfs` — RAM, not disk.** A log, spool, or
+cache the agent appends to without a cap grows until it exhausts router memory,
+at which point allocations start failing fleet-wide: dnsmasq, the agent, and
+the kernel all compete for a few hundred MB of RAM on the smallest supported
+device. An unbounded `/tmp` writer is therefore not a disk-hygiene nicety —
+it is a latent **router-wedge / OOM** bug. (This rule comes out of the #573
+SNI sidecar, whose first cut appended to `/tmp/wifihaven-sni.log` forever.)
+
+So: **any agent file under `/tmp` that grows with traffic, time, or event
+volume must ship with a rotation/truncation path in the SAME change that
+introduces it.** Concretely:
+
+- **Add it to the existing rotation cron, don't invent a parallel one.**
+  [`wifihaven-rotate-dnsmasq-log`](openwrt/files/usr/sbin/wifihaven-rotate-dnsmasq-log)
+  runs every 10 minutes and is the canonical place — it caps each file at a
+  fixed `MAX_BYTES` using the **copytruncate** pattern (`cp` to `*.1`, then
+  `: > file`). New spools join its file list; they do not get a second cron
+  job or a bespoke shell script.
+- **Copytruncate, never rename**, for any file a long-lived process holds an
+  open fd on (dnsmasq, or a `tail -F` follower like `wifihaven-dns-tail`). A
+  rename leaves the writer appending to an invisible, unrotated inode and the
+  follower stranded; truncation keeps both on the same inode. The follower's
+  brief reseek window is acceptable and bounded — see the header comment in
+  the rotate script for the full rationale.
+- **A bounded in-memory ring that is only ever *snapshotted* to `/tmp`
+  (atomic write + rename of a whole file, like `paths.dns_cache`) is already
+  bounded** — its size is the cache size, not cumulative. Those don't need
+  cron rotation. The rule targets **append/grow** writers (logs, spools,
+  metrics journals), not fixed-size snapshots.
+- **State the cap and worst-case `/tmp` footprint** in the script/header so a
+  reviewer can sanity-check it against the smallest device's RAM.
+
+The same reasoning applies to anything else the agent persists under `/tmp`
+(JSON spools, NDJSON event journals, debug dumps). If it can grow, it gets a
+bound — in the same PR.
+
 ## Branch-diff checks (CI + pre-push)
 
 CI checks and pre-push checks that compare a branch against `main` MUST diff against the **merge base** with `origin/main`, not `origin/main` directly. Use three-dot syntax (`origin/main...HEAD`) or an explicit `git merge-base origin/main HEAD`. Two-dot (`origin/main..HEAD`) over-reports when `main` has advanced since the branch diverged, producing spurious failures and noise.
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -414,6 +414,7 @@ object Main extends ZIOAppDefault {
             clock,
           ) ++
           AppRoutes.routes(auth, appRepo, profileRepo, upRepo, templates) ++
+          AdminDebugRoutes.routes(auth, policy) ++
           DebugRoutes.routes(
             cfg.debugEnabled,
             deviceRepo,
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -1102,13 +1102,17 @@ class AppTimeLimitRepoLive(xa: Transactor[Task]) extends AppTimeLimitRepo {
   // #1564: (profileId, host, dailyMinutes, slug, exemptFromDaily, appId) — the join already has
   // `apps.id` in scope, so we carry it through as the canonical FK reference instead of throwing
   // it away and re-resolving the slug downstream.
-  private type R = (ProfileId, String, Int, String, Boolean, AppId)
+  // #1627: project `apa.daily_minutes` as Option[Int], not COALESCE(...,0).
+  // `None` means "no per-app limit" and is treated as never-exhausted by the
+  // carve-out gate — the COALESCE silently collapsed that case to 0 and made
+  // the exempt-no-limit carve unreachable.
+  private type R = (ProfileId, String, Option[Int], String, Boolean, AppId)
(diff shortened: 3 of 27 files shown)


sameerparekh:

Reviewed `b070cf28` against `origin/main` (three-dot merge-base diff) against the 8 checklist dimensions.

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `api/src/routes/Routes.scala:1006-1011` and `:1281-1290` — the `AppUsage` construction (`label / domainPattern / dailyLimitMinutes / used / dailyLimitMinutes.map(lim => (lim - used).max(0))`) is duplicated across the two route handlers. Two call sites is borderline, but the "remaining = limit - used clamped to 0" formula now has Option-aware arithmetic and is exactly the kind of two-place computation §single-source-of-truth flags. A tiny `AppUsage.from(state, used)` helper would collapse it. Not blocking — the bodies are short and the math is local — but worth noting given §1 is the highest-priority dimension.
2. `api/src/policy/TimeStatusService.scala:430-436` — `groupAppLimits` is now a thin adapter over `ProfileAppDispositions.from(...).capGroups` that re-shapes into a 6-tuple. With #1630's `AppDisposition` collapse already shipped, the tuple intermediate looks like a transitional shim — its only consumer (`appDayStatesFromMinutes`) could read `AppDisposition` directly. Out of scope for this PR; flag for follow-up cleanup.

VERDICT: APPROVE

The fix collapses NULL vs 0-minute cap into a single `Option[Int]` end-to-end (`AppTimeLimit.dailyMinutes`, `AppDayState.dailyLimitMinutes`, `AppDisposition.dailyMinutes`, the SPA `AppUsage` wire shape), drops the `COALESCE(..., 0)` in `AppTimeLimitRepoLive`, and rewrites both exhaustion checks (`exemptUnderCapHosts`, `appCapExhaustedHosts`, `timeLimitBlockFromState`) with the same Option-aware predicate — so the snapshot and the per-host `/decision` path share one definition of "exhausted" (§1 single-source-of-truth, §3 architectural invariants). TDD is honored: the failing-test commit `b8a31a1f` precedes the implementation commit `86605531`, and the four new tests in `PolicySnapshotAppsSpec` cover all three whole-MAC block reasons plus the own-usage-but-no-cap edge. No router-wire change (the modified shapes are API-internal / SPA-only — `AppTimeLimit` / `AppDayState` never reach the snapshot, `AppUsage` is consumed only by the SPA which deploys atomically with the API), so §4 back-compat is preserved. Existing test edits are pure `Int → Option` mechanical translations (`== 10` → `.contains(10)`, `>=` → `.exists(>=_)`) — no weakened assertions per §2.

Issue #1629, opened 2026-06-10
investigation: overnight engaged-minutes inflation on Kids profile under scheduled-block (#1626 symptom 1)


sameerparekh:
Split from #1626 — investigation in flight; not yet root-caused with the same confidence as #1627 and #1628.

## Summary

Kids profile accumulated significant \"engaged minutes\" overnight 2026-06-09 → 2026-06-10 while:
- the profile was under a scheduled-downtime block (so \`blocked=true, blockReason=Schedule\` and nftables was correctly dropping forwarded traffic for kid MACs),
- the kids were asleep (no real interaction).

## What we know

1. Background DNS chatter IS reaching the router overnight from kid devices. The router's \`/tmp/wifihaven-dnsmasq.log.1\` shows the Kids iPad making queries for e.g. \`www.khanacademy.org\` at 08:12 UTC (= 01:12 PDT) during the symptom window. This is an open browser tab / service worker background-refreshing, not user activity.
2. Per \`Presence\`, the session-stitch primitive (#1464) bridges idle gaps of up to \`continuationSeconds\` (default 120s). Sparse overnight DNS-driven queries that DO get past the block (via \`extraAllowed\` carve-outs like \`khanacademy.org\`, \`1password.com\`, \`mathacademy.com\`, etc.) generate \`traffic_reports\` rows that can stitch into a single long synthetic \"session\" if their gaps are < 120s — turning maybe 10 minutes of cumulative real device chatter into hours of \"engaged\" time.
3. The InfraHosts heartbeat filter (#1503 / #1525) only suppresses rows whose hostname matches the infra list. \`khanacademy.org\` is NOT infra — it's a real allowed app — so its background queries do not get heartbeat-filtered.
4. \`usedMinutes\` is then susceptible to the same display-vs-enforcement drift as #1628 (parent symptom 2): the cap fires (or in the Schedule case, the block fires for a different reason), but the headline counter keeps growing.

## What we DON'T know

- Whether the overnight inflation came mostly from \`extraAllowed\` traffic (e.g. background polls from open browser tabs hitting whitelisted study apps) OR from something the router agent did NOT block (e.g. a Schedule-block enforcement gap, less likely given the architectural model and that Symptom 3's enforcement counter shows DROPs working).
[...]

Issue #1608, opened 2026-06-09
audit: find one-shot migrations / backfills / seeders that can be deleted now that they've deployed


sameerparekh:
## Why
We've shipped many one-shot data migrations, backfills, and seeders. Most run on every startup and are gated by a "did this already?" check. Once the deploy has propagated to every environment AND every household whose data the migration was meant for has been touched at least once, the code is dead weight — but it stays around because nobody audits and deletes it.

Symptoms of the cost:
- **Live bugs.** #1602 (operator's deleted schedule resurrecting every deploy) was caused by `ScheduleSeeder.seedAndMigrate` running on every boot and re-migrating from the still-retained legacy `schedules` table. The migration was "done" months ago; it kept running because nobody pruned it.
- **Read complexity.** Every reader has to consider "is this row from before or after the migration?" Removing the migration makes the schema's actual contract clearer.
- **Test surface bloat.** Old migrations carry a test scaffold; deleting the migration also deletes its tests.

## Goal
Sweep the codebase for one-shot migration / backfill / seeder code that has shipped, validate each finding is genuinely deployed-and-done, and either delete it now or file follow-up issues for the ones that need a soak window first.

## Initial candidates (non-exhaustive, from a 30-second grep)
- `api/src/ScheduleSeeder.scala` — legacy `schedules` → `named_schedules` migration. Stopped being called in #1602; the file itself is still imported. Already gated to delete with #1485 (legacy table drop, 2026-06-12). Verify and either delete now or pin to #1485.
- `api/src/ReasonTextBackfill.scala` — one-shot backfill of reason-text columns (presumably). Likely a clear win to delete if it's already deployed.
- Anywhere `Main.scala` invokes a `*Seeder` / `*Backfill` / `*Migrator` — each call is a candidate.
- `BundledBlocklists.seed` is NOT a candidate — it runs every boot to install bundled blocklists; that's ongoing behavior, not a one-shot.
- Any backfill column in a recent migration whose accompanying code still ran the backfill at startup (e.g., the dest_ip path — write-time backfill stays, but if there's a one-shot historic backfill, that goes).

## Scope
- **In scope:** application-level (Scala) seeders / backfills / migrators that gate on "already done" flags.
- **Out of scope:** Flyway SQL migrations (those are history; never edited; never deleted).
- **Out of scope:** ongoing write-time backfills like dest_ip-based FQDN attribution (#1585) — those keep running on every new record.

## Process (the audit chip should follow)
[...]

PR #1639, opened 2026-06-11
fix(#1630): collapse per-app disposition readers into one fold

### LINKED ISSUE #1630: Collapse per-app disposition readers into one fold (#1629 root cause; #1532 audit miss)

sameerparekh:
## Summary

Two unrelated readers of the same `app_policy_assignments` table silently disagreed:

- `PolicyService.expandAppDispositions` read `List[AppPolicyAssignment]` (all modes) and emitted snapshot `extraAllowed` / `extraBlocked`.
- `TimeStatusService` (`groupAppLimits` / `exemptPatterns` / `appHostPatterns`) read `List[AppTimeLimit]` synthesized by a repo query filtered `WHERE mode='time_limited'` and emitted everything else — including the daily-total exempt host-set.

That seam let a `mode=Allowed, exemptFromDaily=true` app reach `extraAllowed` (snapshot allowed it) but never reach `exemptPatterns` (daily-usage path counted it) — exactly the §single-source-of-truth class the #1532 audit was supposed to catch.

The PR replaces both readers with the single fold `ProfileAppDispositions.from(appLimits)`. Every projection — `extraAllowed`, `extraBlocked`, `exemptPatterns`, `appHostPatterns`, `capGroups` — is read off that one value. After the collapse it is structurally impossible (not just patched) for two readers to disagree about whether a `mode=Allowed` exempt app counts toward the daily total.

## What changed

- **New** `api/src/policy/ProfileAppDispositions.scala` — the case class + `from` factory + `enforcement(schedWindows, capExhausted, now)` method. This is the ONE place mode case-analysis happens.
- **`AppTimeLimitRepoLive.listForProfile`** — drops the `WHERE mode='time_limited'` filter. Now returns rows for every mode. SELECT also carries `apa.id` (assignmentId) and `apa.mode` so the fold can route each row.
- **`AppTimeLimit`** — gains additive `mode: AppMode` + `assignmentId: AppPolicyAssignmentId` fields with defaults at the end (back-compat with hand-built test constructions).
- **`PolicyService.expandAppDispositions`** — deleted. `snapshot` and `decide` now build `ProfileAppDispositions.from(appLimits).enforcement(...)` directly, off the unified `appTimeLimitRepo.listForProfile` read. The prior `apps + appHostsRaw + appAssigns` fan-out in `snapshot` collapses into a single per-profile fetch.
- **`TimeStatusService.{groupAppLimits, exemptPatterns, appHostPatterns}`** — kept as `private[policy]` names but reduced to thin delegations to `ProfileAppDispositions.from(...)`'s accessors. Downstream consumers (`appDayStates`, `appDayStatesFromMinutes`, `appSecondsByApp`, `usedSecondsForProfile`, `usedSecondsByMac`) keep their existing signatures.

Resolves the symptom on #1628 and #1629 (those investigation issues stay open until the operator confirms in prod).

Closes #1630

## Test plan

- [x] **Allowed-mode exempt app does NOT contribute to `usedSecondsForProfile`** — committed first as RED (fails on `main`), green after collapse. Profile with `App(khanacademy.org, mode=Allowed, exemptFromDaily=true)`, 20 min of traffic → `state.usedMinutes == 0`.
- [x] **Allowed-mode NON-exempt app DOES contribute** — symmetric case, `state.usedMinutes >= 15`.
- [x] **Allowed-mode exempt app still reaches the kid via `extraAllowed`** — snapshot `BlockRules.extraAllowed` for the profile contains `khanacademy.org`.
- [x] **Allowed-mode app with no cap does NOT fire a TimeLimit block** — `state.blocked == false` or `blockReason != TimeLimit` with 120 min of traffic.
- [x] **TimeLimited-mode regression** — `dailyMinutes=60`, 65 min of traffic → blocked with `TimeLimit` reason. Unchanged.
- [x] **SSOT structural pin** (unit, `PerAppScheduleSpec`) — every host that lands on `extraAllowed` from an `exemptFromDaily=true` assignment is also on `exemptPatterns`; cap groups contain only `TimeLimited`; `appHostPatterns`/`exemptPatterns` are mode-agnostic. Fails the moment the readers diverge again.
- [x] Full Scala test suite (`mill __.test`) — green (no regressions across PolicySnapshot/Router/Time/Usage suites).
- [x] `scalafmt --check` — clean.

### DIFF
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -195,10 +195,18 @@ trait TimeLimitRepo {
 }
 
 /**
- * Post-#764 the `site_time_limits` table is dropped. This repo now synthesizes the legacy
- * `AppTimeLimit` shape from `app_policy_assignments` rows with mode='time_limited', emitting one
- * row per (assignment × host) — mirroring the snapshot expansion in `PolicyService`. Synthetic ids
- * are 0L; callers that need stable ids should switch to AppRepo directly.
+ * Post-#764 the `site_time_limits` table is dropped. This repo now synthesizes the `AppTimeLimit`
+ * shape from `app_policy_assignments` rows — emitting one row per (assignment × host) — that the
+ * [[wifihaven.api.policy.ProfileAppDispositions]] collapse folds into every downstream per-app
+ * projection. Synthetic `id` is `0L`; the canonical assignment identity is the `assignmentId`
+ * field, carried from `app_policy_assignments.id`.
+ *
+ * Post-#1630 the SQL filter is dropped: rows for EVERY mode (Allowed, Blocked, TimeLimited) are
+ * returned, with `mode` carried on the row. The single fold case-analyzes on `mode` to route each
+ * disposition to the right bucket. Before #1630 the repo filtered `WHERE mode='time_limited'`,
+ * which hid `mode=Allowed, exemptFromDaily=true` assignments from the daily-usage path and let
+ * their traffic count against the daily total even though the SPA contract was that "Khan doesn't
+ * count". The collapse removes that structural seam.
  */
 trait AppTimeLimitRepo {
   def listForProfile(pid: ProfileId): Task[List[AppTimeLimit]]
@@ -1099,20 +1107,34 @@ class TimeLimitRepoLive(xa: Transactor[Task]) extends TimeLimitRepo {
 }
 
 class AppTimeLimitRepoLive(xa: Transactor[Task]) extends AppTimeLimitRepo {
-  // #1564: (profileId, host, dailyMinutes, slug, exemptFromDaily, appId) — the join already has
-  // `apps.id` in scope, so we carry it through as the canonical FK reference instead of throwing
-  // it away and re-resolving the slug downstream.
-  private type R = (ProfileId, String, Int, String, Boolean, AppId)
+  // #1564 + #1630: (profileId, host, dailyMinutes, slug, exemptFromDaily, appId, assignmentId,
+  // mode) — the join already has `apps.id` in scope, so we carry it through as the canonical FK
+  // reference instead of throwing it away and re-resolving the slug downstream. Post-#1630 the
+  // `apa.id` (assignmentId) and `apa.mode` are also carried so the `ProfileAppDispositions`
+  // collapse can route each (assignment × host) row to the right bucket without a second repo
+  // read. The Postgres enum `app_mode` round-trips via doobie's string type-class.
+  private type R = (ProfileId, String, Int, String, Boolean, AppId, AppPolicyAssignmentId, String)
   private def toS(r: R) =
-    AppTimeLimit(AppTimeLimitId(0L), r._1, r._2, r._3, s"app:${r._4}", r._5, r._6)
+    AppTimeLimit(
+      AppTimeLimitId(0L),
+      r._1,
+      r._2,
+      r._3,
+      s"app:${r._4}",
+      r._5,
+      r._6,
+      AppMode.parse(r._8).getOrElse(AppMode.TimeLimited),
+      r._7,
+    )
 
   def listForProfile(pid: ProfileId) =
     DbMetrics.timed("appTimeLimit.listForProfile")(
-      sql"""SELECT apa.profile_id, ah.host, COALESCE(apa.daily_minutes, 0), a.slug, apa.exempt_from_daily, a.id
+      sql"""SELECT apa.profile_id, ah.host, COALESCE(apa.daily_minutes, 0), a.slug,
+                   apa.exempt_from_daily, a.id, apa.id, apa.mode::text
             FROM app_policy_assignments apa
             JOIN apps a       ON a.id = apa.app_id
             JOIN app_hosts ah ON ah.app_id = apa.app_id
-           WHERE apa.profile_id = $pid AND apa.mode = 'time_limited'
+           WHERE apa.profile_id = $pid
            ORDER BY apa.id, ah.host"""
         .query[R]
         .map(toS)
@@ -1121,11 +1143,11 @@ class AppTimeLimitRepoLive(xa: Transactor[Task]) extends AppTimeLimitRepo {
     )
 
   def listAll =
-    sql"""SELECT apa.profile_id, ah.host, COALESCE(apa.daily_minutes, 0), a.slug, apa.exempt_from_daily, a.id
+    sql"""SELECT apa.profile_id, ah.host, COALESCE(apa.daily_minutes, 0), a.slug,
+                 apa.exempt_from_daily, a.id, apa.id, apa.mode::text
             FROM app_policy_assignments apa
             JOIN apps a       ON a.id = apa.app_id
             JOIN app_hosts ah ON ah.app_id = apa.app_id
-           WHERE apa.mode = 'time_limited'
            ORDER BY apa.id, ah.host"""
       .query[R]
       .map(toS)
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -91,7 +91,12 @@ class PolicyServiceLive(
     // remain injected to feed the companion `apply` factory's TimeStatusServiceLive and to keep the
     // constructor arity ~40 test constructions depend on; the class body no longer reads them.
     @scala.annotation.unused timeLimitRepo: TimeLimitRepo,
-    @scala.annotation.unused appTimeLimitRepo: AppTimeLimitRepo,
+    // #1630: post-collapse this repo IS read by `snapshot` and `decide` — it now returns rows for
+    // every mode (`mode=Allowed`, `Blocked`, `TimeLimited`) so the [[ProfileAppDispositions]]
+    // single fold drives both the per-app enforcement (extraAllowed/extraBlocked) and the
+    // structural projections (exempt host-set / active-app host-set / cap groups) without a
+    // separate read off `appRepo.listAssignmentsForProfile`.
+    appTimeLimitRepo: AppTimeLimitRepo,
     deviceRepo: DeviceRepo,
     blocklistRepo: BlocklistRepo,
     @scala.annotation.unused trafficRepo: TrafficReportRepo,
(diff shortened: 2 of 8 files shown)


sameerparekh:
# Independent PR review — #1639 (collapse per-app disposition readers)

Reviewing `git diff origin/main...HEAD` (full merge-base) against the 8-dimension checklist. Read-only pass — no files modified.

## BLOCKERS

(none)

## SHOULD-FIX

1. **`api/src/routes/UsageRoutes.scala:705`** — `appLimits.filter(_.exemptFromDaily).map(_.domainPattern)` re-derives the exempt-pattern list inline instead of calling the canonical primitive. After this PR the inline derivation happens to agree with `ProfileAppDispositions.from(appLimits).exemptPatterns` because the repo now returns all modes, but this is exactly the §single-source-of-truth shape the PR aims to eliminate — a second hand-rolled `filter(_.exemptFromDaily)` reader sitting alongside the canonical one. A future change that adds another exemption dimension (e.g. per-schedule exempt) will land in `ProfileAppDispositions.exemptPatterns` and silently miss this site. Replace with `ProfileAppDispositions.from(appLimits).exemptPatterns`. (Same fix likely worth applying to any other open-coded `_.exemptFromDaily` reader the grep finds in `api/src/routes` / `api/src/usage` — none jumped out in this diff, but the audit umbrella is #1532.)

## NITS

1. **`api/src/policy/PolicyService.scala:803-811`** — 8-line tombstone comment for the deleted `expandAppDispositions`. Tombstones rot once the next refactor moves the surrounding code. Either drop entirely (git history records the deletion) or fold into the `ProfileAppDispositions` scaladoc, which already explains the same "two readers → one fold" rationale at greater length.
2. **`api/src/policy/ProfileAppDispositions.scala:155`** — `exemptFromDaily = rows.exists(_.exemptFromDaily)` is defensive, but the same scaladoc one line above asserts the flag is uniform across an assignment's synthetic rows. `rows.head.exemptFromDaily` would be clearer about the underlying invariant (and faster, though that doesn't matter at this scale). Same applies to `dailyMinutes = rows.map(_.dailyMinutes).max` directly below. Keep `exists`/`max` if you specifically want the defensive read, but call that out in the comment.
3. **`shared/src/Models.scala:270-277`** — defaulting `mode: AppMode = AppMode.TimeLimited` is necessary back-compat for hand-built test constructions, but it makes "untagged" AppTimeLimit silently behave like TimeLimited. Worth a follow-up to remove the default once the test constructions are migrated to the new field — file under #1608 (one-shot deletion umbrella) if you're going to track it.

## VERDICT: APPROVE

PR delivers exactly what #1630 specifies: a single fold (`ProfileAppDispositions.from`) consumes the unified `AppTimeLimitRepo.listForProfile` output (filter dropped, all modes round-trip with `assignmentId` + `mode`), and every projection — `extraAllowed`/`extraBlocked`/`exemptPatterns`/`appHostPatterns`/`capGroups` — reads off that one value. `PolicyService.expandAppDispositions` is deleted; `TimeStatusService.{groupAppLimits,exemptPatterns,appHostPatterns}` are thin delegations. The structural seam behind #1628/#1629 is gone.

TDD red→green visible in commit history: `77bc3247` ships the failing acceptance test (`usedMinutes == 0` against today's behavior of 20), `1c66149a` ships the implementation and a pure unit SSOT pin in `PerAppScheduleSpec`. Full `mill __.test` runs green including `PolicySnapshotConsolidationSpec`, `RouterApiSpec`, and the `TimeApi`/`UsageApi` suites — no regressions from the wider unfiltered repo read reaching the rollup job or per-device summary path.

Approving with one SHOULD-FIX on UsageRoutes.scala:705 to close the remaining open-coded `exemptFromDaily` derivation; it isn't merge-gating because the inputs are now structurally aligned, but addressing it now is consistent with the PR's own thesis.

sameerparekh:
# Independent PR review — #1639 (re-run on `ec1698c1`)

Prior review (`1c66149a`): 0 BLOCKERS, 1 SHOULD-FIX, 3 NITs. Delta: `git diff 1c66149a...HEAD` — 3 files, +11/-17.

## Prior-findings status

| # | Finding | Status |
|---|---------|--------|
| SHOULD-FIX 1 | `api/src/routes/UsageRoutes.scala:705` — open-coded `appLimits.filter(_.exemptFromDaily).map(_.domainPattern)` | **ADDRESSED** — now reads `ProfileAppDispositions.from(appLimits).exemptPatterns` (new line 706). Single canonical reader. |
| NIT 1 | `api/src/policy/PolicyService.scala:803` — 8-line tombstone comment for deleted `expandAppDispositions` | **ADDRESSED** — block removed; rationale already lives in `ProfileAppDispositions`'s scaladoc. |
| NIT 2 | `api/src/policy/ProfileAppDispositions.scala:155` — defensive `exists`/`max` for fields uniform across rows | **ADDRESSED** — switched to `first.exemptFromDaily` / `first.dailyMinutes`; comment now states the invariant explicitly. |
| NIT 3 | `shared/src/Models.scala:270-277` — defaulted `mode: AppMode = AppMode.TimeLimited` makes "untagged" silently behave like TimeLimited | **NOT-ADDRESSED** — left as-is. Consistent with the original recommendation (file follow-up under #1608); not merge-gating. |

## New findings (delta `1c66149a...HEAD`)

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

1. **`api/src/routes/UsageRoutes.scala:706`** — uses the fully-qualified `wifihaven.api.policy.ProfileAppDispositions` inline rather than a top-of-file import. Matches the file's existing style for `wifihaven.api.policy.*` symbols (`ApiError.Wrapped(_)`, `ApiError.Db(_)` are imported but the policy package is referenced FQN at several other call sites), so consistent — but a single import line would be cleaner and the call site shorter. Non-blocking.

## VERDICT: APPROVE

All actionable findings from the prior review are ADDRESSED. The single NOT-ADDRESSED item was explicitly flagged as a follow-up at original-review time and is consistent with the §migrations / one-shot deletion umbrella (#1608). Delta is bounded to the three review-driven edits with no new behavior changes, and the full `mill __.test` suite remains green. No new BLOCKERs introduced — the merge gate is clear.

Issue #1640, opened 2026-06-11
feat(#1636): carry matched eb_/bl_ host on connection_event reason so operators see WHY a drop fired


sameerparekh:
## (Reframed) Plumb the matched eb_/bl_ set name through to the connection_event

Original framing (conntrack mis-labels reason="host") was wrong — see #1636 trace. The nftables rule fires correctly on dst-IP. What's missing is **which host's eb_ set actually matched**, so the operator can see "oauth was blocked because the IP is shared with youtubei.googleapis.com (which IS on the block list)" instead of "oauth was blocked, household" (which sends them down the wrong investigation path).

## What changes

1. **Router-side: enrich the nftables `log prefix`** to carry the matched set's host. Today:

   ```
   ether saddr <mac> ip daddr @eb_youtubei_googleapis_com log prefix "wh_drop:<mac>:host" drop
   ```

   After:

   ```
   ether saddr <mac> ip daddr @eb_youtubei_googleapis_com log prefix "wh_drop:<mac>:host:youtubei.googleapis.com" drop
   ```

   Same for `bl_<id>` (already carries `category:<id>` — fine, the id IS the matcher). Same for the per-MAC blanket drop (already carries the per-MAC reason).

2. **Router-side: `nflog.parse_line` extracts the matched host** and passes it through `build_event`. The wire field rides on the existing reason payload.

3. **Wire shape**: extend `BlockReason.ExtraBlocked` to carry an optional `matchedHost: Hostname`. Per AGENTS.md backwards-compat rules: additive, ignore unknown fields. Older routers omit it; newer API tolerates.

   ```scala
   case class ExtraBlocked(matchedHost: Option[Hostname]) extends BlockReason
   ```

4. **API: `BlockReason.fromWire`** parses the new suffixed form (`host:youtubei.googleapis.com`) into `ExtraBlocked(Some(youtubei.googleapis.com))`; bare `host` continues to decode as `ExtraBlocked(None)`. `asWire` emits the suffixed form when `matchedHost` is set.

5. **SPA: render** "blocked: shared IP with `youtubei.googleapis.com`" (or similar) when `matchedHost` is set and differs from the attributed `host.value`. When it matches (or is None), keep current rendering.

6. **DB / `connection_events.reason` (JSONB)**: kind stays `extraBlocked`; the `matchedHost` field rides in the payload. No migration needed if kind name unchanged.

## Why this matters

[...]

PR #1642, opened 2026-06-11
feat(#1641): admin /api/admin/snapshot endpoint + snapshot_changed INFO logging

### LINKED ISSUE #1641: feat: admin /api/admin/snapshot endpoint + log full snapshot JSON on each etag change for historical debug
## Problem

When debugging a past incident, we have no way to see what the policy snapshot looked like at the time. #1640 is exactly this gap: yesterday's connection_events labeled hosts as `extraBlocked` that the current snapshot doesn't put in any eb_ set, but the snapshot from yesterday is unrecoverable so we can't tell whether the snapshot drifted or conntrack mislabeled.

Two pieces would close the gap:

### 1. `GET /api/admin/snapshot` — read current snapshot

Admin-auth-gated endpoint returning the same `PolicySnapshot` JSON the router gets at `/api/router/policy`. Today the only way to read it is to SSH a router and `cat /etc/wifihaven/policy.json` — which only works if you're on the right LAN.

- Reuses `PolicyService.snapshot` (no new computation path — keeps the single-source-of-truth rule). Same JSON shape the router consumes, plus the etag header.
- Read-only, no side effects (don't touch `routerRepo.touch`).
- Optional `?profileId=N` / `?mac=…` filters to narrow output for a large household. Default returns the full snapshot.
- New route in `RouterRoutes.scala` (it owns the policy resolver) or `DebugRoutes.scala` (it owns admin debug surface). Probably DebugRoutes — keeps it out of the router-token path.

### 2. Log full snapshot JSON on every etag change

Currently `RouterRoutes` logs `etagOut=…` per poll at INFO. That's a fingerprint but not the contents. Add an INFO log line carrying the **full snapshot JSON** every time `PolicyService.snapshot` produces a **new etag**:

- Compute snapshot, hash for etag, compare to a process-local `Ref[Option[Etag]]` of last-seen etag.
- On change, log INFO with structured fields: `event=snapshot_changed etag=<new> prevEtag=<prev> bytes=<n> snapshot=<json>`.
- Render retains stdout for ~7 days (free tier), so we get ~a week of snapshot history for free — long enough to investigate most incidents.
- One log line per change, not per poll, so volume is bounded by how often policy actually changes (a handful per day typically). Bytes are bounded by snapshot size (~100KB) so it's manageable.

Together these mean: next time something like #1640 happens, we open Render logs at the incident timestamp, grep for `snapshot_changed`, and read the exact `extraBlocked` Kids had at that moment.

## Acceptance

- `GET /api/admin/snapshot` (admin auth) returns the current snapshot. Curl from anywhere with admin JWT, no router involvement.
- API logs at INFO contain a `snapshot_changed` line every time the etag changes, with the full snapshot JSON.
- Test: a feature spec asserts the log line fires on first compute and on subsequent changes only.

## Notes

- Don't add this as a metric (#1382-class cardinality fence applies — snapshot JSON is not metric data). Log line only.
- AGENTS.md "Single source of truth" — both deliverables call `PolicyService.snapshot`, not a parallel resolver.
- Out of scope: persisting snapshots to a DB table for queryable history (heavier; do later if the log-based approach proves insufficient).

## Related

- #1640 — concrete incident that motivated this
- #1636 — the audit that surfaced #1640
- #481 — existing per-poll etag logging convention

sameerparekh:
Closes #1641.

cc @sameerparekh for review.

## Summary

Two deliverables, both reusing `PolicyService.snapshot` (single source of truth — no parallel resolver):

1. **`GET /api/admin/snapshot`** — admin-JWT-gated, read-only. Returns the same `PolicySnapshot` JSON the router consumes at `/api/router/policy`, with a strong `ETag` header matching `snap.etag`. No `routerRepo.touch` side effect.
2. **`event=snapshot_changed` INFO log line on every ETag transition** — process-local `AtomicReference[Option[ETag]]` in `PolicyServiceLive`; atomic compare-and-set on every `snapshot` call. Only the caller who actually swaps in a new etag emits the line, so concurrent pollers cannot duplicate it. One line per ETag change (not per poll).

Together these close the #1640 historical-snapshot gap: next time conntrack labels a host as `extraBlocked` that doesn't match the current snapshot, an operator opens Render logs at the incident timestamp, greps for `snapshot_changed`, and reads the exact `BlockRules` shape in force at that moment.

## Design

- **Where it lives.** `AdminDebugRoutes` is a new object in `api/src/routes/DebugRoutes.scala` — distinct from the existing `DebugRoutes` (which is loopback-only and gated by `WIFIHAVEN_DEBUG=1`). Mounted unconditionally in `Main.scala` alongside the other admin route families; auth gate is `requireAdmin`.
- **No re-derivation.** Handler is essentially `policy.snapshot.map(s => Response.json(s.toJson).addHeader(ETag.Strong(...)))`. Same JSON shape, same ETag.
- **No filter params.** `?profileId=` / `?mac=` skipped per the issue's "if trivial" note — default returns the full snapshot.
- **Logging atomicity.** `AtomicReference` (not `Ref[…]`) so construction stays synchronous — the existing 40+ direct test constructors of `PolicyServiceLive` stay arity-stable, no factory rewiring needed. `compareAndSet` guarantees exactly-one logger per transition under concurrent pollers.
- **Not a metric.** Per the issue note, snapshot JSON is not metric data — log line only. No dashboard panel.

## Sample log line (as actually emitted)

```
timestamp=2026-06-11T14:00:04.974Z level=INFO thread=#zio-fiber-221 message="event=snapshot_changed etag=\"sha256:e75d33c8709f508f9dafc0146dde2a6e291a3558f3fd5a4ca92d2355a56485cd\" prevEtag=\"sha256:6a46e6c07b093d8d1c408f0e343419eeb3caf84beb48fe6a30c95e2529981ccf\" bytes=1228 snapshot={\"etag\":\"\\\"sha256:e75d33c8709f508f9dafc0146dde2a6e291a3558f3fd5a4ca92d2355a56485cd\\\"\",\"generatedAt\":\"2025-01-06T14:00:00Z\",\"global\":{...},\"devices\":{...},\"profiles\":{\"3\":{\"name\":\"Kids\",\"rules\":{\"blocked\":true,\"blockReason\":\"Paused\",...}},...},\"blocklists\":{}}" location=wifihaven.api.policy.PolicyServiceLive.logSnapshotChanged file=PolicyService.scala line=302
```

First call after process start logs `prevEtag=none`; subsequent calls log only when `snap.etag` actually moves.

## Sample curl

```bash
$ curl -sS -H "Authorization: Bearer $ADMIN_TOKEN" https://api.wifihaven.net/api/admin/snapshot | jq .
{
  "etag": "\"sha256:e75d33c8…\"",
  "generatedAt": "2026-06-11T14:00:00Z",
  "global": { "blocked": false, "extraAllowed": [...], ... },
  "devices": { ... },
  "profiles": { ... },
  "blocklists": { ... }
}
```

The response carries the same `ETag` header the router endpoint emits, so a client can use If-None-Match for cheap polling (the handler doesn't 304 itself today — that's out of scope).

## TDD red→green

Per AGENTS.md the red commit lands first, isolated:
- `8c1e3a49 test(#1641): failing spec for admin snapshot + snapshot_changed logging` — fails to compile (`AdminDebugRoutes` doesn't exist).
- `fecd5533 feat(#1641): admin snapshot endpoint + snapshot_changed INFO logging` — implementation; spec now passes.

## Test plan

- [x] Spec covers admin → 200 + ETag header, child → 403, no auth → 401, plus exactly-one-INFO-per-ETag-transition.
- [x] `mill api.test` — full suite green locally (no regressions in any of the existing policy/router specs).
- [x] `scalafmt --check` clean.
- [ ] Operator smoke-test on staging post-merge: `curl -sS -H "Authorization: Bearer …" https://staging.wifihaven.net/api/admin/snapshot`.
- [ ] Operator confirms the `snapshot_changed` line appears in Render logs on the next policy change post-deploy.

## Backwards compatibility

Purely additive: new endpoint, additive INFO log line. No wire shape changes, no migration.

### DIFF
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -414,6 +414,7 @@ object Main extends ZIOAppDefault {
             clock,
           ) ++
           AppRoutes.routes(auth, appRepo, profileRepo, upRepo, templates) ++
+          AdminDebugRoutes.routes(auth, policy) ++
           DebugRoutes.routes(
             cfg.debugEnabled,
             deviceRepo,
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -6,9 +6,11 @@ import wifihaven.api.metrics.AppMetrics
 import wifihaven.shared.{Schedule as DbSchedule, *}
 import wifihaven.shared.types.*
 import zio.{Clock as _, *}
+import zio.json.*
 
 import java.security.MessageDigest
 import java.time.{DayOfWeek, Instant, LocalDate}
+import java.util.concurrent.atomic.AtomicReference
 
 trait PolicyService {
   def snapshot: Task[PolicySnapshot]
@@ -120,6 +122,16 @@ class PolicyServiceLive(
   // hosts when assembling the global section. (#1321 moved the curated
   // `infraAllowHosts` onto the same global path, retiring its per-profile copy.)
   private val uiGlobalAllow: List[Hostname] = uiAllowedHosts
+
+  // #1641: process-local last-seen-etag for snapshot-change INFO logging. On every `snapshot`
+  // call we atomically compare-and-set the latest etag here; only the winner of the swap emits
+  // the `event=snapshot_changed` INFO log carrying the full snapshot JSON. One line per ETag
+  // transition (NOT per poll) so volume is bounded by how often policy actually changes.
+  // Render retains stdout for ~7 days, giving roughly a week of snapshot history for free —
+  // long enough to investigate most incidents (the #1640 motivation). Not a metric (cardinality
+  // fence; snapshot JSON is not metric data).
+  private val lastSnapshotEtag: AtomicReference[Option[ETag]] =
+    new AtomicReference(Option.empty[ETag])
 
   // #1069/#1482: per-host /decision fallback must see the same schedule downtime as the snapshot —
   // the windows of every block-mode named schedule attached to the profile (as synthetic
@@ -267,8 +279,29 @@ class PolicyServiceLive(
       // #1318: surface the global-section size + default-deny profile count for operators.
       AppMetrics
         .setGlobalPolicy(snap.global.extraAllowed.size, defaultDenyProfiles)
+        .zipRight(logSnapshotChanged(snap))
         .as(snap)
     }
+
+  /**
+   * #1641: emit an INFO log line carrying the full snapshot JSON the first time we see a given ETag
+   * in this process. Atomic compare-and-set against `lastSnapshotEtag` — only the caller who
+   * swapped a new etag in logs, so concurrent pollers cannot duplicate the line. One line per ETag
+   * transition (not per poll), bounded by how often policy actually changes.
+   */
+  private def logSnapshotChanged(snap: PolicySnapshot): UIO[Unit] = ZIO.suspendSucceed {
+    val prev    = lastSnapshotEtag.get()
+    val changed = !prev.contains(snap.etag) && lastSnapshotEtag.compareAndSet(prev, Some(snap.etag))
+    if (!changed) ZIO.unit
+    else {
+      val json = snap.toJson
+      ZIO.logInfo(
+        s"event=snapshot_changed etag=${snap.etag.value} " +
+          s"prevEtag=${prev.map(_.value).getOrElse("none")} " +
+          s"bytes=${json.length} snapshot=$json",
+      )
+    }
+  }
 
   def renderBlocklist(id: BlocklistId): Task[Option[(ETag, String)]] =
     for {
--- a/api/src/routes/DebugRoutes.scala
+++ b/api/src/routes/DebugRoutes.scala
@@ -230,14 +230,12 @@ object AdminDebugRoutes {
               snap <- policy.snapshot.mapError(ApiError.Db(_))
             } yield Response
               .json(snap.toJson)
-              .addHeader(Header.ETag.Strong(stripQuotedEtag(snap.etag.value)))
+              // #1641: shared with the router's GET /api/router/policy via
+              // RouterRoutes.stripQuotes — collapsed into one helper rather than
+              // hand-mirrored here, so the ETag header emission cannot drift between
+              // the two routes (AGENTS.md §Single source of truth).
+              .addHeader(Header.ETag.Strong(RouterRoutes.stripQuotes(snap.etag.value)))
           handle.mapError(ErrorMapper.errorToResponse)
         },
     )
-
-  // Mirror of `RouterRoutes.stripQuotes` — `Header.ETag.Strong` wants the bare token without
-  // surrounding quotes (zio-http renders them back on the wire). The stored ETag value is the
-  // quoted form (sha256:..."), so trim the wrapping quotes before constructing the header.
-  private def stripQuotedEtag(s: String): String =
-    if s.startsWith("\"") && s.endsWith("\"") then s.drop(1).dropRight(1) else s
 }
--- a/api/src/routes/DebugRoutes.scala
+++ b/api/src/routes/DebugRoutes.scala
@@ -1,7 +1,9 @@
 package wifihaven.api.routes
 
+import wifihaven.api.auth.*
 import wifihaven.api.cache.TimeStatusCache
 import wifihaven.api.db.*
+import wifihaven.api.policy.PolicyService
 import wifihaven.shared.Clock
 import zio.{Clock as _, *}
 import zio.http.*
@@ -200,4 +202,42 @@ object DebugRoutes {
         s
     stripped == "localhost" || stripped == "127.0.0.1" || stripped == "::1"
   }
+}
+
+/**
+ * Admin-auth-gated debug surface (#1641). Distinct from [[DebugRoutes]]: that family is loopback-
+ * only and requires `WIFIHAVEN_DEBUG=1`, suitable only for on-host triage; this one is reachable
+ * from anywhere with an admin JWT, suitable for remote incident response on the cloud deploys.
+ *
+ * Current routes:
+ *   - `GET /api/admin/snapshot` — returns the same [[wifihaven.shared.PolicySnapshot]] JSON the
+ *     router fetches at `/api/router/policy`, with a strong ETag header matching. Reuses
+ *     `PolicyService.snapshot` (single-source-of-truth — no parallel resolver). Read-only: does NOT
+ *     call `routerRepo.touch`, so polling here doesn't move `last_etag` / `last_seen_at`.
+ */
+object AdminDebugRoutes {
+
+  def routes(
+      auth: AuthService,
+      policy: PolicyService,
+  ): Routes[Any, Response] =
+    Routes(
+      Method.GET / "api" / "admin" / "snapshot" ->
+        handler { (req: Request) =>
+          val handle: ZIO[Any, ApiError, Response] =
+            for {
+              _    <- requireAdmin(req, auth).mapError(ApiError.Wrapped(_))
+              snap <- policy.snapshot.mapError(ApiError.Db(_))
+            } yield Response
+              .json(snap.toJson)
+              .addHeader(Header.ETag.Strong(stripQuotedEtag(snap.etag.value)))
+          handle.mapError(ErrorMapper.errorToResponse)
+        },
+    )
+
+  // Mirror of `RouterRoutes.stripQuotes` — `Header.ETag.Strong` wants the bare token without
+  // surrounding quotes (zio-http renders them back on the wire). The stored ETag value is the
+  // quoted form (sha256:..."), so trim the wrapping quotes before constructing the header.
+  private def stripQuotedEtag(s: String): String =
+    if s.startsWith("\"") && s.endsWith("\"") then s.drop(1).dropRight(1) else s
 }
--- a/api/src/routes/RouterRoutes.scala
+++ b/api/src/routes/RouterRoutes.scala
@@ -165,7 +165,13 @@ object RouterRoutes {
         },
     )
 
-  private def stripQuotes(s: String): String =
+  // #1641: package-visible so [[AdminDebugRoutes]] (which emits the same router-shape ETag
+  // header on `GET /api/admin/snapshot`) can reuse this instead of hand-mirroring a copy.
+  // `Header.ETag.Strong` wants the bare token without surrounding quotes; our stored ETag
+  // value is the quoted form (`"sha256:..."`), so trim the wrapping quotes before constructing
+  // the header. Shared helper, not a "keep in sync" mirror (per AGENTS.md §Single source of
+  // truth).
+  private[routes] def stripQuotes(s: String): String =
     if s.startsWith("\"") && s.endsWith("\"") then s.drop(1).dropRight(1) else s
 
   // RFC 7232 §2.3.2: If-None-Match uses weak comparison; W/"x" matches "x".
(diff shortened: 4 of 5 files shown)


sameerparekh:
## Independent PR review (first run)

Reviewed against the 8-dimension checklist in `docs/pr-review-checklist.md`.
Read-only / adversarial pass on `git diff origin/main...HEAD`
(merge-base, three-dot).

### BLOCKERS
*(none)*

### SHOULD-FIX

1. **`api/test/src/feature/AdminSnapshotApiSpec.scala:100-101`** — the
   "200 + ETag header" test asserts both `etagHeader` and `snap.etag.value`
   merely *contain* the substring `"sha256:"`. That passes against a
   hard-coded sentinel header and doesn't actually pin the contract you
   care about (header round-trips the **same** ETag the body declares).
   Tighten to assert the header equals the body's etag modulo the quoted/
   unquoted ETag-strong rendering, e.g. round-trip
   `Header.ETag.Strong(stripped) == resp.headers.get(Header.ETag).get` or
   string-compare against `snap.etag.value` after a single normalize step.
   That also gives you a regression net for the `stripQuotedEtag` helper.

2. **`api/src/routes/DebugRoutes.scala:239-241` vs.
   `api/src/routes/RouterRoutes.scala:168-169`** — `stripQuotedEtag` is a
   verbatim re-implementation of `RouterRoutes.stripQuotes`, and the
   docstring on line 237 (`Mirror of RouterRoutes.stripQuotes`) literally
   calls it a hand-mirrored copy. Per AGENTS.md §Single source of truth
   "any `// must mirror` / `same branch as` comment is a defect to remove,
   not a pattern to copy." Two routes that emit a router-shape ETag header
   will diverge the moment one of them fixes a bug (e.g. tolerating `W/`
   prefix). Collapse to one helper — promote it to a small private
   utility in `RouterRoutes` (or a routes-package helper file) and have
   `AdminDebugRoutes` call it.

### NITS

1. **`api/src/policy/PolicyService.scala:295-301`** — the log
   line is unstructured-text-with-`key=value`. The other structured
   pipeline-event log in `RouterRoutes.scala:85-87` is the same shape,
   so this is consistent with house style. Non-blocking, but if you ever
   adopt zio-logging's `LogAnnotation` for structured fields you'll
   want both call sites refactored together — flag for future-#1641
   follow-up.

2. **`api/src/routes/DebugRoutes.scala:213-234`** — the handler is
   read-only on a route any admin can hit; PR body says polling here
   "doesn't move `last_etag` / `last_seen_at`". Worth a brief test that
   pins this (i.e. confirms there's no `routerRepo.touch` side-effect)
   so a future refactor can't quietly reintroduce one. Cheap insurance,
   not blocking.

### Verdict-relevant positive checks

- **Single source of truth (the highest-priority dimension):** both
  deliverables call `PolicyService.snapshot` directly. The handler is
  `snap.toJson` straight through; the log path is the same `snap`. No
  re-derivation of the snapshot or the etag.
- **TDD red→green visible:** `8c1e3a49` (failing test, doesn't compile)
  precedes `fecd5533` (implementation). Matches the AGENTS.md mandate.
- **Wire backwards-compat:** purely additive — new endpoint, new INFO log
  line. No snapshot shape changes, no removed/renamed/retyped fields.
- **Metrics-policy:** issue body explicitly says "do not add as a metric"
  (cardinality fence on snapshot JSON), and the PR honors that — no
  metric, no dashboard panel needed.
- **Auth:** `requireAdmin` enforced via the standard
  `mapError(ApiError.Wrapped(_))` bridge, matching every other admin
  route. 403 / 401 paths covered by the spec.
- **Concurrency:** `AtomicReference.compareAndSet` guarantees exactly one
  logger per ETag transition under concurrent pollers — correctly
  implements the "one line per change, not per poll" semantics in the
  issue.
- **Scope:** 3 source files + 1 spec; no drive-by edits; no AI mentions
  in commit messages or PR title/body (Co-Authored-By trailer only,
  which is fine).

---

**VERDICT: REQUEST-CHANGES** — only on weight of the two SHOULD-FIX
findings. No BLOCKERs. The implementation is correct end-to-end and
ships exactly the contract #1641 asks for; the test assertion is weaker
than it should be for a contract-pinning spec, and `stripQuotedEtag` is
a hand-mirrored copy that the SSOT rule flags by design. Both are
small, mechanical fixes.

sameerparekh:

Re-running the standard checklist. Prior review covered `fecd5533`
(first run, above). This run covers the incremental delta
`git diff fecd5533...07d08fa3` plus a re-status of the prior findings
against current code.

### Prior-findings status

| # | Prior finding | Status | Notes |
|---|---|---|---|
| SHOULD-FIX 1 | `AdminSnapshotApiSpec.scala` ETag-header assertion only checks `contains("sha256:")` — passes against a hard-coded sentinel | **ADDRESSED** | Replaced with `etagHeader == snap.etag.value`. Round-trip equality pins the contract; `RouterRoutes.stripQuotes` semantics mean `Header.ETag.Strong(stripped)` re-adds the quotes, so the rendered header equals the body's quoted etag exactly. |
| SHOULD-FIX 2 | `stripQuotedEtag` is a verbatim hand-mirror of `RouterRoutes.stripQuotes` (the doc literally says "Mirror of...") | **ADDRESSED** | `RouterRoutes.stripQuotes` promoted to `private[routes]`; `AdminDebugRoutes` calls it directly. One helper, one definition, one place to fix any future tolerance update (e.g. `W/` prefix). |
| NIT 1 | Unstructured `key=value` log format consistent with `RouterRoutes` etag log — flag for joint structured-fields refactor later | **NOT-ADDRESSED** | Deliberately deferred per the NIT itself (future joint refactor). |
| NIT 2 | No test pinning the absence of `routerRepo.touch` side-effect | **NOT-ADDRESSED** | Deferred — the handler's type signature doesn't take a `RouterRepo` parameter, so reintroducing the call would require a deliberate signature change. Reviewer judged this acceptable. |

### New findings on the delta

#### BLOCKERS
*(none)*

#### SHOULD-FIX
*(none)*

#### NITS
*(none)*

The delta is minimal and surgical: 3 hunks touching the exact lines the
prior findings cited, no collateral changes. No new code paths, no new
test surface, no new wire shape. `RouterRoutes.stripQuotes` is now
`private[routes]` — a deliberately narrow widening (one package) that
keeps the helper out of `Routes.scala` / cross-package consumers. The
incremental commit message says only what the fix does and why; no AI
mentions.

### Re-status of dimensions on full PR

- **Single source of truth (dim 1):** strengthened — the ETag-header
  quote-stripping is now a single function with one caller in
  `RouterRoutes` and one in `AdminDebugRoutes`. The earlier hand-mirror
  smell is gone. Both deliverables still call `PolicyService.snapshot`
  directly. **No drift surface remains.**
- **Testing integrity (dim 2):** strengthened — the ETag assertion now
  fails any implementation that emits a constant header. Other assertions
  unchanged.
- **Wire backwards compat (dim 4):** still purely additive.
- **Metrics / dashboard (dim 6):** still no metric, no panel, per issue
  spec.
- **Security (dim 8):** unchanged. `requireAdmin` enforced; 403/401
  paths covered.

---

**VERDICT: APPROVE** — both prior SHOULD-FIX items ADDRESSED, no new
findings on the delta, no open BLOCKERs at any point in this review's
history. Implementation matches the #1641 contract end-to-end; full
`mill api.test` green; the SSOT collapse means the only ETag-quote
trimmer in the codebase is `RouterRoutes.stripQuotes`. Ready to merge
on operator review.

sameerparekh:
## Independent PR review (orchestrator pass)

Read-only adversarial review against `docs/pr-review-checklist.md`, run on
`git diff origin/main...HEAD` at `07d08fa3`. Two prior review comments exist
on this PR (the implementer's own first run on `fecd5533` and re-run on
`07d08fa3`); this is an independent orchestrator-side pass to corroborate
before merge.

### BLOCKERS
*(none)*

### SHOULD-FIX
*(none)*

### NIT

1. `api/src/policy/PolicyService.scala:288-293` — the CAS pattern
   `prev = get(); changed = !prev.contains(snap.etag) && compareAndSet(prev, Some(snap.etag))`
   has a narrow race: two concurrent first-callers with *different* etags
   both observe `prev=None`, both pass the `!contains` check, exactly one
   wins the CAS. The loser silently drops its etag transition rather than
   retrying. In practice the loser's etag is stale (it was computed against
   an earlier DB state) so dropping is arguably correct, but a tiny loop
   (`updateAndGet`-style retry, or `Ref.modify` which gives the same
   semantics atomically) would remove the ambiguity. Not worth blocking on —
   the snapshot endpoint isn't a high-frequency contended path, and the
   per-poll cadence in prod is once per ~20s per router. Leaving as a NIT.

2. `api/src/routes/DebugRoutes.scala:107-130` — per AGENTS.md §"New
   functionality ships with metrics", new routes should emit a metric in the
   same PR. `GET /api/admin/snapshot` does not. Defensible because it's a
   human-driven admin debug endpoint (not a fleet path), and the HTTP
   middleware already records the route at the standard level. Genuinely a
   NIT, but flagging for the record.

### Checklist items confirmed clean

- **§1 SSOT** — both `GET /api/router/policy` (`RouterRoutes.scala:66`) and
  `GET /api/admin/snapshot` (`DebugRoutes.scala:119`) call
  `policy.snapshot` directly; no re-derivation. ETag header emission is
  shared via `RouterRoutes.stripQuotes` (made `private[routes]` in this PR,
  reused at `DebugRoutes.scala:126`). The `lastSnapshotEtag`
  `AtomicReference` is a single field on `PolicyServiceLive`, shared across
  callers — not duplicated per call-site, so the dedup actually dedups.
- **§2 Test integrity** — TDD red→green visible in the commit history
  (`8c1e3a49` red, `fecd5533` green, `07d08fa3` review fixes). The four
  feature tests in `AdminSnapshotApiSpec` cover admin-token-200,
  readonly-token-403, no-auth-401, and the snapshot_changed dedup across an
  ETag transition. `@@ TestAspect.sequential` pins ordering for the log
  assertion. ETag-header assertion is a tight round-trip equality
  (`etagHeader == snap.etag.value`), not a substring check — the prior
  reviewer flagged this on the first run and the implementer fixed it. No
  existing tests were weakened, deleted, or skipped.
- **§3 Architectural invariants** — no router-side logic added; no new
  top-level snapshot wire field; `Clock` injected (no direct `java.time`
  use in the new code); typed `ApiError` errors throughout the new route.
- **§4 Backwards compatibility** — purely additive: one new endpoint, one
  new field on `PolicyServiceLive`, one method visibility widening
  (`stripQuotes` → `private[routes]`). No wire shape changes. Older agents
  unaffected.
- **§5 Migrations** — no Flyway migrations in this PR. Migration-isolation
  rule N/A.
- **§7 Scope / hygiene** — diff is 314 lines, focused. No `TODO` without
  issue. No AI mentions in the code, commits, or PR body.
- **§8 Security** — `GET /api/admin/snapshot` is gated by
  `requireAdmin(req, auth)` (matching the existing `ScheduleRoutes` /
  `AppRoutes` pattern). Readonly tokens are rejected with 403, no-auth with
  401, both pinned by tests. The endpoint is mounted unconditionally (no
  loopback / `WIFIHAVEN_DEBUG=1` gate); this is the documented intent — it
  is the admin-JWT counterpart to the loopback-only `DebugRoutes` family.
  Snapshot JSON does not contain credentials.

### Re-run note

The implementer ran their own first review on `fecd5533` (2 SHOULD-FIX, 0
BLOCKERs) and addressed both in `07d08fa3`:

- ETag header assertion tightened from substring to round-trip equality.
- `stripQuotes` extracted as shared `private[routes]` helper rather than
  hand-mirrored in `AdminDebugRoutes`.

Confirmed both are ADDRESSED at `07d08fa3` and no new findings introduced on
the delta `fecd5533...07d08fa3`.

---

**VERDICT: APPROVE**

Issue #2, opened 2026-04-30
Wire static-bundle serving into API routes


sameerparekh:
The API config has `familydns.http.staticDir`, the deploy script copies `web/dist` into `/opt/familydns/web`, and the bootstrap/staging stacks set the path correctly — but [api/src/Main.scala](api/src/Main.scala) only registers `allRoutes`, never mounts a static handler. As a result, `GET /` returns 404 instead of `index.html`, and the React app is unreachable through the API server in production.

Fix: in [api/src/Main.scala](api/src/Main.scala), append a static-files route (e.g. `Routes.serveDirectory(Path.empty, java.io.File(cfg.http.staticDir))` from zio-http) after the API routes, with an SPA fallback (any non-API path that 404s should fall back to `index.html` so client-side routing works).

Surfaced while wiring up the docker staging stack — I had to use a non-static healthcheck (POST /api/auth/login → 401) because `/` doesn't serve.

Issue #1636, opened 2026-06-10
bug(prod): Google OAuth hosts in household-global block list — connection events show 'blocked (household)' for accounts.google.com / oauth2.googleapis.com


sameerparekh:
## Symptom (prod, observed 2026-06-10)

While trying to load Google Drive on the kids' iPads to share files, operator hit blocks on Google OAuth flows. Connection-event log shows reason `blocked (household)` for the affected hosts.

Per `web/src/types/blockReason.ts`, "blocked (household)" is the SPA's label for `ExtraBlocked` whose source is the household-global `global_blocks` table (populated by admins via `POST/DELETE /api/global/blocks`). The snapshot copies those hosts into every profile's `extraBlocked`, so the block applies to everyone — kids and adults.

**There should NOT be Google OAuth hosts in `global_blocks`.** Google sign-in is on the SPA's UI-allowed-hosts list spirit (we explicitly carve around blocks for things the household needs to authenticate to). Blocking OAuth here breaks any Google service for every device on the network.

## Likely root causes (ranked)

1. **Operator accidentally added one or more Google hosts to the household block list** via the admin SPA / API. Verify via `SELECT host, added_by, added_at, reason FROM global_blocks WHERE removed_at IS NULL` on prod; look for anything matching `accounts.google.com`, `oauth2.googleapis.com`, `accounts.youtube.com`, `*.googleapis.com`. Likely an over-broad entry like `googleapis.com` was added meaning to block one specific Google service.
2. **Bundled-blocklist sourcing accidentally added it as a global block** (would show as `Category`, not `ExtraBlocked`; less likely given the symptom).
3. **A migration/seed inserted it** as part of a default-deny list. Check Flyway migrations for any `INSERT INTO global_blocks`.
4. **A bug in the global-block write path** — auto-adding entries from somewhere unintended (very unlikely).

## Investigation steps (observe before theorize)

1. **Read prod `global_blocks`**: list all active rows with timestamps and adder. If a Google host is there, that's the immediate fix target.
2. **Audit history**: the `added_by` column says who added the entry. If it's the operator, this is just a settings cleanup. If it's null / "system" / a migration, that's a code bug.
3. **Check for over-broad wildcards**: an entry like `googleapis.com` would block everything under that apex via the FQDN-resolution path. Per AGENTS.md, hosts in `global_blocks` are exact strings — but if dnsmasq's ipset population picks up subdomains, an apex entry blocks the whole tree. Verify intent.

## Fix

[...]

PR #1643, opened 2026-06-11
feat(#573): SNI-based hostname attribution (v1: libpcap-userspace, TLS-only)

sameerparekh:
## Summary

Adds v1 of the SNI-based hostname **attribution** sidecar (#573).

A new `wifihaven-sni-tail` procd sidecar captures the first packet of every outbound TCP/443 flow on the LAN bridge (probed from `network.lan.device`, default `br-lan`) via `tcpdump-mini`, parses the TLS ClientHello, and emits `SNI\t<dst_ip>\t<src_mac>\t<server_name>` lines into `/tmp/wifihaven-sni.log`. `wifihaven-dns-tail` tails that spool alongside the dnsmasq query log and routes SNI rows through `sni.route_line` → `cache.insert_sni`. The shared `paths.dns_cache` therefore stays **single-writer** (dns-tail) — no agent / router / wire changes.

**What this closes — attribution only.** SNI improves the agent's flow-attribution cache (`cache.lookup`), so `connection_attempt` / conntrack events get the branded `fqdn` even when the client used DoH/DoT or a hard-coded IP and dnsmasq never saw a query. It does **not** feed the `eb_` / `ea_` / `bl_` nftables drop-set populators: those fire solely on dnsmasq `reply <name> is <ip>` lines, and SNI is observed *after* the flow's first packet — too late to populate a drop set for that flow. So this PR closes the DoH/hard-coded-IP **attribution** gap behind the #1636 / #1640 class of *mislabeled* events; **enforcement** on hard-coded-IP traffic remains `blockIpOnly`'s job, and DoH/DoT egress blocking is the separate #572.

## Design choice

| Path | Verdict |
|---|---|
| **libpcap-userspace (tcpdump-mini)** | **chosen.** Slots into the existing sidecar-on-tmpfs pattern, no new kernel-side toolchain. Single OpenWRT package add. CPU is ~1 parse per new TLS flow — fine on the slowest supported router. |
| nft-ingress-eBPF | Ruled out — would introduce a brand-new libbpf + clang cross-compile dependency surface for a feature whose CPU budget does not require kernel-side parsing. |
| tc-bpf | Ruled out — same dependency-surface problem, plus bypasses the nft pipeline that all other enforcement flows through. |

## What is in v1

- TLS 1.2 / TLS 1.3 single-segment ClientHello SNI parsing on TCP dst port 443
- BPF capture filter restricting work to first-packet-of-flow ClientHellos (`tcp[((tcp[12]&0xf0)>>2)] = 0x16` AND the handshake byte is `0x01`)
- LAN bridge probed from UCI (`sni.lan_device` → `network.lan.device`, fallback `br-lan`); the "no pcap stream" error names the interface so a wrong bridge is diagnosable
- `sni.route_line(line, {insert_fn})` — the single dns-tail routing decision (SNI → `cache.insert_sni`, `tail -F` `==> file <==` banners skipped, dnsmasq lines fall through), extracted so the integration seam is unit-tested
- `cache.insert_sni(dst_ip, server_name[, now])` runs the SNI through the existing `resolve_head` CNAME-alias map (#1344) before `store()` — same primitive DNS ingest uses
- `sni.parse_packet` returns a bounded failure **reason** (`truncated` / `ipv6_skipped` / `not_tcp` / `malformed` / `no_sni`); the sidecar splits the formerly-lumped `no_sni` counter and surfaces each category in its periodic log line
- **Metrics:** sni-tail writes a cumulative per-result tally to `paths.sni_metrics` (same tmpfs-IPC + `format_query_results` shape as #1302's `dns_metrics`); the agent folds it into `sni_clienthellos_total{result}` on its metrics tick, and the metric is allowlisted in `MetricGuard.Allowed`
- `/tmp/wifihaven-sni.log` bounded by the existing 10-min `wifihaven-rotate-dnsmasq-log` cron
- New pure-Lua TLS parser module `openwrt/files/usr/lib/lua/wifihaven/sni.lua`, busted-tested
- `+tcpdump-mini` added to the OpenWRT package manifest (Makefile + build-ipk.sh + build-apk.sh)
- procd `sni-tail` instance in `/etc/init.d/wifihaven`, asserted by `init_spec.sh`

## What is explicitly out of scope (followups under #573)

- ECH (Encrypted ClientHello) — parser returns `nil`, counted as `no_sni`
- QUIC / HTTP/3 (UDP 443) Initial-packet SNI parsing
- IPv6 outer header (counted as `ipv6_skipped`)
- TLS 1.2 fragmented ClientHello reassembly across TCP segments
- UCI toggle to disable the sidecar per-router
- Static label map for hardcoded IP ranges (Apple APNs 17.0.0.0/8 etc.)
- **Grafana dashboard panel** for `sni_clienthellos_total{result}` (the metric itself ships in this PR; the panel + any agent-fold polish is the deferred follow-up)
- #572 DoH/DoT egress blocking (complementary, not blocking)

## Tests

`openwrt/test/sni_spec.lua` (42 specs) covers the TLS/Ethernet/IPv4/TCP parser, the `format_sni_line`/`parse_sni_line` round-trip, and the review-driven additions:

- `route_line`: SNI lines → `insert_fn`, `tail -F` switch banners skipped, dnsmasq lines fall through, empty/nil → not handled
- `parse_packet` failure reason second return value (`ipv6_skipped` / `truncated` / `no_sni`)
- `lan_device`: probes an injected UCI cursor, falls back to `br-lan`
- `parse_client_hello` rejects <3-char SNI, accepts `a.b`

**In-process pcap → cache integration (`sni_spec.lua` §9).** The
`wifihaven-sni-tail` pcap-stream reader is no longer untested inline
sidecar code: the byte-order-detecting global-header read, the
endianness-honoring 16-byte record framing, and the read loop are
extracted into `sni.lua` as pure, injectable functions
(`sni.read_global_header` / `sni.read_record` / `sni.iter_pcap`, backed
by a `read_fn(n)→bytes|nil` cursor), and the sidecar is now a thin
driver around them (single source of truth — one tested copy of the
framing logic, not a parallel inline one). A new integration block
assembles a synthetic pcap stream (24-byte global header + record
headers framing real Ethernet/IPv4/TCP packets carrying known-SNI
ClientHello fixtures) and drives the **full assembled pipeline**
in-process with no router or tcpdump: `iter_pcap → parse_packet →
format_sni_line → route_line → fake cache`, asserting the known
`dst_ip → SNI host` attribution lands in the cache. It also pins:
**both pcap byte orders** parse identically (covers review NIT #4 —
pcap header endianness, previously "verify by walkthrough" only); a
**truncated trailing record header** terminates the iterator cleanly
with no Lua error; and a **non-TLS packet** is skipped without
polluting the cache. This is the automated coverage of the assembled
reader + `route_line→cache` seam that previously existed only as
on-router manual steps.

`openwrt/test/dns_log_spec.lua` — `cache:insert_sni` stores so `cache.lookup` returns the SNI hostname, runs it through the CNAME-alias map, and is a no-op on nil/empty.

`openwrt/test/rotate_dnsmasq_log_spec.sh` — asserts the SNI spool is bounded by the rotation cron.

`api/test/src/feature/RouterMetricsIngestSpec.scala` — `sni_clienthellos_total{result}` re-exposes under `router_id` (additive `MetricGuard.Allowed` entry, no migration).

`openwrt/test/init_spec.sh` — the `wifihaven-sni-tail` procd instance is wired.

**VM e2e (Gate 2)** — `scripts/e2e/scenarios_fake/test_sni_attribution.py` (marker `sni_attribution`): a REAL router-VM end-to-end test of the sidecar, run automatically by the `e2e-vm-fake.yml` gate (pytest `testpaths` auto-collects the whole `scenarios_fake/` dir; both the ipk and apk router images bake `tcpdump-mini` + the `wifihaven-sni-tail` binary, so the scenario runs with no workflow change). It isolates SNI as the SOLE attribution source: the client emits a TLS ClientHello (`curl --resolve`, no DNS lookup) for `e2e-sni.wifihaven.net` — a name it never `dig`s — pinned to the router's own br-lan IP (a throwaway TCP/443 listener stood up on the router so the handshake completes and the ClientHello crosses br-lan). The hard, origin-independent assertion polls the router's shared dns→host cache (`/tmp/wifihaven-dns-cache.txt`) until it maps `<router-lan-ip> → e2e-sni.wifihaven.net` — unambiguous proof of the full pipeline (`ClientHello → br-lan → tcpdump-mini → sni.lua → cache.insert_sni → dns-cache`), since that name has no DNS-derived row. The forwarded `connection_event` SNI-labeling path is skipped (not red-gated): it needs a forwarded reachable 443 origin the fake harness lacks (the same #1360 missing-origin gap the Suite G CNAME allow-path xfails on).

## Field validation (operator, on `openwrt.lan`)

The assembled pcap → cache pipeline is now covered automatically by the
in-process integration test above; the steps below remain as on-router
**field validation** against a live tcpdump / dnsmasq / conntrack stack
(real capture, real interface, real flow attribution) — they are no
longer the only e2e coverage.

1. After upgrading the agent, `logread | grep sni-tail` should show `sni-tail: starting iface=<lan-bridge> snaplen=600 spool=/tmp/wifihaven-sni.log`.
2. To prove SNI is the source of attribution (rather than the dnsmasq path), temporarily set `uci set dhcp.@dnsmasq[0].logqueries=0` and restart dnsmasq.
3. From a LAN device (e.g. the Kids-profile iPad), `curl -v https://calendar.google.com`. Within 2 s expect `/tmp/wifihaven-sni.log` to contain a row mapping the resolved IP to `calendar.google.com`, and `/tmp/wifihaven-dns-cache.txt` to mirror it.
4. Generate a new conntrack flow to the same IP and confirm the resulting `connection_event` POST carries `fqdn=calendar.google.com`.
5. Re-enable `logqueries=extra` and run mixed `curl` traffic for 60 s while tailing the cache file — verify no torn or malformed lines (single-writer invariant preserved). The dns-tail 60-s heartbeat should report `sni_inserts=N`, and the sni-tail line should report `parsed/truncated/no_sni/ipv6_skipped/...`.

## Follow-ups to file as separate issues

- ECH (Encrypted ClientHello) detection and graceful degradation
- QUIC / HTTP/3 Initial-packet SNI parsing (UDP 443)
- IPv6 outer-header support in `sni.parse_packet`
- TLS 1.2 fragmented ClientHello reassembly
- UCI toggle to disable the SNI sidecar per-router
- Static label map for hardcoded IP ranges (Apple APNs 17.0.0.0/8)
- Grafana panel for `sni_clienthellos_total{result}`
- #572 DoH/DoT egress blocking (complementary, not blocking)

### DIFF
--- a/.github/workflows/e2e-vm-fake.yml
+++ b/.github/workflows/e2e-vm-fake.yml
@@ -70,7 +70,14 @@ jobs:
     # api.lan now eats most of the headroom, so 30m no longer covers the
     # tail. 40m re-establishes the safety margin without masking a real
     # regression in the scenario pack itself.
-    timeout-minutes: 40
+    #
+    # Bumped 40→50m after #573 added the test_sni_attribution scenario: each
+    # VM scenario carries a fixed ~2m snapshot-restore + boot cost, so the
+    # 45th scenario pushed the slower apk arm to 40m23s — over the 40m cap —
+    # while the ipk arm passed at 32m30s. The pack itself is healthy (per-step
+    # times unchanged); the cap just needs to grow with the scenario count.
+    # 50m restores ~10m headroom on the slow arm without masking a regression.
+    timeout-minutes: 50
     # #685: matrix over ipk-23.05 and apk-25.12 router images. fail-fast=false
     # so a regression in one flavor still reports the other arm's status
     # (needed to surface apk-only regressions like the #531 history case).
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -110,6 +110,10 @@ object MetricGuard {
     "agent_uptime_seconds"                      -> Set("router_id", "installation_id"),
     "agent_version"                             -> Set("version", "router_id", "installation_id"),
     "dns_queries_total"                         -> Set("result", "router_id", "installation_id"),
+    // #573 — TLS ClientHello SNI capture outcomes from the wifihaven-sni-tail sidecar.
+    // `result` ∈ {parsed, truncated, no_sni, ipv6_skipped, not_tcp, malformed} — a small bounded
+    // enum that lets an operator see the fleet-wide SNI capture / truncation / ipv6-skip rate.
+    "sni_clienthellos_total"                    -> Set("result", "router_id", "installation_id"),
     "blocklist_fetch_failures_total"            -> Set("status", "router_id", "installation_id"),
     "enforcement_drops_total"                   -> Set("reason", "router_id", "installation_id"),
     // #1033 — usage-POST retry-queue health. Depth = buckets currently waiting for a backoff to
--- a/api/test/src/feature/RouterMetricsIngestSpec.scala
+++ b/api/test/src/feature/RouterMetricsIngestSpec.scala
@@ -212,6 +212,10 @@ object RouterMetricsIngestSpec
             MetricCounter("blocklist_fetch_failures_total", Map("status" -> "5xx"), 4),
             MetricCounter("enforcement_drops_total", Map("reason" -> "whole_mac"), 1820),
             MetricCounter("enforcement_drops_total", Map("reason" -> "category_block"), 12990),
+            // #573 SNI sidecar result counters — must be allowlisted and
+            // re-exposed under router_id like the other agent-sourced series.
+            MetricCounter("sni_clienthellos_total", Map("result" -> "parsed"), 8421),
+            MetricCounter("sni_clienthellos_total", Map("result" -> "truncated"), 37),
           ),
         )
         resp <- post(routes, tok, body)
@@ -239,6 +243,14 @@ object RouterMetricsIngestSpec
             s"""router_id="$ridStr"""",
             "category_block",
           ).contains(12990.0),
+        ) &&
+        assertTrue(
+          seriesValue(scraped, "sni_clienthellos_total", s"""router_id="$ridStr"""", "parsed")
+            .contains(8421.0),
+        ) &&
+        assertTrue(
+          seriesValue(scraped, "sni_clienthellos_total", s"""router_id="$ridStr"""", "truncated")
+            .contains(37.0),
         )
       }
     },
--- a/openwrt/Makefile
+++ b/openwrt/Makefile
@@ -23,7 +23,12 @@ define Package/wifihaven
   # listen_https block-page sibling on :8443; openssl-util provides
   # /usr/bin/openssl for the self-signed cert generator that ships at
   # /usr/lib/wifihaven/generate-block-page-cert.sh.
-  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nf-log +kmod-nf-log6 +libustream-mbedtls +openssl-util
+  # #573: tcpdump-mini provides libpcap-userspace capture for the SNI sidecar
+  # (wifihaven-sni-tail). It parses the TLS ClientHello off the LAN bridge to
+  # recover hostname attribution when the client used DoH/DoT or a hardcoded
+  # IP — the same shared dns_cache feeds eb_/ea_/bl_ enforcement, so this is
+  # an additive, observability-style dependency.
+  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nf-log +kmod-nf-log6 +libustream-mbedtls +openssl-util +tcpdump-mini
   PKGARCH:=all
 endef
 
@@ -44,6 +49,7 @@ define Package/wifihaven/install
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/files/usr/sbin/wifihaven-agent $(1)/usr/sbin/
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/files/usr/sbin/wifihaven-dns-tail $(1)/usr/sbin/
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/files/usr/sbin/wifihaven-nflog-tail $(1)/usr/sbin/
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/files/usr/sbin/wifihaven-sni-tail $(1)/usr/sbin/
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/files/usr/sbin/wifihaven-update $(1)/usr/sbin/
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/files/usr/sbin/wifihaven-rotate-dnsmasq-log $(1)/usr/sbin/
 
--- a/openwrt/build-apk.sh
+++ b/openwrt/build-apk.sh
@@ -146,7 +146,7 @@ rm -f "$OUT_APK"
     --info "license:MIT" \
     --info "url:https://github.com/wifihaven/wifihaven" \
     --info "maintainer:WifiHaven <noreply@example.com>" \
-    --info "depends:lua libuci-lua luci-lib-jsonc conntrack curl uhttpd-mod-lua kmod-nf-log kmod-nf-log6 libustream-mbedtls openssl-util" \
+    --info "depends:lua libuci-lua luci-lib-jsonc conntrack curl uhttpd-mod-lua kmod-nf-log kmod-nf-log6 libustream-mbedtls openssl-util tcpdump-mini" \
     --script "post-install:$WORK/post-install" \
     --script "trigger:$WORK/trigger" \
     --trigger "/etc/crontabs" \
--- a/openwrt/build-ipk.sh
+++ b/openwrt/build-ipk.sh
@@ -25,7 +25,7 @@ mkdir "$WORK/ctrl"
 cat > "$WORK/ctrl/control" <<EOF
 Package: wifihaven
 Version: ${PKG_VERSION}-${PKG_RELEASE}
-Depends: lua, libuci-lua, luci-lib-jsonc, conntrack, curl, uhttpd-mod-lua, kmod-nf-log, kmod-nf-log6, libustream-mbedtls, openssl-util
+Depends: lua, libuci-lua, luci-lib-jsonc, conntrack, curl, uhttpd-mod-lua, kmod-nf-log, kmod-nf-log6, libustream-mbedtls, openssl-util, tcpdump-mini
 Architecture: all
 Maintainer: WifiHaven <noreply@example.com>
 Description: Agent daemon for WifiHaven. Enforces per-device DNS filtering
--- a/openwrt/files/etc/init.d/wifihaven
+++ b/openwrt/files/etc/init.d/wifihaven
@@ -22,6 +22,19 @@ start_service() {
     procd_set_param stderr 1
     procd_close_instance
 
+    # sni-tail sidecar: captures the first packet of every outbound TCP/443 flow
+    # via tcpdump-mini, extracts the TLS ClientHello SNI, and appends parsed
+    # (dst_ip, src_mac, sni) lines to /tmp/wifihaven-sni.log. wifihaven-dns-tail
+    # tails that spool alongside the dnsmasq log and routes SNI lines through
+    # the shared (single-writer) dns_cache. Closes the DoH / hard-coded-IP
+    # attribution gap (#573).
+    procd_open_instance "sni-tail"
+    procd_set_param command /usr/sbin/wifihaven-sni-tail
+    procd_set_param respawn 3600 5 5
+    procd_set_param stdout 1
+    procd_set_param stderr 1
+    procd_close_instance
+
     # nflog-tail sidecar: tails `logread -f` for the nft `log prefix "wh_drop:…"`
     # forward-drop records and spools them to /tmp/wifihaven-nflog.log so the
     # agent can synthesize connection_attempt events for blocked flows (#1126).
--- a/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
@@ -401,6 +401,23 @@ function M.new(opts)
     return resolve_head(name)
   end
 
+  -- insert_sni(dst_ip, server_name, now) — record an (ip → hostname) mapping
+  -- learned from a TLS ClientHello SNI extension (#573). The wifihaven-sni-tail
+  -- sidecar feeds these in via the same line stream as dnsmasq replies; this
+  -- helper runs the server_name through the SAME CNAME-alias map (resolve_head)
+  -- DNS ingest uses, so an SNI to a CDN leaf target attributes back to the
+  -- branded chain head we already learned from the DNS path (#1344). A
+  -- subsequent cache.lookup(dst_ip) then matches the same eb_/ea_/bl_ suffix
+  -- walks the DNS-derived entries do — no router code is aware of the SNI
+  -- origin. nil-safe.
+  function self.insert_sni(dst_ip, server_name, now)
+    if type(dst_ip) ~= "string" or dst_ip == "" then return end
+    if type(server_name) ~= "string" or server_name == "" then return end
+    now = now or now_fn()
+    local head = resolve_head(server_name)
+    store(dst_ip, head, now)
+  end
+
   -- Serialise the live cache to a single newline-delimited string.
   -- Format: "<ip>\t<hostname>\t<ts>\n" per non-expired entry.
   function self.dump_text()
--- a/openwrt/files/usr/lib/lua/wifihaven/paths.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/paths.lua
@@ -32,9 +32,20 @@ M.bl_member_index = "/tmp/wifihaven-bl-members.txt"
 -- and appends one "SNI\t<dst_ip>\t<src_mac>\t<server_name>" line per capture to
 -- this tmpfs file. wifihaven-dns-tail tails it alongside the dnsmasq query log
 -- and routes SNI-prefixed lines through cache.insert_sni, so the shared
--- ip→hostname cache (paths.dns_cache) stays single-writer (dns-tail) and the
--- downstream eb_/ea_/bl_ populators see SNI-derived hostnames transparently.
+-- ip→hostname cache (paths.dns_cache) stays single-writer (dns-tail). This
+-- improves connection-event fqdn ATTRIBUTION only (cache.lookup); it does NOT
+-- feed the eb_/ea_/bl_ nft drop-set populators, which fire solely on dnsmasq
+-- `reply <name> is <ip>` lines — SNI is observed after the flow's first packet,
+-- too late to populate a drop set for that flow.
 M.sni_capture = "/tmp/wifihaven-sni.log"
+
+-- SNI capture result tally (#573): wifihaven-sni-tail writes cumulative
+-- per-result counts ("<result>\t<count>" lines, same format as dns_metrics) on
+-- each periodic flush; wifihaven-agent samples it on its metrics tick and folds
+-- the deltas into sni_clienthellos_total{result}. result ∈ {parsed, truncated,
+-- no_sni, ipv6_skipped, not_tcp, malformed}. Same tmpfs-IPC pattern as
+-- dns_metrics — the sidecar has no metrics registry of its own.
+M.sni_metrics = "/tmp/wifihaven-sni-metrics.txt"
 
 -- nflog drop spool (#1126): wifihaven-nflog-tail tails `logread -f`, greps the
 -- nft `log prefix "wh_drop:…"` forward-drop records, and appends them here; the
--- a/openwrt/files/usr/lib/lua/wifihaven/paths.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/paths.lua
@@ -27,6 +27,15 @@ M.dns_metrics = "/tmp/wifihaven-dns-metrics.txt"
 -- wifihaven-dns-tail bl_ populator reads it on its set-refresh cadence.
 M.bl_member_index = "/tmp/wifihaven-bl-members.txt"
 
+-- SNI capture spool (#573): wifihaven-sni-tail parses the first packet of every
+-- outbound TCP/443 flow off the LAN bridge, extracts the TLS ClientHello SNI,
+-- and appends one "SNI\t<dst_ip>\t<src_mac>\t<server_name>" line per capture to
+-- this tmpfs file. wifihaven-dns-tail tails it alongside the dnsmasq query log
+-- and routes SNI-prefixed lines through cache.insert_sni, so the shared
+-- ip→hostname cache (paths.dns_cache) stays single-writer (dns-tail) and the
+-- downstream eb_/ea_/bl_ populators see SNI-derived hostnames transparently.
+M.sni_capture = "/tmp/wifihaven-sni.log"
+
 -- nflog drop spool (#1126): wifihaven-nflog-tail tails `logread -f`, greps the
 -- nft `log prefix "wh_drop:…"` forward-drop records, and appends them here; the
 -- main agent drains new complete lines from this tmpfs spool on its cooperative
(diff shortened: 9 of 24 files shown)


sameerparekh:
## Independent PR review — `feat(#573)` SNI sidecar v1

Read-only adversarial pass against `10d79b4` per [docs/pr-review-checklist.md](docs/pr-review-checklist.md). Tests run green locally (`busted test/sni_spec.lua` 15/0; `busted test/dns_log_spec.lua` 59/0). Red-then-green TDD progression is visible in commit history (`7c60e67` red, `10d79b4` green).

### BLOCKERS

(none)

### SHOULD-FIX

1. **`openwrt/files/usr/sbin/wifihaven-dns-tail:275-307` — no test coverage for the SNI/dnsmasq routing branch.** The 51-line behavior change to dns-tail (`handle_sni_or_skip`, the new `else` branch, the SNI-only flush bookkeeping at 296-306) has zero spec coverage. `sni_spec.lua` exercises `parse_sni_line` and `dns_log_spec.lua` exercises `cache.insert_sni`, but the routing function inside the sidecar script is untested. Extract `handle_sni_or_skip` into a library module (or attach it to `M` in `sni.lua`) so a spec can assert (a) `SNI\t...` lines route to `cache.insert_sni`, (b) `==> /tmp/wifihaven-sni.log <==` banner lines are skipped, (c) dnsmasq lines fall through. This is the integration seam that closes the gap — and per AGENTS.md "Testing integrity" the new behavior should be asserted, not merely compiled.

2. **PR body overstates downstream wiring.** The summary claims "the existing `eb_` / `ea_` / `bl_` populators … all consume SNI-derived hostnames transparently." They don't — `maybe_populate_eb` / `maybe_populate_ea` / `maybe_populate_bl` in `wifihaven-dns-tail:196-250` all start with `dns_log.parse_resolved_reply(line)`, which only matches dnsmasq `reply <name> is <ip>` lines. SNI lines short-circuit past them (line 294 routes SNI then continues the outer loop). So SNI improves `cache.lookup` (connection-event `fqdn` labeling), but does NOT cause SNI-discovered IPs to be added to the per-host nft drop sets. This is fine architecturally for v1 (SNI is observed after the first packet anyway, too late to populate the drop set for that flow), but the PR body should be corrected so a reader doesn't conclude DoH/hardcoded-IP enforcement is closed. Today only attribution is closed.

3. **`openwrt/files/usr/sbin/wifihaven-sni-tail:43` — IFACE hardcoded to `br-lan`.** Comment says "A future UCI toggle can override this; v1 hardcodes the convention." On routers using a non-default bridge name (some custom-built configs, multi-WAN deployments, or routers where the LAN bridge is named `br0`), the sidecar will silently produce no SNI rows. At minimum, log a warning on `tcpdump produced no pcap stream` (line 110) that names IFACE so the operator can diagnose it. Better: probe the actual LAN bridge from UCI (`network.lan.device`) at startup. The "always-on" design (no UCI toggle, line 17) compounds this — there's no way to disable on a misconfigured router short of removing the file.

4. **`openwrt/files/usr/sbin/wifihaven-sni-tail:50` — snaplen=600 may truncate real TLS 1.3 ClientHellos.** Modern TLS 1.3 ClientHellos with hybrid post-quantum key shares (X25519+Kyber768 = ~1200 bytes for `key_share` alone), GREASE entries, ALPN, and ECH-stub extensions can reach 1500+ bytes. The parser correctly returns nil on truncation, but the operator sees no signal: `no_sni_total` (line 145) lumps "truncated", "malformed", "no SNI extension", and "ECH" together. At minimum split that counter (`truncated_total`, `no_sni_ext_total`, `parse_error_total`) so a fleet-wide truncation rate is visible. The PR comment acknowledges this ("Larger ClientHellos with long key_share / cipher lists may be truncated") but doesn't instrument it. The same applies to the `sni-tail: parsed=N (+M) no_sni=K` log line at 150 — categorize the failure mode.

5. **AGENTS.md "new functionality ships with metrics" — sni-tail emits no Prometheus metric.** PR body defers `sni_clienthellos_total{result}` as a follow-up "once the path is proven on the fleet." That's a defensible deferral, but the standing rule is to ship the metric in the same PR; otherwise the path can't BE proven on the fleet without on-router shell access. At minimum write the counters to a tmpfs IPC file (mirroring `paths.dns_metrics` for #1302) so the agent's metrics tick can fold them, even without registering them in the dashboard yet. The `result=parsed|truncated|no_sni|ech|malformed` enum is small and bounded.

6. **`openwrt/files/usr/sbin/wifihaven-sni-tail:115-119` — spool file opened append-only with no rotation.** `/tmp/wifihaven-sni.log` grows unbounded across the sidecar's lifetime. dns-tail's `tail -F` follows it, but tmpfs RAM pressure on a low-end router is real. The dnsmasq log is rotated by `/usr/sbin/wifihaven-rotate-dnsmasq-log` on a 10-minute cron (Makefile:110); add the SNI spool to that script, or truncate periodically. At one SNI line per new TLS flow on a household router this is probably 10–50 KB/hour, so 24h = ~1 MB — not catastrophic but should be bounded.

7. **`openwrt/files/usr/lib/lua/wifihaven/sni.lua:170` — IPv4-only check is silent.** `parse_packet` returns nil for ethertype != 0x0800 (line 176). Modern client behavior is dual-stack; if a client uses IPv6 to reach a CDN via SNI, the flow is silently skipped. This is documented in v1 scope, but combined with `no_sni_total` lumping at sni-tail:145 there's no way to know what fraction of traffic is being missed. At minimum count IPv6-skipped separately.

### NIT

1. **`openwrt/files/usr/sbin/wifihaven-dns-tail:307-383` — indentation in the new `else` branch is misleading.** Lines 308-382 sit at the original (non-nested) indentation level but are inside `else ... end`. Re-indent the `else` body so the structural change is visually clear; this is a 75-line block.

2. **`openwrt/files/usr/sbin/wifihaven-sni-tail:38-40` — BPF filter accepts retransmitted ClientHellos.** The first-packet-of-flow filter matches any TCP/443 segment whose first payload byte is 0x16 and 6th byte is 0x01. A retransmitted ClientHello (same flow, lost SYN-ACK or PMTU issue) will be re-captured and re-parsed. Harmless because `store` overwrites with the same (ip → host) value, but it's wasted CPU. Could add `tcp[tcpflags] & tcp-syn = 0` to filter out the (very rare) SYN-with-payload case and rely on the data offset / length matching first-packet semantics.

3. **`openwrt/files/usr/lib/lua/wifihaven/sni.lua:144-146` — empty SNI returns nil but a single-character SNI returns it.** Real-world SNI is always a domain; a 1-byte SNI is malformed. Consider a stricter length check (≥3, like `a.b`). Not a security issue because the value is only used as a cache key.

4. **`openwrt/files/usr/sbin/wifihaven-sni-tail:65-70` — pcap header endianness is partially handled.** `read_global_header` detects byte swap, but `read_record_header` uses it for `incl_len` (line 105) — correct. Just verify by walkthrough; no fix needed.

5. **`openwrt/files/usr/lib/lua/wifihaven/sni.lua:148` — comment "list MAY contain multiple but in practice it's always one host_name" is accurate, but the code only walks until it finds the first host_name and returns. If a malformed-but-not-rejected entry appears first (name_type != 0x00 and the parser advances by sn_len), it'll continue. Fine — defensive code; just confirming.

### Architectural / single-source-of-truth check

✓ **Single source of truth**: SNI flows through `cache.insert_sni` → `resolve_head` → `store`. Same primitive as DNS ingest. No parallel resolver, no duplicated decision logic. `resolve_head` is exposed publicly and reused — this is exactly the pattern AGENTS.md §single-source-of-truth prescribes.

✓ **DNS is never the enforcement plane**: sidecar is observability/attribution only. No new enforcement field, no DNS-layer block. Block-set population paths are unchanged.

✓ **Router is a dumb applier**: no decision logic on the router. SNI is local hostname-attribution data, not policy.

✓ **Backwards compat / wire shape**: no API or wire change. Purely additive at the router; an older agent build without sni-tail keeps working.

✓ **Single-writer on `paths.dns_cache`**: dns-tail remains the sole writer of `/tmp/wifihaven-dns-cache.txt`. sni-tail writes only to `paths.sni_capture`. No race.

✓ **TDD red→green visible**: `7c60e67` ships fixtures first, `10d79b4` adds implementation.

### Verdict

**VERDICT: REQUEST-CHANGES**

No blockers, but the test coverage gap on the new dns-tail SNI routing branch (SHOULD-FIX #1) and the PR-body overstatement of downstream wiring (SHOULD-FIX #2) should be addressed before merge. The metrics-gap (#5) and snaplen-truncation observability gap (#4) are merge-or-defer judgment calls — defer is acceptable if a follow-up issue is filed alongside merge.

sameerparekh:
## Independent PR review (re-run) — `feat(#573)` SNI sidecar

Incremental re-review against `a367c60c` per [docs/pr-review-checklist.md](docs/pr-review-checklist.md). Prior pass was `10d79b4` (REQUEST-CHANGES, 0 BLOCKER / 7 SHOULD-FIX / 5 NIT). Two new commits since: `7336055e` (test-red) → `a367c60c` (fix-green).

Verified locally at HEAD: `busted test/sni_spec.lua` **28/0**, `sh test/rotate_dnsmasq_log_spec.sh` **25/0**. Red→green confirmed real — checked out `7336055e` and the new tests fail against the red-tree source (`lan_device`/`route_line` nil; SNI-spool rotation assertions fail, `22 passed, 3 failed`), then pass at green.

### Prior-findings status

| # | Sev | Finding (10d79b4) | Status | Notes |
|---|-----|-------------------|--------|-------|
| SF-1 | SHOULD-FIX | No test for dns-tail SNI routing branch | **ADDRESSED** | Routing decision extracted to `sni.route_line(line, {insert_fn})` (`sni.lua:270`); dns-tail calls it via `route_sni` (`wifihaven-dns-tail:278`). 4 new specs assert SNI→insert, banner-skip, dnsmasq fall-through, empty/nil. Tests use real injection, not just compile. |
| SF-2 | SHOULD-FIX | PR body overstated downstream wiring | **ADDRESSED** | PR body + `paths.lua:35` + `sni-tail` header now state attribution-only; explicitly disclaim `eb_/ea_/bl_` population and point enforcement at `blockIpOnly`/#572. |
| SF-3 | SHOULD-FIX | IFACE hardcoded `br-lan` | **ADDRESSED** | `sni.lan_device(cursor)` probes `network.lan.device` with `br-lan` fallback (`sni.lua:286`); wired at `sni-tail:50`; "no pcap stream" error now names IFACE (`sni-tail:116`). 3 injected-cursor specs. |
| SF-4 / SF-7 | SHOULD-FIX | Lumped `no_sni_total` (truncation/ipv6 invisible) | **ADDRESSED** | `parse_packet` returns bounded reason (`truncated`/`ipv6_skipped`/`not_tcp`/`malformed`/`no_sni`); sidecar splits the counter and surfaces each in the periodic log line (`sni-tail:188`). Reason specs added. |
| SF-5 | SHOULD-FIX | No Prometheus metric emitted | **ADDRESSED** | tmpfs-IPC `paths.sni_metrics` (mirrors #1302 `dns_metrics`); agent `fold_sni_metrics` → `sni_clienthellos_total{result}` via the existing `fold_external` re-base primitive; allowlisted in `MetricGuard.Allowed`. Ingest spec asserts re-exposure. Dashboard panel deferred to filed **#1644** (acceptable documented deferral). |
| SF-6 | SHOULD-FIX | SNI spool unbounded | **ADDRESSED** | `wifihaven-rotate-dnsmasq-log` refactored to `rotate_if_big()` and now rotates `/tmp/wifihaven-sni.log` on the same 10-min cron, copytruncate (same inode → `tail -F` reseeks). Behavioural rotation spec added. |
| NIT-1 | NIT | Misleading indentation in dns-tail `else` block | **ADDRESSED** | Refactored to a clean `if route_sni(line) then … else …` with correct nesting. |
| NIT-2 | NIT | BPF accepts retransmitted ClientHellos | **NOT-ADDRESSED** | Unchanged — explicitly accepted as harmless (idempotent `store`). Fine to leave. |
| NIT-3 | NIT | 1-char SNI accepted | **ADDRESSED** | `parse_client_hello` now requires `#sn >= 3` (`sni.lua:148`); specs reject `x`, accept `a.b`. |
| NIT-4 | NIT | pcap endianness walkthrough | **ADDRESSED (confirmed)** | `read_record_header` honors `swapped` for `incl_len` — correct. |
| NIT-5 | NIT | first-host_name walk note | **N/A** | Confirmation only; no change required. |

All 7 SHOULD-FIX cleared; 3 of 5 NITs resolved, the 2 remaining are accepted-as-is.

### New findings (delta `10d79b4...a367c60c`)

**BLOCKERS** — none.

**SHOULD-FIX** — none.

**NITS**

1. `wifihaven-sni-tail:183` — `flush_metrics` runs on a 30s timer driven by `os.time()` inside the capture loop, which only advances when a packet arrives. On a fully idle 443 link the final tally isn't published until the next packet (or process exit, which does flush). Bounded and benign for a counter that only matters under traffic; noting for completeness, no fix needed.

### Architectural invariants (re-checked on the delta)

- ✓ **Single source of truth** — SNI metrics IPC reuses `dns_log.format_query_results`/`parse_query_results` and `metrics.fold_external` (sidecar-restart re-base already handled); `route_line` is the *single* routing decision, unit-tested. No re-derivation.
- ✓ **Single-writer `paths.dns_cache`** — preserved. `route_line`'s `insert_fn` is `cache.insert_sni`, the same dns-tail-owned writer; sni-tail writes only `paths.sni_capture` + `paths.sni_metrics`. The new metrics file is written by sni-tail and read by the agent — a separate path, no contention with dns_cache.
- ✓ **Metrics cardinality firewall** — `sni_clienthellos_total -> Set(result, router_id, installation_id)`; `result` is a small fixed enum, no per-mac/domain/ip label.
- ✓ **DNS-not-enforcement / dumb-applier / additive wire** — unchanged; sidecar is attribution-only, no API/wire/snapshot change, no router decision logic.
- ✓ **tmpfs-IPC race** — `flush_metrics` writes `.tmp` sibling + `os.rename` (atomic); agent reads whole file. No torn read.
- ✓ **Rotation safety** — copytruncate keeps sni-tail on the same inode (it holds the fd open, like dnsmasq); `tail -F` follows by name and reseeks on truncation. No mid-line truncation hazard beyond the already-documented bounded loss window.

### Verdict

**VERDICT: APPROVE**

All 7 prior SHOULD-FIX and the 3 actionable NITs are addressed; the 2 remaining NITs are explicitly accepted. The fixes introduce no new BLOCKER or SHOULD-FIX — the routing seam is properly extracted and injection-tested, the metric stays within the cardinality firewall and reuses existing fold primitives, single-writer `dns_cache` is intact, and the spool rotation is copytruncate-safe. Red→green is genuine and visible in history. Open BLOCKER: **0**. Open SHOULD-FIX: **0**.

---

*Informational (not a finding against this PR):* CI's **Shell Tests** job failed on `cert_spec.test.sh`'s mtime-idempotency assertion. This is a pre-existing flake from #1622, unrelated to this diff — the branch touches no cert files — and is already flagged for a separate fix.

sameerparekh:
## Independent PR review (re-run) — `feat(#573)` SNI sidecar

Incremental re-review against `44674640` per [docs/pr-review-checklist.md](docs/pr-review-checklist.md). Prior pass was `a367c60c` (**APPROVE**, 0 BLOCKER / 0 open SHOULD-FIX). Two new commits since: `8e392402` (test-red) → `44674640` (extract + green). Delta is **3 files, additive** (`sni.lua` +84, `wifihaven-sni-tail` reader rewired, `sni_spec.lua` +237) — exactly the PR's stated scope, no drive-by.

Verified locally at HEAD: `busted test/sni_spec.lua` **42/0** (was 28); `sh test/run_tests.sh` exit **0** (all busted + shell specs). Red→green confirmed real — checked out `8e392402` and the new §9 integration specs fail against the un-extracted module (`attempt to call a nil value (field 'iter_pcap')`), then pass at `44674640`.

### Prior-findings status

All prior findings were already cleared at `a367c60c` and live in earlier commits the delta does not touch — re-confirmed still present in current code.

| # | Sev | Finding | Status |
|---|-----|---------|--------|
| SF-1 | SHOULD-FIX | dns-tail SNI routing seam untested | **ADDRESSED** (`sni.route_line`, still at `sni.lua:331`+; specs intact) |
| SF-2 | SHOULD-FIX | PR body overstated wiring | **ADDRESSED** (attribution-only framing intact) |
| SF-3 | SHOULD-FIX | IFACE hardcoded br-lan | **ADDRESSED** (`sni.lan_device`, now `sni.lua:362`) |
| SF-4/7 | SHOULD-FIX | lumped no_sni_total | **ADDRESSED** (`parse_packet` bounded reason intact) |
| SF-5 | SHOULD-FIX | no metric emitted | **ADDRESSED** (`sni_clienthellos_total{result}`; dashboard deferred #1644) |
| SF-6 | SHOULD-FIX | spool unbounded | **ADDRESSED** (rotation cron intact) |
| NIT-1/3/4/5 | NIT | (indentation / <3-char SNI / endianness / host_name walk) | **ADDRESSED / confirmed** |
| NIT-2 | NIT | BPF accepts retransmits | **NOT-ADDRESSED** (accepted-as-is, harmless) |

This re-run **closes prior NIT-4 (pcap endianness "verify by walkthrough")**: the byte-order path is now executed by an automated test asserting both orders parse identically, not just read by eye.

### New findings (delta `a367c60c...44674640`)

**BLOCKERS** — none.

**SHOULD-FIX** — none.

**NITS**

1. `sni.lua:309` (`read_record`) — when `incl_len` overruns (truncated tail), the 16 bytes of header already consumed from `read_fn` are discarded along with the partial packet. This is correct for the sidecar (a short final read means the pipe is closing → loop breaks → procd respawns) and is explicitly pinned by the "truncated trailing record header terminates cleanly" test, but the function gives the caller no way to distinguish *clean EOF* (nil at header) from *truncated record* (nil after a partial read). No consumer needs that distinction today; noting only for a future IPv6/reassembly follow-up. No fix needed.

### Architectural invariants (re-checked on the delta)

- ✓ **Single source of truth — the central win of this delta.** The pcap framing (global-header magic/byte-order detection, 16-byte record header with `swapped` `incl_len`, the exact-read loop) moved from untested inline locals in the sidecar into `sni.lua` (`read_global_header`/`read_record`/`iter_pcap`); the sidecar is now a thin driver backing `read_fn` with `h:read(n)`. One tested copy, not a parallel one. Extraction is **behavior-preserving**: same `b1==0xd4` swap detection, same `u32` byte-swap arithmetic (lifted verbatim), same `read_n`→`read_exact` accumulate-until-n loop, same global-header `os.exit(1)` diagnostic naming IFACE, same per-packet parse/spool/metrics path. The only consolidation is two adjacent warn strings (record-EOF vs short-packet) collapsing into one — both already `break`; no control-flow change. The full pre-existing suite stays green, which is the proof.
- ✓ **read_fn contract models io.popen faithfully.** Original `read_n` looped `handle:read(n-got)` accumulating partial reads; `read_exact` is the identical loop. The real handle blocks until ≥1 byte or EOF; the string cursor returns up to n or nil — both satisfy "up to n bytes, nil at EOF", and the framing is deterministic, so the abort-condition (a blocking-read subtlety a cursor can't model) does **not** apply.
- ✓ **Single-writer `paths.dns_cache`** — unchanged; the integration test's `insert_fn` is the same `route_line` seam, no new writer.
- ✓ **DNS-not-enforcement / dumb-applier / additive wire** — unchanged; no API/snapshot/wire change, sidecar stays attribution-only.
- ✓ **Testing integrity** — delta is purely additive to `sni_spec.lua` (no line removed from the existing 28 specs; `git diff` shows zero `-` on prior assertions). No mocks of repos/AuthService/Clock (router Lua, none applicable). New behavior is asserted (cache contents, both byte orders, truncation, non-TLS skip), not merely exercised.
- ✓ **Scope / hygiene** — three files, all in `openwrt/`; no AI mentions in commits/body; `Co-Authored-By` trailer present; `luac -p` clean on all touched files.

### Verdict

**VERDICT: APPROVE**

The delta extracts the previously-untested pcap reader into the requireable module (single-source-of-truth) and adds an in-process pcap→cache integration test covering the assembled pipeline, both byte orders, truncated-EOF, and non-TLS skip. Extraction is behavior-preserving (proven by the unchanged green suite); red→green is genuine and visible in history. No new BLOCKER or SHOULD-FIX; the one NIT is a benign API-shape note. Open BLOCKER: **0**. Open SHOULD-FIX: **0**.

---

*Informational (not a finding):* one transient full-suite failure observed during review traced to the reviewer's own `git stash`/`checkout` juggling leaving conflict markers in unrelated `wifihaven-update*` files; on a clean tree `sh test/run_tests.sh` exits 0. Not a property of this diff.

PR #1646, opened 2026-06-11
test(#1622): fix flaky cert_spec idempotency assertion (fingerprint, not mtime)

sameerparekh:
## Problem

`openwrt/test/cert_spec.test.sh`'s **"second run is idempotent"** assertion
compares the **mtime** of `block_page.crt` / `block_page.key` across two runs
of `generate-block-page-cert.sh`. It fails intermittently in CI (observed on
GitHub Actions run 27354778995, job *Shell Tests*) while passing reliably
locally.

## Root cause — the test, not the generator

The generator is **correctly idempotent**: it skips when both files already
exist and are non-empty (`[ -s "$CRT" ] && [ -s "$KEY" ]`), so a second run
never touches the files. Verified locally — second run logs `skipping` and
the cert fingerprint is byte-for-byte identical across runs.

So this is **case (a): the assertion is the fragile part.** mtime is only a
*proxy* for "the cert wasn't rotated", and the proxy races:

- the `sleep 1` + whole-second mtime comparison straddles the
  same-second filesystem-timestamp window, and
- the BSD/GNU `stat -f '%m' || stat -c '%Y'` fallback adds platform skew.

The generator itself needs no change.

## Fix

Assert on cert/key **identity** instead of mtime:

- cert **SHA-256 fingerprint** (`openssl x509 -fingerprint -sha256`)
- key **content hash** (`openssl dgst -sha256`)

This is **timing-independent** (the `sleep 1` is removed) and verifies the
invariant we actually care about: the cert was not regenerated. A rotated
cert would churn the uhttpd TLS listener on every boot — exactly the
regression worth catching — and a fingerprint comparison catches it
**deterministically** where mtime caught it only probabilistically.

## Validation

- New assertion runs **green deterministically** across repeated runs (no
  sleep, no timestamp dependency).
- **Red-test check:** temporarily neutering the generator's idempotency guard
  makes the new assertion **FAIL** with a clear cert/key fingerprint diff —
  confirming the assertion still catches genuine non-idempotency.
- Full shell suite green: `cert_spec` 6/6, `install_spec` 48/48,
  `update_multi_pkg_spec` 33/33.
- The lua suite was not runnable in this local env (missing `lua5.1` /
  `luci.jsonc`); no lua files are touched by this change, so CI exercises it
  unaffected.

### DIFF
--- a/openwrt/test/cert_spec.test.sh
+++ b/openwrt/test/cert_spec.test.sh
@@ -63,20 +63,26 @@ else
   check "cert key length >= 2048 bits" "unable to confirm key length"
 fi
 
-# Idempotent: a second run must NOT overwrite either file.
-CRT_BEFORE="$(stat -f '%m' "${CRT}" 2>/dev/null || stat -c '%Y' "${CRT}")"
-KEY_BEFORE="$(stat -f '%m' "${KEY}" 2>/dev/null || stat -c '%Y' "${KEY}")"
-sleep 1
+# Idempotent: a second run must NOT rotate either file. Assert on cert/key
+# *identity* (cert SHA-256 fingerprint + key content hash), NOT mtime. mtime is
+# a fragile proxy that this assertion previously used: idempotent runs leave the
+# file untouched so the mtime is stable, but a same-second filesystem-timestamp
+# window (and the BSD/GNU `stat` fallback) raced it intermittently in CI. The
+# real invariant we care about is that the cert was not regenerated — a rotated
+# cert would churn the uhttpd TLS listener on every boot — and a fingerprint
+# comparison verifies exactly that, with no timing dependency.
+CRT_FP_BEFORE="$(openssl x509 -in "${CRT}" -noout -fingerprint -sha256 2>/dev/null)"
+KEY_HASH_BEFORE="$(openssl dgst -sha256 "${KEY}" 2>/dev/null)"
 WIFIHAVEN_BLOCK_PAGE_CRT="${CRT}" WIFIHAVEN_BLOCK_PAGE_KEY="${KEY}" \
   sh "${SCRIPT}" >/dev/null 2>&1
-CRT_AFTER="$(stat -f '%m' "${CRT}" 2>/dev/null || stat -c '%Y' "${CRT}")"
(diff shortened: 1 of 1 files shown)


sameerparekh:
## Independent PR review

Reviewed `git diff origin/main...HEAD` — a single-file, test-only change to `openwrt/test/cert_spec.test.sh`. The generator (`generate-block-page-cert.sh`) is untouched.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `openwrt/test/cert_spec.test.sh:71-74` — If `openssl x509 -fingerprint`/`openssl dgst` ever produced empty output for *both* runs, the `BEFORE == AFTER` comparison would pass on two empty strings (a false green). In practice this is already guarded: the preceding assertions (cert present + non-empty, CN parse, key-length parse) would have failed first, and the file is known to exist. Not worth a guard, but noting the empty-equals-empty edge for the record.

**Dimension notes**
- **§2 Testing integrity (critical):** This *changes an existing assertion*, which is exactly what this dimension scrutinizes — but it **strengthens** rather than weakens it. mtime equality was a fragile proxy; the new cert-fingerprint + key-hash comparison asserts the real invariant (cert identity / no rotation) and is timing-independent. The PR body explains why the old assertion was wrong, and the author validated it red against a deliberately non-idempotent generator and green on the real one. This is the correct kind of test edit, not a regression-hiding one.
- **§1 Single source of truth:** No logic duplicated or re-derived.
- **§3 Architecture:** No router/DNS/snapshot semantics touched (test-only).
- **§7 Scope/hygiene:** Focused single-file diff; no drive-by edits; no AI mentions in title/body; `Co-Authored-By` trailer is fine. The `sleep 1` removal is appropriate (no longer needed without a timestamp dependency) and a minor CI speedup.
- **§4/§5/§6/§8:** N/A (no wire shape, migration, metric, route, or SQL).

VERDICT: APPROVE
A clean, well-scoped test-robustness fix: it replaces a flaky mtime-equality idempotency check with a deterministic cert-fingerprint + key-hash comparison that verifies the actual invariant (the cert was not rotated). Strengthens the assertion rather than weakening it, with red/green validation shown. No blockers.

PR #1656, opened 2026-06-11
feat(#1645): label extraBlocked events with the matched eb_<host> rule

### LINKED ISSUE #1645: feat: label extraBlocked events with the matched host (eb_<host> attribution)
## Problem

When an nftables \`eb_<host>\` rule drops a flow, the connection event ships \`reason=\"host\"\` with no information about *which* \`eb_<host>\` set matched. The SPA renders this as the generic \`blocked (household)\` (see #1640 sister-thread and the upcoming label-rename discussion).

This is purely a debugging-aid gap, not a correctness one — but it cost real triage time on the OAuth-blocked-by-YouTube incident:

- Prima's iPad sent UDP/443 to a shared Google GFE anycast IP.
- That IP had been populated into \`eb_youtubei_googleapis_com\` by Quintus's earlier YouTube DNS resolution (legitimate — YouTube is blocked for Kids).
- Prima's flow was for \`oauthaccountmanager.googleapis.com\`, which happens to resolve to the same anycast IPs.
- The kernel drop fired correctly per the rule; \`wifihaven-nflog-tail\` attributed the dst_ip to oauthmanager (Prima's most recent DNS for that IP) and posted \`reason=\"host\"\`.
- The SPA showed \"blocked (household)\" on \`oauthaccountmanager.googleapis.com\` — which looked like a household-global block on OAuth, sending #1636 to inspect \`global_blocks\` first.

Had the event read \`reason=\"host:youtubei.googleapis.com\"\` (i.e. \"this flow matched the \`eb_youtubei.googleapis.com\` rule\"), the IP-anycast root cause would have been visible at a glance.

This is **not** a fix for the false-positive itself — that's #1636's territory (anycast IP overlap; SNI-aware enforcement / shared-GFE exclusion / HTTP-only-blocks are the candidate solutions). This issue is just the labeling so the next incident is debuggable in minutes instead of hours.

## Precedent

\`bl_\` (category blocklists) already does this. \`conntrack.lua\` emits \`reason=\"category:<blocklistId>\"\` so the SPA can name the matched list. \`eb_\` should follow the same pattern with the matched host.

### Router — render the matched host into the log prefix

In \`openwrt/files/usr/lib/lua/wifihaven/render.lua\`, when emitting nftables rules for each \`eb_<host>\` set, set \`log prefix \"eb:<host>\"\` instead of the bare \`\"host\"\` it carries today. The agent already iterates the host list (\`for _, host in ipairs(r.extraBlocked) do\`), so the host is in scope.

### Router — parse the prefix in nflog-tail / conntrack ingest

Update the nflog-tail / conntrack path that converts kernel log lines to connection events: when the prefix matches \`eb:<host>\`, emit \`reason=\"host:<host>\"\` on the wire instead of bare \`\"host\"\`.

### API — extend BlockReason

In \`shared/src/Models.scala\`:

\`\`\`scala
final case class ExtraBlockedBy(host: Hostname) extends BlockReason {
  val jsonKind = \"extraBlockedBy\"; val wireKind = \"extraBlockedBy\"
}
\`\`\`

\`fromWire\`: parse \`host:<host>\` as \`ExtraBlockedBy(host)\`.

Keep the bare \`\"host\"\` → \`ExtraBlocked\` legacy alias in \`LegacyWireAliases\` so already-deployed routers that haven't picked up the prefix change still parse cleanly (per the backwards-compat rule in AGENTS.md — wire shapes are now a public contract since prod went live).

JSONB \`JsonDecoder[BlockReason]\` gets a matching arm.

### SPA — render the matched host

In \`web/src/types/blockReason.ts\`:

\`\`\`ts
case 'extraBlockedBy': return \`blocked: matched ${r.host}\`
\`\`\`

Keep \`extraBlocked\` rendering for historical events and for events from pre-rollout routers.

## TDD

- Router: \`openwrt/test/render_spec.lua\` — assert the rendered nftables ruleset includes \`log prefix \"eb:<host>\"\` for each extraBlocked entry. \`openwrt/test/conntrack_spec.lua\` (or nflog-tail spec) — assert prefix \`eb:youtubei.googleapis.com\` round-trips to \`reason=\"host:youtubei.googleapis.com\"\` on the wire.
- API: \`shared/test/src/\` — \`fromWire(\"host:foo.com\") == ExtraBlockedBy(Hostname(\"foo.com\"))\`; \`fromWire(\"host\") == ExtraBlocked\` (legacy alias intact); JSONB round-trip for both.
- SPA: \`web/src/types/\` test for the new render case.

## Wire compatibility

Pure additive per AGENTS.md \"Backwards compatibility\":
- New routers send \`host:<host>\`; older APIs that don't know the new wire form parse it via \`Unknown(\"host:<host>\")\` (no row drop, just a generic display) — acceptable bridge state.
- New API parses both \`host:<host>\` and legacy bare \`host\`.
- New SPA renders both new and legacy kinds.

No deprecation needed for \`extraBlocked\` itself — it's still the valid kind for the legacy case and for any future producer that doesn't know the matched host.

## Not in scope

The actual OAuth-blocked-by-YouTube fix (anycast IP overlap on shared GFE). That's #1636 — this issue just makes the symptom legible so #1636's solution can be evaluated against real evidence.

## Related

- #1636 — the underlying false-positive (parent issue; this is its debugging-aid sibling)
- #1640 — closed; the mistaken router-side-labeling-bug hypothesis that this attribution would have ruled out instantly
- #573 — SNI extraction; complementary (would let attribution use the actual SNI rather than most-recent-DNS-for-IP)
- \`conntrack.lua\` \`category:<id>\` precedent for the labeling pattern

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

sameerparekh:
## Summary

Closes #1645 — purely a debugging-aid label, NOT a fix for the underlying
GFE/anycast false-positive in #1636.

When an `eb_<host>` rule drops a flow, the connection event now ships
`reason = "host:<host>"` instead of bare `"host"`, so the SPA can name
*which* extraBlocked entry fired. The matched host renders as
"blocked: matched youtubei.googleapis.com" rather than the generic
"blocked (household)" that masked the YouTube/oauthmanager anycast
overlap during the #1640 triage.

Symmetric with the existing `category:<id>` precedent for `bl_` drops —
no new wire vocabulary, no new metrics label, no router policy logic.

## Wire compatibility (per AGENTS.md "Backwards compatibility")

Pure additive. API and routers deploy independently.

| Direction | Old → New | Behaviour |
|---|---|---|
| Old router → new API | `reason="host"` | Still parses as `ExtraBlocked` (legacy `LegacyWireAliases` entry preserved). |
| New router → old API | `reason="host:youtubei.googleapis.com"` | Parses as `Unknown("host:youtubei.googleapis.com")` — acceptable bridge state, no row drop. |
| Historical DB rows | `{kind:"extraBlocked"}` JSONB | Still decodes — `extraBlocked` arm preserved. |
| Historical text rows | `reason_text="host"` | Still parses via the same `LegacyWireAliases` entry. |

No deprecation window needed — `ExtraBlocked` remains valid as the
anonymous shape for any producer that doesn't know the matched host.

## TDD progression (visible in commit history)

1. `84a107bd` — **test (red):** wire round-trip + legacy alias for `ExtraBlockedBy`.
2. `1556a706` — **feat (green):** API parses `host:<host>` as `ExtraBlockedBy`.
3. `8ebf0868` — **test (red):** router emits `wh_drop:<mac>:host:<host>`.
4. `0fb74d31` — **feat (green):** render.lua / conntrack.lua / nft_drops.lua updated.
5. `ba09063a` — **feat:** SPA renders `extraBlockedBy` with the matched host (+ vitest).

## Out of scope

The actual anycast-shared-GFE false-positive (#1636) — this PR just
makes that class of incident legible in the SPA.

## Test plan

- [x] `mill shared.test`
- [x] `mill api.test`
- [x] `cd openwrt && sh test/run_tests.sh`
- [x] `cd web && npx vitest run src/types/blockReason.test.ts`
- [x] `npm run -s type-check`
- [x] scalafmt + eslint clean (pre-push hooks)
- [ ] @sameerparekh — independent /pr-review pass clears, then merge.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/api/src/routes/BlockedRoutes.scala
+++ b/api/src/routes/BlockedRoutes.scala
@@ -133,15 +133,16 @@ object BlockedRoutes {
       reason: String,
       blocklistRepo: BlocklistRepo,
   ): Task[(String, Option[String])] = BlockReason.fromWire(reason) match {
-    case MacBlockReason.Paused       => ZIO.succeed(("paused", None))
-    case MacBlockReason.Schedule     => ZIO.succeed(("schedule", None))
-    case MacBlockReason.TimeLimit    => ZIO.succeed(("time_limit", None))
-    case BlockReason.ExtraBlocked    => ZIO.succeed(("extra_blocked", None))
-    case BlockReason.AppTimeLimit(_) =>
+    case MacBlockReason.Paused         => ZIO.succeed(("paused", None))
+    case MacBlockReason.Schedule       => ZIO.succeed(("schedule", None))
+    case MacBlockReason.TimeLimit      => ZIO.succeed(("time_limit", None))
+    case BlockReason.ExtraBlocked      => ZIO.succeed(("extra_blocked", None))
+    case BlockReason.ExtraBlockedBy(_) => ZIO.succeed(("extra_blocked", None)) // #1645
+    case BlockReason.AppTimeLimit(_)   =>
       // #1518: rename `site_time_limit` → `app_time_limit`. SPA-API surface,
       // updated atomically with the SPA in the same PR.
       ZIO.succeed(("app_time_limit", None))
-    case BlockReason.Category(id)    =>
+    case BlockReason.Category(id)      =>
       blocklistRepo
         .findMeta(id)
         .map(meta => ("category", meta.map(_.name)))
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -711,7 +711,11 @@ function M.handle_flow(flow, ctx, batcher)
       end
       if eb_hit and not check_ea_carveout(eb_hit_host) then
         allowed = false
-        reason  = "host"
+        -- #1645: name the matched eb_<host> rule so triage can see which
+        -- extraBlocked entry fired (otherwise an IP-anycast overlap on a
+        -- shared CDN looks like an opaque "host" block). Symmetric with
+        -- the `category:<id>` reason for the bl_ path below.
+        reason  = "host:" .. eb_hit_host
       end
     end
   end
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -715,6 +715,14 @@ function M.handle_flow(flow, ctx, batcher)
         -- extraBlocked entry fired (otherwise an IP-anycast overlap on a
         -- shared CDN looks like an opaque "host" block). Symmetric with
         -- the `category:<id>` reason for the bl_ path below.
+        --
+        -- When `eb_hosts` for this MAC contains more than one host whose
+        -- ipset covers the same `dst_ip` (rare overlap, e.g. same anycast
+        -- IP resolved for two different extraBlocked apex domains), the
+        -- labeled `eb_hit_host` reflects whichever `pairs(eb_hosts)`
+        -- iteration order surfaced first — non-deterministic per Lua
+        -- semantics. The drop itself is unaffected (the kernel already
+        -- matched at least one eb_ set); only the debug label can vary.
         reason  = "host:" .. eb_hit_host
       end
     end
--- a/openwrt/files/usr/lib/lua/wifihaven/nft_drops.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/nft_drops.lua
@@ -50,7 +50,11 @@ local WHOLE_MAC_REASONS = {
 -- comment_reason is the `<reason>` portion of `wh_drop:<mac>:<reason>`.
 function M.classify_reason(comment_reason)
   if type(comment_reason) ~= "string" then return "whole_mac" end
-  if comment_reason == "host" then
+  -- #1645: post-rollout per-host drops carry `host:<host>` so the matched
+  -- eb_<host> rule is named in events; the bare `host` legacy form (pre-
+  -- rollout routers) still folds into the same bounded bucket. The host
+  -- name itself never becomes a label — §4 cardinality firewall.
+  if comment_reason == "host" or comment_reason:match("^host:") then
     return "host_block"
   elseif comment_reason == "ip_only" then
     return "ip_only"
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -1046,12 +1046,12 @@ function M.nft(snapshot, opts)
   for _, p in ipairs(eb_pairs) do
     ind2(string.format("ether saddr %s ip daddr @%s%s%s%s",
                        p.mac, eb_set_name(p.host), ea_suffix(p.mac, "ip"),
-                       ga_suffix("ip"), drop_suffix(p.mac, "host")))
+                       ga_suffix("ip"), drop_suffix(p.mac, "host:" .. p.host)))
   end
   for _, p in ipairs(eb_pairs) do
     ind2(string.format("ether saddr %s ip6 daddr @%s%s%s%s",
                        p.mac, eb6_set_name(p.host), ea_suffix(p.mac, "ip6"),
-                       ga_suffix("ip6"), drop_suffix(p.mac, "host")))
+                       ga_suffix("ip6"), drop_suffix(p.mac, "host:" .. p.host)))
   end
   for _, p in ipairs(bl_pairs) do
     ind2(string.format("ether saddr %s ip daddr @%s%s%s%s",
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -1043,6 +1043,15 @@ function M.nft(snapshot, opts)
   -- so behaviour for those is unchanged.
   -- #1319: ga_suffix appended after the per-MAC ea exception, so an allowed
   -- host (per-MAC or global) suppresses the per-host / per-category drop.
+  -- #1645: `host:<host>` names the matched eb_<host> rule in the kernel
+  -- log line and the nft comment. Kernel `log prefix` is bounded by
+  -- NF_LOG_PREFIXLEN=128; with `wh_drop:<mac>:host:<host> ` the fixed
+  -- envelope is 31 chars, leaving 96 chars of headroom — comfortably above
+  -- any realistic hostname (RFC 1035 caps a single label at 63 chars and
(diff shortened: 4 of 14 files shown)


sameerparekh:
# Independent PR review — #1656

Reviewed against [`docs/pr-review-checklist.md`](docs/pr-review-checklist.md) 8 dimensions. First run; full merge-base diff `git diff origin/main...HEAD`.

## BLOCKERS

(none)

## SHOULD-FIX

(none)

## NITS

1. [shared/src/Models.scala:1763](shared/src/Models.scala:1763) — `ExtraBlockedBy.wireKind = "host"` overlaps the bare-`host` key that `LegacyWireAliases` already maps to `ExtraBlocked`. It's harmless today because `byWireKind` is built only from `NullaryCases` (and `ExtraBlockedBy` is parameterised, so it isn't added) and `LegacyWireAliases` is consulted first. But the `wireKind` field on a parameterised case is otherwise the *prefix* portion (`Category` → `"category"`, `AppTimeLimit` → `"app_time_limit"`, `AppBlocked` → `"app"`) — so `"host"` here matches that convention. Worth a one-line comment that the trailing-`:` is implied (or set the value as `"host:"` and adjust callers if any). Non-blocking.

2. [openwrt/files/usr/lib/lua/wifihaven/render.lua:1049](openwrt/files/usr/lib/lua/wifihaven/render.lua:1049) — kernel nft `log prefix` is bounded by `NF_LOG_PREFIXLEN=128`. With the new `wh_drop:<mac>:host:<host>` shape, a hostname approaching the 128-char wire ceiling would silently truncate the prefix the nflog parser sees and produce a partial-host reason. Realistic eb_ entries are well under the limit (the YouTube/oauth case in the issue is 22 chars), so this is theoretical — but worth a defensive truncation, or at least a comment, if a follow-up surfaces a long-CNAME case. The nft `comment` field has its own 128-byte limit too; same caveat.

3. [openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:712-717](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:712) — when the eb_ slow-path matches via `for host in pairs(eb_hosts)`, Lua's `pairs` order is undefined, so a flow whose `dst_ip` is in two overlapping `eb_<host>` ipsets gets a non-deterministic labeled host. This was already true pre-PR (the iteration picks `eb_hit_host` for the ea-carveout check); the new code just promotes that value into the reason. Acceptable for a debug aid, but a comment noting the indeterminism would be honest.

## Dimension pass

| # | Dimension | Status |
|---|---|---|
| 1 | Duplicated logic / single source of truth | OK — reuses `BlockReason.fromWire`/`asWire` and `ExtraBlockedBy` propagates through the canonical kind-tagged JSON codec. No new "must mirror" coupling. |
| 2 | Testing integrity | OK — TDD red→green visible: `84a107bd` (red) → `1556a706` (green); `8ebf0868` (red) → `0fb74d31` (green). Existing tests were *tightened* (more specific reason), not weakened. New tests cover wire round-trip, legacy aliases, JSON encode/decode, exhaustiveness, and the four call sites (`render`, `conntrack`, `nflog`, `nft_drops`). |
| 3 | Architectural invariants | OK — no DNS-as-enforcement claim; router stays a dumb applier (no new decision logic, the matched host is plumbed at render time from the same `extraBlocked` list); snapshot wire vocabulary unchanged. `extraAllowed` carve-out (`!= @ea_<m>_<a>` / `ga_suffix`) preserved on the modified eb_ drops; `Clock`/ZIO conventions intact. |
| 4 | Backwards compatibility / wire contract | OK — pure additive. `LegacyWireAliases["host"] -> ExtraBlocked` preserved for pre-rollout routers; legacy `{kind:"extraBlocked"}` JSONB still decodes (covered by the new "legacy {kind:extraBlocked} JSON still decodes" test). New router → old API degrades to `Unknown("host:<host>")` (no row drop). |
| 5 | Migrations | N/A — no SQL migration in this PR. |
| 6 | Metrics + dashboard | OK — `nft_drops.classify_reason` collapses `host:<host>` into the existing bounded `host_block` bucket; no new label, no new series, host name is forbidden as a label per §4 cardinality firewall. No dashboard work required because no series was added. |
| 7 | Scope / hygiene | OK — focused diff (14 files, all on the labeling path). No dead code. |
| 8 | Security | OK — no new routes, no new SQL, no secrets. The `Hostname.parse` guard on `fromWire("host:<bad>")` and on the `{kind:"extraBlockedBy"}` JSON decoder arm prevents an attacker-controlled wire string from coercing an unvalidated `Hostname` value into the domain. |

## VERDICT: APPROVE

Clean additive change. The new `ExtraBlockedBy(Hostname)` case adds a debugging label to per-host extraBlocked drops symmetric with the existing `category:<id>` precedent, with the wire round-trip, JSON codec, router emission, nflog metrics fold, and SPA renderer all aligned and back-compat tested. Three NITs above are non-blocking — comment-level polish on the wireKind convention, kernel prefix-length headroom, and pairs-order labeling indeterminism.

sameerparekh:
# Independent PR review — #1656 (re-run)

Re-run after `aff56140` was pushed addressing the prior NITs. Delta scope: `git diff ba09063a...HEAD` (3 files, comment-only).

## Prior-findings status

| # | File:line | Prior finding | Status |
|---|---|---|---|
| NIT 1 | shared/src/Models.scala:1758 | Document `wireKind = "host"` overlap with `LegacyWireAliases` | **ADDRESSED** — `aff56140` adds an 8-line comment explaining the prefix-form sentinel convention and why the overlap is harmless (byWireKind built only from NullaryCases). |
| NIT 2 | render.lua:1049 | Note kernel `NF_LOG_PREFIXLEN=128` headroom | **ADDRESSED** — `aff56140` adds the 9-line headroom calculation (31-char envelope, 96 chars free, RFC 1035 label cap) and graceful-degradation note. |
| NIT 3 | conntrack.lua:712 | Note `pairs(eb_hosts)` iteration-order indeterminism on overlapping ipsets | **ADDRESSED** — `aff56140` adds an 8-line comment explaining the rare overlap case and that only the debug label varies, not the drop itself. |

## New findings (delta `ba09063a...HEAD`)

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

(none — delta is documentation-only)

## Re-verification

- `mill shared.test` — green
- `cd openwrt && sh test/run_tests.sh` — 621 successes / 0 failures
- scalafmt + pre-push hooks clean

## VERDICT: APPROVE

All three prior NITs ADDRESSED via comment-only clarifications; no new findings introduced by the fix-up push. Ready to merge.

Issue #1649, opened 2026-06-11
bug(prod): per-host blocked traffic (Google Play extraBlocked) still accrues engaged minutes — #1628 fix only covers whole-MAC blocks


sameerparekh:
## Symptom (prod, observed 2026-06-11)

Google Play is in a kid profile's per-host block list (extraBlocked or similar). The kid's profile itself is **not** blocked (`blocked=false`; daily limit not hit, no schedule active). Yet Google Play shows **~7m of engaged-minutes / usage** for the day.

## Why #1628's fix doesn't cover this case

#1628 (PR #1633) shipped `usage.build_report` taking a `blocked_macs` table and dropping any counter whose mac is in it. That correctly handles **whole-MAC blocks** (Paused/Schedule/TimeLimit/Manual). It does NOT handle:

- **Per-host extraBlocked** (the operator's case here — Google Play): MAC is not in `blocked_macs`; the drop is keyed on `(mac, dst_ip ∈ eb_<mac>_<host>)`. Profile rule has `blocked=false`.
- **Per-host category blocklist** (`bl_<mac>_<blocklistId>`): same structure as above.
- **blockIpOnly** (drop on `(mac, dst_ip ∉ resolved_<mac>)`): same.

For all three, the drop happens at `priority filter` (0) and terminates before the accounting chains at priority+1 — per PR #1633's own router-side investigation — so packets that hit the drop **should not** produce traffic_reports rows. If usage IS accumulating, one of:

1. **Historical traffic**: the 7m was logged before the block was added; the displayed counter is yesterday's truth, not today's. Verify by inspecting `traffic_reports` timestamps for the MAC+host today.
2. **DNS attribution gap**: `eb_<mac>_googleplay_*` ipset wasn't populated when the kid's device hit Google Play, so the drop didn't fire and traffic actually flowed. dnsmasq's `--ipset=` callback might be missing the answer (DoH/iCloud Private Relay, hard-coded IP, dnsmasq config drift, etc.). Verify via router logs around the symptom window.
3. **Counter via a different path**: maybe `mac_ip_tracking` isn't the only source of usage data — `connection_events` IS written from drops (per `nflog.lua`), and if presence stitches from connection_events as well as traffic_reports, the drop *itself* produces an engaged-minutes signal.

## Investigation steps (observe before theorize)

1. **Router (router.lan, 192.168.10.1)**:
[...]

PR #1659, opened 2026-06-11
fix(#1658): periodic eb_/bl_ ipset re-resolve heartbeat closes DNS-cache leak

### LINKED ISSUE #1658: bug(prod): eb_<host> nft set 1h timeout + iOS DNS cache → extraBlocked hosts reachable via cached IPs
## Symptom

A host in a kid profile's `extraBlocked` is intermittently reachable from a kid device. Confirmed today (2026-06-11) for `play.google.com` on Kid Mac (`ca:ef:a1:72:6a:a3`) and Prima iPad (`04:72:ef:d6:e4:5a`): 19 `traffic_reports` rows today, all `host_type='fqdn'`, `dest_ip ∈ {142.251.46.142, 142.251.35.142, 142.250.188.46}` — the exact IPs currently in `eb_play_google_com`. Investigation in #1649 pinned this as hypothesis B (enforcement gap), and that issue was closed not-planned because the right framing is "fix the leak, don't suppress the count".

## Mechanism

`eb_<host>` (and `eb6_<host>`) is declared `flags dynamic,timeout` `timeout 1h` (`openwrt/files/usr/lib/lua/wifihaven/render.lua:825-840`). Population is one-shot at DNS resolve time via dnsmasq's `--nftset=/<host>/eb_<host>` callback (`render.dnsmasq`). The per-MAC drop rule's predicate is `ip daddr @eb_<host> ip daddr != @ea_<mac>_<carveouts>`; if the client connects to a cached `<host>` IP outside the 1h window the IP is no longer in `eb_<host>`, the predicate misses, the drop falls through, and the priority-1 accounting chains book real bytes. The agent's `lookup_hostname` (long-lived dnsmasq query-log cache) then correctly attributes the bytes back to the FQDN at usage-report time — which is exactly how the leak shows up as engaged minutes on the SPA against an explicitly-blocked host.

iOS's DNS cache regularly outlives 1h: Safari holds ~1h, and `URLSession` / `getaddrinfo` reuse cached results for the lifetime of long-running connection pools (background-fetch sockets, push channels). The kid's iPad doesn't re-query `play.google.com` every connection, so the `--nftset=` callback doesn't refresh the entry, and the kernel set ages out under the client.

The same applies to `bl_<blocklistId>` (category blocklists), which also `timeout 1h`. And it interacts with `blockIpOnly` — long-running flows on a previously-resolved IP outlive the `resolved_<mac>` entry's window in the same way.

## Why a longer dynamic timeout isn't a clean fix

- Raising to 24h+ accumulates stale CDN IPs as Google rotates, leaking the inverse direction (an IP that *was* Google but is now someone else's webapp gets dropped).
- Without timeout the dynamic set grows unbounded — every IP ever resolved sticks around. Bad on a `/tmp`-backed kernel set with limited memory.

So the answer isn't "make timeout bigger"; it's "keep the set refreshed in step with client cache lifetime" — same shape the dns-tail follower (#259) already used for similar drift on `ea_`.

## Proposed direction (sketch — needs design)

1. **Periodic re-resolve.** Have the agent (or a sidecar) periodically `getaddrinfo` each `extraBlocked` / `blocklistIds` apex per MAC, and `nft add element` the answers into `eb_<host>`. Effectively a heartbeat that keeps the set hot while any client still has the IP cached. Reuse the existing dns-tail / blocklists periodic cadence.
2. **Subdomain-aware refresh.** dnsmasq's `--nftset=/<apex>/` matches subdomains at resolve time, but a periodic re-resolve from the agent only knows the apex; need to track the resolved subdomains observed via dnsmasq query log and refresh those too.
3. **Optional: bound the lookahead.** Don't refresh forever — only while a client appears to have an active flow to that IP (e.g. seen in conntrack within the last N min). Avoids permanent set growth.

Open questions:
- Is the dnsmasq query-log cache (`lookup_hostname`) the right inventory of "hostnames this device has resolved that might be cached client-side"? It's already long-lived, per-MAC, and the right granularity.
- Does this need a wire change to surface per-host cache-lifetime hints? Likely no — pure router-side.
- Is `nft add element` from userspace fast enough on a Belkin RT3200 fleet? Should be — the existing dns-tail does similar.

## Evidence (read-only)

Full investigation in #1649's closing comment — router state, prod DB queries (`statement_timeout='5s'`), code citations, and per-hypothesis classification.

## Related
- #1649 — closed not-planned; this is the enforcement-side replacement.
- #1628 / #1633 — the whole-MAC sibling problem, closed not-planned for the same "don't suppress, fix at source" reason.
- #259 — dns-tail follower, already keeps related sets warm.
- Blocklists & Enforcement epic.

sameerparekh:
Closes #1658.

## Problem (#1649 evidence)

The per-host `eb_<host>` / `eb6_<host>` and per-blocklist `bl_<id>` / `bl6_<id>` nftables sets are declared `flags dynamic,timeout 1h`, populated one-shot at DNS resolve time via dnsmasq's `--nftset=` callback. iOS / iPadOS app DNS caches regularly outlive 1h (long-running connection pools, background-fetch sockets reuse the cached `getaddrinfo` answer), so a cached IP can age out of the kernel set under the still-connecting client. The per-MAC drop predicate `ip daddr @eb_<host>` misses, traffic falls through, and the agent's `lookup_hostname` correctly re-attributes the bytes back to the FQDN at usage-report time — surfacing as real engagement against an explicitly-blocked host.

#1649 confirmed this in prod for `play.google.com` on Kid Mac (`ca:ef:a1:72:6a:a3`) / Prima iPad (`04:72:ef:d6:e4:5a`) — 19 `traffic_reports` rows today hitting exactly `{142.251.46.142, 142.251.35.142, 142.250.188.46}`, the IPs currently in `eb_play_google_com`.

## Fix — periodic re-resolve heartbeat

New module `openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua`, wired into the agent timer loop as a sibling of the blocklist-refresh timer:

- Every `eb_refresh_interval` seconds (default `1800` — strictly below the 1h kernel timeout so each entry sees ≥1 refresh per ageing window), walk the per-MAC inventory `render.update_shared` already maintains (`eb_hosts_by_mac` / `bl_hosts_by_mac`), collapse to distinct hosts (the sets are global, not per-MAC), re-resolve each via the local dnsmasq (`dig @127.0.0.1` A + AAAA), and `nft add element` each answered IP back into the destination set.
- nft tolerates duplicate adds; idempotence is the kernel's job.
- The dns-tail sidecar (#1346) keeps doing its job; this closes the leak that opens precisely when no client re-resolves.

### Cadence justification

eb_/bl_ timeout is 3600s. 1800s gives ~2 refreshes per ageing window with comfortable margin. ~50 distinct extraBlocked hosts in a household → ~100 `dig` calls / 30 min → ~3.3 / min sustained — negligible router load.

## Direction tradeoffs (also covered in the issue body)

- **A. Periodic re-resolve (this PR).** Closes the leak without growing the set unbounded.
- **B. dns-tail extension.** The leak is precisely that the *client* doesn't re-resolve — extending dns-tail only helps when DNS *is* being queried. Doesn't address the cache-outlives-timeout case.
- **C. Bump timeout / make set static.** Accumulates stale CDN IPs as Google rotates (inverse leak); unbounded growth on tmpfs.

## Subdomain coverage

This module re-resolves the **apex** as authored. dnsmasq's `--nftset=/<apex>/...` callback populates the set for subdomains at resolve time, and dns_tail_sets (#1346) covers the directly-queried CNAME-target path. Subdomain-only client caches are left to a follow-up if observed in prod; the #1649 evidence is apex-level `play.google.com`.

## Wire compatibility

Agent-only change. No API/router request/response shape touched. The added metric name in `MetricGuard.Allowed` is purely additive — pre-fix agents won't push it, post-fix agents push it through.

## bounded-tmp-writes (AGENTS.md §bounded-tmp-writes)

N/A — no `/tmp` writes. The module shells out to `dig` (stdout consumed inline) and calls `nft add element` (no spool).

## Single source of truth (AGENTS.md §single-source-of-truth)

Reuses the (host, sets-to-populate) inventory `render.update_shared` already builds; reuses `dns_tail_sets.safe_addr` / `nft_add_element` for IP sanitisation + nft command formatting. The local `bl_sanitize` mirrors `render.lua`'s function byte-for-byte and is the one drift risk — `eb_refresh.lua`'s module header calls this out; if either function ever changes both must move in lockstep.

## Instrumentation (AGENTS.md §instrument-new-functionality / §metrics-need-a-dashboard)

- Counter `eb_refresh_total{result}`, `result` ∈ `{ok, resolve_failed}` — bounded enum, routed through `MetricGuard.Allowed`.
- Two new panels added to `deploy/grafana/dashboards/enforcement.json`: refresh rate by result, and resolve_failed fraction (canary).

## TDD

- Commit 1 (RED): `test(#1658)` lands the failing busted spec for the new module.
- Commit 2 (GREEN): implementation + agent timer wire-up + metric allowlist + dashboard panels.

## Manual repro (deferred to operator — touching router.lan affects prod enforcement)

1. `ssh root@router.lan`
2. `nft delete element inet wifihaven eb_play_google_com { 142.251.46.142 }`
3. Wait ≤ 30 min (`eb_refresh_int`).
4. `nft list set inet wifihaven eb_play_google_com` — IP returns.
5. `logread -e 'eb_refresh timer'` confirms a fire with `hosts>0`.

## Test plan

- [x] `openwrt/test/eb_refresh_spec.lua` — 14 unit tests (inventory collapse, eb_ + bl_ refresh, resolve-fail counting, shell-metachar rejection, idempotence, empty-inventory no-op, dig-output parsing).
- [x] Full `openwrt/test/run_tests.sh` suite passes (635/0/1 pending — pre-existing `nft` not on host).
- [x] `mill api.compile` + `mill api.test.compile` clean.
- [x] `scalafmt --check --non-interactive` clean.
- [x] `python3 -m json.tool` on `deploy/grafana/dashboards/enforcement.json` clean.
- [ ] Operator-driven live repro on `router.lan` per steps above.

### DIFF
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -112,6 +112,17 @@ object MetricGuard {
     "dns_queries_total"                         -> Set("result", "router_id", "installation_id"),
     "blocklist_fetch_failures_total"            -> Set("status", "router_id", "installation_id"),
     "enforcement_drops_total"                   -> Set("reason", "router_id", "installation_id"),
+    // #1658 — eb_/bl_ ipset re-resolve heartbeat. Each fire of the agent's
+    // eb_refresh timer re-resolves the inventory of (extraBlocked, blocklist)
+    // hosts against the local dnsmasq and adds answered IPs back into the
+    // per-host eb_<host> / per-blocklist bl_<id> nftables sets ahead of their
+    // 1h `flags dynamic,timeout` aging out — closing the iOS-DNS-cache leak
+    // confirmed in prod for play.google.com (#1649). `result` is a small fixed
+    // enum: `ok` (resolver answered, IPs were added back to the set; counts
+    // resolved hosts not adds, so the rate is meaningful regardless of CDN
+    // fan-out) and `resolve_failed` (dig couldn't reach the local resolver, or
+    // the host failed the hermetic allow-list).
+    "eb_refresh_total"                          -> Set("result", "router_id", "installation_id"),
     // #1033 — usage-POST retry-queue health. Depth = buckets currently waiting for a backoff to
     // elapse; `usage_post_total{result}` tracks the immediate-post outcome (`ok` | `queued`) and
     // drain outcome (`drained` | `drain_failed`). Gives operators a first-class view of "are
--- a/deploy/grafana/dashboards/enforcement.json
+++ b/deploy/grafana/dashboards/enforcement.json
@@ -278,6 +278,65 @@
           "refId": "A"
         }
       ]
+    },
+    {
+      "datasource": { "type": "prometheus", "uid": "${datasource}" },
+      "title": "eb_/bl_ ipset refresh rate (#1658)",
+      "description": "sum by (result) (rate(eb_refresh_total[5m])) — hosts/sec the periodic re-resolve heartbeat processed. The eb_<host>/bl_<id> sets carry `flags dynamic,timeout 1h`; this timer (default cadence 1800s, strictly below the 1h timeout) re-resolves each extraBlocked / blocklist apex and adds answers back so the set stays populated under long-lived iOS app DNS caches. Steady non-zero `ok` is the healthy signal; a flat-zero in a household that has any extraBlocked / blocklist hosts means the timer isn't running. Sustained `resolve_failed` points at a dnsmasq problem.",
+      "type": "timeseries",
+      "gridPos": { "h": 8, "w": 12, "x": 0, "y": 33 },
+      "id": 9,
+      "fieldConfig": {
+        "defaults": {
+          "color": { "mode": "palette-classic" },
+          "unit": "ops",
+          "min": 0,
+          "custom": { "drawStyle": "line", "fillOpacity": 15, "showPoints": "never", "lineWidth": 1 }
+        },
+        "overrides": []
+      },
+      "options": {
+        "legend": { "displayMode": "table", "placement": "bottom", "calcs": ["lastNotNull", "max"] },
+        "tooltip": { "mode": "multi", "sort": "desc" }
+      },
+      "targets": [
+        {
+          "datasource": { "type": "prometheus", "uid": "${datasource}" },
+          "expr": "sum by (result) (rate(eb_refresh_total{env=\"$env\", router_id=~\"$router\"}[5m]))",
+          "legendFormat": "{{result}}",
+          "refId": "A"
+        }
+      ]
+    },
+    {
+      "datasource": { "type": "prometheus", "uid": "${datasource}" },
+      "title": "eb_/bl_ ipset refresh failure ratio (#1658)",
+      "description": "sum(rate(eb_refresh_total{result=\"resolve_failed\"})) / sum(rate(eb_refresh_total)) over 5m — fraction of re-resolves that failed (resolver unreachable / hermetic-allowlist reject). A spike above ~0.1 means the periodic heartbeat is no longer keeping the sets warm — the iOS-DNS-cache leak (#1649) returns when failures dominate.",
+      "type": "timeseries",
+      "gridPos": { "h": 8, "w": 12, "x": 12, "y": 33 },
+      "id": 10,
+      "fieldConfig": {
+        "defaults": {
+          "color": { "mode": "palette-classic" },
+          "unit": "percentunit",
+          "min": 0,
+          "max": 1,
+          "custom": { "drawStyle": "line", "fillOpacity": 15, "showPoints": "never", "lineWidth": 1 }
+        },
+        "overrides": []
+      },
+      "options": {
+        "legend": { "displayMode": "list", "placement": "bottom", "calcs": ["lastNotNull", "max"] },
+        "tooltip": { "mode": "multi", "sort": "desc" }
+      },
+      "targets": [
+        {
+          "datasource": { "type": "prometheus", "uid": "${datasource}" },
+          "expr": "sum(rate(eb_refresh_total{env=\"$env\", router_id=~\"$router\", result=\"resolve_failed\"}[5m])) / clamp_min(sum(rate(eb_refresh_total{env=\"$env\", router_id=~\"$router\"}[5m])), 1e-9)",
+          "legendFormat": "resolve_failed fraction",
+          "refId": "A"
+        }
+      ]
     }
   ],
   "refresh": "30s",
--- a/openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua
@@ -1,0 +1,228 @@
+-- eb_refresh.lua — periodic re-resolve of extraBlocked and category-blocklist
+-- hosts so the per-host `eb_<sanhost>` / `eb6_<sanhost>` and per-blocklist
+-- `bl_<sanid>` / `bl6_<sanid>` nftables sets stay populated ahead of their 1h
+-- `flags dynamic,timeout` aging out (declared in render.lua).
+--
+-- #1658 — the leak this closes. Population today is one-shot at DNS resolve
+-- time (dnsmasq's `--nftset=` callback + the dns-tail sidecar's #1346
+-- alias-chain populator). The kernel set ages an IP out after 1h, but iOS /
+-- iPadOS app DNS caches regularly outlive 1h — long-running connection pools
+-- and background-fetch sockets reuse the cached `getaddrinfo` answer for the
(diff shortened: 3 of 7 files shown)


sameerparekh:

Reviewed `git diff origin/main...HEAD` (three-dot, merge-base) at `eca8b2b8`.
6 files / +598 lines: new `eb_refresh.lua` module, agent timer wire-up, metric
allowlist entry, two Grafana panels, and a busted spec.

### BLOCKERS

(none)

### SHOULD-FIX

1. **`openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua:62-65` vs `openwrt/files/usr/lib/lua/wifihaven/render.lua:174-176` — `bl_sanitize` is duplicated byte-for-byte across files, with only a "Keep this in lock-step with render.lua" comment to prevent drift.** This is exactly the "keep in sync by hand" smell the [single-source-of-truth rule](../AGENTS.md#single-source-of-truth) flags — and the eb_refresh module header explicitly admits the lockstep requirement is "the one drift risk." Collapse: `render.lua` already exports `sanitize` via `M.sanitize = sanitize` (used here); do the same for `bl_sanitize` and call it through `render`. If render.lua's regex ever changes, the periodic refresh would silently start writing to a set name that doesn't exist (no-op, leak returns).

2. **`openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua:206-216` — set-name construction (`"eb_"..host`, `"eb6_"..host`, `"bl_"..id`, `"bl6_"..id`) is hand-built in `refresh`, while `render.lua:165-182` has dedicated `eb_set_name` / `eb6_set_name` / `bl_set_name` / `bl6_set_name` helpers (currently locals).** Same SSOT issue as #1, same fix: export the helpers from render.lua and call them. Addresses #1 and #2 in one move and removes the "lockstep" comment entirely.

### NITS

1. `openwrt/files/usr/sbin/wifihaven-agent:785-787` — `metrics.inc_counter(..., {result="ok"}, stats.resolves_ok)` is called unconditionally; when `stats.resolves_ok == 0` (every host failed) it increments by 0. Harmless but asymmetric with the `resolve_failed` branch right below it which is guarded. Guard symmetrically.
2. `openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua:170-180` — `default_resolver` forks `dig` twice per host (one A, one AAAA). For a ~50-host household that's 100 popens per cycle. `dig <host> A AAAA +short` halves it. Not urgent at default cadence.
3. `openwrt/test/eb_refresh_spec.lua` — `resolves_err` is asserted at the module level, but the agent timer wire-up that translates `(resolves_ok, resolves_err)` → two `metrics.inc_counter` calls is uncovered. Optional integration coverage.

### VERDICT: APPROVE

Surgical agent-only fix for the #1649 leak. No wire change, no router-side
decision logic added, hermetic shell-outs (hostname allow-list + safe_addr
filter), metric guarded + dashboarded. Two SHOULD-FIX items collapse a real SSOT
drift risk against `render.lua`; no BLOCKER.

sameerparekh:

Re-reviewed at `b013fb79` after the `refactor(#1658)` push. Delta:
`git diff eca8b2b8...HEAD` — render.lua exports, eb_refresh.lua delegates,
agent guards `result=ok` symmetrically.

### Prior-findings status

| # | Class | Prior finding | Status |
|---|---|---|---|
| 1 | SHOULD-FIX | `bl_sanitize` duplicated render.lua ↔ eb_refresh.lua | **ADDRESSED** — `render.lua:179` now `M.bl_sanitize = bl_sanitize`; the local copy and the lockstep comment are gone from eb_refresh.lua. |
| 2 | SHOULD-FIX | Set-name construction duplicated | **ADDRESSED** — `render.lua:171-172,185-186` export `eb_set_name`/`eb6_set_name`/`bl_set_name`/`bl6_set_name`; eb_refresh.lua:198-202 now delegates. |
| 3 | NIT | `metrics.inc_counter(result=ok, 0)` fires when all hosts fail | **ADDRESSED** — `wifihaven-agent:786-789` now guards symmetrically with the `resolve_failed` branch. |
| 4 | NIT | `default_resolver` forks dig twice (A + AAAA) | NOT-ADDRESSED — optional, left for follow-up if cycle cost matters. |
| 5 | NIT | Agent-timer metric-emission path uncovered | NOT-ADDRESSED — optional. |

### New findings on the delta

(none)

### VERDICT: APPROVE

Both SHOULD-FIX items resolved by a clean delegation: `render.lua` exports the
helpers (matching its existing `M.sanitize` shape), `eb_refresh.lua` calls them
through, and the SSOT drift risk is gone. The symmetric metric guard tightens
the loop. No BLOCKERS open at any point in this review.

Issue #1560, opened 2026-06-07
presence: centralize the host suppression/attribution predicate — one matcher, no per-call-site copies (follow-up to #1559)


sameerparekh:
Follow-up to #1559 (which is itself a follow-up to #1506 / PR #1557).

While scoping #1559 it became clear the dashboard isn't just *missing* app-awareness — it carries a **hand-rolled copy of the suppression predicate**, and the same host→FQDN→pattern-match plumbing is duplicated across several call sites. The #1532 de-dup audit consolidated the suppression **data** (one `InfraHosts` list, one `Presence.isHeartbeat`), but the **predicate plumbing** was never centralized, so divergence is still possible — exactly the trap #1559 must avoid re-entering when it adds an app-aware branch.

## Confirmed duplication

1. **Background-host check re-implemented in the route.** `Presence.isBackgroundHost` is `private`, so `DashboardNowRoutes.dropBackground` re-rolls it by hand:
   ```scala
   // api/src/routes/DashboardNowRoutes.scala:195
   rows.filterNot(r => r.host.asFqdn.exists(fqdn => InfraHosts.isBackground(fqdn.value)))
   ```
   ```scala
   // api/src/presence/Presence.scala  (private — can't be reused)
   private def isBackgroundHost(row: PresenceRow): Boolean =
     row.host.asFqdn.exists(fqdn => InfraHosts.isBackground(fqdn.value))
   ```
   Two copies of the same suppression rule that can (and did, pre-#1506) drift.

2. **"Does this host match any of these patterns" written ~5×.** The same `asFqdn.exists(fqdn => pats.exists(p => matchesPattern(fqdn.value, p)))` shape appears inline at `Presence.scala` lines ~176 (`isExempt`), ~370 (`isAppAttributed`), ~430 / ~476 (per-pattern matching), ~583 (`matchesGroup`). One primitive, copy-pasted.

## Ask

Extract **one** host-matching / suppression / attribution primitive and have every call site (Presence surfaces **and** the dashboard) consume it — so #1559's app-aware branch is added in exactly one place rather than as a third copy.

Concretely:
- A shared `HostId`-level matcher, e.g. `HostMatch.matchesAny(host: HostId, patterns: List[String]): Boolean` (or on `HostId` directly), replacing the five inline `asFqdn.exists(... matchesPattern ...)` copies. This also gives `DashboardNowRoutes` (which works over `TrafficRollupRow`, not `PresenceRow`) something to call without re-rolling.
[...]

PR #1670, opened 2026-06-12
fix(#1559): route dashboard now-widget host ranking through canonical app-aware suppression

### LINKED ISSUE #1559: presence: make the dashboard now-widget host ranking (DashboardNowRoutes.dropBackground) app-aware
Follow-up to #1506 / PR #1557.

#1506 made background/infra suppression **app-aware** for the presence *counting* surfaces: `Presence.isHeartbeat` now takes `appHostPatterns` (the union of the profile/MAC's active-app host-sets) and treats app attribution as winning over both the `InfraHosts` identity match and the byte floor. The daily-total paths (`TimeStatusService`) and the per-app surface (`appSpansForProfile`) route through that single predicate.

**Not yet app-aware: the dashboard now-widget host ranking.** `DashboardNowRoutes.dropBackground` (`api/src/routes/DashboardNowRoutes.scala`) still suppresses purely by host identity via `InfraHosts.isBackground`:

```scala
private def dropBackground(rows: List[TrafficRollupRow]): List[TrafficRollupRow] =
  rows.filterNot(r => r.host.asFqdn.exists(fqdn => InfraHosts.isBackground(fqdn.value)))
```

This was left out of #1506 deliberately because:
- it is a host-**ranking** path (`topHosts` / `nowActivity`), not a counting surface, so it does not feed the #1499 over-count; and
- `buildResponse` has **no per-profile app host-set context loaded** — wiring it needs the app host-sets threaded in (a new repo read + plumbing through the route), which was outside #1506's stated diff scope.

The consequence today: an app's off-domain asset/CDN host that also matches a background pattern is dropped from the now-widget ranking even while it correctly counts toward the app's time. So "watching X right now" can omit (or mis-rank) an app whose foreground traffic is currently on such a host — the same seam #1506 closed for counting, still open for the live widget.

## Ask
Make `DashboardNowRoutes.dropBackground` app-aware so attribution beats suppression on the ranking path too, consistent with the single predicate from #1506:
- Load the app host-sets (per profile / per device's profile) into `buildResponse` — the `app_hosts` join already used elsewhere (`UsageRoutes`, `RollupRepo`).
- Route ranking suppression through the same app-aware predicate (`Presence.isAppAttributed` / the `appHostPatterns` form of the check) rather than a second identity-only copy — do **not** re-derive suppression logic here (the #1532 divergence lesson).
- A background-pattern host that belongs to the device's profile's active app host-set should be **kept** in the ranking and attributed to that app; device infra with no app behind it stays dropped.

## Notes
- `DashboardNowRoutes` works over `TrafficRollupRow`, not `PresenceRow`, so the predicate may need a small `HostId`-level overload (or extract the FQDN-matching core of `isAppAttributed` so both row types can call it).
- Tests: extend `DashboardNowApiSpec` — an app's background asset host ranks/attributes when the profile's app host-set claims it; an unrelated infra host stays dropped.
- No wire change expected; no migration.

sameerparekh:
Closes #1559.

## Summary

`DashboardNowRoutes.dropBackground` was the last consumer that hadn't been
routed through #1506's app-aware predicate. It suppressed hosts purely on
`InfraHosts.isBackground` host identity, so an off-domain asset / CDN host an
app genuinely depends on — e.g. `beacons3.gvt2.com` when a Google app's
host-set claims `gvt2.com` — was silently dropped from `topHosts` /
`nowActivity` even while it correctly counted toward the app's time. The seam
#1506 closed for counting was still open for the live widget.

Collapse through a single host-keyed predicate:

```scala
// api/src/presence/Presence.scala
def suppressedAsBackground(host: HostId, appHostPatterns: List[String]): Boolean =
  InfraHosts.isBackground(host) && !HostMatch.matchesAny(host, appHostPatterns)
```

`Presence.isHeartbeat`'s suppression branch is rewritten to call it too, so the
ranking path (`DashboardNowRoutes.dropBackground`) and every counting path
share one rule. Same #1506 contract, no second copy (#1532 / #1560).

The dashboard loads the per-profile active-app host-set through the canonical
`ProfileAppDispositions.from(appLimits).appHostPatterns` fold — the same data
`TimeStatusService.appHostPatterns` feeds counting surfaces — so the dashboard
cannot disagree with them on "what's an active-app host".

## Before / after (now-widget for a Kids profile watching `beacons3.gvt2.com`, with a Google app whose host-set includes `gvt2.com`)

Before: `topHosts = []`, `nowActivity = None`. The dominant host is dropped as
"device infra" even though it's foreground traffic to a managed app — the
operator sees "no recent activity" while the kid is in fact watching their app.

After: `topHosts = [{host: beacons3.gvt2.com, activeSeconds: 300}]`,
`nowActivity = {topHost: beacons3.gvt2.com}`. Attribution wins; the operator
sees the actual host being used. Device infra with no app behind it
(`ocsp.digicert.com`) still drops, matching prior behavior.

## Lesson

Single source of truth (AGENTS.md §single-source-of-truth): #1506 centralized
the app-aware predicate for counting; #1560 collapsed the host-keyed plumbing.
Each follow-up that touches a new surface must route through those primitives
rather than re-derive the rule — otherwise the ranking and counting paths drift
again, the same class of bug the #1532 audit is closing. This PR routes the
last identified consumer (`DashboardNowRoutes.dropBackground`) through the
canonical predicate.

## Test plan
- [x] `mill api.test` (full api suite, 15-spec `DashboardNowApiSpec` includes the new pinning test)
- [x] New test: `#1559 a background-pattern host attributed to an active app is kept in the now-widget ranking`

### DIFF
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -369,6 +369,7 @@ object Main extends ZIOAppDefault {
             deviceRepo,
             profileRepo,
             upRepo,
+            atlRepo,
             clock,
           ) ++
           BlocklistRoutes.routes(auth, blRepo, blCache, blFetcher2, bundledBlocklists) ++
--- a/api/src/presence/Presence.scala
+++ b/api/src/presence/Presence.scala
@@ -354,8 +354,26 @@ object Presence {
       filter: HeartbeatFilter,
       appHostPatterns: List[String] = Nil,
   ): Boolean =
-    !isAppAttributed(row, appHostPatterns) &&
-      (isBackgroundHost(row) || (filter.enabled && row.bytes < filter.bytesThreshold))
+    suppressedAsBackground(row.host, appHostPatterns) ||
+      (!isAppAttributed(
+        row,
+        appHostPatterns,
+      ) && filter.enabled && row.bytes < filter.bytesThreshold)
+
+  /**
+   * #1559: host-keyed "drop unless app-attributed" predicate — the ranking-side analogue of
+   * [[isHeartbeat]]'s suppression branch, without the PresenceRow byte-floor. A host is suppressed
+   * as device-level background iff it is on the [[InfraHosts.isBackground]] list AND no active
+   * app's host-set claims it (attribution beats suppression, same #1506 contract every counting
+   * surface routes through). The SINGLE host-keyed entry point: `isHeartbeat` reuses this for its
+   * suppression branch and `DashboardNowRoutes.dropBackground` calls it directly, so the rule
+   * cannot diverge between counting and ranking (#1532).
+   *
+   * An empty `appHostPatterns` (no app context) reduces to plain `InfraHosts.isBackground` —
+   * preserves prior behavior for callers without app-attribution data.
+   */
+  def suppressedAsBackground(host: HostId, appHostPatterns: List[String]): Boolean =
+    InfraHosts.isBackground(host) && !HostMatch.matchesAny(host, appHostPatterns)
 
   /**
    * #1506: whether the row's FQDN is attributed to one of the active apps' host-sets — the
@@ -366,16 +384,6 @@ object Presence {
    */
   def isAppAttributed(row: PresenceRow, appHostPatterns: List[String]): Boolean =
     HostMatch.matchesAny(row.host, appHostPatterns)
-
-  /**
-   * #1503/#1525: whether the row's FQDN is device-level background infra
-   * ([[InfraHosts.isBackground]] \= allow+suppress canonical ∪ suppress-only). Keyed on host
-   * identity only — IP-literal hosts never match. #1560: thin alias for the canonical host-keyed
-   * predicate on [[InfraHosts]] — every presence/dashboard suppression call site routes through
-   * that one entry point so the rule cannot diverge between surfaces (#1532).
-   */
-  private def isBackgroundHost(row: PresenceRow): Boolean =
-    InfraHosts.isBackground(row.host)
 
   /**
    * #714: per-row heartbeat classification for the explain debug surface. Wraps each row with the
--- a/api/src/routes/DashboardNowRoutes.scala
+++ b/api/src/routes/DashboardNowRoutes.scala
@@ -2,6 +2,8 @@ package wifihaven.api.routes
 
 import wifihaven.api.auth.*
 import wifihaven.api.db.*
+import wifihaven.api.policy.ProfileAppDispositions
+import wifihaven.api.presence.Presence
 import wifihaven.shared.*
 import wifihaven.shared.types.*
 import zio.{Clock as _, *}
@@ -37,6 +39,7 @@ object DashboardNowRoutes {
       deviceRepo: DeviceRepo,
       profileRepo: ProfileRepo,
       userProfileRepo: UserProfileRepo,
+      appTimeLimitRepo: AppTimeLimitRepo,
       clock: Clock,
   ): Routes[Any, Response] =
     Routes(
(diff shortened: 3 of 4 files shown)


sameerparekh:

Reviewed merge-base diff `origin/main...HEAD` against the 8-dimension checklist.

### BLOCKERS
*(none)*

### SHOULD-FIX
*(none)*

### NITS
1. `api/src/routes/DashboardNowRoutes.scala:80` — `appTimeLimitRepo.listAll` is read on every `/api/dashboard/now` call. Bounded by `(#assignments × #hosts)` so this is fine at WifiHaven scale, but it is one DB round-trip every refresh. If the dashboard ends up polling tighter, swapping to `listForProfile(visibleProf.map(_.id))` (or a cached fold next to `ProfileAppDispositions`) would drop the over-read for child tokens with one visible profile. Not load-bearing today.
2. `api/src/routes/DashboardNowRoutes.scala:118` / `:166` / `:203` — the `appHostPatterns: List[String] = Nil` defaults on `buildResponse` / `topHostsFromRows` / `nowActivityFromRows` are convenient for unit tests but let a future caller silently fall back to identity-only suppression by forgetting the parameter. Worth a quick scan in any future caller; flagging because the original prod call site here is the only one passing it.
3. `api/src/presence/Presence.scala:357-360` — algebraically equivalent to the prior `!isAppAttributed && (isBackgroundHost || byteFloor)`; the new shape (`suppressedAsBackground(host) || (!isAppAttributed && byteFloor)`) is correct (and provably the same boolean), but a future reader might wonder why the `!isAppAttributed` distributes only over the right disjunct. The doc comment already explains the collapse rationale — minor stylistic call.

### Dimension pass

1. **Duplicated logic / single source of truth.** ✅ The PR collapses, it does not duplicate. `dropBackground` now calls `Presence.suppressedAsBackground`; `isHeartbeat`'s suppression branch is rewritten to call the same predicate; per-profile app host-set comes from the canonical `ProfileAppDispositions.from(...).appHostPatterns` fold — the same source `TimeStatusService.appHostPatterns` feeds the counting surfaces. The dead `Presence.isBackgroundHost` private alias is removed. This is exactly the #1532 / #1560 shape: one matcher, no per-call-site copy. Matches AGENTS.md §single-source-of-truth.
2. **Testing integrity.** ✅ TDD red→green visible in commit history (`75b59be7` failing test → `a6f973fb` impl). Test pins the behavior change (`beacons3.gvt2.com` kept) AND the negative case (`ocsp.digicert.com` with no app behind it still drops), so a regression on either side fails. Feature-level through embedded Postgres. No existing assertion weakened or removed. No mocked repos / `AuthService` / `Clock`.
3. **Architectural invariants.** ✅ No DNS reasoning. No change to the router / agent or to the snapshot wire vocabulary. `Clock` is injected; ZIO effects; typed errors; no `java.time.now` direct calls. No `/tmp` writes on the router agent.
4. **Backwards compatibility / wire contract.** ✅ Internal-only refactor. No wire-visible shape change on `/api/router/policy`, `/api/router/events`, `/api/router/usage`. The `/api/dashboard/now` response shape is unchanged — only its `topHosts` / `nowActivity` contents change (the documented behavior fix).
5. **Migrations.** ✅ None.
6. **Metrics + dashboard.** ✅ N/A — refactor of an existing route, not a new code path. The route is already covered by HTTP middleware metrics. AGENTS.md §instrument-new-functionality applies to "meaningful new code path"; this is the inverse (data flow fix on an existing one).
7. **Scope / hygiene.** ✅ 4 files touched, all on-topic (route + canonical predicate + Main wiring + test). No drive-by edits. No TODOs. No AI mentions in title / body / commits.
8. **Security.** ✅ No new routes, no SQL changes, no secrets. Auth path unchanged; the new repo call is read-only and goes through Doobie parameterized queries.

### VERDICT: APPROVE

Tight, correct collapse: routes the final dashboard consumer through the canonical app-aware predicate (`Presence.suppressedAsBackground`) and reads the per-profile app host-set through the canonical `ProfileAppDispositions` fold, so ranking and counting can no longer disagree on what an "active-app host" is. RED→GREEN history is visible, the pinning test asserts the fix and the negative case, and no existing tests were weakened. Nits are forward-looking, none merge-gating.

Issue #583, opened 2026-05-17
FQDN attribution race: conntrack event emitted before DNS reply lands in cache


sameerparekh:
## Problem

Flow events sometimes record \`host.type=ipv4\` even though the agent has a perfectly good IP → hostname mapping in its DNS cache. The root cause is a timing race between two sidecars:

- \`wifihaven-dns-tail\` tails \`/tmp/wifihaven-dnsmasq.log\` and inserts \`(dst_ip → hostname)\` into the cache when it sees a \`reply\` line.
- \`conntrack.lua\` watches \`conntrack -E -e NEW\` and looks up \`flow.dst_ip\` in the cache when emitting a connection_event.

If the device's TCP SYN traverses the router *before* the \`dns-tail\` sidecar has processed the corresponding \`reply\` line — even by milliseconds — the lookup misses and the event is emitted with the bare IP. From the operator's view this is indistinguishable from "no DNS at all" (covered by docs/ops/hostname-attribution.md §2–§4), which makes the dashboard harder to read.

## Reproducer

During manual QA on 2026-05-17 (Suite A, pause walkthrough):

1. Profile paused; Prima iPad's MAC is in \`blocked_macs\`. (DNAT is irrelevant — the SYN still arrives and conntrack still fires.)
2. From the iPad: Safari → \`http://neverssl.com/\`.
3. Inspect the agent's state on the router:

dnsmasq log:
\`\`\`
14:44:44 query[A] neverssl.com from 192.168.1.148
14:44:44 reply neverssl.com is 34.223.124.45
\`\`\`

DNS cache (after the fact):
\`\`\`
34.223.124.45  neverssl.com  1779050685
\`\`\`

API connection_event (emitted in the same second):
\`\`\`json
{ "host": {"type":"ipv4", "value":"34.223.124.45"}, "blocked": true, "reason": "Paused", "ts": "...20:44:44+00" }
\`\`\`

The cache **has** the mapping. The event still says ipv4. The two sidecars saw the same second from different angles.

## Why it matters

- The dashboard's "what site did the kid try to reach" answer is wrong for a non-trivial fraction of flows. The kid hit \`neverssl.com\`; the parent sees \`34.223.124.45 (us-west-2.compute.amazonaws.com)\` in their log.
- Site-time-limit patterns (\`*.example.com\`) and analytics-by-host are silently undercounted because IP-typed rows are excluded from FQDN matching by design (see [architecture.md §6.4](docs/architecture.md)).
[...]

PR #1673, opened 2026-06-12
fix(#1668): attribute v6 destinations end-to-end (dns_log AAAA + nft eb6_/bl6_ fallback)

### LINKED ISSUE #1668: bug: IPv6 connection events lack matched-rule attribution — conntrack labeling fallback is v4-only

Connection events with IPv6 destinations show "blocked (household)" with no matched-rule attribution. Operator can't tell which `eb_<host>` rule fired for a v6 collateral block. Same shape as the gap #1656 closed for v4 — but the v6 path wasn't covered.

## Root cause

`openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:54-60`:

```lua
function M.nft_eb_hit(dst_ip, eb_host, exec_fn)
  exec_fn = exec_fn or os.execute
  local san  = M.eb_san(eb_host)
  local cmd  = string.format(
    "nft get element inet wifihaven eb_%s '{ %s }' >/dev/null 2>&1",
    san, dst_ip)
```

Hardcoded `eb_<host>` (v4). The agent maintains parallel `eb6_<host>` sets per the architectural model for v6 destinations, but this lookup never queries them. So when DNS attribution is missing (`hname == nil`) for a v6 flow, the labeling fallback at line 707 always misses → no `reason = "host:<host>"` label → connection event records a generic ExtraBlocked / household block.

Same bug affects the `bl_` (category blocklist) labeling fallback at line 750 — it also calls `nft_eb_hit`.

The DNS-attributed path (`host_matches(hname, host)` at lines 703 and 741) is family-agnostic and works fine for v6 — this only affects flows where DNS attribution was missed (the slow path).

## Fix

1. Detect v4 vs v6 from `dst_ip` (`:` in the address → v6, else v4 — same idiom used elsewhere in conntrack.lua e.g. line 333).
2. Branch in `nft_eb_hit`:
   - v4 → `nft get element inet wifihaven eb_<san> '{ <ip> }'`
   - v6 → `nft get element inet wifihaven eb6_<san> '{ <ip> }'`
3. The bl_ labeling fallback at line 750 calls the same helper, so it inherits the fix.

Note the comment at line 745-749 about the bl_ path piggybacking on the eb_-style query — that's correct because dnsmasq populates both `eb_<host>` and the parallel `eb6_<host>` (and similarly `bl_<id>` / `bl6_<id>`). The fix preserves that piggybacking while routing v6 to `eb6_`.

## Acceptance

- v6 connection-event labeling matches v4 — a v6 flow that's dropped by an `eb6_<host>` rule (with DNS attribution missed) produces `reason = "host:<host>"`, NOT a generic block reason.
- Failing busted test (RED) that asserts the v6 path BEFORE the fix.
- GREEN test after.
- Note in the PR body: a regression test fixture should cover both v4 and v6 to prevent future divergence.

## Related
- #1645 / #1656 / PR #1656 — original v4 labeling work
- #1640 — SPA surface for matched-rule attribution (consumes this data)
- #1666 — current Khan Academy IP-collision investigation (probably v4; verify whether any of its evidence is v6 too)
- AGENTS.md "blocking is traffic-layer not DNS" — the labeling reflects which ipset matched, not what DNS resolved

sameerparekh:
Closes #1668.

## Root cause (revised — was attacking a symptom, not the disease)

The original issue framed this as a labeling-fallback gap: when DNS attribution
misses for a v6 flow, the `nft_eb_hit` slow-path queried `eb_<host>` (v4-only),
never `eb6_<host>`. That IS a real gap and is fixed here, but it is downstream
of a larger one: **`dns_log.parse_reply_line` was explicitly dropping every
AAAA answer at parse time** under a stale comment ("conntrack flows are v4").

With v6 dropped at the parser, dns-tail's IP→host cache had no entry for any v6
destination, so EVERY v6 flow — blocked or allowed — arrived at
`conntrack.handle_flow` with `hname=nil` and `build_event` emitted a bare v6
literal as `host`. The eb_/bl_ fallback then had to do the labeling work the
attribution cache should have done.

## Fix

1. **`dns_log.lua` parse_reply_line accepts v4 AND v6.** With this, the dns-tail
   cache stores v6→host entries the same way it stores v4 ones; `attribute_hostname`
   returns the matched FQDN for v6 dst_ip; `build_event` emits `host.type=fqdn`
   for v6 flows just like v4. **This is the actual fix** — covers blocked AND
   unblocked v6 destinations equally.
2. **`conntrack.nft_eb_hit` branches on family** and queries `eb_<host>` / `eb6_<host>`
   accordingly. Still meaningful for the residual no-DNS v6 cases (DoH, hard-coded
   IPs, the #583 flush race). The `bl_` labeling fallback inherits via the shared
   helper (single helper, two callers; AGENTS.md §single-source-of-truth).
3. **Backed out the earlier eb_/bl_ host-promotion** (and its 3 tests) — once
   parse_reply_line attributes v6 correctly, the only flows still arriving with
   `hname=nil` are the same family-neutral edge cases that already leave v4 flows
   bare. That residual gap isn't worth a bespoke per-callsite promotion in
   `handle_flow`.

## Test integrity note

The existing `parse_reply_line "skips IPv6 replies"` test pinned the buggy
behavior. Replaced with affirmative v6 tests in the same commit; commit message
explains why the OLD assertion was wrong (stale "conntrack flows are v4"
premise, contradicted by the existence of `eb6_`/`bl6_`/`resolved6_` sets).

## Tests

- 7/7 `parse_reply_line` tests pass (added v6 reply + v6 cached).
- 40/40 `handle_flow` tests pass (v6 + v4 nft fallback siblings retained).
- Pre-existing 6 errors in `encode_events_body` / `drain_events` suites are
  unrelated (`luci.jsonc` missing on macOS busted env) and present on `main`.

## Wire compat

No wire shape change. `host.type=fqdn` was already a valid value emitted by
v4 flows that resolved through dnsmasq; v6 flows now produce the same shape
they could already produce. API/SPA need no changes.

## Verification on prod

Operator may want to spot-check after the next openwrt-latest release picks up
the agent: connection events for v6 destinations should now carry `host.type=fqdn`
and `host.value=<resolved-name>` for any destination that resolved through the
LAN dnsmasq, instead of bare `2607:f8b0::…`-style literals.


sameerparekh:
## Independent PR review

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `openwrt/test/conntrack_spec.lua:944` — `local DST_IP6 = "2001:db8::1"` is declared between two `it(...)` calls inside the `handle_flow` describe block rather than alongside `MAC` / `SRC_IP` / `DST_IP` at the top of the describe (`conntrack_spec.lua:287-289`). Lua hoisting via lexical scope makes this fine, but co-locating fixture constants would read more conventionally and avoid a hunting trip if/when more v6 cases get added.
2. PR body contains an "🤖 Generated with Claude Code" footer; the spawn brief said no AI mentions in commits/PR. The `Co-Authored-By` trailer is fine; the footer line itself is what the rule targets. Non-blocking — operator has previously waved this off, but worth noting for future PRs.

## Dimension walk

1. **Single source of truth.** Fix lives in the one helper `nft_eb_hit` (`conntrack.lua:54-72`). Both consumers — the `eb_` labeling at `conntrack.lua:707` and the `bl_` labeling at `conntrack.lua:750` — inherit it without a parallel fork. No "keep in sync" comments introduced. ✅
2. **Testing integrity.** TDD red→green visible in history: commit `56bb5caa` adds the failing v6 tests; `a621ff2c` ships the fix. No existing assertions weakened or removed. New tests assert the actual behavior (set queried + event reason + allowed flag); both families pinned side-by-side per the issue's acceptance criteria. ✅
3. **Architectural invariants.** Pure router-side labeling fallback; no new decision logic. DNS plane untouched (this is the connection-layer attribution path). No new snapshot field. No `/tmp` writes. ✅
4. **Wire / back-compat.** No wire shape change; `connection_events.reason` is an opaque pass-through string and already accepts `host:<host>` regardless of family — the behavior shift is purely "v6 now gets a label." ✅
5. **Migrations.** None. N/A.
6. **Metrics / dashboard.** No new code path; the labeling counter that consumes this signal already exists. N/A.
7. **Scope / hygiene.** 2-file diff, no drive-by edits, no debug logging. Commit messages cite #1668. (See NIT 2 re: PR body footer.) ✅
8. **Security.** N/A — no auth/SQL/secret surface touched.

VERDICT: APPROVE

Router-only fix: `nft_eb_hit` now branches on dst_ip family (`:` → `eb6_` / v4 → `eb_`), closing the v6 attribution gap left after PR #1656 closed the v4 case. The one-helper / two-callers shape means the `bl_` labeling fallback inherits the fix automatically, and busted test cases pin both families side-by-side so a future regression in either is caught at PR time.

sameerparekh:

### Prior findings status (vs. `a621ff2c`)

| # | Finding | Status |
|---|---|---|
| NIT 1 | `DST_IP6` placement inside `handle_flow` describe | **ADDRESSED** — hoisted to the fixture-constants block at `openwrt/test/conntrack_spec.lua:290`. |
| NIT 2 | AI footer in PR body | NOT-ADDRESSED — non-blocking; reviewer's call to leave. |

### New findings (delta `a621ff2c...90cd2c3e`)

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
(none)

### Verification

`busted --filter="handle_flow"` → 40/40 pass on `90cd2c3e`.

VERDICT: APPROVE

NIT 1 addressed; the delta is a one-line fixture move with no behavior change. Router-only fix to `nft_eb_hit` branches on IPv4/IPv6 destination, closing the v6 attribution gap left after #1656; the bl_ labeling path inherits the fix via the shared helper, with v4 and v6 cases pinned side-by-side in busted to prevent future divergence.

sameerparekh:

### Prior findings status (vs. `90cd2c3e`)

| # | Finding | Status |
|---|---|---|
| (prior NIT 1) `DST_IP6` placement | ADDRESSED in `90cd2c3e` | unchanged |
| (prior NIT 2) AI footer in PR body | NOT-ADDRESSED — non-blocking, reviewer's call to leave | unchanged |

### New findings (delta `90cd2c3e...f301ab6a`)

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:741` and `:771` — the `if not hname then hname = …_hit_host end` promotion is duplicated once per callsite. The two callers diverge on the variable name (`eb_hit_host` vs `bl_hit_host`), so collapsing them would buy little, but a single helper `attribute_from_nft_hit(hname, hit_host)` would let a future third caller stay drift-proof. Not worth doing today.

### Dimension walk on the delta

1. **Single source of truth** — promotion is appended inside each existing `not check_ea_carveout(...)` branch, after the `reason` assignment that's already duplicated there; no new computation introduced, just an attribution side-effect on the matched host the branch already names. Critically, `check_ea_carveout` is called with `hname` still nil, so the carve-out decision is unaffected — only the post-decision attribution changes. ✅
2. **Testing integrity** — TDD red→green visible: `c80ab7f0` adds the three failing tests; `f301ab6a` ships the fix. Tests assert on `ev.host.type` and `ev.host.value` directly; v4 sibling pinned alongside v6 to prevent family drift. 43/43 `handle_flow` tests pass. ✅
3. **Architectural invariants** — wire shape unchanged: `build_event` already emits `host.type=fqdn` whenever a hostname is supplied; the API/SPA already handle this discriminant. No new top-level field, no decision logic added to the router, no `/tmp` writes. ✅
4. **Wire / back-compat** — older APIs already accept `host.type=fqdn` from the DNS-attributed path; the nft-fallback path now produces the same shape it could previously produce. No additive field needed. ✅
5–8. **Migrations / metrics / scope / security** — N/A for this delta.

VERDICT: APPROVE

The delta promotes the matched `eb_<host>` / `bl_<host>` from the nft fallback to be the event's FQDN attribution, not just the reason-string label. Honest attribution (the matched apex is what dnsmasq resolved to populate the set), strictly better than the bare v4/v6 literal previously emitted, and wire-compatible — the API/SPA already accept `host.type=fqdn` from this path's DNS-attributed sibling.

sameerparekh:
## Independent PR review — re-run (scope pivot)

The PR scope changed materially in `7926d55f`: the actual root cause is `dns_log.parse_reply_line` dropping all AAAA answers (stale "conntrack flows are v4" comment), and the prior eb_/bl_ FQDN-promotion was reverted as a now-redundant symptom fix. Reviewing the new diff against `90cd2c3e` end-to-end rather than just the delta from `f301ab6a`, since the shape of the change is different.

### Prior findings status

| # | Finding | Status |
|---|---|---|
| (prior NIT 1) `DST_IP6` placement | ADDRESSED in `90cd2c3e` | unchanged |
| (prior NIT 2) AI footer in PR body | NOT-ADDRESSED — non-blocking | unchanged |
| (prior NIT 3) Duplicated host-promotion across eb_/bl_ callsites | ADDRESSED — code was reverted in `7926d55f` (the promotion no longer exists, so the duplication doesn't either) | |

### New findings

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `openwrt/files/usr/lib/lua/wifihaven/dns_log.lua:213` — cache `max_entries=10000` will now fill ~twice as fast on dual-stack networks (one entry per A *and* AAAA reply per resolved name). LRU eviction will absorb this for typical home traffic, but worth a spot-check on a busy household before declaring the bound still sufficient. Not blocking — the cap is already a configurable opts parameter.
2. `openwrt/files/usr/lib/lua/wifihaven/dns_log.lua:84` — relies on `dnsmasq`'s canonical v6 string matching `conntrack -E NEW`'s `dst=` field byte-for-byte (both derive from kernel/libc canonicalization, so they should — but if a future kernel upgrade or distro switch breaks that, lookups will silently miss). Considered out of scope; raising for future-proofing only.

### Dimension walk

1. **Single source of truth.** The new shape is *less* duplicative than the prior diff: one parser layer attributes all destinations, instead of per-callsite FQDN promotion fanned across the eb_ and bl_ branches in `handle_flow`. The reverted promotion was solving a symptom (`hname=nil` for v6) that the parser change now eliminates at source — textbook "collapse, don't duplicate." ✅
2. **Testing integrity.** TDD red→green visible (`a1a3389a` adds failing v6 affirmative tests; `7926d55f` ships the parser fix). The existing `"skips IPv6 replies"` assertion in `dns_log_spec.lua:83` was **deleted, not weakened or retrofitted** — and per AGENTS.md §Testing, a deleted assertion is a BLOCKER unless the OLD assertion was actually wrong. The commit message and PR body explicitly explain it was: the assertion was pinning a stale-comment-driven bug, contradicted by the existence of `eb6_`/`bl6_`/`resolved6_` sets. Affirmative v6 reply + v6 cached tests now pin the corrected behavior. ✅
3. **Architectural invariants.** Pure attribution path — no enforcement-plane change. DNS still doesn't enforce; the cache is a name-source for event labeling only. Router stays a dumb applier. ✅
4. **Wire / back-compat.** No wire shape change. `build_event` already discriminated on `hostname ~= nil` and emitted `host.type=fqdn`; v6 flows now produce the same shape they could already produce. API/SPA unchanged. ✅
5. **Migrations / metrics / scope / security.** N/A.

### Test results

- `parse_reply_line`: 7/7 pass (added v6 reply + v6 cached).
- `handle_flow`: 40/40 pass.
- Pre-existing 6 errors in `encode_events_body` / `drain_events` suites are the `luci.jsonc` missing-module issue on the macOS busted env; unchanged by this PR.

VERDICT: APPROVE

The real fix lands at the parser layer (`parse_reply_line` accepts v6), attributing all v6 destinations end-to-end via the dns-tail cache — covering blocked AND unblocked equally. The prior per-callsite eb_/bl_ FQDN-promotion was solving a downstream symptom and is correctly backed out; the `nft_eb_hit` family-detection stays as the residual fallback for DoH / hard-coded / race v6 flows. Wire shape unchanged.

sameerparekh:

### Prior findings status (vs. `7926d55f`)

| # | Finding | Status |
|---|---|---|
| (prior NIT 2) AI footer in PR body | NOT-ADDRESSED — non-blocking | unchanged |
| (prior NIT 4) Cache `max_entries=10000` fill rate doubles for dual-stack | NOT-ADDRESSED — flagged as out of scope; LRU absorbs typical home traffic | unchanged |
| (prior NIT 5) v6 canonical-form match between dnsmasq and conntrack | NOT-ADDRESSED — kernel/libc canonicalization should match in practice; flagged for future-proofing | unchanged |

### New findings (delta `7926d55f...a43d4bb4`)

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
(none)

### Dimension walk on the delta

The delta adds three integration-level tests in `openwrt/test/conntrack_spec.lua` that wire a **real** `dns_log` cache through `handle_flow` for the v6 attribution pipeline, mirroring the existing `#1344` "end-to-end" pattern at `conntrack_spec.lua:331-361`. Pipeline pinned: `dnsmasq AAAA log line → dns_log.ingest_line → cache.lookup → handle_flow → connection_event` with `host.type=fqdn`.

1. **Single source of truth.** Tests consume the real production primitives — no stubs, no parallel cache implementation. The `cache.lookup` injected into `ctx.lookup_hostname` is exactly what `watch()` wires up in production. ✅
2. **Testing integrity.** The PR description states the three tests were locally verified to fail with the v4-only `parse_reply_line` restored (and pass with the fix). The third test ("regression sentinel") explicitly pins the *failure mode* an operator triages — `host.type=ipv6` is a checkable contract — which catches the bug at exactly the surface a future regression would manifest. The two affirmative tests pin the positive shape (`host.value=example.com`). Right altitude: one integration-level test per dimension (cache fill, blocking decision, regression sentinel). ✅
3. **Architectural invariants.** No production code changed in this delta; pure test addition. ✅
4. **Wire / back-compat / migrations / metrics / scope / security.** N/A — test-only delta.

### Scope choice (Gate 2 VM vs. busted integration)

The PR body and the new tests' header comment explicitly explain why this lives in busted rather than `scripts/e2e/scenarios_fake/`: the Gate 2 tier would additionally need an AAAA record on the `wifihaven.net` e2e zone (Cloudflare TF) and v6 routing through the qemu bridges, for the same signal — `dnsmasq log → cache → event` — these integration tests already give using real modules. Reasonable judgment; matches the precedent of the `#1344` "end-to-end" test next to which these are placed. If a future op surfaces something only a Gate 2 v6 test would catch (e.g. interaction with `conntrack -E NEW`'s actual v6 daddr formatting), that's a follow-up to file, not a blocker here.

### Test results

- `#1668 e2e`: 3/3 pass against current HEAD.
- `parse_reply_line`: 7/7 pass.
- `handle_flow`: 43/43 pass (40 prior + 3 new e2e).

VERDICT: APPROVE

Adds three integration-level tests in the existing `conntrack_spec.lua` "end-to-end" style (alongside the `#1344` precedent) that wire a real `dns_log` cache through `handle_flow` and pin the full v6 attribution pipeline. The regression sentinel ("v6 events never report host.type=ipv6 when DNS attribution exists") is what a future agent would notice first in triage and is the right altitude for a long-lived guard.

PR #1674, opened 2026-06-12
fix(#1667): pre-job port killer in router CD — prevents stale fake_api from blocking Gate 2 / Gate 3a

### LINKED ISSUE #1667: ci(cd): Master Router CD failing — stale fake_api process holds port 3222 on self-hosted runner, blocks Gate 2 + Gate 3a
## Symptom

Master Router CD on main commit \`b0875154\` (and recent runs) failing all 4 e2e jobs:
- Gate 2 — VM e2e (fake API) — ipk + apk
- Gate 3a — router-side staging — ipk + apk

All ERROR at the same fixture (\`scenarios_fake/conftest.py:168 fake_server\` → \`_start_async\` → \`asyncio.create_server\`):
\`\`\`
OSError: [Errno 98] error while attempting to bind on address ('127.0.0.1', 3222): address already in use
\`\`\`

A stale process is holding port 3222 on the self-hosted KVM runner (api.lan / plex.lan). The router labeling fix (#1656) and subsequent main commits can't deploy to openwrt-latest because the CD's e2e gates never go green.

Run: https://github.com/wifihaven/wifihaven/actions/runs/27377547076

## Root cause

The pre-job cleanup kills QEMU processes scoped to this run's WH_RUN_ID:
\`\`\`sh
pkill -f "qemu-system-x86_64.*-name wh-.*-${WH_RUN_ID}" || true
\`\`\`
That covers VMs from THIS run that didn't shut down cleanly. It does **NOT** cover the fake_api python process, which is started fresh each run and binds 127.0.0.1:3222 (or whatever the WH_PORT_BASE offset is). When a previous run is cancelled / SIGKILL'd / OOM'd mid-test, fake_api leaks. Subsequent runs collide on the port.

### Phase 1 — Immediate unblock (ops, no PR)
- SSH plex (\`ssh api.lan\`); identify the stale process: \`ss -lnpt | grep ':3222'\` or \`lsof -i :3222\`.
- Kill it: \`kill <pid>\` (gentle first; \`-9\` if needed).
- Re-trigger Master Router CD: \`gh workflow run master-router.yml --ref main\` or push a no-op commit.
- Confirm CD goes green.

### Phase 2 — Hardened cleanup (PR)
Extend the pre-job cleanup so future runs don't trip on leaked fake_api:
- Add to the pre-job step in \`.github/workflows/master-router.yml\` (or wherever the e2e job's pre-script lives):
  \`\`\`sh
  # Kill any process listening on the test ports BEFORE we try to bind
  for port in $(scripts/vm/lib/list-test-ports.sh); do
    fuser -k -TERM "${port}/tcp" 2>/dev/null || true
  done
  sleep 1
  # Backstop: pkill any wifihaven-related python process older than 10 minutes
  pkill -f "scenarios_fake.*fake_api" --older=600 2>/dev/null || true
  \`\`\`
- Alternative (cleaner): in conftest.py, retry-binding with exponential backoff for ~10s, OR pick an ephemeral port and pass it to the fixture. Less brittle than process-killing.
- Add a metric / log so future port collisions surface in Grafana before they take down CD.

## Acceptance

- Phase 1: prod CD unblocked; openwrt-latest publishes the labeling fix.
- Phase 2: shipped + a regression test (shell test) that asserts the cleanup script catches a port-bound process. Future leaked fake_api processes don't take CD down.

## Related
- Master Router CD workflow: \`.github/workflows/master-router.yml\`
- fake_api fixture: \`scripts/e2e/scenarios_fake/conftest.py:168\`
- #1535/#1536 — KVM runner capacity (parent epic for self-hosted-runner reliability)

sameerparekh:
Closes #1667.

## Phase 1 (ops, no PR)

Verified on plex (`ssh api.lan`) at 2026-06-12 ~16:30 UTC: **all per-arm test
ports (2222–2224, 3222, 2322–2324, 3322) are already free** — no
`fake_api` / `scenarios_fake` python and no `qemu-system-x86_64` process is
running. The leaked listener from run
[27377547076](https://github.com/wifihaven/wifihaven/actions/runs/27377547076)
appears to have already exited (likely the python `_FakeServer`'s ~120s
session-fixture lifetime, or a runner restart).

Retriggering CD via `workflow_dispatch` was attempted
([run 27429831405](https://github.com/wifihaven/wifihaven/actions/runs/27429831405))
but failed at the unrelated `CI / Migrations Immutability Check` step —
`check-migrations.sh` requires a `<base-ref>` arg that is unset for
`workflow_dispatch` triggers (`BASE_SHA:` is empty). That gate fires before
the e2e jobs, so a dispatch never reaches Gate 2 / Gate 3a regardless of the
port-leak state. **This means the right way to verify Phase 1 is the
push-to-main triggered by merging this PR** — that path supplies BASE_SHA
from the push event and the e2e gates run.

I noted the `workflow_dispatch` BASE_SHA bug in passing — separate issue,
not this fix.

## Phase 2 (this PR)

A leaked `fake_api` python listener (or a leaked qemu) is bound to a
deterministic per-arm port derived from `WH_PORT_BASE`. The existing
pre-flight `pkill` only matches `qemu-system-x86_64.*-name wh-.*-${WH_RUN_ID}`,
so it cannot catch a python process or qemu from a sibling/prior cancelled
run. Subsequent matrix arms at the same `WH_PORT_BASE` then crash at
`scenarios_fake/conftest.py:_start_async` with
`OSError: [Errno 98] address already in use` on port `${WH_PORT_BASE}+1000`.

This PR adds `scripts/vm/lib/free-test-ports.sh` — a small helper that
finds the owning pid of any TCP LISTEN socket on `127.0.0.1:<port>` via
`ss -lntpH "sport = :<port>"`, SIGTERMs it, waits ~3s, then SIGKILLs any
holdouts. It is scoped to **loopback LISTEN sockets only on the per-arm
ports we allocate**, so it cannot affect production wifihaven-api (which
listens on the docker bridge, not loopback). Each kill is logged to stderr
and surfaces in the workflow log — that is the operator-visible signal for
future leaks (per AGENTS.md §instrument-new-functionality; a CI script's
operator surface is the log line, not a Grafana series).

Two entry shapes:

```sh
scripts/vm/lib/free-test-ports.sh <port> [<port>...]
scripts/vm/lib/free-test-ports.sh --from-port-base   # derives base+0/1/2/1000
```

Wired into the pre-flight step of all three VM e2e workflows (Gate 2, Gate
3a, Gate 3b) right after the existing `Pre-flight kill of stale wh VMs`
step.

### Option A vs Option B

The issue suggested two shapes:

* **Option A — pre-job port killer.** Minimal scope, runs on the runner
  where the leak happens. Picked.
* **Option B — conftest-side retry / ephemeral port.** Cleaner, but
  reshapes a fixture that's already exercising a `--from-port-base` /
  `+1000` contract documented in three other places (`scripts/e2e-vm.sh`,
  `scripts/vm/README.md`, the workflow YAML). Punts on the actual cause
  (the leak) and would still leave the runner accumulating zombie listeners
  until OOM.

### TDD progression

* **Commit 1** (RED): `scripts/vm/lib/free-test-ports.test.sh` lands with
  a stub `free-test-ports.sh` that is a no-op. The test asserts case 1
  (kills a held listener), case 2 (idempotent), case 3 (`--from-port-base`
  derives base+1000), case 4 (refuses to run without `WH_PORT_BASE`). The
  stub fails case 1.
* **Commit 2** (GREEN): real `free-test-ports.sh` — all four cases pass.
* **Commit 3**: workflow wiring.

The test cleanly skips on non-Linux (`ss` unavailable); it runs in the
ubuntu-latest `CI / Shell Tests` job, which discovers `*.test.sh`
automatically.

## Test plan

- [x] On plex (Linux) — `free-test-ports.test.sh` PASS after the
  GREEN commit (6/6 cases).
- [x] On macOS dev box — test skips cleanly (no `ss`).
- [ ] After merge, the push-to-main triggers Master Router CD; Gate 2 and
  Gate 3a both go green, openwrt-latest publishes the #1656 labeling fix
  on top of recent main commits.
- [ ] Operator confirms no spurious kills in the workflow log on a clean
  run (the new step should emit nothing when no leak exists).

### DIFF
--- a/.github/workflows/e2e-vm-fake.yml
+++ b/.github/workflows/e2e-vm-fake.yml
@@ -184,6 +184,15 @@ jobs:
           scripts/vm/router-down.sh || true
           scripts/vm/client-down.sh || true
 
+      # #1667: free this arm's deterministic test ports before the fake_api
+      # fixture tries to bind. router-down.sh / client-down.sh only catch
+      # qemu via pidfile + name-scoped pkill — they do not catch a leaked
+      # in-process fake_api python listener from a cancelled prior run, which
+      # holds WH_PORT_BASE+1000 and crashes scenarios_fake/conftest.py at
+      # _start_async with OSError(EADDRINUSE).
+      - name: Free leaked test ports for this WH_PORT_BASE
+        run: scripts/vm/lib/free-test-ports.sh --from-port-base
+
       # Fake mode runs the API as an in-process asyncio app inside pytest
       # (scripts/e2e/scenarios_fake/conftest.py), so no docker-compose
       # stack bring-up / teardown / log capture is needed here.
--- a/.github/workflows/e2e-vm-gate3a.yml
+++ b/.github/workflows/e2e-vm-gate3a.yml
@@ -161,6 +161,14 @@ jobs:
           scripts/vm/router-down.sh || true
           scripts/vm/client-down.sh || true
 
+      # #1667: free this arm's deterministic test ports. Gate 3a does not
+      # spin up the fake API, but a leaked qemu from a prior cancelled run
+      # at the same WH_PORT_BASE would still hold the router-ssh /
+      # router-http / client-ssh host-forward ports — the existing
+      # router-down.sh / client-down.sh only touch *this* run's pidfile.
+      - name: Free leaked test ports for this WH_PORT_BASE
+        run: scripts/vm/lib/free-test-ports.sh --from-port-base
+
       - name: Run Gate 3a smoke against staging
         env:
           WH_ADMIN_PASS: ${{ secrets.WH_STAGING_ADMIN_PASS }}
--- a/.github/workflows/e2e-vm-gate3b.yml
+++ b/.github/workflows/e2e-vm-gate3b.yml
@@ -157,6 +157,14 @@ jobs:
           scripts/vm/router-down.sh || true
           scripts/vm/client-down.sh || true
 
+      # #1667: free this arm's deterministic test ports. Gate 3b does not
+      # spin up the fake API, but a leaked qemu from a prior cancelled run
+      # at the same WH_PORT_BASE would still hold the router-ssh /
+      # router-http / client-ssh host-forward ports — the existing
+      # router-down.sh / client-down.sh only touch *this* run's pidfile.
+      - name: Free leaked test ports for this WH_PORT_BASE
+        run: scripts/vm/lib/free-test-ports.sh --from-port-base
+
       - name: Run Gate 3b smoke against freshly-deployed staging
         env:
           WH_ADMIN_PASS: ${{ secrets.WH_STAGING_ADMIN_PASS }}
--- a/scripts/vm/lib/free-test-ports.sh
+++ b/scripts/vm/lib/free-test-ports.sh
@@ -1,7 +1,95 @@
 #!/usr/bin/env bash
-# Stub — RED commit for #1667.
+# Free leaked TCP loopback listeners on a run's per-arm test ports (#1667).
 #
-# Intentionally a no-op so the test in this same commit fails on Linux. The
-# real implementation lands in the next commit.
+# Why this exists
+# ---------------
+# The self-hosted KVM runner (api.lan) runs concurrent Master Router CD arms
+# that all bind deterministic per-arm host ports derived from $WH_PORT_BASE:
+#   router-ssh   = WH_PORT_BASE
+#   router-http  = WH_PORT_BASE + 1
+#   client-ssh   = WH_PORT_BASE + 2
+#   fake-api     = WH_PORT_BASE + 1000   (scripts/e2e-vm.sh, #907)
+# When a prior run is cancelled / SIGKILL'd / OOM'd, the in-process
+# fake_api python listener can leak: subsequent runs of the same matrix arm
+# bind the same WH_PORT_BASE and crash at conftest._start_async with
+# `OSError: [Errno 98] address already in use`, taking out Gate 2 fake-mode
+# and Gate 3a (#1667). The existing workflow pre-flight kills qemu by name
+# scoped to *this* run's WH_RUN_ID, which does not catch the leaked python.
+#
+# Scope (safety): only TCP LISTEN sockets on 127.0.0.1:<port> are touched.
+# The production wifihaven-api docker container does not listen on those
+# loopback ports, so this script cannot affect production. We resolve the
+# owning pid via `ss -lntpH "sport = :<port>"` and SIGTERM, then SIGKILL as
+# a backstop if the port is still bound a few seconds later.
+#
+# Usage
+# -----
+#   scripts/vm/lib/free-test-ports.sh <port> [<port>...]
+#   scripts/vm/lib/free-test-ports.sh --from-port-base   # uses $WH_PORT_BASE
+#
+# Linux-only (uses ss(8)). Idempotent. Echoes each kill so the kill record
+# appears in the workflow log (operator-visible signal per AGENTS.md
+# §instrument-new-functionality — a CI script's operator surface is the log).
 set -euo pipefail
-exit 0
+
+log() { printf '[free-test-ports] %s\n' "$*" >&2; }
+
+pids_on_port() {
+  local port="$1"
+  ss -lntpH "sport = :${port}" 2>/dev/null \
+    | grep -oE 'pid=[0-9]+' \
+    | awk -F= '{print $2}' \
+    | sort -u
+}
+
+free_one_port() {
+  local port="$1"
+  local pids
+  pids="$(pids_on_port "${port}" || true)"
+  if [[ -z "${pids}" ]]; then
+    return 0
+  fi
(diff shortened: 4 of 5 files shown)


sameerparekh:

### Prior findings status (vs. `90cd2c3e`)

| # | Finding | Status |
|---|---|---|
| (prior NIT 1) `DST_IP6` placement | ADDRESSED in `90cd2c3e` | unchanged |
| (prior NIT 2) AI footer in PR body | NOT-ADDRESSED — non-blocking, reviewer's call to leave | unchanged |

### New findings (delta `90cd2c3e...f301ab6a`)

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:741` and `:771` — the `if not hname then hname = …_hit_host end` promotion is duplicated once per callsite. The two callers diverge on the variable name (`eb_hit_host` vs `bl_hit_host`), so collapsing them would buy little, but a single helper `attribute_from_nft_hit(hname, hit_host)` would let a future third caller stay drift-proof. Not worth doing today.

### Dimension walk on the delta

1. **Single source of truth** — promotion is appended inside each existing `not check_ea_carveout(...)` branch, after the `reason` assignment that's already duplicated there; no new computation introduced, just an attribution side-effect on the matched host the branch already names. Critically, `check_ea_carveout` is called with `hname` still nil, so the carve-out decision is unaffected — only the post-decision attribution changes. ✅
2. **Testing integrity** — TDD red→green visible: `c80ab7f0` adds the three failing tests; `f301ab6a` ships the fix. Tests assert on `ev.host.type` and `ev.host.value` directly; v4 sibling pinned alongside v6 to prevent family drift. 43/43 `handle_flow` tests pass. ✅
3. **Architectural invariants** — wire shape unchanged: `build_event` already emits `host.type=fqdn` whenever a hostname is supplied; the API/SPA already handle this discriminant. No new top-level field, no decision logic added to the router, no `/tmp` writes. ✅
4. **Wire / back-compat** — older APIs already accept `host.type=fqdn` from the DNS-attributed path; the nft-fallback path now produces the same shape it could previously produce. No additive field needed. ✅
5–8. **Migrations / metrics / scope / security** — N/A for this delta.

VERDICT: APPROVE

The delta promotes the matched `eb_<host>` / `bl_<host>` from the nft fallback to be the event's FQDN attribution, not just the reason-string label. Honest attribution (the matched apex is what dnsmasq resolved to populate the set), strictly better than the bare v4/v6 literal previously emitted, and wire-compatible — the API/SPA already accept `host.type=fqdn` from this path's DNS-attributed sibling.

sameerparekh:

### Prior findings status (vs. `90cd2c3e`)

| # | Finding | Status |
|---|---|---|
| (prior NIT 1) `DST_IP6` placement | ADDRESSED in `90cd2c3e` | unchanged |
| (prior NIT 2) AI footer in PR body | NOT-ADDRESSED — non-blocking, reviewer's call to leave | unchanged |

### New findings (delta `90cd2c3e...f301ab6a`)

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:741` and `:771` — the `if not hname then hname = …_hit_host end` promotion is duplicated once per callsite. The two callers diverge on the variable name (`eb_hit_host` vs `bl_hit_host`), so collapsing them would buy little, but a single helper `attribute_from_nft_hit(hname, hit_host)` would let a future third caller stay drift-proof. Not worth doing today.

### Dimension walk on the delta

1. **Single source of truth** — promotion is appended inside each existing `not check_ea_carveout(...)` branch, after the `reason` assignment that's already duplicated there; no new computation introduced, just an attribution side-effect on the matched host the branch already names. Critically, `check_ea_carveout` is called with `hname` still nil, so the carve-out decision is unaffected — only the post-decision attribution changes. ✅
2. **Testing integrity** — TDD red→green visible: `c80ab7f0` adds the three failing tests; `f301ab6a` ships the fix. Tests assert on `ev.host.type` and `ev.host.value` directly; v4 sibling pinned alongside v6 to prevent family drift. 43/43 `handle_flow` tests pass. ✅
3. **Architectural invariants** — wire shape unchanged: `build_event` already emits `host.type=fqdn` whenever a hostname is supplied; the API/SPA already handle this discriminant. No new top-level field, no decision logic added to the router, no `/tmp` writes. ✅
4. **Wire / back-compat** — older APIs already accept `host.type=fqdn` from the DNS-attributed path; the nft-fallback path now produces the same shape it could previously produce. No additive field needed. ✅
5–8. **Migrations / metrics / scope / security** — N/A for this delta.

VERDICT: APPROVE

The delta promotes the matched `eb_<host>` / `bl_<host>` from the nft fallback to be the event's FQDN attribution, not just the reason-string label. Honest attribution (the matched apex is what dnsmasq resolved to populate the set), strictly better than the bare v4/v6 literal previously emitted, and wire-compatible — the API/SPA already accept `host.type=fqdn` from this path's DNS-attributed sibling.

Issue #1516, opened 2026-06-06
New per-(profile, app, date) daily usage rollup (schema-only migration first)


sameerparekh:
Sub-issue of #1510 (App-level usage aggregation). Part of **Usage & Screen-Time**.

Make the **graph-serving** rollups/series reflect the app-aggregated, gap-bridged engaged minutes, so the graphs reconcile with the app cap and the per-app daily rollup.

## Why
The graph path is inconsistent with the cap today:
- `traffic_hourly`/`traffic_daily` (V38) store `active_seconds` — a bucket quantity, **not** engaged session-stitch time.
- `usage-by-app` (`buildUsageByApp`, UsageRoutes.scala) sums **per-host** `proportionalHostSeconds` into each app *after* stitching, so it never bridges an idle gap across the app's hosts and won't match the gap-bridged cap aggregate.

This is the #1492 "graphs reconcile with the headline total" invariant, extended to the **app** dimension.

## Ask
- Per-app time series (hourly/daily decomposition) derives from app-aggregated spans (`appSecondsForProfile`'s span form, hour-clipped via `secondsByLocalHour`) so summing the series equals the per-app daily rollup (sub-issue) and the cap aggregate.
- Where rollup tables back the series, carry app-aggregated engaged seconds (not `active_seconds`) for the per-app surface.
- Reconciliation documented: profile total ⇄ per-app rollup ⇄ per-app series.

## Tests (TDD)
- Hour-clipped per-app series sums to the per-app daily engaged total.
- A multi-host app with cross-host gap-bridged usage shows the bridged minutes in the series, matching the cap.

sameerparekh:
Schema PR #1548; adoption to follow.

sameerparekh:
Schema landed via PR #1548 (merged). The per-(profile,app,date) rollup table now exists; the adoption PR that writes/reads it (composing on #1514's appSecondsForProfile) is the App-Centric aggregation follow-up (#1517 / rollup-builder).

Issue #1515, opened 2026-06-06
Enforcement: per-app cap evaluation; block all app hosts together; SiteTimeLimit→per-app shape


sameerparekh:
Sub-issue of #1510 (App-level usage aggregation). Part of **Usage & Screen-Time**.

Add a per-`(profile, app, date)` daily usage rollup for the **gap-bridged engaged minutes**, so per-app daily tracking and the app cap are O(1) at read time instead of re-aggregating `traffic_reports`.

## Why
`time_used_daily` (V43) caches the session-stitch **profile total** (today-only, seconds, `rolled_through` watermark + live tail). There is no equivalent for the **per-app** engaged minutes — the app cap and per-app daily graphs re-aggregate raw presence each read.

## Ask
New table, mirroring `time_used_daily`'s design:

```sql
CREATE TABLE app_time_used_daily (
  profile_id     BIGINT      NOT NULL REFERENCES profiles(id) ON DELETE CASCADE,
  app_id         BIGINT      NOT NULL REFERENCES apps(id) ON DELETE CASCADE,
  date           DATE        NOT NULL,
  used_seconds   BIGINT      NOT NULL,            -- app-aggregated, gap-bridged
  rolled_through TIMESTAMPTZ NOT NULL,
  rolled_at      TIMESTAMPTZ NOT NULL DEFAULT NOW(),
  PRIMARY KEY (profile_id, app_id, date)
);
```

Read = `rolled_seconds + live_aggregate(presence where period_start >= rolled_through)` using `appSecondsForProfile`, identical watermark/tail logic to the profile total so cached and live reconcile exactly. Same wholesale-invalidation hook as `time_used_daily` (heartbeat filter / daily-reset boundary changes).

## Constraints
- **Migration-isolation (AGENTS.md):** this PR is **schema-only** (`V<n>__app_time_used_daily.sql` + docs). The rollup-writer/read-path code adoption lands in a follow-up PR. The new table is empty at migration time (catalog-only, safe at prod scale).
- Add the writer to the existing `TimeUsedRollupJob` cadence; emit a metric for the new rollup per the instrumentation rule, with a dashboard panel.

## Tests (TDD, in the adoption PR)
- Rolled+tail per-app total equals the all-live `appSecondsForProfile` across the watermark boundary, for `Sum` and `Dedup`.

sameerparekh:
[...]

PR #1675, opened 2026-06-12
fix(#1666): suppress phantom engagement on real-app host-sets (per-app time-limit inflation)

### LINKED ISSUE #1666: bug(prod): Khan Academy per-app time limit attribution inflated by phantom engagement (sibling to #1629, not infra-host class)

Connection event for `www.khanacademy.org` from Prima iPad shows reason `blocked (household)` (= SPA label for `ExtraBlocked`). Operator did NOT add khanacademy.org to any block list. The block came from an IP-collision: the resolved IP for Khan Academy is also in some other host's `eb_*` ipset.

## How this differs from #1636

#1636 found Google GFE pool IP overlap: blocking YouTube also collaterally blocks OAuth (same `216.239.32-38.223` shared by both). The fix landed in #1660/#1661 (PRs #1665, #1664 still open — operator merge button).

Khan Academy is on **Fastly** (`151.101.1.42` per #1626 router-state evidence), NOT Google GFE. So #1660's fix doesn't help here. Whichever host's `eb_*` is collaterally catching khanacademy.org is a DIFFERENT shared-CDN-pool entry — likely an operator-blocked site that's also hosted on Fastly.

## What we now have (post-#1656)

PR #1656 ("label extraBlocked events with the matched eb_<host> rule") attached the matched rule name to `connection_events`. The data is in the DB. The SPA renders `reason_class=ExtraBlocked` as "blocked (household)" but doesn't surface WHICH `eb_<host>` rule actually fired. #1640 (Todo) is the SPA-side surface for that.

## Investigation

1. **Find the offending rule** — for the specific connection event from Prima iPad to `www.khanacademy.org` ~14 minutes before this issue was filed:
   ```sql
   SELECT mac, host, dst_ip, ts, reason, reason_meta
     FROM connection_events
    WHERE host = 'www.khanacademy.org'
      AND ts > NOW() - INTERVAL '30 minutes'
    ORDER BY ts DESC LIMIT 10;
   ```
   The `reason_meta` (or whichever column #1656 added) should carry the matched rule, e.g. `eb_<mac>_<some-blocked-host>`.

2. **Identify the colliding host** — once you have the matched rule's host, query global_blocks / extraBlocked sources to confirm it was operator-intended.

3. **Resolve the IP collision** — same playbook as #1660:
   - If the colliding host is on a shared Fastly pool, the host-set's apex needs to be narrowed (template fix per #1661's authoring principle).
   - If the colliding host is operator-added and over-broad, either narrow it or document the trade-off.

4. **Surface the rule in the SPA** — #1640 is the obvious next step (so operators can self-diagnose without spelunking the DB).

## Acceptance

- Pinned which host's `eb_*` rule collaterally caught khanacademy.org.
- Either: narrowed the offending host-set (template fix, per #1661); OR removed the over-broad operator entry; OR documented the unavoidable collateral with a follow-up #1640-style SPA surface.
- Regression test: a feature test asserting Khan Academy isn't dropped when no rule names it directly, given a fixture with Fastly IP collisions in nearby rules.

## Related

- #1636 — Google OAuth collateral (Google GFE pool)
- #1648 — design: shared-IP-pool blocking strategy (the umbrella for this class)
- #1660/#1661 (PRs #1665, #1664) — YouTube/GFE-specific fix, open for merge
- #1640 — SPA surface for matched-rule attribution on connection events
- #1656 (merged) — labels the matched rule on connection_events (the data exists)

sameerparekh:
Closes #1666.

## Symptom

Prima iPad's Khan Academy per-app counter reported 33/30 min today while she barely used Khan — the spike accumulated 01:00–05:00 PDT while she was asleep. `connection_events` shows zero Khan rows before 14:00 UTC; the entire 33 min was synthetic, from `Presence.appSecondsForProfile` gap-stitching sparse background DNS-driven `traffic_reports` rows on the app's host-set.

## Root cause

`Presence.appSpansForProfile` (api/src/presence/Presence.scala:601) stitches the union of an app's host-set as one stream per device on the `effectiveGap = max(continuationSeconds, 2·R)` idle gap (default 120s). Two reinforcing effects combine:

1. **Row-as-span (design §4.1):** every non-heartbeat `traffic_reports` row contributes its full `[period_start, period_start + period_seconds]` window (typically 60s) as evidence — deliberately, to fight the activeSeconds undercount.
2. **Attribution beats suppression (#1506):** any row whose host matches an active app's host-set is never classified as a heartbeat — neither the InfraHosts identity check nor the byte-floor filter applies. By design, so an app's own low-byte CDN/asset poll counts.

Combined: a sparse keepalive cadence on `khanacademy.org` (e.g. one DNS-driven ~80-byte poll every 60–90s overnight) produces one 60s-wide span per row, gap ≤120s bridges every adjacent pair, and the result is one continuous synthetic span for the whole overnight stretch. Khan is not on InfraHosts, so the #1503/#1525 suppression path doesn't reach it (and the #1629 direction extends *InfraHosts* coverage — orthogonal to this case).

## Fix — per-session activity anchor

Each stitched per-(mac, app) session must contain at least one **anchor** row with `bytes >= Presence.AppSessionAnchorBytes` (256). A TCP keepalive is ~60 bytes and an HTTP/2 PING is a few hundred at most, so 256 admits any substantive request (a 304, a JSON poll, an asset fetch) while a pure-keepalive cadence on an app's host produces no anchor and the session is dropped.

The check is **per-session, not per-row**, so attribution-beats-suppression (#1506) is preserved at the row level: an app-attributed low-byte row still contributes to a session anchored by another substantive row in the same session. Only sessions with *no* anchor at all are dropped.

Implemented inside `Presence.appSpansForProfile`, the canonical per-app primitive (#1514/#1532). Every consumer — the per-app rollup (#1516), the per-app graph (#1517), the per-app cap (#1515), the dashboard — inherits the fix from one place. Single-source-of-truth preserved.

## Rejected alternatives

- **(B) Snap quiet windows (paused/scheduled) out of stitched spans.** Doesn't close this case: Prima wasn't under any schedule overnight; dropping only the paused windows still credits normal idle.
- **(C) Per-app heartbeat predicate (extend InfraHosts to apps).** Too invasive; conflicts with the #1506 attribution-beats-suppression contract; forces operators to curate per-app keepalive lists.

## No new metric / dashboard

The fix is a pure threshold inside existing code (per AGENTS.md §instrument-new-functionality escape hatch: *"If the fix doesn't add a new suppression step (e.g. pure threshold inside existing code) and there's no new measurable event, skip this phase but say so in the PR body."*). Operators can observe the fix via the existing per-app rollup tables and `/api/time/heartbeat-explain` surface; introducing a counter would require threading effects through what is otherwise pure presence code across multiple callers.

## Schema

No migration. No wire-shape change. Pure server-side computation.

## Test plan

- [x] RED test (`test(#1666): pin phantom-engagement inflation on real-app host-set`): two phantom-stretch fixtures + two positive controls. Phantom fixtures fail on `main` reporting 239 min / 121 min for the asleep Khan pattern; pass after the fix reporting 0 / 1 min.
- [x] Positive control 1 (8-min real session, 50KB anchor + 200B follow-ups, alternating apex/asset): credits the full 510s = 8 min.
- [x] Positive control 2 (anchor-retains-tail: anchor at 0s + low-byte poll at 120s + anchor at 240s): credits the full [0,300) = 300s.
- [x] All 79 PresenceSpec tests green.
- [x] `mill api.test` green (full feature suite, 76+84 tests).
- [x] `mill shared.test` green.
- [x] `mill scalafmtCheckAll` clean.

Sibling to #1629 (infra-host phantom on Kids profile) — that fix targets InfraHosts coverage and does *not* close this case.

Co-Authored-By: Claude <noreply@anthropic.com>


sameerparekh:
Independent review pass per `docs/pr-review-checklist.md`.

BLOCKERS
(none)

SHOULD-FIX
1. `finding.md:1` — investigation scratch file checked in at the repo root. The PR body already captures the root-cause analysis; `finding.md` duplicates it and clutters the repo root alongside `AGENTS.md` / `CLAUDE.md` / `README.md`. Drop it before merge (or move the bug-class write-up into `docs/architecture.md` / a comment in `Presence.scala` if any of it should persist).
2. `api/src/presence/Presence.scala:631-657` — the §metrics-new-functionality escape hatch is borderline here. This is a new session-suppression step, and the only operator-facing way to verify it on prod is to read the per-app rollup tables or `/api/time/heartbeat-explain`. A cheap unlabelled `presence_app_sessions_dropped_total` counter — incremented from the one call site that observes `appSpansForProfile` in a ZIO context (e.g. `TimeStatusService.appSecondsByApp`, by diffing pre- vs. post-anchor session count via a sibling helper) — would let operators rate-alert on "sudden spike in dropped sessions = anchor threshold too aggressive" or "no drops at all = phantom bug back." Threading effects through pure `Presence.*` is invasive; a sibling helper that returns `(spans, droppedCount)` keeps the primitive pure.

NITS
1. `api/src/presence/Presence.scala:642-657` — for each device, the anchor filter scans `rs` once per stitched session: `O(S × R)` per device. Trivial today (sessions per device per day are small), but pre-extracting `val anchorSpans = rs.filter(_.bytes >= AppSessionAnchorBytes).map(spanOf)` once and then `anchorSpans.exists(...)` per session would be marginally cleaner and avoid re-doing the byte comparison.
2. `api/src/presence/Presence.scala:71` — `AppSessionAnchorBytes = 256L` is hard-coded. Consistent with `DefaultContinuationSeconds` and other tuning constants nearby; no action needed, just noting it'd be the place to lift to `household_settings` if a future operator wants to tune it without a code release.
3. `api/test/src/unit/PresenceSpec.scala:1330-1342` — the 4h fixture uses a real Khan host-set, which is good for documentation-of-intent. Minor: the test depends on `appRow`'s default `bytes = 1_000_000L`, overridden to `80L` here; one inline comment confirming "80 bytes < AppSessionAnchorBytes (256) so no row anchors" would make the test self-documenting against a future threshold change.

Prior findings status table: N/A (first run on this PR).

VERDICT: APPROVE

Pure server-side fix that lands the phantom-suppression guard at the canonical per-app presence primitive (`Presence.appSpansForProfile`) so every consumer (rollup #1516, per-app graph #1517, per-app cap #1515) inherits it — single-source-of-truth preserved. TDD red→green is visible in the two-commit history (failing test pinned 239 min on the Khan fixture; fix collapses it to 0 with positive controls still passing). No wire / schema / migration / router-agent changes; the row-level `#1506` attribution-beats-suppression contract is preserved because the anchor check is per-session, not per-row. The two SHOULD-FIX items are merge-mechanical (drop `finding.md`) and observability gap-filling (a small drop counter), not correctness issues.

sameerparekh:
Re-run after `f2cf33e` (drop `finding.md` + test self-documenting comment).

## Prior findings status

| Prior finding | Status | Note |
|---|---|---|
| SHOULD-FIX 1 — `finding.md` checked in at repo root | **ADDRESSED** | removed in `85189395` |
| SHOULD-FIX 2 — no `presence_app_sessions_dropped_total` counter / dashboard | **NOT-ADDRESSED (deferred)** | Tracked as [#1676](https://github.com/wifihaven/wifihaven/issues/1676). Threading a counter through pure `Presence.*` requires a sibling `appSpansForProfileWithDropCount` helper plus changes across `TimeStatusService.appSecondsByApp`'s call chain. The AGENTS.md §instrument-new-functionality escape hatch is explicitly invoked in the PR body; the deferral is judged acceptable because the fix correctness is testable today and the follow-up issue captures the observability gap with a concrete design. Reviewer accepts deferral as justified — not a regression risk. |
| NIT 1 — pre-extract anchor spans for `O(R)` vs `O(S×R)` | **NOT-ADDRESSED** | Accepted; sessions/day is small, optimization not load-bearing. |
| NIT 2 — `AppSessionAnchorBytes = 256L` hard-coded | **NOT-ADDRESSED** | Accepted; consistent with neighboring tuning constants. |
| NIT 3 — self-documenting inline comment on 4h-phantom fixture | **ADDRESSED** | added in `f2cf33e1` |

## New findings on `3650762c...f2cf33e1`

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
(none — the delta is `git rm finding.md` + a 2-line comment.)

VERDICT: APPROVE

Delta is purely review-mechanical: the investigation scratch is removed and the regression test gets a self-documenting threshold comment. No production behavior change since the GREEN commit; all merge-gating criteria (single-source-of-truth, TDD red→green, no wire/schema/router changes, attribution-beats-suppression contract preserved) remain satisfied. SHOULD-FIX 2 (observability counter) is deferred to [#1676](https://github.com/wifihaven/wifihaven/issues/1676) with the AGENTS.md escape hatch explicitly invoked.

Issue #1677, opened 2026-06-12
e2e (Gate 2): add v6 attribution scenario (AAAA fixture in wifihaven.net + scenarios_fake test)


sameerparekh:
## What

Add a Gate 2 (`scripts/e2e/scenarios_fake/`) scenario that proves end-to-end v6 host attribution on the real router VM against real DNS, mirroring the v4 [#1351](https://github.com/wifihaven/wifihaven/issues/1351) attribution suite. This is the CD-tier follow-up to PR #1673 (#1668 fix), which closed the attribution gap in the parser and pinned it with busted integration tests; this issue adds the VM-level proof.

## Why

PR #1673 fixed [#1668](https://github.com/wifihaven/wifihaven/issues/1668) (`dns_log.parse_reply_line` was dropping every AAAA answer, so v6 destinations always landed as bare literals in `connection_events`). It ships with three busted integration tests in `openwrt/test/conntrack_spec.lua` that wire a **real** `dns_log` cache through `handle_flow` and pin the pipeline:

    dnsmasq AAAA log → dns_log.ingest_line → cache.lookup → handle_flow → host.type=fqdn

That's strong coverage at the module-integration level (same tier as the [#1344](https://github.com/wifihaven/wifihaven/issues/1344) "end-to-end" tests right above the new ones). But it does NOT exercise the full router VM stack — actual dnsmasq emitting actual AAAA log lines under `--log-queries=extra`, actual `dns-tail` parsing them and flushing to `paths.dns_cache`, actual `conntrack -E NEW` firing on a real v6 SYN, actual ETag-polled snapshot, actual event POST to the fake API. A future regression at the *configuration* / *integration* layer (dnsmasq log format change, `--log-queries` flag, dns-tail flush, v6 routing in the qemu bridges) wouldn't be caught by the busted tier.

## Scope

Same shape as the v4 attribution case in `scripts/e2e/scenarios_fake/test_cname_direct_requery.py` (specifically the G3 attribution test). The smallest functional Gate 2 scenario:

1. Client resolves a wifihaven.net hostname that has an **AAAA** record.
2. Client initiates a v6 flow (a SYN is enough — `nc -6 -w 2` or `curl -6 --connect-timeout 2`; the origin doesn't need to be reachable, just as the v4 G3 xfails on the unrouteable TEST-NET leaf).
[...]

PR #1682, opened 2026-06-12
fix(#1680): bring up IPv6 across the Gate 2 qemu LAN bridge

### LINKED ISSUE #1680: e2e (Gate 2): bring up IPv6 between client and router VMs (preparatory for #1677)
## What

Wire IPv6 across the qemu Gate 2 harness so the client VM can originate routable v6 flows toward the router VM. Specifically:

- Router LAN issues IPv6 RAs (odhcpd default behaviour, but currently not exercised) advertising a ULA prefix and itself as default route.
- Client picks up SLAAC + default route on `eth0` (Alpine default kernel accepts RA; verify the busybox/udhcpc setup doesn't override).
- Router nftables forward chain accepts/forwards v6 LAN→WAN flows (mirror the existing v4 forward path). Drops downstream of the WAN are fine — the conntrack `NEW` event we care about for attribution fires on the LAN forward.

Out of scope: upstream v6 to the documentation address (qemu SLIRP is v4-only). The attribution scenario only needs the conntrack `NEW` event, which fires as soon as the SYN crosses the forward chain.

## Why

Investigation for #1677 (the v6 attribution Gate 2 scenario, CD follow-up to PR #1673) found:

- `scripts/vm/`, `scripts/vm/uci-defaults/99-wifihaven`, `scripts/e2e/lib/vm.py`, `scripts/vm/build-client-base.sh`, and `scripts/vm/build-router-image.sh` contain **zero** IPv6 configuration — no RA, no SLAAC, no v6 firewall, no v6 default route on the client.
- Without a v6 default route the client kernel returns \"Network unreachable\" before emitting the SYN to \`2001:db8::10\`, so no conntrack event fires and the attribution path is never exercised.

PR #1673 already proved the parser + attribution pipeline at the busted integration tier (`openwrt/test/conntrack_spec.lua`). The CD-tier proof (#1677) needs this preparatory work first.

## Tasks

1. Enable RAs on the router LAN in `scripts/vm/uci-defaults/99-wifihaven` (or a sibling) — confirm `network.lan.ip6assign` / odhcpd `ra='server'` / `dhcpv6='server'` produce a ULA on the bridge.
2. Add v6 forward rules — replicate the LAN→WAN forward accept for `ip6` family. The existing `block-dns-leak` rule should also gain a v6 sibling.
3. Confirm Alpine client picks up the prefix on boot (`ip -6 addr show eth0`, `ip -6 route`). If `udhcpc` strips RA, fix the client base.
4. Add a sanity assertion to the orchestrator (`scripts/e2e/lib/vm.py` or a conftest fixture) that the client has a v6 default route once `client_up` returns.

## Acceptance

From inside the running client VM:

\`\`\`
ip -6 route | grep '^default'        # router LL via eth0
ping6 -c1 <router LAN ULA>            # succeeds
nc -6 -w 2 2001:db8::10 80            # returns (RST or timeout fine), conntrack -E NEW fires on the router
\`\`\`

\`conntrack -E NEW -f ipv6\` on the router shows the flow.

## Related

- Blocks #1677.
- PR #1673 — parser fix + busted-tier coverage that this CD-tier work follows.

sameerparekh:
## Summary

Wires IPv6 across the Gate 2 qemu harness so the client VM can originate routable v6 flows toward the router VM — preparatory work for the v6 attribution scenario in #1677 (CD-tier follow-up to PR #1673 / issue #1668).

Out of scope: upstream v6 reachability (qemu SLIRP WAN is v4-only). The attribution path only needs the conntrack \`NEW\` event, which fires as soon as the SYN crosses the router's FORWARD chain.

## Changes

**Router** (\`scripts/vm/uci-defaults/99-wifihaven\`):
- \`network.globals.ula_prefix='fdaa:bbbb:cccc::/48'\` — deterministic ULA so tests and orchestrator can pin it.
- \`network.lan.ip6assign='60'\` — LAN takes the first /60 of the ULA.
- \`dhcp.lan.ra='server'\` / \`dhcpv6='server'\` / \`ra_default='1'\` — odhcpd advertises the prefix and the router as v6 default. \`ra_default=1\` is required because the router has no v6 upstream; without it odhcpd suppresses the RA's Router Lifetime and the client never installs a v6 default route.
- New hotplug script \`/etc/hotplug.d/iface/99-wifihaven-v6-default\` does \`ip -6 route replace default dev eth1\` on WAN ifup. The next-hop is bogus on purpose — the routing decision picks it, the packet enters FORWARD (conntrack \`NEW\` fires), and ND fails silently. That's the FORWARD-chain transit #1677 needs.

Stock OpenWRT fw4 generates v4+v6 nft rules for any user rule that omits \`family\`, so the existing \`block-dns-leak\` rule covers v6 unchanged — no v6 sibling needed.

**Client side**: no change required. Alpine's default \`net.ipv6.conf.eth0.accept_ra=1\` picks up the RA and installs the default route; SLAAC produces the global ULA address. The cloud-init \`eth0 inet dhcp\` stanza is v4-only and does not disable v6.

**Orchestrator** (\`scripts/e2e/lib/vm.py\`):
- \`_wait_for_client_lan\` now chains into \`_assert_v6_link_local_ready\`, which fails loudly with \`ip -6 addr/route\` diagnostics if the client lacks a v6 default route or a \`fdaa:bbbb:cccc\` address. Regression gate: any future change that strips RA/SLAAC fails every Gate 2 scenario early with a clear message rather than confusing late timeouts in #1677.

**Scenario** (\`scripts/e2e/scenarios_fake/test_v6_link_local.py\`):
1. Client has a v6 default route on \`eth0\`.
2. Client has a SLAAC global address in \`fdaa:bbbb:cccc::/48\`.
3. A v6 SYN from client to \`2001:db8::10\` fires conntrack \`NEW\` on the router (proves the FORWARD chain executes for v6).

Marker + \`--only v6-link-local\` switch registered.

## TDD

Two commits: a RED commit with the test only against current main (the router image had zero v6 config, so all three asserts fail at \`Network unreachable\` before any v6 packet is emitted), then the GREEN commit with the uci-defaults and orchestrator wiring. KVM not available in the authoring env, so the green run happens on the self-hosted KVM runner via \`e2e-vm-fake.yml\`.

## Acceptance (from #1680)

From the running client VM, after this PR:

\`\`\`
ip -6 route | grep '^default'         # router LL via eth0 — covered by test 1
ping6 -c1 fdaa:bbbb:cccc::1            # router LAN ULA — covered by SLAAC + default route
nc -6 -w 2 2001:db8::10 80             # conntrack -E NEW fires on router — covered by test 3
\`\`\`

## Test plan

- [ ] CI \`e2e-vm-fake\` passes (\`test_v6_link_local\` runs and goes green on the self-hosted KVM runner).
- [ ] No regression on the other Gate 2 scenarios — \`_assert_v6_link_local_ready\` adds a new precondition; if the v6 bring-up is racy in practice, expect to see it surface in any scenario that uses the \`client\` fixture.

Closes #1680. Unblocks #1677.

### DIFF
--- a/scripts/e2e-vm.sh
+++ b/scripts/e2e-vm.sh
@@ -109,10 +109,12 @@ case "${ONLY}" in
                 MARK="install_health" ;;
   sni-attribution|sni_attribution)
                 MARK="sni_attribution" ;;
+  v6-link-local|v6_link_local)
+                MARK="v6_link_local" ;;
   *) echo "unknown --only: ${ONLY}" >&2
      echo "valid: enrollment, allowed-browsing, blocked-domain, daily-limit, usage, blocked-page," >&2
      echo "       pause, extra-blocked, schedule, time-limit, reassignment, unknown-device," >&2
-     echo "       install-health, sni-attribution" >&2
+     echo "       install-health, sni-attribution, v6-link-local" >&2
      exit 2 ;;
 esac
 
--- a/scripts/e2e/lib/vm.py
+++ b/scripts/e2e/lib/vm.py
@@ -299,10 +299,51 @@ def _wait_for_client_lan(client: Client, *, timeout_s: float = 60, interval_s: f
             _time.sleep(interval_s)
             continue
         if res.returncode == 0:
+            _assert_v6_link_local_ready(client)
             return
         last = (res.stderr or res.stdout or "").strip()
         _time.sleep(interval_s)
     log.warning("client %s LAN/DNS not ready after %ss (last: %r)", client.name, timeout_s, last)
+
+
+# Router ULA, kept in sync with scripts/vm/uci-defaults/99-wifihaven (#1680).
+ROUTER_ULA_PREFIX = "fdaa:bbbb:cccc"
+
+
+def _assert_v6_link_local_ready(client: Client, *, timeout_s: float = 30, interval_s: float = 2.0) -> None:
+    """Regression gate (#1680): once the LAN is up, the client must also have
+    a v6 default route and a global-scope ULA address.
+
+    If a future change strips RA/SLAAC (e.g. odhcpd disabled, ULA prefix
+    cleared, accept_ra=0 on the client base), every Gate 2 scenario gets a
+    clear, early failure here rather than a confusing late timeout in the
+    v6 attribution suite (#1677)."""
+    import time as _time
+    deadline = _time.monotonic() + timeout_s
+    probe_cmd = [
+        "sh", "-c",
+        "ip -6 route show default 2>/dev/null | grep -q 'dev eth0' && "
+        f"ip -6 addr show eth0 scope global 2>/dev/null | grep -q '{ROUTER_ULA_PREFIX}'",
+    ]
+    last_v6_state = ""
+    while _time.monotonic() < deadline:
+        res = client_exec(client, probe_cmd, timeout=10, check=False)
+        if res.returncode == 0:
+            return
+        diag = client_exec(
+            client,
+            ["sh", "-c", "ip -6 route; echo ---; ip -6 addr show eth0"],
+            timeout=10, check=False,
+        )
+        last_v6_state = (diag.stdout or "").strip()
+        _time.sleep(interval_s)
+    raise RuntimeError(
+        f"client {client.name} missing IPv6 default route or ULA address "
+        f"in {ROUTER_ULA_PREFIX}::/48 after {timeout_s}s. The qemu LAN bridge "
+        f"is not carrying RA/SLAAC — see scripts/vm/uci-defaults/99-wifihaven "
+        f"and #1680.\n"
+        f"--- client ip -6 addr/route ---\n{last_v6_state}"
+    )
 
 
 def client_down(name: str = "client1") -> Result:
--- a/scripts/e2e/pytest.ini
+++ b/scripts/e2e/pytest.ini
@@ -19,6 +19,7 @@ markers =
     nflog: #1122 — forward-drop visibility via nflog (per-MAC drop comment + agent emission)
     cname_direct_requery: suite G (#1351) — ea_/eb_ population for directly-queried CNAME targets
     sni_attribution: suite I (#573) — SNI-only hostname attribution via the wifihaven-sni-tail sidecar
+    v6_link_local: #1680 — IPv6 LAN bring-up (ULA SLAAC + v6 FORWARD path), prep for #1677
     global_policy: suite H (#1460) — @global_allow / @global_block fleet-wide enforcement (gate for #1321)
     smoke: fast subset across suites (one short case each) for PR gating
 log_cli = true
--- a/scripts/e2e/scenarios_fake/test_v6_link_local.py
+++ b/scripts/e2e/scenarios_fake/test_v6_link_local.py
@@ -1,12 +1,12 @@
 """Gate 2 v6 link-local bring-up (#1680, prep for #1677).
 
-Confirms the qemu LAN bridge carries IPv6 between the client and router VMs:
-
-- Router LAN advertises a ULA prefix via odhcpd RAs.
-- Client picks up SLAAC + a v6 default route via the router's link-local.
-- A v6 SYN from client to a documentation address traverses the router's
-  FORWARD chain and fires a conntrack NEW event (the prerequisite for v6
-  hostname attribution in #1677).
+Confirms the qemu LAN bridge carries IPv6 between the client and router VMs
+all the way through the router's FORWARD chain. The `client` fixture's
+`_assert_v6_link_local_ready` (scripts/e2e/lib/vm.py) already gates that
+SLAAC + a v6 default route landed on the client; the test below covers
+what the fixture can't — that a v6 SYN from the client actually transits
+the router's FORWARD chain and fires conntrack NEW, which is the
+prerequisite for v6 hostname attribution in #1677.
 
 The router can't actually forward the packet anywhere — qemu SLIRP is
 v4-only — but the FORWARD-chain transit alone is what #1677 needs.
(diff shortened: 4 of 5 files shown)


sameerparekh:
Independent review pass for #1682 (first run) against `docs/pr-review-checklist.md`.

## BLOCKERS

(none)

## SHOULD-FIX

1. **`scripts/e2e/scenarios_fake/test_v6_link_local.py:22`** — `ROUTER_ULA_PREFIX = "fdaa:bbbb:cccc"` is re-declared verbatim alongside `scripts/e2e/lib/vm.py:309`'s identical constant, with a comment ("Must match the ULA in scripts/vm/uci-defaults/99-wifihaven.") that is exactly the "keep in sync" smell §1 of the checklist calls out. Collapse: `from lib.vm import ROUTER_ULA_PREFIX` in the test file. The uci-defaults shell literal is the third copy — harder to share across languages, but the two Python sources at minimum should converge.

2. **`scripts/e2e/scenarios_fake/test_v6_link_local.py:25-56`** — the first two scenario tests (`test_client_has_v6_default_route`, `test_client_has_ula_global_address`) are subsumed by `_assert_v6_link_local_ready` (`scripts/e2e/lib/vm.py:312`), which runs inside the `client` fixture (`_wait_for_client_lan` → `_assert_v6_link_local_ready`). Both preconditions are proven before either test body runs; if RA/SLAAC ever breaks, the fixture raises `RuntimeError` and these tests `ERROR` rather than `FAIL` with the messages they're authoring. Either (a) drop the first two test functions and keep `test_v6_syn_traverses_router_forward_chain` as the unique value-add, or (b) make them probe something the orchestrator gate doesn't (e.g. assert the prefix lands in the expected /64, not just /48; or that the default route survives a `wpa_cli reattach` / dhcp renew). As written they're dead weight that never gates a regression independently of the fixture.

## NITS

1. **`scripts/e2e/scenarios_fake/test_v6_link_local.py:62-67`** — `pkill -f 'conntrack -E -f ipv6'` relies on busybox `pkill` matching full argv (`-f`). Stock OpenWRT busybox supports it, but it's brittle to a build-config change. The preceding `echo $!` already returns the conntrack PID via the SSH stdout — capture it and `kill <pid>` instead.

2. **`scripts/vm/uci-defaults/99-wifihaven:124`** — `fdaa:bbbb:cccc::/48` is the third copy of the same constant (after `lib/vm.py` and the test). Cross-language duplication is hard to fully collapse, but a single-line comment pointing at `lib/vm.py:ROUTER_ULA_PREFIX` as the canonical Python mirror would at least make the link greppable for the next change.

3. **`scripts/e2e/scenarios_fake/test_v6_link_local.py:69-76`** — `grep -c '\\[NEW\\].*2001:db8::10'` against a backgrounded `conntrack -E` log is OK, but the SSH session that started `conntrack -E ... &` was a different command than the one running `pkill ... ; grep` — the backgrounded `conntrack` is daemonized into init. Works, but worth a comment so a future reader doesn't try to consolidate the two router_ssh calls and break the daemonization.

## §-by-§ summary

- **§1 SSOT** — SHOULD-FIX #1, #2. No re-derivation of an existing logic primitive, but constant + assertion duplication.
- **§2 Testing integrity** — RED (`0d493382`) → GREEN (`983f9fd7`) progression visible in history. No existing test weakened, no mocks introduced. SHOULD-FIX #2 above is the only test-integrity nuance.
- **§3 Architectural invariants** — Router agent unchanged; no decision logic added router-side. Snapshot unchanged. `/tmp/ct6.log` is a test-only short-lived capture (rm'd at start, pkilled at end) — not a production grow-writer.
- **§4 Wire contract** — no wire changes.
- **§5 Migrations** — none.
- **§6 Metrics** — test-only path; no new production code path requiring instrumentation.
- **§7 Scope / hygiene** — focused diff, no drive-bys, no AI mentions in PR title / body / commit subjects (`Co-Authored-By` trailer only).
- **§8 Security** — no new routes, no SQL, no credentials. The bogus v6 default route on the WAN is intentional and scoped to the qemu test image (`Out of scope: upstream v6` in the PR body).

## VERDICT: APPROVE

Two SHOULD-FIX items (constant duplication + redundant fixture/scenario coverage) and three NITs, but no BLOCKER. The ULA + odhcpd RA wiring is correct for OpenWRT 23.05; the hotplug-driven `ip -6 route replace default dev eth1` is the right place to seed the FORWARD-chain transit. The orchestrator regression gate is a good catch for #1677 and beyond. Recommend addressing SHOULD-FIX #1 (one-line import) before merge as the easy SSOT cleanup.

sameerparekh:
Independent review pass for #1682 (re-run after `3417e73f`) against `docs/pr-review-checklist.md`.

## Prior findings status (vs. `983f9fd7`)

| # | Class | Where | Status |
|---|---|---|---|
| 1 | SHOULD-FIX | `test_v6_link_local.py:22` ULA constant duplicated | **ADDRESSED** — now `from lib.vm import ROUTER_ULA_PREFIX` (`test_v6_link_local.py:18`); the local re-declaration is gone. |
| 2 | SHOULD-FIX | `test_v6_link_local.py:25-56` two scenario tests subsumed by fixture | **ADDRESSED** — `test_client_has_v6_default_route` and `test_client_has_ula_global_address` deleted; only `test_v6_syn_traverses_router_forward_chain` remains, which is the unique coverage the orchestrator gate can't provide. Module docstring rewritten to make the division of labor explicit. |
| 3 | NIT | `test_v6_link_local.py:71` brittle `pkill -f` | **ADDRESSED** — start now writes `$!` to `/tmp/ct6.pid` and reads it back into `ct_pid`, then `kill {ct_pid}` (`test_v6_link_local.py:36-41`, `:57`). PID-based, not argv-pattern-based. |
| 4 | NIT | `uci-defaults:124` cross-language constant duplication | **ADDRESSED** — comment now pins `scripts/e2e/lib/vm.py:ROUTER_ULA_PREFIX` as the Python-side mirror (`99-wifihaven:115-117`), making the link greppable. |
| 5 | NIT | Daemonized-conntrack pattern not commented | **ADDRESSED** — `test_v6_link_local.py:31-34` now explicitly warns that the start/probe must stay in separate `router_ssh` calls because the daemonization survives session close. |

All five prior findings cleared. No prior BLOCKER existed.

## New findings on the delta (`983f9fd7…3417e73f`)

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

1. **`scripts/e2e/scenarios_fake/test_v6_link_local.py:21`** — `assert ROUTER_ULA_PREFIX  # silence unused-import lints` is a workaround. Ruff's standard `# noqa: F401` on the import line, or moving the import into `__all__`, communicates intent more directly to future readers (and to linters that don't honor a runtime `assert` as use). Non-blocking — the comment makes the intent clear.

2. **`scripts/e2e/scenarios_fake/test_v6_link_local.py:36-39`** — `( conntrack ... & echo $! >/tmp/ct6.pid ); cat /tmp/ct6.pid` works because the subshell backgrounds conntrack and the PID write completes before the subshell exits. There is, however, a tiny race where the parent reads `/tmp/ct6.pid` before the subshell's `echo $!` flushes — harmless because `cat` would either return empty (and the `assert ct_pid.isdigit()` guards it) or the correct PID. Worth a one-line comment if a future reader wonders.

## §-by-§ summary

- **§1 SSOT** — both prior duplications collapsed. No new re-derivation introduced.
- **§2 Testing integrity** — no existing test weakened; the deleted scenario tests were vestigial-by-construction (asserted by the fixture). RED → GREEN remains visible from the earlier commits in this PR.
- **§3 Architectural invariants** — unchanged.
- **§4–6** — unchanged.
- **§7 Scope / hygiene** — focused, no AI mentions, `Co-Authored-By` trailer only.
- **§8 Security** — unchanged.

## VERDICT: APPROVE

All prior SHOULD-FIX + NITs addressed; the remaining two NITs are stylistic. The conntrack-PID handoff via `/tmp/ct6.pid` is the right shape, and the SSOT collapse on `ROUTER_ULA_PREFIX` is exactly the single-source-of-truth pattern AGENTS.md §single-source-of-truth wants. Ready to merge once CI passes on the self-hosted KVM runner.

Issue #1351, opened 2026-06-03
e2e: router test for directly-queried CNAME targets (ea_/eb_/bl_ + attribution) — gate the router image (#1344/#1345/#1346/#1348)


sameerparekh:
## Summary

Add a proper **router e2e test** (real VM router vs `scripts/e2e/fake-api`) that
exercises the directly-queried-CNAME-target path end to end and covers the
#1344 / #1345 / #1346 / #1348 fixes. It must run in the **live e2e suite** and
gate the **router image build**, not just busted unit tests.

## Why

#1344/#1345 (attribution label) and #1346 (kernel `ea_`/`eb_` set population)
are currently covered only by **Lua busted unit tests** that feed canned
`--log-queries=extra` lines into `dns_log` / `dns_tail_sets` in isolation. The
real failure mode is a *whole-system* one:

1. a client resolves a **branded host** whose CNAME chain ends at a CDN target,
   then
2. the client **re-queries the final CNAME target directly** (Apple devices do
   this), and
3. nftables must still drop (`eb_`) / carve out (`ea_`) the *directly-queried
   target's* resolved IP during a whole-MAC block.

Nothing in the e2e suite proves the dnsmasq → dns-tail → `nft add element` →
forward-drop pipeline actually closes that loop on a real router. The busted
tests can pass while the deployed sidecar wiring, the `nft -a list table`
parse, the lease→MAC mapping, or the procd service is broken (cf. #1334, where
the category path was a silent prod no-op despite green unit tests).

## Scope

Add scenario(s) under `scripts/e2e/scenarios_fake/` (mirror
`test_extra_blocked.py` / `test_02_allowed_browsing.py` style; snapshot-driven
via `SnapshotBuilder`, observed via `_observers.py`). Suggested file:
`test_cname_direct_requery.py`.

The harness needs a **controllable CNAME chain** in the LAN resolver — do
**not** pin real CDN edge hostnames (they rotate; see #1337). Stand up a
fixed `brand.test → CNAME mid.cdn.test → CNAME edge.cdn.test → A <ip>` chain
(fake upstream / dnsmasq `cname=`/`host-record` fixtures, or the fake-api
serving a controlled zone — pick what fits the existing VM DNS setup) so the
"direct re-query of the leaf" step is deterministic and the leaf IP is known.

### Cases to cover

**#1346 `eb_` (extraBlocked — silent-bypass, higher severity):**
[...]

PR #1684, opened 2026-06-12
e2e(#1677): Gate 2 scenario for v6 attribution + AAAA fixture on e2e-edge

### LINKED ISSUE #1668: bug: IPv6 connection events lack matched-rule attribution — conntrack labeling fallback is v4-only

### LINKED ISSUE #1677: e2e (Gate 2): add v6 attribution scenario (AAAA fixture in wifihaven.net + scenarios_fake test)

sameerparekh:
## Summary

CD-tier follow-up to PR #1673 (which closed #1668 — `dns_log.parse_reply_line`
was dropping every AAAA answer pre-fix, so v6 destinations always landed in
`connection_events` as bare v6 literals). #1673 pinned the parser fix with
busted-integration tests in `openwrt/test/conntrack_spec.lua`; this PR adds
the VM-level proof on a real router VM with real dnsmasq, real `dns-tail`,
real `conntrack -E NEW` on a real v6 SYN, and a real event POST to the fake
API.

Two changes that land together:

1. **`infra/cloudflare/main.tf`** — adds an `AAAA` on the existing
   `e2e-edge.wifihaven.net` leaf (`2001:db8::10`, RFC 3849 documentation
   space). Pairing the AAAA with the existing A produces the realistic
   dual-stack production shape: one name returning both families. Applied
   on merge by `master-cloudflare.yml` (#1357/#1358); CI's
   `cloudflare-terraform` job already validates `fmt + validate` on PRs.
   The knob lives in `variables.tf` as `e2e_edge_ip6`.

2. **`scripts/e2e/scenarios_fake/test_v6_attribution.py`** — one Gate 2
   scenario: client `dig AAAA e2e-edge.wifihaven.net` to populate the
   dns_log cache, then `nc -6 e2e-edge.wifihaven.net 80` to fire
   `conntrack -E NEW` with destination `2001:db8::10`, then
   `wait_event_with_host_attribution` for a `connection_event` whose
   `host.type=fqdn` and `host.value` contains `e2e-edge.wifihaven.net`.
   Also asserts `host.value` is not a bare v6 literal even when `type=fqdn`
   (belt-and-braces against an attribution leak).

Builds on #1680 (qemu LAN bridge ULA + SLAAC + v6 FORWARD path); the
`_assert_v6_link_local_ready` gate added there already proves the client
has a v6 default route before this scenario runs.

## Notes

- **Propagation window after merge**: the AAAA TTL is 300s. The first
  Gate 2 run on the merge commit may need a re-run if `dig AAAA` hits
  cache misses; the scenario surfaces a clear `requires terraform apply`
  message rather than hanging silently. This matches the v4 #1351
  pattern.
- **Independent rollback**: API and Cloudflare deploy independently
  per `AGENTS.md` §"SPA hosting split". Removing the AAAA from the zone
  rolls back the scenario's fixture without touching the API or router
  agent.

## Test plan

- [ ] `cloudflare-terraform` CI job (fmt + validate) passes.
- [ ] `master-cloudflare.yml` applies the AAAA on merge.
- [ ] Self-hosted KVM runner picks up the new scenario via
      `scripts/e2e-vm.sh --mode=fake --only v6-attribution` after
      propagation completes.
- [ ] RED proof, locally: temporarily revert the v6 acceptance in
      `openwrt/files/usr/lib/lua/wifihaven/dns_log.lua` (the `or is_ipv6(value)`
      branch from PR #1673) and rebuild the router image — the scenario
      fails with `host.type=ipv6` and the bare literal as `host.value`.
      Documented in the test docstring; not run in the work environment
      because KVM is not available.

Closes #1677.


sameerparekh:
**Independent PR review — first run** (reviewed-sha=`9bdf0748`).

PR adds a Cloudflare AAAA on `e2e-edge.wifihaven.net` (`2001:db8::10`, RFC 3849) and a Gate 2 scenario `test_v6_attribution.py` that proves a v6 SYN to a dual-stack wifihaven.net host attributes to its FQDN end-to-end on a real router VM — the CD-tier follow-up to PR #1673 / #1668. Builds on the v6 LAN bring-up from #1680.

## BLOCKERS

(none)

## SHOULD-FIX

1. **`scripts/e2e/scenarios_fake/test_v6_attribution.py:124-198` — RED progression is "AAAA-absent", not "parser-broken".** The commit history shows a RED test commit before the AAAA-record commit, which makes the test fail at `_wait_for_aaaa_resolution` (no answers). That proves the *test wiring* is sound, not that the test pins the #1668 regression. The actual #1668-pinning failure mode — `dns_log.parse_reply_line` dropping AAAA so `host.type=ipv6` with a bare literal — requires running with PR #1673's `dns_log.lua` v6 acceptance reverted, which can't be done in this work env (no KVM). The PR body and the test docstring both describe this honestly and the busted-tier coverage in PR #1673 already pins the regression at module level, so this is acceptable — but consider opening a follow-up issue to fold a one-shot "run the scenario once with the v6 acceptance temporarily reverted on a KVM runner, capture the failure, attach it to the PR" check into the release process. Otherwise this CD scenario can only catch a v6 *attribution* regression at the layer below — not the specific `parse_reply_line` failure mode.

## NITS

1. **`scripts/e2e/scenarios_fake/test_v6_attribution.py:88-92` — `_fire_v6_syn` builds the `nc` invocation via `sh -c` with f-string interpolation of `LEAF_HOST`.** `LEAF_HOST` is a module constant so there is no injection surface, but `client_exec(client, ["nc", "-6", "-w", "2", LEAF_HOST, "80"], check=False, timeout=10)` is one line shorter and avoids the shell entirely. The existing `test_v6_link_local.py` uses the same `sh -c` pattern, so this is consistent — leave it for symmetry if you prefer.
2. **`scripts/e2e/scenarios_fake/_observers.py:131-146` — `dig_ipv6_answers` largely duplicates `lib.traffic.dns_query`** by re-shelling `dig` directly so it can pass `AAAA` as a qtype. Could be collapsed by extending `dns_query(client, host, *, qtype: str = "A", timeout_s=5)`; not material here because `dig_ipv4_answers` already deviates from `dns_query` in the same module. If `dns_query` ever grows a sibling caller wanting qtype, take the collapse opportunity then.
3. **`infra/cloudflare/variables.tf:27-31` — `e2e_edge_ip6` default `"2001:db8::10"` is already canonical**, so Cloudflare won't produce a perpetual TF diff. If anyone overrides this with a non-canonical form (e.g. `2001:0DB8::10`), Cloudflare's API normalizes on round-trip and `terraform plan` would show drift forever. Worth a one-line warning in the description that the value must already be in canonical lowercase, no leading zeros.
4. **`scripts/e2e/scenarios_fake/test_v6_attribution.py:185-198` — `wait_event_with_host_attribution` times out silently if attribution lands as the bare v6 literal** (because the substring match fails). The post-timeout `assert ev is not None` paths to a clear error message, but a richer diagnostic dumping the most-recent few events for this MAC at the timeout would shorten the next regression-debug loop. The v4 G3 attribution test has the same gap, so this is a pre-existing pattern issue, not a fresh one.

## Dimension coverage notes

- **§1 single-source-of-truth** — pure additive helper + test, no decision/computation duplication.
- **§2 testing integrity** — RED → GREEN commits visible; no existing tests weakened, deleted, or skipped; the new test asserts the actual behavior (`host.type == "fqdn"`, value not a v6 literal); no forbidden mocks.
- **§3 architectural invariants** — no enforcement logic added; no agent decision logic; no new snapshot field (uses the existing minimal vocabulary unchanged); no `/tmp` writers.
- **§4 wire backwards-compatibility** — no snapshot, ingest, or admin-API shape changes.
- **§5 migrations** — none.
- **§6 metrics + dashboard** — N/A; PR adds a test, not a new emitting code path.
- **§7 scope / hygiene** — diff focused on six files all in the v6-attribution path; `Co-Authored-By` trailer is fine, no AI / Claude mention in PR title/body.
- **§8 security** — no credentials; RFC 3849 documentation address; Terraform variable has a safe default; no new auth surface.

## VERDICT: APPROVE

CD-tier scenario + Cloudflare AAAA fixture for v6 hostname attribution, additive across the board, with RED → GREEN history. No BLOCKERS; SHOULD-FIX is an observability suggestion about regression-pinning rigor, not a correctness issue. NITs are stylistic and consistent with prior siblings.

sameerparekh:
**Independent PR review — re-run** (reviewed-sha=`f796936d`, prior=`9bdf0748`).

Delta since prior review: one file, `infra/cloudflare/variables.tf` description-only edit addressing NIT 3.

## Prior-findings status

|---|---|---|
| SHOULD-FIX 1 | RED progression is "AAAA-absent" not "parser-broken" — capture a parser-reverted RED on KVM | **PARTIAL** — filed as follow-up #1685 ("e2e: fold a one-shot revert-and-fail check into the v6 attribution Gate 2 release process"), not folded into this PR. Intentional: the fix is a release-runbook change, not a scenario change, so it doesn't belong on this PR's diff. |
| NIT 1 | `_fire_v6_syn` uses `sh -c` instead of arglist | NOT-ADDRESSED — explicitly left for symmetry with the existing `test_v6_link_local.py` pattern. |
| NIT 2 | `dig_ipv6_answers` could collapse into `dns_query(qtype=…)` | NOT-ADDRESSED — pre-existing duplication; deferred per "if `dns_query` ever grows a sibling caller…". |
| NIT 3 | `e2e_edge_ip6` description should warn about Cloudflare API canonicalization drift | **ADDRESSED** — `infra/cloudflare/variables.tf:29` now carries the canonical-form warning (`NOTE: pass the value in canonical form …`). |
| NIT 4 | `wait_event_with_host_attribution` could dump recent events on timeout | NOT-ADDRESSED — pre-existing pattern shared with the v4 G3 test. |

## NEW findings (delta `9bdf0748...f796936d`)

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

(none — the delta is a single description-string addition with no functional surface)

## VERDICT: APPROVE

Description-only follow-up addressing NIT 3, no functional changes. Prior verdict (APPROVE) stands: no BLOCKER was ever open, no new BLOCKER was introduced. Outstanding NITs are either intentional symmetry with sibling tests or deferred to natural-collapse opportunities; the only SHOULD-FIX is tracked at #1685.

Issue #573, opened 2026-05-17
Recover hostname attribution from TLS SNI when DNS isn't available


sameerparekh:
## Problem

Several real-world flow patterns leave the agent without a hostname for an otherwise legitimate connection:

- Long-lived sockets that predate the agent — DNS happened on a previous network or before the agent started.
- HTTP/3 / QUIC connection reuse — one DNS lookup spawns many host requests.
- Encrypted DNS (covered separately by the DoH-blocking issue).
- Apple APNs and other hardcoded-IP services (unrecoverable here, separate problem).

See [docs/ops/hostname-attribution.md](docs/ops/hostname-attribution.md) §2 and §4.

For TLS-bearing flows, the **SNI in the unencrypted ClientHello** carries the hostname the client is asking for. Reading SNI doesn't break TLS, doesn't require a custom CA, and reliably recovers the hostname even when our dnsmasq log has no mapping.

## Proposal

Add an SNI-extraction sidecar to the OpenWRT agent that:

1. Mirrors the first ~512 bytes of each new outbound TLS flow (TCP/443) to a userspace handler — either via:
   - `nft hook ingress` + a small eBPF program that emits a perf event with `(mac, dst_ip, dst_port, sni_string)` for the first packet of a connection, OR
   - a `tc-bpf` filter on the LAN bridge, OR
   - libpcap with BPF filter `tcp dst port 443 and (tcp[((tcp[12] & 0xf0) >> 2) + 5] == 0x01)` (ClientHello) — simpler but heavier.
2. Parses the SNI extension from the ClientHello.
3. Adds the `(dst_ip, sni)` pair to the same DNS cache the dnsmasq log feeds, with a short TTL.
4. The existing attribution path picks it up automatically.

### Tradeoffs

- **CPU**: ~1 packet per new TLS flow, small parse. Should be cheap on modern OpenWRT hardware but worth measuring on the slowest supported router (#X TBD).
- **Encrypted ClientHello (ECH)**: when ECH is widely deployed, this technique stops working for ECH-enabled flows. ECH adoption is still low (mid-2026), so this buys years of utility.
- **0-RTT QUIC**: the SNI is similarly visible in the QUIC initial packet; QUIC parser is more involved but doable.
- **Trust model**: this is observation only. We don't modify, terminate, or interfere with the TLS handshake. Operators with privacy concerns can disable the sidecar.

## Out of scope

- TLS interception / MITM.
- Decrypting payloads.
[...]

PR #1686, opened 2026-06-12
fix(#1683): require wifihaven.dns_tail_sets, not bare dns_tail_sets — unbreaks Master Router CD

### LINKED ISSUE #1683: ci(cd): Master Router CD failing — agent inside VM doesn't poll /api/router/policy after enrollment (post-#1667 unblock, new failure mode)
## Symptom

Master Router CD on main commit \`a29a55bb\` failing all 4 e2e jobs (Gate 2 + Gate 3a × ipk/apk) with:
\`\`\`
TimeoutError: no policy fetch after id=0 in 180s
\`\`\`

#1667's Phase 1 port-killer worked — the fake API now binds cleanly on 127.0.0.1:3222 and enrollment exchanges the router token. But the router agent inside the VM never polls \`/api/router/policy\` for 180 seconds → fixture times out → all 49 tests ERROR.

Run: https://github.com/wifihaven/wifihaven/actions/runs/27431880817

## Root-cause candidates (ranked by risk)

Recent router-agent commits since the last green Gate 2 run (need to bisect, but suspect list ordered by code-volume / runtime-dependency risk):

1. **77a7d51f feat(#573): SNI-based hostname attribution (libpcap-userspace, TLS-only)** — adds a NEW userspace packet-capture component to the agent. If libpcap / pcap kernel access / capabilities aren't present in the OpenWRT image, the agent may fail to start or hang at init.
2. **a29a55bb fix(#1668): v6 conntrack attribution** — touched the eb6_/bl6_ lookup path; smaller surface but possible startup-time syntax / module-load issue.
3. **b0875154 fix(#1658): periodic ipset re-resolve heartbeat** — introduced a new timer loop; could deadlock the agent if scheduled wrong.
4. **284763ec feat(#1645): connection_event labeling** — landed earlier; lower risk now.

## Investigation steps

1. **Pull artifacts** from the failed run if any (console.log, agent log, /etc/wifihaven/policy.json snapshot). The current job uploads on failure if scripts/vm/.run/**/console.log exists — check whether it does.
2. **Bisect the agent**: build the OpenWRT image at the pre-#573 commit (b0875154) and run Gate 2 locally on plex; if green, the bisect target is the #573 commit. Then build at #1645 (284763ec) and check; etc.
3. **Inspect the VM directly**: if the operator can SSH into a Gate 2 VM during failure (or via the e2e harness's interactive mode), get \`logread | grep -i wifihaven\` to see whether the agent even started, and if so, why it didn't poll.
4. **Check for missing OpenWRT package deps**: the libpcap pcap-userspace component from #573 — does \`build-ipk.sh\` / \`build-apk.sh\` declare libpcap as a runtime dependency? If not, the package installs but the binary can't load the lib.

## Fix paths

- **If #573 is the culprit**:
  - Quick: revert PR #1643 (commit 77a7d51f) and re-deploy. Re-introduce #573 with proper packaging in a follow-up.
  - Better: add the missing libpcap runtime dep to the OpenWRT package manifests and verify Gate 2 goes green.
- **If #1668 is the culprit**:
  - Lower-risk revert (small change); fix the regression in a follow-up.
- **If #1658 is the culprit**:
  - Inspect the timer loop for a tight-loop / deadlock; fix the scheduling.

## Acceptance

- Pinned which commit broke the agent startup with evidence (agent logs or successful pre-revert Gate 2 run).
- Fix or revert lands; openwrt-latest publishes; routers pick up #1645/#1658/#1668's accumulated improvements.
- A regression test added that catches "agent doesn't poll policy within N seconds of boot" earlier in the CI pipeline (Gate 1 / smoke) — currently this kind of regression only surfaces at Gate 2's 3-minute timeout.

## Related
- #1667 — port-collision unblock (Phase 1 worked; this is the NEXT failure exposed)
- #573 / PR #1643 — SNI attribution (top suspect)
- #1668 / #1658 / #1645 — other recent agent changes

sameerparekh:
Closes #1683.

## Symptom

Master Router CD has been red since PR #1659 (#1658) landed. All 4 e2e arms
(Gate 2 × {ipk,apk} + Gate 3a × {ipk,apk}) fail with:

```
TimeoutError: no policy fetch after id=0 in 180s
```

The fake API receives the router-token enrollment but the agent never polls
`/api/router/policy`. Latest failing run before this fix:
https://github.com/wifihaven/wifihaven/actions/runs/27434620060

## Root cause

Reproduced locally on plex by booting the freshly-built router image and
SSHing in:

```
logread | grep -i wifihaven | tail
…
daemon.err wifihaven-agent[2881]: /usr/bin/lua: /usr/lib/lua/wifihaven/eb_refresh.lua:57:
                                  module 'dns_tail_sets' not found:
…
daemon.info procd: Instance wifihaven::agent s in a crash loop 6 crashes, 0 seconds since last crash
```

PR #1659 introduced `eb_refresh.lua` with:

```lua
local dns_tail_sets = require("dns_tail_sets")        -- bare name
local render        = require("wifihaven.render")     -- qualified
```

The bare `require("dns_tail_sets")` works in `openwrt/test/run_tests.sh`,
which sets `LUA_PATH` to include **both** `./files/usr/lib/lua/?.lua` and
`./files/usr/lib/lua/wifihaven/?.lua`, so either spelling resolves. On a
real router lua's default `LUA_PATH` is only `/usr/share/lua/?.lua;
/usr/lib/lua/?.lua` — the bare form looks for `/usr/lib/lua/dns_tail_sets.lua`,
which doesn't exist (it lives at `/usr/lib/lua/wifihaven/dns_tail_sets.lua`).
The agent crashes at startup → procd respawns it 6× → procd marks it
crash-looped → no `/api/router/policy` poll ever happens → every Gate 2 /
Gate 3a scenario errors at fixture setup.

Top suspect from #1683 (#573 / `tcpdump-mini`) was a red herring — the SNI
dep was packaged correctly. The bisect target was the lower-risk #1659 change.

## Fix

Single character: `require("dns_tail_sets")` → `require("wifihaven.dns_tail_sets")`.

## Test

TDD red→green:

1. **Red commit** (`a7bd3c5`): adds `openwrt/test/lua_module_paths_spec.sh`
   which mirrors the production `LUA_PATH` (only `./files/usr/lib/lua/?.lua` —
   NOT the `wifihaven` subdir) and asserts every module under
   `files/usr/lib/lua/wifihaven/` resolves cleanly via
   `require('wifihaven.X')`. Wired into `run_tests.sh`. Before the fix it
   reports `FAIL: require('wifihaven.eb_refresh')` with the same "module
   not found" error the router crashed on. This pins the regression so any
   future bare `require` in a wifihaven module fails CI immediately, well
   before Gate 2's 3-minute timeout.
2. **Green commit** (`6522564`): applies the require fix; the spec turns
   green.

## Verified

Built `openwrt-wifihaven-ipk-23.05.img` on plex against this branch and ran
`scripts/e2e-vm.sh --mode=fake --only allowed-browsing`:

```
================= 1 passed, 44 deselected in 67.78s (0:01:07) ==================
```

Full Gate 2 fake suite running in parallel; will confirm before merge.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/eb_refresh.lua
@@ -54,7 +54,7 @@
 -- `eb_set_name` / `eb6_set_name` / `bl_set_name` / `bl6_set_name` exports so
 -- the (host|id) → nft-set-name mapping has a single source of truth.
 
-local dns_tail_sets = require("dns_tail_sets")
+local dns_tail_sets = require("wifihaven.dns_tail_sets")
 local render        = require("wifihaven.render")
 
 local M = {}
--- a/openwrt/test/lua_module_paths_spec.sh
+++ b/openwrt/test/lua_module_paths_spec.sh
@@ -1,0 +1,54 @@
+#!/bin/sh
+# Verify every module under files/usr/lib/lua/wifihaven/ loads cleanly under
+# the LUA_PATH that exists on a real router — i.e., only the qualified
+# `wifihaven.X` form is reachable from /usr/lib/lua/, NOT the bare `X` form.
+#
+# The default test harness (test/run_tests.sh) sets LUA_PATH to include both
+# ./files/usr/lib/lua/?.lua AND ./files/usr/lib/lua/wifihaven/?.lua, so
+# require("dns_tail_sets") works alongside require("wifihaven.dns_tail_sets").
+# On the installed router, lua's default LUA_PATH is just /usr/share/lua/?.lua
+# and /usr/lib/lua/?.lua, so a bare require("dns_tail_sets") fails with
+# "module not found" and the wifihaven-agent crash-loops at startup.
+#
+# This caught #1683: PR #1659 (eb_refresh.lua, #1658) used
+# `require("dns_tail_sets")` instead of `require("wifihaven.dns_tail_sets")`,
+# which passed locally but broke every Gate 2 / Gate 3a run because the
+# agent could not load.
+set -e
+
+PASS=0; FAIL=0
+ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+MOD_DIR="$ROOT/files/usr/lib/lua/wifihaven"
+
+[ -d "$MOD_DIR" ] || { printf "MISSING: %s\n" "$MOD_DIR"; exit 1; }
+
+# Mirror the production LUA_PATH: only the qualified `wifihaven.X` form is
+# reachable. Bare names must use require("wifihaven.X") to resolve.
+PROD_LUA_PATH="$ROOT/files/usr/lib/lua/?.lua;$ROOT/test/shim/?.lua;$ROOT/test/shim/?/init.lua;$(lua -e 'print(package.path)')"
+
+check() {
+  if [ "$2" = "ok" ]; then
+    printf "  PASS: %s\n" "$1"; PASS=$((PASS + 1))
+  else
+    printf "  FAIL: %s — %s\n" "$1" "$2"; FAIL=$((FAIL + 1))
+  fi
+}
+
+for f in "$MOD_DIR"/*.lua; do
+  name="$(basename "$f" .lua)"
+  # `init` files under a subdir are loaded as the parent module name, skip.
+  case "$name" in
+    init) continue ;;
+  esac
+  err="$(LUA_PATH="$PROD_LUA_PATH" lua -e "require('wifihaven.$name')" 2>&1 || true)"
+  if [ -z "$err" ]; then
+    check "require('wifihaven.$name') loads under production LUA_PATH" ok
+  else
+    # Trim to the first line for readability.
+    first="$(printf "%s" "$err" | head -n1)"
+    check "require('wifihaven.$name') loads under production LUA_PATH" "$first"
+  fi
+done
+
+printf "\n%d passed, %d failed\n" "$PASS" "$FAIL"
+[ "$FAIL" -eq 0 ]
--- a/openwrt/test/run_tests.sh
+++ b/openwrt/test/run_tests.sh
@@ -37,6 +37,7 @@ sh test/boot_skeleton_spec.sh
 sh test/update_spec.sh
 sh test/update_multi_pkg_spec.sh
 sh test/install_spec.sh
+sh test/lua_module_paths_spec.sh
 sh test/nft_log_dep_spec.sh
 sh test/agent_spec.sh
 sh test/rotate_dnsmasq_log_spec.sh


sameerparekh:
## Independent PR review — PR #1686

Reviewed merge-base diff `git diff origin/main...HEAD` (three-dot). Two commits: a red regression spec, then the one-line require fix.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. [openwrt/test/lua_module_paths_spec.sh:39](openwrt/test/lua_module_paths_spec.sh#L39) — the `case "$name" in init) continue ;;` guard is dead code today (no `init.lua` exists under `files/usr/lib/lua/wifihaven/`). Harmless and forward-defensive, but worth a comment noting it's a future guard rather than handling a present file.
2. [openwrt/test/lua_module_paths_spec.sh:23](openwrt/test/lua_module_paths_spec.sh#L23) — silently assumes host `lua` is on `$PATH` (used both for `print(package.path)` and the per-module `lua -e require`). Same implicit assumption as `run_tests.sh` and the other `*_spec.sh` files, so consistent; a one-liner `command -v lua >/dev/null || { echo "SKIP: lua not installed"; exit 0; }` would harden CI portability, but not gating.

### Dimension pass

1. **Duplicated logic / SSOT** — the fix conforms `eb_refresh.lua` to the same `require("wifihaven.X")` convention already used by `render`, `dns-tail`, and every other consumer of these modules; reduces drift surface rather than adding it. ✓
2. **Testing integrity** — red→green is visible in history (commit `a7bd3c5` adds a spec that fails on the bug; commit `6522564` flips it green). Spec is generic: it iterates every module under `files/usr/lib/lua/wifihaven/` and asserts each loads under a production-shaped LUA_PATH, so a future bare `require` in any wifihaven module will fail CI well before Gate 2's 3-minute timeout. No existing test weakened, no assertion deleted, no fixture retrofitted. ✓
3. **Architectural invariants** — no DNS/enforcement/snapshot/wire/`/tmp`-writer surface touched. Router-agent module-path hygiene only. ✓
4. **Backwards compatibility** — wire shapes unchanged. ✓
5. **Migrations** — none. ✓
6. **Metrics + dashboard** — N/A; this is a startup-crash fix, not a new functional path. ✓
7. **Scope / hygiene** — focused (3 files, +55 / -1), no drive-bys, no AI mentions in PR title/body/commits (the `Co-Authored-By` trailer pattern is fine per project convention; this PR has none and that's also fine). ✓
8. **Security** — N/A. ✓

### VERDICT: APPROVE

The bug is real and reproduced (agent crash-loops at startup with `module 'dns_tail_sets' not found`, blocking every Gate 2 / Gate 3a poll). The fix is the minimum diff that resolves it (one-char module qualifier), and the regression test it ships generalizes to **all** wifihaven Lua modules so this class of `require`-path mismatch fails CI on the next attempt rather than at the 3-minute VM timeout. No BLOCKERs, no SHOULD-FIX.

PR #1689, opened 2026-06-12
fix(#1687): Gate 3b accepts any fc00::/7 ULA (drops hardcoded prefix from #1680)

### LINKED ISSUE #1687: ci(cd): Gate 3b fails on last-published router — harness hardcodes ULA prefix from #1680, old routers use auto-generated ULA
## Symptom

Master API/UI CD on main commit \`7ce197f2\` failing both Gate 3b jobs (ipk + apk):

\`\`\`
RuntimeError: client client1 missing IPv6 default route or ULA address in fdaa:bbbb:cccc::/48 after 30s.
The qemu LAN bridge is not carrying RA/SLAAC — see scripts/vm/uci-defaults/99-wifihaven and #1680.
--- client ip -6 addr/route ---
fdf8:cd30:bf17::/64 dev eth0 proto ra metric 1002 ...
\`\`\`

Run: https://github.com/wifihaven/wifihaven/actions/runs/27437472753

The router IS issuing RAs and the client IS getting a SLAAC address + default route — just under prefix \`fdf8:cd30:bf17::/48\`, not the hardcoded-expected \`fdaa:bbbb:cccc::/48\`.

## Root cause

PR #1682 (closing #1680) did two things together:
1. Added \`/etc/uci-defaults/99-wifihaven\` that sets \`network.globals.ula_prefix='fdaa:bbbb:cccc::/48'\` on first boot.
2. Updated the e2e harness (\`scripts/e2e/lib/vm.py:310\` \`ROUTER_ULA_PREFIX = "fdaa:bbbb:cccc"\`) to assert the client lands in that prefix.

Gate 3b explicitly boots the **last-published** router image (openwrt-latest, asymmetric-deploy test). That image was built BEFORE #1682 merged, so its uci-defaults don't set the deterministic prefix → OpenWRT falls back to auto-generated random ULA (\`fdf8:cd30:bf17\` etc., re-randomized per build). The harness assertion fails.

Compounding: Master Router CD is also red right now (see #1683 — agent doesn't poll policy after enrollment), so a new openwrt-latest won't publish to bring routers onto the deterministic prefix. Gate 3b stays red indefinitely until either Router CD recovers OR the harness loosens.

## Fix

**Loosen the harness assertion** — the actual test is "v6 traffic flows between client and router," not "v6 lands on a specific prefix." Hardcoding the prefix is a brittle implementation detail.

\`scripts/e2e/lib/vm.py:310\`:
\`\`\`python
- ROUTER_ULA_PREFIX = "fdaa:bbbb:cccc"
\`\`\`

Replace with either:
- (preferred) **Pull the actual ULA from the router** at fixture-setup time: \`ssh root@<router> uci get network.globals.ula_prefix\`; assert the client landed inside that prefix.
- (simpler) **Accept any ULA** (the IANA ULA range is fc00::/7): assert the client has at least one IPv6 address in \`fc00::/7\` AND a default-route via the router. Drop the specific-prefix check entirely.

The simpler form is sufficient — the test's purpose is "v6 plumbed end-to-end," and any ULA satisfies that.

## Acceptance

- Gate 3b ipk + apk pass on the current main without changing the last-published router image.
- The check still catches a genuine "client has no v6 at all" failure (the original intent of #1680).
- A small follow-up note in \`docs/architecture.md\` or wherever the e2e gates are documented: harness assumptions about router state must be derived from the live router, not hardcoded.

## Related

- #1680 / PR #1682 — introduced the hardcoded ULA in the harness
- #1683 — Master Router CD red (agent doesn't poll policy); blocks the alternative path of "wait for openwrt-latest to update"
- AGENTS.md §single-source-of-truth — hardcoded-expected ULA in two places (uci-defaults AND harness) is the divergence shape we collapse on sight

sameerparekh:
Closes #1687.

## Root cause

PR #1682 (closing #1680) added `/etc/uci-defaults/99-wifihaven` to write a deterministic ULA prefix `fdaa:bbbb:cccc::/48` on freshly built routers AND hardcoded that same prefix in `scripts/e2e/lib/vm.py:ROUTER_ULA_PREFIX` so the Gate 2 client-v6 assertion could pin it. Gate 3b explicitly boots the **last-published** router image (asymmetric-deploy test) — that image was built before #1682 merged, so its uci-defaults never write the deterministic prefix and OpenWRT falls back to a per-build random ULA (`fdf8:cd30:bf17::/48` in the failing run). v6 IS plumbed end-to-end; the harness was checking the wrong invariant.

The fix: drop the hardcoded prefix and accept any ULA in `fc00::/7` (RFC 4193) — that's what "client got a v6 ULA from the router" actually means. The new pure function `v6_ula.client_has_ula_address` is unit-covered (`scripts/e2e/lib/test_v6_ula.py`); `_assert_v6_link_local_ready` now AND-gates a `dev eth0` default route with the loosened ULA check.

## Compounding with #1683

Master Router CD is also red (#1683 — unblocked by 7db78029, still rolling out). Without this PR, Gate 3b would stay red until a new openwrt-latest with the deterministic ULA publishes; with it, API/UI CD recovers immediately without waiting on the router CD lane.

## #1680's regression catch is preserved

The original purpose of the assertion — "client got a v6 default route AND a global-scope ULA from the router via RA/SLAAC" — is still pinned, just by membership in fc00::/7 instead of by string match. The unit suite has explicit cases for:

- accepts deterministic prefix (Gate 2)
- accepts random ULA (Gate 3b)
- rejects no v6 at all (the #1680 regression shape)
- rejects link-local only
- rejects GUA only

## §single-source-of-truth note

The earlier hardcoded prefix lived in two places: `scripts/vm/uci-defaults/99-wifihaven` and `scripts/e2e/lib/vm.py`. The fix collapses them — neither side needs to know the literal value any more, because the assertion describes the functional property (any ULA) rather than mirroring a config value. Updated the uci-defaults comment to match.

## Test plan

- [x] `python3 -m pytest scripts/e2e/lib/test_v6_ula.py` — 6 passed (deterministic, random, fc00 half, no-v6, link-local-only, GUA-only)
- [ ] Master API/UI CD passes Gate 3b on the next run against this branch's merge
- [ ] Gate 2 still green (new routers — deterministic prefix is still a ULA, so the loosened check passes)

### DIFF
--- a/scripts/e2e/lib/test_v6_ula.py
+++ b/scripts/e2e/lib/test_v6_ula.py
@@ -1,0 +1,72 @@
+"""Unit cover for the #1687 fix: the Gate 3b client-v6 check must accept
+ANY ULA (fc00::/7 per RFC 4193), not a hardcoded prefix.
+
+Before #1687, the harness asserted the client landed in `fdaa:bbbb:cccc::/48`
+(the prefix `/etc/uci-defaults/99-wifihaven` writes on a freshly built
+router). That broke Gate 3b, which boots the last-published router image —
+older builds use OpenWRT's auto-generated random ULA, so the check failed
+even though v6 was fully plumbed end-to-end.
+
+The check's real intent (from #1680) is "client got a v6 ULA from the
+router via RA/SLAAC" — any ULA satisfies that. Run standalone:
+
+    python3 -m pytest scripts/e2e/lib/test_v6_ula.py
+"""
+from __future__ import annotations
+
+from v6_ula import client_has_ula_address
+
+
+# Real `ip -6 addr show eth0` output shape, abridged.
+def _addr_output(*global_addrs: str) -> str:
+    lines = [
+        "2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 state UP qlen 1000",
+        "    inet6 fe80::5054:ff:fe12:3456/64 scope link",
+        "       valid_lft forever preferred_lft forever",
+    ]
+    for a in global_addrs:
+        lines.append(f"    inet6 {a} scope global dynamic")
+        lines.append("       valid_lft 1800sec preferred_lft 1800sec")
+    return "\n".join(lines)
+
+
+def test_accepts_deterministic_ula_from_99_wifihaven():
+    """The Gate 2 case: new router image, deterministic prefix from
+    scripts/vm/uci-defaults/99-wifihaven."""
+    out = _addr_output("fdaa:bbbb:cccc::1/64")
+    assert client_has_ula_address(out) is True
+
+
+def test_accepts_random_ula_from_last_published_router():
+    """The Gate 3b case (#1687): older router image, OpenWRT-generated
+    random ULA. Must still pass — any ULA is fine."""
+    out = _addr_output("fdf8:cd30:bf17::1/64")
+    assert client_has_ula_address(out) is True
+
+
+def test_accepts_fc00_half_of_the_range():
+    """fc00::/7 is fc00::/8 plus fd00::/8. The fd00 half is the only one
+    in practical use today, but the check must not reject fc00 either."""
+    out = _addr_output("fc12:3456:7890::1/64")
+    assert client_has_ula_address(out) is True
+
+
+def test_rejects_no_v6_at_all():
+    """Preserve #1680's intent: a client with no v6 ULA must still fail
+    the check (catches a genuine RA/SLAAC regression)."""
+    out = "2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP>\n    inet6 fe80::1/64 scope link"
+    assert client_has_ula_address(out) is False
+
+
+def test_rejects_only_link_local():
+    """Link-local fe80::/10 is not a ULA — must not satisfy the check."""
+    out = _addr_output()  # only the link-local line
+    assert client_has_ula_address(out) is False
+
+
+def test_rejects_only_gua():
+    """A global unicast (2000::/3) is not a ULA — must not satisfy.
+    (In practice the qemu LAN bridge never sees a GUA, but the check
+    should reject it cleanly rather than spuriously pass.)"""
+    out = _addr_output("2001:db8::1/64")
+    assert client_has_ula_address(out) is False
--- a/scripts/e2e/lib/test_v6_ula.py
+++ b/scripts/e2e/lib/test_v6_ula.py
@@ -34,34 +34,34 @@ def test_accepts_deterministic_ula_from_99_wifihaven():
     """The Gate 2 case: new router image, deterministic prefix from
     scripts/vm/uci-defaults/99-wifihaven."""
     out = _addr_output("fdaa:bbbb:cccc::1/64")
-    assert client_has_ula_address(out) is True
+    assert client_has_ula_address(out)
 
 
 def test_accepts_random_ula_from_last_published_router():
     """The Gate 3b case (#1687): older router image, OpenWRT-generated
     random ULA. Must still pass — any ULA is fine."""
     out = _addr_output("fdf8:cd30:bf17::1/64")
-    assert client_has_ula_address(out) is True
+    assert client_has_ula_address(out)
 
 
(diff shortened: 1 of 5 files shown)


sameerparekh:
## Independent PR review — #1689

Scope: `git diff origin/main...HEAD` against the 8 dimensions in `docs/pr-review-checklist.md`.

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

1. `scripts/e2e/lib/vm.py:315` — `from .v6_ula import client_has_ula_address` is a mid-file import. There is already one other mid-file import in this file (line 19, `from .sh import …`), so this matches local precedent — but if a cleanup pass ever consolidates imports at the top, this one should join it.
2. `scripts/e2e/lib/vm.py:328-338` — the probe now makes two SSH calls per iteration (route + addr) instead of one `&&`-joined shell pipeline. Functionally equivalent and easier to read; the extra SSH round-trip per 2s tick is unmeasurable. Worth keeping as-is — flagging only so a future cleanup doesn't "optimize" back to the inline `grep -q` form and re-introduce the brittle string-match.
3. `scripts/e2e/lib/test_v6_ula.py:38,45,52,58,64,71` — `assert x is True/False` works but `assert x` / `assert not x` is the conventional pytest style. Cosmetic.

### Dimension-by-dimension

- **§1 SSOT.** *Improves* SSOT — collapses the two hardcoded mirrors (`vm.py:ROUTER_ULA_PREFIX` + the uci-default value) onto a single functional invariant ("any ULA"). No new keep-in-sync coupling introduced. ✅
- **§2 Testing integrity.** TDD red→green is visible in commit history (`1564fca0` RED, `283ecf7c` GREEN). The regression test for #1680's intent ("no v6 at all" still fails the check) is retained explicitly (`test_rejects_no_v6_at_all`). No existing assertion was weakened — only a hardcoded string-prefix check was replaced by a more general predicate covered by 6 cases incl. negative ones. Bug-fix-with-failing-test confirmed. ✅
- **§3 Architectural invariants.** Test-harness only. No router/agent decision logic added; snapshot wire vocabulary untouched; no DNS-as-enforcement reasoning. ✅
- **§4 Wire contract.** N/A — no wire-visible shapes change. ✅
- **§5 Migrations.** None. ✅
- **§6 Metrics + dashboard.** N/A — no emitted series added. ✅
- **§7 Scope / hygiene.** Focused diff (4 files, all e2e harness). No AI/Claude mentions in commits or PR body. Branch diffed three-dot vs `origin/main`. ✅
- **§8 Security.** N/A — no auth, SQL, or secret surfaces touched. ✅

### VERDICT: APPROVE

Replaces a hardcoded ULA prefix with an `fc00::/7` membership check, unblocking Gate 3b on the last-published router image without weakening #1680's "v6 plumbed end-to-end" guarantee. Two prior hardcoded mirrors (harness + uci-default) collapse to one functional predicate, covered by 6 unit cases including the original "no v6 at all" regression shape. Test-only / e2e-harness scope — no wire, snapshot, or enforcement code touched.

sameerparekh:
## Independent PR review — #1689 (re-run)

Prior reviewed-sha: `283ecf7c`. Delta: `git diff 283ecf7c...HEAD` — `scripts/e2e/lib/test_v6_ula.py` only, `is True`/`is False` → bare `assert` / `assert not`.

### Prior findings status

| # | Class | File:line | Status |
|---|-------|-----------|--------|
| 1 | NIT | `scripts/e2e/lib/vm.py:315` mid-file import | NOT-ADDRESSED (intentional — matches local precedent at line 19) |
| 2 | NIT | `scripts/e2e/lib/vm.py:328-338` two-SSH probe shape | NOT-ADDRESSED (intentional — keeps brittle inline `grep -q` from coming back) |
| 3 | NIT | `scripts/e2e/lib/test_v6_ula.py` `is True`/`is False` style | ADDRESSED (commit `72dbf2e7`) |

No prior BLOCKERs or SHOULD-FIXes.

### New findings from delta

(none)

### VERDICT: APPROVE

NIT-only delta: tightens the unit-test idiom to bare `assert` / `assert not`. All 6 ULA cases still pass. No new findings; no open BLOCKER.

PR #1690, opened 2026-06-12
fix(#1688): complete v6 attribution end-to-end (Gate 2 v6 tests pass)

### LINKED ISSUE #1688: ci(cd): #1677 v6 e2e tests fail in Gate 2 — v6 attribution + conntrack NEW path incomplete after #1668
## Symptom

Master Router CD on main commit \`7db78029\` (post-#1683 unblock) failing 2 of the new v6 e2e tests added by PR #1684 (#1677):

\`\`\`
FAILED scenarios_fake/test_v6_attribution.py::test_v6_destination_attributes_to_fqdn_not_bare_literal
       — TimeoutError: timed out waiting for connection_event for 02:e2:fa:27:60:3a with host
         attribution containing 'e2e-edge.wifihaven.net'
FAILED scenarios_fake/test_v6_link_local.py::test_v6_syn_traverses_router_forward_chain
       — Failed: no conntrack NEW for v6 SYN to 2001:db8::10 (grep returned '0').
\`\`\`

Run: https://github.com/wifihaven/wifihaven/actions/runs/27439506443

#1683 (agent doesn't poll policy) IS fixed (commit 7db78029, missing \`wifihaven.\` namespace on \`dns_tail_sets\` require). The remaining failures are real v6 functionality gaps that #1668's PR #1673 was supposed to close — but didn't, fully.

## Hypothesis

PR #1673 (fix #1668) tightened the conntrack labeling fallback to query \`eb6_\` for v6 destinations. PR #1684 added new e2e tests to validate the v6 attribution path end-to-end. The tests show:

1. **v6 conntrack NEW events aren't being captured.** \`test_v6_syn_traverses_router_forward_chain\` greps the router's conntrack table for v6 SYN entries and gets 0. Either:
   - nftables v6 forward chain doesn't allow the test traffic through (drops before conntrack NEW fires)
   - \`conntrack -E\` family selection misses v6 (the watcher started with -f ipv4 and misses v6 events)
   - Routing isn't set up correctly for v6 in the qemu harness (router not actually forwarding v6 LAN→WAN)

2. **v6 connection events don't carry FQDN attribution.** \`test_v6_destination_attributes_to_fqdn_not_bare_literal\` waits for a connection_event whose host is the actual FQDN, not the bare v6 literal. If conntrack NEW isn't firing (test 1), no event is emitted at all → timeout. If conntrack NEW IS firing for v6, the event might be created but with \`host=<v6 literal>\` instead of the resolved FQDN — that's a v6-side gap in \`attribute_hostname\` / the dns-cache lookup path.

Tests 1 and 2 might share root cause (no v6 conntrack events) or be two distinct gaps.

## Investigation steps

Boot a Gate 2 VM locally on plex (api.lan / 192.168.10.43) at current main; reproduce the failure; then:

1. **Verify v6 forward chain.** \`nft list chain inet wifihaven wifihaven_block_v6\` (or whichever v6 chain exists) — does it have an accept rule for the test SYN? Does \`forwarding\` work for v6 in OpenWRT (\`/proc/sys/net/ipv6/conf/all/forwarding\` = 1)?
2. **Verify conntrack -E captures v6.** \`ssh\` into the router VM; \`pgrep -af conntrack\` to see how the watcher was started; verify family covers v6. Lua agent code: \`openwrt/files/usr/lib/lua/wifihaven/conntrack.lua\` and \`openwrt/files/usr/sbin/wifihaven-agent\`.
3. **Verify dns-tail captures AAAA.** \`cat /tmp/wifihaven-dns-cache.txt\` — does it have entries for v6 destinations? \`openwrt/files/usr/lib/lua/wifihaven/dns_log.lua\` should handle AAAA records (per the PR title "dns_log AAAA + nft eb6_/bl6_ fallback"). Verify.
4. **Trace one v6 flow end-to-end.** From the client VM, \`curl -6\` (or analogous) to the test destination; watch the router for the SYN, conntrack NEW event, dns-cache lookup, attribute_hostname call.

## Fix

Whatever step 1-4 surfaces. Likely one of:
- nft v6 forward chain accept rule missing.
- conntrack watcher not subscribing to family=ipv6.
- dns_log AAAA parser not writing to the cache file.
- attribute_hostname not querying v6 entries from the cache.

Each is a small, targeted fix. PR ships with a regression-test update so future failures don't slip past.

## Acceptance

- Both failing Gate 2 tests pass.
- Master Router CD goes green; the accumulated #1645/#1658/#1668 fixes ship to openwrt-latest.
- A small regression test pinning each fix (one for conntrack v6 NEW detection, one for AAAA-aware attribution).

## Related
- #1683 — agent-doesn't-poll (fixed in 7db78029)
- #1668 / PR #1673 — v6 conntrack labeling (partial fix; this issue is its e2e completion)
- #1677 / PR #1684 — added the failing tests

sameerparekh:
Closes #1688.

## What was broken

Two gaps prevented v6 destinations from producing connection_events end to end after #1668's parser fix:

1. **v6 conntrack hooks weren't loaded.** The wifihaven package depends on `+conntrack` but not `+kmod-nf-conntrack6`, so on a fresh OpenWRT image the kernel's v6 conntrack hooks (via `nf_defrag_ipv6`) are absent. `conntrack -E -e NEW` then sees 0 v6 flow events — both the agent's watcher AND `test_v6_link_local`'s own conntrack capture come up empty. This explains the test 1 failure (`no conntrack NEW for v6 SYN to 2001:db8::10`).
2. **is_wan_bound rejected every v6 flow.** The helper checked `flow.src_ip:sub(1, #lan_prefix)` against the v4 lan_prefix only, so a v6 source (e.g. `fdaa:bbbb:cccc::42`) never matched and the watch loop silently dropped every v6 conntrack NEW event before `handle_flow` ran. This explains the test 2 failure (event for `02:e2:fa:…` never arrives).

## Fix

- `openwrt/Makefile`: add `+kmod-nf-conntrack6` to DEPENDS.
- `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua`: make `is_wan_bound` family-aware off the src address; accept an optional `lan_prefix_v6` parameter. When unset, v6 flows stay filtered (preserves today's prod behaviour for routers that have not authored a v6 LAN).
- `openwrt/files/usr/sbin/wifihaven-agent` + `openwrt/files/etc/config/wifihaven`: thread a new `lan_prefix_v6` UCI option through to `conntrack.watch`.
- `scripts/vm/uci-defaults/99-wifihaven`: seed `lan_prefix_v6=fdaa:bbbb:cccc:` on the Gate 2 router VM; belt-and-braces force-enable `net.ipv6.conf.{all,br-lan,eth1}.forwarding=1` from the WAN-ifup hotplug so the FORWARD chain reliably sees the SYN even when fw4's v6 zone wiring is stripped down.

Per AGENTS.md §single-source-of-truth, `is_wan_bound` stays a single helper with one family-fork inside it — no v4-vs-v6 forking is sprinkled across call sites.

## Tests

Regression in `openwrt/test/conntrack_spec.lua`:
- Accepts v6 LAN→public when `lan_prefix_v6` is set.
- Rejects v6 LAN→LAN.
- Rejects every v6 flow when `lan_prefix_v6` is unset/empty/nil (prod default).

Local busted: `343 successes / 0 failures / 0 errors / 1 pending` (the pending is the pre-existing `nft -c` check that requires nft on the host).

## Test plan

- [ ] CI Gate 2 ipk: `test_v6_attribution.py::test_v6_destination_attributes_to_fqdn_not_bare_literal` passes.
- [ ] CI Gate 2 ipk: `test_v6_link_local.py::test_v6_syn_traverses_router_forward_chain` passes.
- [ ] CI Gate 2 apk: both v6 tests pass.
- [ ] Master Router CD turns green so accumulated #1645/#1658/#1668/#1683/#1688 ship to openwrt-latest.

### DIFF
--- a/openwrt/Makefile
+++ b/openwrt/Makefile
@@ -28,7 +28,11 @@ define Package/wifihaven
   # recover hostname attribution when the client used DoH/DoT or a hardcoded
   # IP — the same shared dns_cache feeds eb_/ea_/bl_ enforcement, so this is
   # an additive, observability-style dependency.
-  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nf-log +kmod-nf-log6 +libustream-mbedtls +openssl-util +tcpdump-mini
+  # #1688: kmod-nf-conntrack6 loads the v6 conntrack hooks (via nf_defrag_ipv6
+  # on modern kernels). Without it `conntrack -E -e NEW` sees 0 v6 flow events,
+  # so the agent never emits connection_events for v6 destinations and the
+  # Gate 2 v6 attribution suite (scripts/e2e/scenarios_fake/test_v6_*.py) fails.
+  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nf-log +kmod-nf-log6 +kmod-nf-conntrack6 +libustream-mbedtls +openssl-util +tcpdump-mini
   PKGARCH:=all
 endef
 
--- a/openwrt/files/etc/config/wifihaven
+++ b/openwrt/files/etc/config/wifihaven
@@ -21,6 +21,13 @@ config wifihaven 'wifihaven'
 	# LAN subnet prefix used to identify outbound (egress) flows
 	option lan_prefix '192.168.1.'
 
+	# Optional v6 LAN ULA prefix used to identify outbound v6 flows (#1688).
+	# Leave unset to filter all v6 flows out of connection_events (today's
+	# prod behaviour). Set to the router's LAN ULA prefix (e.g.
+	# 'fdaa:bbbb:cccc:') to let v6 traffic through the same is_wan_bound
+	# filter the v4 path uses.
+	# option lan_prefix_v6 'fdaa:bbbb:cccc:'
+
 	# How many connection_attempt events to accumulate before flushing
 	option event_batch_size '50'
 
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -796,7 +796,7 @@ function M.parse_conntrack_line(line)
 end
 
 -- ---------------------------------------------------------------------------
--- is_wan_bound(flow, lan_prefix) -> bool
+-- is_wan_bound(flow, lan_prefix, lan_prefix_v6) -> bool
 --
 -- Returns true when the flow is an outbound WAN-destined flow: src_ip is on
 -- the LAN AND dst_ip is NOT on the LAN.  This filters out LAN-internal flows
@@ -804,18 +804,27 @@ end
 -- appear in connection_events — they are noise with no parental-control signal.
 -- (#575)
 --
--- lan_prefix example: "192.168.1."
+-- Family is detected from src_ip — a colon means IPv6, otherwise v4 — so a
+-- single helper handles both stacks (#1688). lan_prefix_v6 is the LAN v6 ULA
+-- prefix (e.g. "fdaa:bbbb:cccc:"); when empty/nil v6 flows are rejected so a
+-- prod router without an authored v6 LAN keeps today's behavior.
+--
+-- lan_prefix    example: "192.168.1."
+-- lan_prefix_v6 example: "fdaa:bbbb:cccc:"
 -- ---------------------------------------------------------------------------
-function M.is_wan_bound(flow, lan_prefix)
-  return flow.src_ip:sub(1, #lan_prefix) == lan_prefix
-     and flow.dst_ip:sub(1, #lan_prefix) ~= lan_prefix
+function M.is_wan_bound(flow, lan_prefix, lan_prefix_v6)
+  local is_v6 = flow.src_ip:find(":", 1, true) ~= nil
+  local prefix = is_v6 and lan_prefix_v6 or lan_prefix
+  if not prefix or prefix == "" then return false end
+  return flow.src_ip:sub(1, #prefix) == prefix
+     and flow.dst_ip:sub(1, #prefix) ~= prefix
 end
 
 -- is_outbound is kept as a backward-compatible alias so existing call sites
 -- outside the watch loop continue to work. New code should use is_wan_bound.
 -- @deprecated use is_wan_bound
-function M.is_outbound(flow, lan_prefix)
-  return M.is_wan_bound(flow, lan_prefix)
+function M.is_outbound(flow, lan_prefix, lan_prefix_v6)
+  return M.is_wan_bound(flow, lan_prefix, lan_prefix_v6)
 end
 
 -- ---------------------------------------------------------------------------
@@ -831,6 +840,8 @@ end
 --   eb_hosts_by_mac   table     shared ref: { mac -> { hostname -> true } }
 --   ea_hosts_by_mac   table     shared ref: { mac -> { hostname -> true } }
 --   lan_prefix     string    default "192.168.1."
+--   lan_prefix_v6  string    optional v6 LAN ULA prefix (e.g. "fdaa:bbbb:cccc:")
+--                            — when unset, v6 flows are filtered out (#1688).
 --   max_batch      int       default 50
 --   flush_interval int       default 10  (seconds)
 --   max_retries    int       default 3
@@ -850,9 +861,10 @@ end
 -- }
 -- ---------------------------------------------------------------------------
 function M.watch(cfg)
-  local log        = cfg.log            or default_log()
-  local lan_prefix = cfg.lan_prefix     or "192.168.1."
-  local max_batch  = cfg.max_batch      or 50
+  local log           = cfg.log            or default_log()
+  local lan_prefix    = cfg.lan_prefix     or "192.168.1."
+  local lan_prefix_v6 = cfg.lan_prefix_v6  or ""
+  local max_batch     = cfg.max_batch      or 50
   local flush_int  = cfg.flush_interval or 10
   local max_retry  = cfg.max_retries    or 3
   local base_delay = cfg.base_delay     or 2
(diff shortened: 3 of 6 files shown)


sameerparekh:
## Independent PR review — first pass

Reviewing PR #1690 against `docs/pr-review-checklist.md` (8 dimensions).
Diff: `git diff origin/main...HEAD` — 6 files, +81/−14.

BLOCKERS
(none)

SHOULD-FIX
1. `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:919` — the agent invokes
   `conntrack -E -e NEW` with no `-f`. The conntrack(8) man page asserts that
   "if `-f` is not passed, the default layer 3 protocol will be IPv4." The PR
   relies on this defaulting NOT applying to `-E` (the netlink subscription is
   family-agnostic in conntrack-tools source), and that is plausibly correct,
   but it's a load-bearing assumption with no in-PR test and no comment
   pinning it. If on the deployed conntrack-tools build the default DOES
   filter to v4, test 2 (`test_v6_destination_attributes_to_fqdn_not_bare_literal`)
   stays red even with `kmod-nf-conntrack6` present and the `is_wan_bound` fix
   landing — i.e. the PR does not actually close #1688. CI will tell us, but
   either drop a comment at line 919 that the family-agnostic subscription is
   the contract relied on, or pre-empt the risk by spawning a parallel
   `conntrack -E -e NEW -f ipv6` reader. Minor preference: comment + CI gate.

NITS
1. `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:815` — family detection
   uses `flow.src_ip:find(":", 1, true) ~= nil`. The existing v6 helper at
   `conntrack.lua:63` uses the same pattern (`dst_ip:find(":", 1, true)`); a
   tiny shared `is_v6(addr)` predicate would collapse the two and pin the
   convention. Not load-bearing — both call sites are short.
2. `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:917` — the watcher log
   line now formats `lan_prefix_v6` with `%q`, which double-quotes the value
   in the log. Harmless but inconsistent with the surrounding `%s` for
   `lan_prefix`; either both quoted or both bare reads cleaner.
3. `scripts/vm/uci-defaults/99-wifihaven:155` — three `echo 1 >/proc/sys/...
   /forwarding` writes happen on every `ifup wan` hotplug. Idempotent and
   harmless, but the surrounding comment frames them as "belt-and-braces"
   without confirming whether stock OpenWRT 23.05 / 25.12 actually leaves
   them at 0 in this harness. A one-line `ip6tables-save | head` or `sysctl
   net.ipv6.conf.all.forwarding` capture in the failure-artifact bundle
   would let a future reviewer confirm which legs of the fix are actually
   load-bearing — purely diagnostic, not blocking.

### Dimension status

- **Duplicated logic / single source of truth** — clean. The family fork
  lives in ONE place (`is_wan_bound`); `is_outbound` alias updated to
  match; single call site at `conntrack.lua:937` threads the new prefix.
  No "keep in sync" smell.
- **Testing integrity** — clean. Four new `is_wan_bound` unit cases pin
  the v6 contract (set/unset prefix × LAN→public / LAN→LAN / non-LAN
  src). No existing tests weakened, deleted, or retrofitted. Local
  busted run: `343 successes / 0 failures / 0 errors / 1 pending`
  (pre-existing `nft -c` skip). The e2e tests at
  `scripts/e2e/scenarios_fake/test_v6_*.py` are the RED gate; the
  fix's job is to turn them green.
- **Architectural invariants** — clean. Router stays a dumb applier; no
  decision logic added to the agent. No DNS-as-enforcer slippage. No
  new snapshot fields. No new `/tmp` writers (the `/proc/.../forwarding`
  writes are fixed-name single-byte sysctl, not append spools).
- **Backwards compatibility / wire contract** — clean. No wire change.
  `lan_prefix_v6` is router-internal UCI; default empty preserves
  today's prod behavior. UCI is owned by a single agent build, so the
  no-deprecation-window rule applies here.
- **Migrations** — N/A.
- **Metrics + dashboard** — N/A in the strict sense (no new emitted
  series). The v6 flow path will start emitting more `connection_event`s
  through existing surfaces; no new metric needed.
- **Scope / hygiene** — focused on #1688. No AI mentions. No dead code.
  PR body has the Test plan checklist.
- **Security** — no creds, no SQL, no new routes. `/proc` writes are
  root-only on a single-tenant router image; no privilege boundary
  crossed.

VERDICT: APPROVE

Three small concerns flagged, none merge-gating. The fix correctly
identifies and addresses two real gaps (`kmod-nf-conntrack6` dep and
v4-only `is_wan_bound`) and ships a unit-level regression for the v6
contract. The remaining risk is one load-bearing conntrack-tools
assumption (SHOULD-FIX 1) that CI will validate on the next Gate 2 run.

sameerparekh:

Reviewing the incremental delta `2f98609...4c61013` (one commit, one file)
against `docs/pr-review-checklist.md`.

### Prior-findings status (from `2f98609`)

| # | Finding | Status |
|---|---|---|
| SHOULD-FIX 1 | `conntrack.lua:919` — conntrack `-E` family-agnostic subscription not documented | **ADDRESSED** — new comment at `conntrack.lua:920–928` pins the contract (`-f` only constrains `-L` table dumps; `-E` opens NFCT_ALL_CT_GROUPS), explains the dep on `kmod-nf-conntrack6`, and points to the Gate 2 e2e tests as the load-bearing gate. |
| NIT 1 | shared `is_v6(addr)` helper | NOT-ADDRESSED (deliberately deferred — non-blocking, two short call sites). |
| NIT 2 | `%q` on `lan_prefix_v6` in log line | **ADDRESSED** — log format now `%s` consistent with surrounding fields. |
| NIT 3 | forwarding-state diagnostic in failure artifacts | NOT-ADDRESSED (out-of-scope diagnostic, non-blocking). |

### New findings on the delta

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
(none)

### Dimension status (delta only)

- **Duplicated logic / single source of truth** — clean. Comment-only +
  one format-string char.
- **Testing integrity** — clean. No tests touched; the prior unit pins
  still hold (local busted: 112 / 0 / 0 / 0).
- **Architectural invariants / wire compat / migrations / metrics /
  scope / security** — clean. Nothing in this delta touches any of them.

VERDICT: APPROVE

Prior SHOULD-FIX + cheap NIT addressed. The two deferred NITs are
non-blocking by design. No new findings in the delta. The fix is ready
to merge as soon as Gate 2 CI confirms both v6 tests are green.

PR #1693, opened 2026-06-13
fix(#1691): use curl -6 not nc -6 in Gate 2 v6 tests (BusyBox nc lacks -6)

### LINKED ISSUE #1691: ci(cd): #1688 fix incomplete — v6 conntrack NEW + FQDN attribution still fail after #1690 shipped kmod-nf-conntrack6 + is_wan_bound v6
## Symptom

Master Router CD on main commit \`98e43a89\` (THE #1688 fix itself) still fails the same two v6 e2e tests added by PR #1684 (#1677):

\`\`\`
FAILED scenarios_fake/test_v6_attribution.py::test_v6_destination_attributes_to_fqdn_not_bare_literal
       — TimeoutError: timed out waiting for connection_event for <mac> with host attribution
         containing 'e2e-edge.wifihaven.net'
FAILED scenarios_fake/test_v6_link_local.py::test_v6_syn_traverses_router_forward_chain
       — Failed: no conntrack NEW for v6 SYN to 2001:db8::10 (grep returned '0').
\`\`\`

Run: https://github.com/wifihaven/wifihaven/actions/runs/27444633516 (Gate 2 apk)

PR #1690 (closing #1688) added three things:
1. \`+kmod-nf-conntrack6\` to \`openwrt/Makefile\` DEPENDS — the v6 conntrack kernel hooks now install.
2. Made \`is_wan_bound\` family-aware off the source address in \`conntrack.lua\`.
3. Added \`lan_prefix_v6\` UCI option, threaded through to \`conntrack.watch\`.

The same failures recurring means at least one of:
- The kmod didn't actually install on the built image (Makefile change not picked up; check \`opkg list-installed | grep conntrack\` inside the booted VM).
- The kmod installed but kernel hooks aren't active without explicit \`modprobe nf_defrag_ipv6\` / \`modprobe nf_conntrack_ipv6\` at boot.
- The conntrack -E -e NEW watcher's netlink subscription, despite the inline comment about NFCT_ALL_CT_GROUPS, isn't actually receiving v6 events on this kernel (verify with \`conntrack -E -f ipv6\` from the VM CLI to prove the kernel CAN dispatch them).
- Something else (PolicyService snapshot? policy.json shape?) is gating v6 flows downstream.

## Investigation steps

Reproduce on plex; SSH into the booted Gate 2 router VM during the failure:

1. \`opkg list-installed | grep -i conntrack\` — confirm \`kmod-nf-conntrack6\` is present.
2. \`lsmod | grep -iE 'conntrack|defrag'\` — are the v6 hooks loaded?
3. \`cat /proc/net/nf_conntrack | grep -c ipv6\` — is the table actually tracking v6?
4. From the client: \`curl -6 http://[2001:db8::10]/\`. On the router, in another shell: \`conntrack -E -f ipv6 -e NEW --timestamp\` — does it print the NEW event?
5. If the kernel sees it but the agent's watcher doesn't — \`pgrep -af conntrack\` and check the spawn arguments; the inline comment claims \`-e NEW\` without \`-f\` is family-agnostic, but verify against this kernel by running the SAME args manually and seeing what comes out.

## Fix paths

Whichever step above identifies the gap:
- Missing kmod: tighten the package build to ensure the kmod is actually included.
- kmod installed but not loaded: add a uci-defaults / init.d \`modprobe nf_conntrack_ipv6 nf_defrag_ipv6\`.
- Watcher subscription wrong: pass \`-f ipv6\` explicitly (the existing inline-comment claim about NFCT_ALL_CT_GROUPS may be wrong on this kernel) OR run a second watcher process for v6.
- Downstream policy gate: trace where the v6 flow gets dropped in PolicyService.

## Acceptance

- Both failing Gate 2 tests pass on a fresh Gate 2 boot.
- Master Router CD goes green; the accumulated #1645 / #1658 / #1668 / #1688 ship to openwrt-latest.
- A regression test (busted or shell) pinning the actual gap (e.g. "is conntrack subscription active for v6").

## Related
- #1688 / PR #1690 — supposed fix that didn't fully close it
- #1668 / PR #1673 — original v6 attribution work
- #1677 / PR #1684 — the failing tests
- #1683 — the OTHER cause of Router CD red (agent doesn't poll); fixed in 7db78029

sameerparekh:
## Symptom

Master Router CD red on main `98e43a89`. Both Gate 2 v6 scenarios fail
with the same errors as before #1690:

- `test_v6_link_local.py::test_v6_syn_traverses_router_forward_chain` —
  `no conntrack NEW for v6 SYN to 2001:db8::10 (grep returned '0')`.
- `test_v6_attribution.py::test_v6_destination_attributes_to_fqdn_not_bare_literal` —
  `TimeoutError: timed out waiting for connection_event for <mac> with host
  attribution containing 'e2e-edge.wifihaven.net'`.

CI run: <https://github.com/wifihaven/wifihaven/actions/runs/27444633516>

## Root cause

The bug is in the test harness, NOT the router agent. The Alpine client
base image's `nc` is BusyBox, whose `nc` applet does NOT implement `-6`:

```
client# nc -6 -w 2 2001:db8::10 80
nc: unrecognized option: 6
BusyBox v1.37.0 (2025-11-21 22:40:56 UTC) multi-call binary.
…
EX=1
```

Both v6 scenarios trail the invocation with `>/dev/null 2>&1; true`, which
swallows the exit-1 → from pytest's perspective the client_exec call
succeeds even though zero packets were emitted. Conntrack correctly
observes zero NEW v6 events because no v6 SYN ever left the client.

The test was a silent no-op from the day it was written (#1684); #1690's
agent-side wiring (`+kmod-nf-conntrack6`, `is_wan_bound` family-fork,
`lan_prefix_v6` threading) is correct and necessary, but it has nothing
to verify against until the client actually sends a v6 SYN.

## Diagnosis evidence (plex, fresh-built openwrt-wifihaven-ipk-23.05.img at 98e43a89)

Rules out the three hypotheses in #1691:

**Bucket E (kmod absent / unloaded) — RULED OUT.**

```
router# opkg list-installed | grep -i conntrack
conntrack - 1.4.8-1
kmod-nf-conntrack - 5.15.167-1
kmod-nf-conntrack-netlink - 5.15.167-1
kmod-nf-conntrack6 - 5.15.167-1
libnetfilter-conntrack3 - 1.0.9-2

router# lsmod | grep -iE 'conntrack|defrag'
nf_conntrack           77824  8 nft_*,nf_conntrack_netlink
nf_conntrack_netlink   32768  0
nf_defrag_ipv4         12288  1 nf_conntrack
nf_defrag_ipv6         16384  1 nf_conntrack

router# cat /proc/net/nf_conntrack | grep ipv6 | wc -l    # v6 conntrack entries
5
```

#1690's Makefile change DID propagate. (Note: `opkg files kmod-nf-conntrack6`
reports no files — on 5.15 kernels v6 conntrack is built into the universal
`nf_conntrack.ko` and the package is effectively a metadependency.)

**Bucket F (watcher subscription) — RULED OUT.**

Both the agent's invocation (`conntrack -E -e NEW` with no `-f`) and the
test's invocation (`conntrack -E -f ipv6 -e NEW`) DO emit v6 NEW events on
this kernel for genuine v6 flows triggered manually. The #1688 inline
comment about NFCT_ALL_CT_GROUPS is correct.

**Bucket G (downstream gate) — RULED OUT.**

By the time `test_v6_link_local` runs in Gate 2 (test [97%]), many prior
tests have pulled in the session-scoped `router_session` fixture, which
enrolls the agent and triggers the first `policy.apply()`. That
atomically replaces the boot-skeleton `inet wifihaven_boot` (priority
`filter - 10`, `policy drop`) with the runtime `inet wifihaven` table
(`policy accept` on forward at priority `filter + 1`). So forwarded v6
NEW packets reach conntrack as expected.

**Confirmed gap (Bucket H — test harness)**:

```
client# nc -6 -w 2 2001:db8::10 80                       # busybox nc
nc: unrecognized option: 6                                # → 0 packets

client# curl -6 -s -m 2 -o /dev/null http://[2001:db8::10]/
                                                          # curl exit 28
router# tcpdump -i br-lan -nn ip6
IP6 fdaa:bbbb:cccc::147.52528 > 2001:db8::10.80: Flags [S], …
router# conntrack -E -e NEW (-f ipv6 also)
[NEW] tcp 6 120 SYN_SENT src=fdaa:bbbb:cccc::147 dst=2001:db8::10
      sport=… dport=80 [UNREPLIED]
```

## Fix

Replace the busybox-`nc -6` invocations with `curl -6` in both Gate 2 v6
scenarios:

- [scripts/e2e/scenarios_fake/test_v6_link_local.py](scripts/e2e/scenarios_fake/test_v6_link_local.py)
- [scripts/e2e/scenarios_fake/test_v6_attribution.py](scripts/e2e/scenarios_fake/test_v6_attribution.py)

`curl` is already in the client base image's package list
([scripts/vm/build-client-base.sh:173](scripts/vm/build-client-base.sh)) and
supports `-6` natively. `-s -m 2` matches the prior `-w 2` SYN-then-bail
semantics; the curl-exit-28 (`Connection timed out`) is swallowed by the
trailing `; true` exactly as before.

The two test docstrings are updated with a "do NOT use `nc -6` here"
guardrail noting BusyBox's gap, so the next person to touch these tests
doesn't quietly reintroduce the regression.

## Test plan

- [x] Local plex repro of failure with `nc -6` (busybox rejects `-6`).
- [x] Local plex repro of success with `curl -6` (v6 SYN traverses br-lan;
      conntrack NEW fires).
- [ ] CI Gate 2 ipk: `test_v6_link_local.py::test_v6_syn_traverses_router_forward_chain` passes.
- [ ] CI Gate 2 ipk: `test_v6_attribution.py::test_v6_destination_attributes_to_fqdn_not_bare_literal` passes.
- [ ] CI Gate 2 apk: both v6 tests pass.
- [ ] Master Router CD turns green so the accumulated agent-side fixes
      (#1645 / #1658 / #1668 / #1688) ship to openwrt-latest.

Closes #1691.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -214,25 +214,51 @@ end
 -- ---------------------------------------------------------------------------
 -- parse_arp_table() -> { ip -> mac }
 --
--- Reads /proc/net/arp at call time.  Called per-event so the MAC table stays
--- fresh without a background refresh loop.
+-- Reads /proc/net/arp at call time (v4) AND `ip -6 neigh` (v6 NDP cache).
+-- The kernel exposes the v4 ARP table as a proc file but the v6 NDP cache
+-- only over netlink, so we shell out for the v6 half. Called per-event so
+-- the MAC table stays fresh without a background refresh loop.
+--
+-- v6 is load-bearing here: without it, the watch loop reaches a v6 conntrack
+-- NEW with `src=<lan-ula>`, asks parse_arp_table for the MAC of that v6
+-- src_ip, gets nil, and emits a connection_attempt event whose `mac` field
+-- is nil. The fake-API Gate 2 attribution suite (#1677) asserts on
+-- `mac == client.mac AND host.value contains LEAF_HOST` — a nil mac fails
+-- the assert and is_wan_bound's #1690 family-fork looks like it didn't
+-- close the loop. (#1691.)
 -- ---------------------------------------------------------------------------
 function M.parse_arp_table()
   local result = {}
+  -- /proc/net/arp columns: IP, HWtype, Flags, HWaddr, Mask, Device
   local f = io.open("/proc/net/arp", "r")
-  if not f then return result end
-  f:read("*l")  -- skip header
-  for line in f:lines() do
-    local ip, _, _, mac = line:match("^(%S+)%s+%S+%s+%S+%s+(%S+)%s+")
-    -- mac field is 4th column; re-match properly
-    local parts = {}
-    for w in line:gmatch("%S+") do parts[#parts + 1] = w end
-    -- /proc/net/arp columns: IP, HWtype, Flags, HWaddr, Mask, Device
-    if #parts >= 4 and parts[4] ~= "00:00:00:00:00:00" then
-      result[parts[1]] = parts[4]
+  if f then
+    f:read("*l")  -- skip header
+    for line in f:lines() do
+      local parts = {}
+      for w in line:gmatch("%S+") do parts[#parts + 1] = w end
+      if #parts >= 4 and parts[4] ~= "00:00:00:00:00:00" then
+        result[parts[1]] = parts[4]
+      end
     end
+    f:close()
   end
-  f:close()
+  -- v6 NDP cache. Format examples:
+  --   `2001:db8::1 dev br-lan lladdr 02:e2:fa:11:22:33 router REACHABLE`
+  --   `fdaa:bbbb:cccc::147 dev br-lan lladdr 02:e2:fa:8e:c2:ce STALE`
+  --   `fe80::2 dev eth1  used 0/0/0 probes 6 FAILED`   (no lladdr — skip)
+  -- The `lladdr <mac>` token is the load-bearing piece; entries in FAILED
+  -- state have no lladdr and we skip them naturally.
+  local pf = io.popen("ip -6 neigh show 2>/dev/null")
+  if pf then
+    for line in pf:lines() do
+      local ip = line:match("^(%S+)")
+      local mac = line:match("lladdr%s+(%S+)")
+      if ip and mac and mac ~= "00:00:00:00:00:00" then
+        result[ip] = mac
+      end
+    end
+    pf:close()
+  end
   return result
 end
 
--- a/openwrt/test/conntrack_spec.lua
+++ b/openwrt/test/conntrack_spec.lua
@@ -81,6 +81,112 @@ describe("arp_lookup_mac", function()
 end)
 
 -- ---------------------------------------------------------------------------
+-- 2b. parse_arp_table: combines /proc/net/arp (v4) AND `ip -6 neigh` (v6 NDP)
+--
+-- Regression for #1691 Bucket G: pre-fix, parse_arp_table only read
+-- /proc/net/arp (v4), so a v6 conntrack NEW with src=<lan-ula> resolved to
+-- mac=nil at handle_flow time, the emitted connection_attempt had a nil
+-- `mac` field, and the Gate 2 attribution suite (which asserts on
+-- mac == client.mac + host contains LEAF_HOST) timed out. The dns-cache
+-- lookup was fine; only the MAC half was missing.
+-- ---------------------------------------------------------------------------
+describe("parse_arp_table", function()
+  -- Helper to stub io.open + io.popen with canned line streams.
+  local function with_stubs(arp_lines, neigh_lines, body)
+    local function reader(lines)
+      local i = 0
+      return {
+        read = function(_, fmt)
+          if fmt == "*l" then i = i + 1; return lines[i] end
+          return nil
+        end,
+        lines = function()
+          return function()
+            i = i + 1
+            return lines[i]
+          end
+        end,
+        close = function() end,
+      }
+    end
+    local saved_open, saved_popen = io.open, io.popen
+    io.open = function(path, _)
+      if path == "/proc/net/arp" and arp_lines then return reader(arp_lines) end
+      return nil
+    end
+    io.popen = function(cmd)
+      if cmd:match("ip %-6 neigh") and neigh_lines then return reader(neigh_lines) end
+      return nil
+    end
+    local ok, err = pcall(body)
+    io.open, io.popen = saved_open, saved_popen
+    assert(ok, err)
+  end
+
+  it("populates v4 entries from /proc/net/arp", function()
+    with_stubs({
+      "IP address       HW type     Flags       HW address            Mask     Device",
+      "192.168.100.42   0x1         0x2         aa:bb:cc:11:22:33     *        br-lan",
+    }, {}, function()
+      local t = conntrack.parse_arp_table()
+      assert.equal("aa:bb:cc:11:22:33", t["192.168.100.42"])
+    end)
+  end)
+
+  it("populates v6 entries from `ip -6 neigh` lladdr tokens", function()
+    with_stubs({
+      "IP address       HW type     Flags       HW address            Mask     Device",
+    }, {
+      "fdaa:bbbb:cccc::147 dev br-lan lladdr 02:e2:fa:8e:c2:ce STALE",
+      "2001:db8::abcd dev br-lan lladdr aa:bb:cc:dd:ee:ff REACHABLE",
+    }, function()
+      local t = conntrack.parse_arp_table()
+      assert.equal("02:e2:fa:8e:c2:ce", t["fdaa:bbbb:cccc::147"])
+      assert.equal("aa:bb:cc:dd:ee:ff", t["2001:db8::abcd"])
+    end)
+  end)
+
+  it("skips v6 NDP entries in FAILED state (no lladdr token)", function()
+    with_stubs({
+      "IP address       HW type     Flags       HW address            Mask     Device",
+    }, {
+      "fe80::2 dev eth1  used 0/0/0 probes 6 FAILED",
+      "2001:db8::10 dev eth1 used 0/0/0 probes 6 FAILED",
+    }, function()
+      local t = conntrack.parse_arp_table()
+      assert.is_nil(t["fe80::2"])
+      assert.is_nil(t["2001:db8::10"])
+    end)
+  end)
+
+  it("merges v4 and v6 in a single lookup table (#1691)", function()
+    with_stubs({
+      "IP address       HW type     Flags       HW address            Mask     Device",
+      "192.168.100.42   0x1         0x2         02:e2:fa:8e:c2:ce     *        br-lan",
+    }, {
+      "fdaa:bbbb:cccc::147 dev br-lan lladdr 02:e2:fa:8e:c2:ce STALE",
+    }, function()
+      local t = conntrack.parse_arp_table()
+      assert.equal("02:e2:fa:8e:c2:ce", t["192.168.100.42"])
+      assert.equal("02:e2:fa:8e:c2:ce", t["fdaa:bbbb:cccc::147"])
+      -- arp_lookup_mac is unchanged — same call site finds either family.
+      assert.equal("02:e2:fa:8e:c2:ce",
+        conntrack.arp_lookup_mac("fdaa:bbbb:cccc::147", t))
+    end)
+  end)
+
+  it("tolerates the v6-neigh popen returning nothing", function()
+    with_stubs({
+      "IP address       HW type     Flags       HW address            Mask     Device",
+      "192.168.100.42   0x1         0x2         aa:bb:cc:11:22:33     *        br-lan",
+    }, nil, function()  -- io.popen returns nil
+      local t = conntrack.parse_arp_table()
+      assert.equal("aa:bb:cc:11:22:33", t["192.168.100.42"])
+    end)
+  end)
+end)
+
+-- ---------------------------------------------------------------------------
 -- 3. Event serialization
 -- ---------------------------------------------------------------------------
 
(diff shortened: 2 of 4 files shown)


sameerparekh:
**Independent PR review** (read-only; per
[`docs/pr-review-checklist.md`](docs/pr-review-checklist.md)).

Diffed against merge base (`git diff origin/main...HEAD`). Two files
touched, both in `scripts/e2e/scenarios_fake/`. Pure test-harness fix:
swap busybox-`nc -6` for `curl -6` so a v6 SYN actually leaves the
client. The root-cause analysis in the PR body is well-evidenced and
correct.

## BLOCKERS

1. **`scripts/e2e/scenarios_fake/test_v6_attribution.py:103`** — the new
   curl URL wraps a **hostname** in URL brackets:
   `curl -6 ... http://[{LEAF_HOST}]/` where `LEAF_HOST =
   "e2e-edge.wifihaven.net"`. Per RFC 3986, square-bracket host syntax
   is reserved for IP-literal (IPv6) hosts only — a hostname inside
   brackets is a malformed URL. curl 8.x rejects it before sending any
   packet:

   ```
   $ curl -6 -v -m 2 -o /dev/null "http://[e2e-edge.wifihaven.net]/"
   curl: (3) bad range in URL position 9:
   http://[e2e-edge.wifihaven.net]/
           ^
   ```

   With `; true` swallowing the exit-3, this is the exact silent no-op
   pattern the PR is trying to fix — the v6 SYN still never leaves the
   client, conntrack still sees zero NEW events, and the test still
   fails with the same TimeoutError as before. Verified on the plex
   client VM that:
   - bare hostname `http://e2e-edge.wifihaven.net/` with `-6` resolves
     AAAA, connects to `[2001:db8::10]:80`, times out with curl exit 28
     (what we want — SYN crossed FORWARD);
   - bracketed hostname fails with curl exit 3 (parse error, no packet).

   Fix: drop the brackets on the hostname form. `curl -6` already forces
   IPv6 resolution + connect, so:

   ```py
   ["sh", "-c", f"curl -6 -s -m 2 -o /dev/null http://{LEAF_HOST}/ ; true"],
   ```

   The literal-IP invocation in [`test_v6_link_local.py:55`](scripts/e2e/scenarios_fake/test_v6_link_local.py)
   correctly keeps the brackets (`http://[2001:db8::10]/`) because that
   IS an IP-literal — only the hostname case is wrong.

## SHOULD-FIX

(none)

## NITS

1. **Two near-duplicate inline curl-incantation comments** in
   [`test_v6_link_local.py:46–54`](scripts/e2e/scenarios_fake/test_v6_link_local.py)
   and [`test_v6_attribution.py:93–104`](scripts/e2e/scenarios_fake/test_v6_attribution.py)
   re-state the same "do NOT use `nc -6` here" guardrail. Mild "keep in
   sync" smell — the next person editing one has to remember to update
   the other. Since `test_v6_attribution.py` already has a `_fire_v6_syn`
   helper, an optional cleanup would factor a shared helper out to
   `lib/vm.py` so both call sites share one source. Non-blocking — the
   inlined copy isn't policy logic, just a comment.

## VERDICT: REQUEST-CHANGES

The PR correctly diagnoses the BusyBox `nc -6` gap and the agent-side
wiring from #1690 is shown to be sound, but the `test_v6_attribution.py`
replacement URL is malformed (hostname inside `[...]`). With the URL
parse error, that test will continue to fail in CI with the identical
TimeoutError. The literal-IP replacement in `test_v6_link_local.py` is
correct.

sameerparekh:
**Independent PR review — re-run on push** (per
[`docs/pr-review-checklist.md`](docs/pr-review-checklist.md)).

Prior review (reviewed-sha=10d7fee0) raised one BLOCKER (bracketed hostname
in the curl URL) and one NIT (duplicated guardrail comment). Status-checking
both against current HEAD `9c3bb1e`, then reviewing the incremental delta
`git diff 10d7fee0...9c3bb1e` against the 8 dimensions.

## Prior findings — status table

| # | Finding | Class | Status | Note |
| - | --- | --- | --- | --- |
| 1 | [`test_v6_attribution.py:103`](scripts/e2e/scenarios_fake/test_v6_attribution.py) — `curl -6 ... http://[{LEAF_HOST}]/` is malformed (square-bracket host syntax is RFC 3986 IP-literal only; curl 8.x errors with exit 3 'bad range in URL', emits no packet, silent no-op). | BLOCKER | **ADDRESSED** | Now [`test_v6_attribution.py:106`](scripts/e2e/scenarios_fake/test_v6_attribution.py): `curl -6 ... http://{LEAF_HOST}/` (no brackets). `curl -6` already forces IPv6-only resolve + connect; verified on plex client VM that this resolves AAAA → connects to `[2001:db8::10]:80` → curl exit 28 (intentional connect-timeout). The literal-IP form in [`test_v6_link_local.py:55`](scripts/e2e/scenarios_fake/test_v6_link_local.py) correctly keeps brackets (`http://[2001:db8::10]/`), which is the right shape for an IP-literal. |
| 2 | Duplicated "do NOT use `nc -6`" guardrail comments across the two scenario files. | NIT | **NOT-ADDRESSED** | Both files still inline the same comment. Non-blocking; the optional cleanup (factor `_fire_v6_syn` into `lib/vm.py`) wasn't taken in this push. |

## Delta review (`10d7fee0...9c3bb1e`)

Three files changed in the new push:

- [`openwrt/files/usr/lib/lua/wifihaven/conntrack.lua`](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua) — `parse_arp_table` extended to also parse `ip -6 neigh show` so v6 src_ip → MAC lookups succeed.
- [`openwrt/test/conntrack_spec.lua`](openwrt/test/conntrack_spec.lua) — busted regression tests for the v6 path.
- [`scripts/e2e/scenarios_fake/test_v6_attribution.py`](scripts/e2e/scenarios_fake/test_v6_attribution.py) — the BLOCKER fix (one-line URL change).

The diagnosis pinned by the new commit is **Bucket G** (downstream gate). The end-to-end pre-fix path was:

1. v6 SYN reaches FORWARD chain (curl works after the BLOCKER fix).
2. Conntrack tracks NEW and the watcher's family-aware `is_wan_bound` (#1690) accepts it.
3. `handle_flow` asks `parse_arp_table` for the MAC of the v6 src_ip — pre-fix, `parse_arp_table` only read `/proc/net/arp` (v4-only), so the lookup returned `nil`.
4. `build_event` emits a `connection_attempt` event with a nil `mac` field — `dns_log` lookup for the dst_ip succeeded (cache had `2001:db8::10 → e2e-edge.wifihaven.net`), so attribution was fine, but the event's mac didn't match `client.mac` and `wait_event_with_host_attribution` timed out.

The new commit reads the v6 NDP cache via `ip -6 neigh show` and merges both families into the same `{ip → mac}` lookup table. Verified on a fresh-built ipk image on plex: Gate 2 v6 pack now `2 passed, 44 deselected in 189.93s` (both `test_v6_attribution` and `test_v6_link_local`).

### Against the 8 dimensions

1. **Duplicated logic / SSoT — clean.** `parse_arp_table` stays the single primitive `handle_flow` calls for MAC lookup; the v6 case is folded INTO that helper, not a parallel v6-only path at the call site. Follows
[`AGENTS.md` §single-source-of-truth](AGENTS.md#single-source-of-truth).
2. **Testing integrity — clean.** The bug-fix commit ships busted coverage that fails without the change:
   [`conntrack_spec.lua:97–164`](openwrt/test/conntrack_spec.lua) pins (a) v4 entries from `/proc/net/arp`, (b) v6 entries from `ip -6 neigh` `lladdr` tokens, (c) skipping `FAILED`-state v6 entries, (d) merged v4+v6 lookup table, (e) graceful fallback when the popen returns nothing. The stubs mock only the external I/O boundary (`io.open`/`io.popen`), in line with AGENTS.md's "mocks — external I/O only" rule. No existing assertions weakened.
3. **Architectural invariants — clean.** Router agent stays a dumb applier. No decision logic added.
4. **Backwards compat — clean.** `parse_arp_table` is internal; no wire shape touched.
5. **Migrations — N/A.**
6. **Metrics / dashboard — N/A.** Extending an existing path; not adding a new emit point.
7. **Scope / hygiene — clean.** Focused diff. No AI mentions in commit messages or PR body. `Co-Authored-By` trailer absent — also fine. Comments well-scoped; the new doctring on `parse_arp_table` motivates the v6 half clearly. Branch-diff uses three-dot (commit log shows the merge-base remained at `origin/main`).
8. **Security — clean.** `io.popen("ip -6 neigh show 2>/dev/null")` invokes `/bin/sh -c` (libc popen behavior). No untrusted input flows to the popen string — the command is a fixed literal — so no shell-injection surface.

### NEW findings

#### BLOCKERS

(none)

#### SHOULD-FIX

(none)

#### NITS

1. [`conntrack.lua:255–264`](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua) — `io.popen("ip -6 neigh show ...")` forks+execs `ip` once per conntrack NEW event, doubling the per-event syscall overhead vs. the prior v4-only `/proc/net/arp` read. Likely fine at typical LAN throughput on a home router, but if the watch loop is profiled later this is the obvious place to introduce a short-TTL cache (e.g. reuse the parsed table across consecutive `tick()`s within ~1s) or move both lookups to a single netlink reader. Non-blocking — the simpler shape is the right one to land first.

## VERDICT: APPROVE

The BLOCKER from the prior review is ADDRESSED in `test_v6_attribution.py:106`,
verified by Gate 2 `2 passed` locally on plex. The new commit also closes the
Bucket G gap that the prior reviewed-sha left lurking — v6 src_ip → MAC was
silently nil because `parse_arp_table` was v4-only — with a single-source-of-
truth fix and pinned busted regressions. No new BLOCKER or SHOULD-FIX. One
NIT noted for follow-up; non-blocking. Safe to merge.

Issue #1679, opened 2026-06-12
Per-allowed-app toggle: "allow during schedule block" (default current = always-allowed)


sameerparekh:
## Background

An app whose per-profile assignment is `mode = Allowed` (a profile-level `extraAllowed` entry) currently overrides **every** drop path — including the per-MAC schedule-block drop. This is the documented `extraAllowed`-beats-`@blocked_macs` contract (see #421 and the `feedback_extraallowed_beats_blocked` operator note: *"admin allow overrides the @blocked_macs drop too, not just eb_/bl_ drops"*).

That's intentional for "schoolwork app" cases — the kid can do Khan even when general internet is bedtime-blocked. But it has a side effect surfaced by #1666 / [#1675](https://github.com/wifihaven/wifihaven/pull/1675):

- Prima's iPad, bedtime block applied correctly at 21:00 MDT (`blocked: true, blockReason: Schedule` in the 02:53 UTC snapshot).
- Khan in her profile `extraAllowed` → `ea_<mac>_<host>` ipset carves Khan IPs out of the per-MAC drop overnight.
- iOS background pings Khan IPs all night, writing sparse `traffic_reports` rows.
- Pre-#1675 those rows gap-stitched into ~21 phantom minutes against her Khan per-app cap, tripping it at 08:40 MDT.

\#1675 fixes the gap-stitch (no more phantom inflation from sparse pings). But there's still a separate, legitimate operator concern: **the operator may not want a "schoolwork" app to be reachable at all during bedtime**, even though the app is allowed during the day.

## Proposal

Add a **per-allowed-app toggle**: "allowed during profile-level schedule blocks" (working name: `allowedDuringScheduleBlock`).

- **Scope**: the toggle is per `AppTimeLimit` / per-app-assignment row (so the same app can be authored differently per profile — e.g. Khan allowed-during-schedule for Prima but blocked-during-schedule for Quintus).
- **Default**: `true` (preserves current behavior — no behavior change for existing assignments on rollout).
- **Semantics** (server-side, in `PolicyService.decide` / snapshot assembly):
  - When the profile resolves to `blocked = true` AND `blockReason = Schedule`:
    - Drop the app's host-set from `extraAllowed` if `allowedDuringScheduleBlock = false`.
    - Keep it if `true`.
[...]

PR #1695, opened 2026-06-13
feat(#1679 schema): add app_policy_assignments.allowed_during_schedule_block

sameerparekh:
Phase 1 of #1679. Schema-only per AGENTS.md §migrations-back-compat.

Adds a per-(profile, app) boolean column `allowed_during_schedule_block` to `app_policy_assignments`, defaulting to `TRUE` — which preserves current behavior on rollout (an existing allowed-app assignment continues to override the per-MAC schedule drop, matching the documented `extraAllowed`-beats-`@blocked_macs` contract from #421).

Code adoption ships in PR 2 after this lands and deploys. In the meantime the column is dormant: server code doesn't read it yet, so behavior is unchanged.

Note on table name: #1679's sketch said `app_time_limits` with a caveat to confirm at design time. The per-(profile, app) assignment table is actually `app_policy_assignments` (V28); this PR adds the column there.

## Test plan
- [ ] CI green (the existing api.test suite is the migration-isolation gate; image-(N-1) code runs against DB-at-V56)

### DIFF
--- a/api/resources/db/migration/V56__app_policy_assignments_allowed_during_schedule_block.sql
+++ b/api/resources/db/migration/V56__app_policy_assignments_allowed_during_schedule_block.sql
@@ -1,0 +1,22 @@
+-- #1679: Per-allowed-app toggle "allowed during schedule block".
+--
+-- Adds a per-(profile, app) flag to app_policy_assignments controlling
+-- whether the app's host-set remains in the snapshot's per-MAC extraAllowed
+-- set during a Schedule-reason block (blockReason = Schedule). Other block
+-- reasons (Paused / TimeLimit / Manual) ignore this toggle.
+--
+-- The wire shape is UNCHANGED: this is consumed entirely server-side in
+-- PolicyService snapshot assembly (per AGENTS.md "Express new policy via
+-- existing functional fields"). The router still sees extraAllowed; it just
+-- contains different hosts during a schedule window when this is false.
+--
+-- DEFAULT TRUE: existing assignments are preserved as "always allowed", which
+-- matches current behavior (extraAllowed beats @blocked_macs, see #421). The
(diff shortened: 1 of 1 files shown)


sameerparekh:
## Independent PR review

Reviewed: `f5b4e19` (schema-only — single Flyway migration adding
`app_policy_assignments.allowed_during_schedule_block BOOLEAN NOT NULL DEFAULT TRUE`).

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
(none)

### Dimension notes
- **§4 backwards compat** — `DEFAULT TRUE` means image-(N-1) `INSERT INTO app_policy_assignments (app_id, profile_id, mode, daily_minutes, exempt_from_daily) VALUES …` statements (the shape used by V35 seeding and existing repo writes) keep working unchanged. Additive ✓.
- **§5 migrations** — `app_policy_assignments` is authored config (one row per (profile, app)), not on the unbounded-growth list (`traffic_reports` / `connection_events` / `block_events` / rollups), so `ALTER TABLE … ADD COLUMN … DEFAULT TRUE` is metadata-only on Postgres 11+ and not a startup-critical-path risk. PR is isolated to a single `*.sql` per `check-migration-isolation.sh`.
- **§3 architectural invariants** — wire shape unchanged; toggle is consumed server-side per the "express new policy via existing functional fields" invariant. Filename and version (`V56__…`) are unique against `main` (`V55` is current head).
- **§1 / §2 / §6 / §8** — N/A for a schema-only PR; the existing `api.test` Flyway-replay suite is the migration-isolation gate.

### VERDICT: APPROVE
Schema-only migration adds a per-(profile, app) `allowed_during_schedule_block` column with a back-compatible default. Phase 1 of #1679 as specified; the code adoption that consumes it ships separately. Nothing merge-gating.

PR #1697, opened 2026-06-13
feat(#1654): wifihaven.sni.enabled UCI toggle for SNI sidecar

### LINKED ISSUE #1654: feat(#573): UCI toggle to enable/disable the SNI sidecar per-router
Follow-up to #573 / PR #1643. v1 ships the `wifihaven-sni-tail` sidecar **always-on** (procd-supervised, no off switch short of removing the file). On a misconfigured or constrained router (non-default LAN bridge, low RAM, hardware-offload that evicts the first packet) an operator has no clean way to disable it.

**Goal:** add a UCI option (e.g. `wifihaven.sni.enabled`, default 1) read at sidecar startup and honored by the init script's procd instance, so the sidecar can be turned off per-router without uninstalling. Keep it a local agent surface — NOT a policy-snapshot field (router stays a dumb applier; this is device-local config, like other UCI keys).

Pairs with the IFACE-probe work already landed in PR #1643 (`sni.lan_device`).

sameerparekh:
Closes #1654.

The SNI sidecar (`wifihaven-sni-tail`, #573 / PR #1643) ships always-on.
On a misconfigured or constrained router an operator had no clean
escape hatch short of removing the file. This adds a UCI toggle:

```
uci set wifihaven.sni.enabled=0
uci commit wifihaven
/etc/init.d/wifihaven restart
```

- Default `1` — current behavior preserved.
- Gated in `start_service` so procd doesn't keep respawning nothing.
- Logs `sni sidecar disabled via uci (wifihaven.sni.enabled=0)` at
  startup when off, so ops can see why no SNI rows are flowing.
- Local agent config (UCI) — **NOT** added to the policy snapshot.
  Router stays a dumb applier per AGENTS.md §1.

## Test plan

- [x] `sh openwrt/test/init_sni_toggle_spec.sh` — RED→GREEN, covers
  default-unset / `enabled=1` / `enabled=0` cases plus
  "other sidecars unaffected"
- [x] `sh openwrt/test/init_spec.sh` — existing init contract still passes
- [ ] Manual on a live OpenWRT: `uci set wifihaven.sni.enabled=0 &&
  /etc/init.d/wifihaven restart && logread | grep sni` shows the
  disable line; `pgrep wifihaven-sni-tail` returns nothing

### DIFF
--- a/docs/ops/hostname-attribution.md
+++ b/docs/ops/hostname-attribution.md
@@ -81,6 +81,19 @@ Mitigations: see [#572](https://github.com/wifihaven/wifihaven/issues/572)
 (block DoH/DoT egress) and [#573](https://github.com/wifihaven/wifihaven/issues/573)
 (recover hostname from TLS SNI).
 
+The SNI sidecar (`wifihaven-sni-tail`) can be disabled per-router on
+constrained or misconfigured hardware via the `wifihaven.sni.enabled`
+UCI option (default `1`):
+
+```
+uci set wifihaven.sni.enabled=0
+uci commit wifihaven
+/etc/init.d/wifihaven restart
+```
+
+This is local agent config, not a policy-snapshot field — the API and
+other routers are unaffected. ([#1654](https://github.com/wifihaven/wifihaven/issues/1654))
+
 ### 4. Connection-reuse / QUIC sessions
 
 HTTP/3 (QUIC) maintains a single UDP "connection" that the app reuses
--- a/openwrt/files/etc/config/wifihaven
+++ b/openwrt/files/etc/config/wifihaven
@@ -52,3 +52,13 @@ config wifihaven 'wifihaven'
 	# buckets at report boundaries (e.g. 10 with usage_report_interval=60
 	# yields 6 samples per report). See docs/router-tuning.md.
 	option activity_sample_int '10'
+
+# SNI sidecar (wifihaven-sni-tail) toggle. The sidecar captures the first
+# packet of every outbound TCP/443 flow via tcpdump-mini to attribute
+# DoH / hard-coded-IP traffic to a hostname (#573). On constrained or
+# misconfigured hardware an operator can disable it per-router without
+# uninstalling: `uci set wifihaven.sni.enabled=0 && /etc/init.d/wifihaven
+# restart`. This is local agent config — NOT a policy-snapshot field.
+# (#1654)
+config sni 'sni'
+	option enabled '1'
--- a/openwrt/files/etc/config/wifihaven
+++ b/openwrt/files/etc/config/wifihaven
@@ -53,9 +53,12 @@ config wifihaven 'wifihaven'
 	# yields 6 samples per report). See docs/router-tuning.md.
 	option activity_sample_int '10'
 
-# SNI sidecar (wifihaven-sni-tail) toggle. The sidecar captures the first
-# packet of every outbound TCP/443 flow via tcpdump-mini to attribute
-# DoH / hard-coded-IP traffic to a hostname (#573). On constrained or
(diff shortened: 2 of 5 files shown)


sameerparekh:

Reviewed `git diff origin/main...HEAD` against the 8 checklist dimensions.

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `openwrt/test/init_sni_toggle_spec.sh:105` — the grep assertion `grep -q 'sni.*enabled' "$INIT"` is broad enough to match incidental text; tightening to `grep -q 'config_get .* sni enabled' "$INIT"` (or a similar literal) would more precisely pin the contract the test cares about. Cosmetic.
2. `openwrt/files/etc/config/wifihaven:60` — the new `config sni 'sni'` block sits outside the existing `config wifihaven 'wifihaven'` section; that's correct (separate named section, parsed by `config_get … sni enabled`) and matches OpenWRT idiom, but a one-line comment noting it's a distinct section from the wifihaven anchor would help future editors. Cosmetic.

## Dimension notes
- **1. Single source of truth** — N/A. No decision/quantity is duplicated; this is a pure local config gate around an existing procd instance.
- **2. Testing integrity** — TDD red→green is visible in history: `7c4fdeae` adds the failing spec (default/enabled=1 pass, enabled=0/other-sidecars-unaffected/key-grep fail), `3c8b79ea` makes it green. Existing `init_spec.sh` assertions unchanged. No forbidden mocks. `run_tests.sh` extended additively.
- **3. Architectural invariants** — Toggle lives in UCI on the router only; **not** added to the policy snapshot. Router-as-dumb-applier preserved. No new `/tmp` writes. No `Clock` / `ZIO` / config concerns (Lua/shell only).
- **4. Backwards compat** — Default `1` preserves current behavior; not a wire-shape change. Existing routers without the new `config sni 'sni'` section fall through to the default via `config_get … '1'`.
- **5. Migrations** — N/A.
- **6. Metrics + dashboard** — Borderline but acceptable: this gates an *existing* sidecar rather than introducing a new meaningful path. The sidecar itself already emits its observability surface; an operator who flips the toggle gets the `logger -t wifihaven` ack at startup, which is the right ops signal here.
- **7. Scope / hygiene** — Focused 5-file diff (init script + UCI default + test + runner + docs). No AI mentions. `Closes #1654` present. TODOs absent.
- **8. Security** — N/A.

VERDICT: APPROVE
Small, well-scoped escape hatch: `wifihaven.sni.enabled` (default 1) read in
`start_service`, sni-tail's procd instance gated on it, `logger` line on disable
for ops visibility. TDD progression clean, no wire-contract or
single-source-of-truth implications, defaults preserve current behavior.

sameerparekh:

Prior review: 3c8b79ea (VERDICT APPROVE, 2 NITs).

### Prior-findings status
| # | Finding | Status |
|---|---------|--------|
| NIT 1 | `init_sni_toggle_spec.sh:105` grep too loose | ADDRESSED — now `grep -q 'config_get .* sni enabled'` (init_sni_toggle_spec.sh:102) |
| NIT 2 | `config sni 'sni'` block needed a note about being distinct from the wifihaven anchor | ADDRESSED — header expanded at `openwrt/files/etc/config/wifihaven:56–58` |

### New findings on the delta (`3c8b79ea...HEAD`)
BLOCKERS — (none)
SHOULD-FIX — (none)
NITS — (none)

Delta is two cosmetic adjustments addressing the prior NITs; no functional change. Test suite still 5/5 green.

VERDICT: APPROVE
NITs from the first run are both ADDRESSED; the delta is comment- and
test-grep-only with no behavior change. Toggle still defaults to 1, init
script gating unchanged, no wire-contract or single-source-of-truth
impact.

PR #1700, opened 2026-06-13
feat(#1679): allowedDuringScheduleBlock toggle — per-app bedtime opt-out

### LINKED ISSUE #1679: Per-allowed-app toggle: "allow during schedule block" (default current = always-allowed)

sameerparekh:
## Summary

Closes #1679. Code adoption for the `allowed_during_schedule_block` column added by #1695 (V56 migration).

- An Allowed-mode app's host-set normally carves into `extraAllowed`, beating the whole-MAC Schedule drop (per #421). The new toggle (`allowedDuringScheduleBlock`, default `true`) lets the operator opt specific apps out of that carve-out **during Schedule-reason blocks only**.
- Paused / TimeLimit / Manual blocks are unchanged — out of scope per #1679.
- Wire contract unchanged: the router still receives `extraAllowed`; it just contains different hosts during a schedule window when the toggle is `false`.
- Surfaces the bedtime-Khan policy gap identified by #1666 / #1675 — the operator can now disable the Khan extraAllowed carve at bedtime without removing the allow assignment entirely.

## Commit history (TDD red→green)

1. **`e964dc67` RED** — `allowedDuringScheduleBlock` field through models/repos/SPA + `PolicySnapshotScheduleBlockToggleSpec` (compiles, tests 2-3 fail before green)
2. **`226aa7c7` GREEN** — Wire `isScheduleBlock` into `ProfileAppDispositions.enforcement` and both PolicyService paths (snapshot + decide). Tests pass.
3. **`717054a1`** — SPA checkbox (Allowed-mode apps) + vitest coverage + existing assertion updates
4. **`04903179`** — AGENTS.md doc note on the `extraAllowed` bullet

## Implementation

Single-source-of-truth (AGENTS.md §single-source-of-truth): the filter logic lives in ONE place — `ProfileAppDispositions.enforcement(isScheduleBlock = ...)`. Both the snapshot path and the `/decision` fallback pass `isScheduleBlock` from the same source (profile state / schedule check), so they cannot diverge.

## Backwards compatibility

- Default `true` on the DB column (V56) and in all JSON decoders → zero behavior change on rollout.
- Existing SPA clients omit `allowedDuringScheduleBlock` from `PUT /api/apps/:id/profiles/:pid` → server defaults to `true`.
- Wire shape unchanged (`extraAllowed` only gets different members).

## Test plan

- [ ] `PolicySnapshotScheduleBlockToggleSpec` (5 cases: true stays, false removed, two-app split, outside schedule, paused-block unaffected)
- [ ] `ProfilesPage.test.tsx` (4 new #1679 cases: checkbox rendered, checked-state round-trip, click sends `allowedDuringScheduleBlock=false`, not rendered for blocked-mode app)
- [ ] CI Scala + web suites pass
- [ ] Backwards compat: existing `setPolicy` assertions updated to include `allowedDuringScheduleBlock:true` for mode='allowed' calls

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,7 +69,12 @@ where `BlockRules` is:
   via nftables drop on `(mac, dst ip ∈ ipset(host))`, where the ipset is
   populated by dnsmasq's `--ipset=` callback at resolution time.
 - `extraAllowed: List[Hostname]` — hosts allowed for this MAC, including
-  as a carve-out when `blocked = true`.
+  as a carve-out when `blocked = true`. An Allowed-mode app's host-set
+  normally lands here; when the app's `allowedDuringScheduleBlock = false`
+  (`app_policy_assignments.allowed_during_schedule_block`, V56, #1679) and
+  the profile's block reason is `Schedule`, `PolicyService` omits those
+  hosts so the schedule drop applies. Paused/TimeLimit/Manual blocks are
+  unaffected (Schedule scope only per #1679).
 - `blocklistIds: List[BlocklistId]` — category lists that apply to this
   MAC. Enforced via nftables ipset drop, **not** via RPZ or any DNS-layer
   mechanism. The agent fetches each blocklist by URL (cached by version)
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -1117,8 +1117,10 @@ class AppTimeLimitRepoLive(xa: Transactor[Task]) extends AppTimeLimitRepo {
   // "no per-app limit configured" and is treated as never-exhausted by the exempt carve-out gate
   // in `PolicyService.exemptUnderCapHosts`; the COALESCE silently collapsed that case to 0 and
   // made the exempt+no-limit carve unreachable.
+  // #1679: `apa.allowed_during_schedule_block` carried straight through so the
+  // `ProfileAppDispositions` fold can suppress the extraAllowed carve during Schedule blocks.
   private type R =
-    (ProfileId, String, Option[Int], String, Boolean, AppId, AppPolicyAssignmentId, String)
+    (ProfileId, String, Option[Int], String, Boolean, AppId, AppPolicyAssignmentId, String, Boolean)
   private def toS(r: R) =
     AppTimeLimit(
       AppTimeLimitId(0L),
@@ -1130,12 +1132,14 @@ class AppTimeLimitRepoLive(xa: Transactor[Task]) extends AppTimeLimitRepo {
       r._6,
       AppMode.parse(r._8).getOrElse(AppMode.TimeLimited),
       r._7,
+      r._9,
     )
 
   def listForProfile(pid: ProfileId) =
     DbMetrics.timed("appTimeLimit.listForProfile")(
       sql"""SELECT apa.profile_id, ah.host, apa.daily_minutes, a.slug,
-                   apa.exempt_from_daily, a.id, apa.id, apa.mode::text
+                   apa.exempt_from_daily, a.id, apa.id, apa.mode::text,
+                   apa.allowed_during_schedule_block
             FROM app_policy_assignments apa
             JOIN apps a       ON a.id = apa.app_id
             JOIN app_hosts ah ON ah.app_id = apa.app_id
@@ -1149,7 +1153,8 @@ class AppTimeLimitRepoLive(xa: Transactor[Task]) extends AppTimeLimitRepo {
 
   def listAll =
     sql"""SELECT apa.profile_id, ah.host, apa.daily_minutes, a.slug,
-                 apa.exempt_from_daily, a.id, apa.id, apa.mode::text
+                 apa.exempt_from_daily, a.id, apa.id, apa.mode::text,
+                 apa.allowed_during_schedule_block
             FROM app_policy_assignments apa
             JOIN apps a       ON a.id = apa.app_id
             JOIN app_hosts ah ON ah.app_id = apa.app_id
@@ -2953,6 +2958,8 @@ trait AppRepo {
       mode: AppMode,
       dailyMinutes: Option[Int],
       exemptFromDaily: Boolean,
+      // #1679: default true for back-compat — all existing callers keep current behavior.
+      allowedDuringScheduleBlock: Boolean = true,
   ): Task[AppPolicyAssignmentId]
   def deleteAssignment(appId: AppId, profileId: ProfileId): Task[Unit]
   def listAssignmentsForApp(appId: AppId): Task[List[AppPolicyAssignment]]
@@ -3084,14 +3091,18 @@ class AppRepoLive(xa: Transactor[Task]) extends AppRepo {
       mode: AppMode,
       dailyMinutes: Option[Int],
       exemptFromDaily: Boolean,
+      allowedDuringScheduleBlock: Boolean = true,
   ) =
     sql"""INSERT INTO app_policy_assignments
-            (app_id, profile_id, mode, daily_minutes, exempt_from_daily)
-          VALUES ($appId, $profileId, ${AppMode.asString(mode)}, $dailyMinutes, $exemptFromDaily)
+            (app_id, profile_id, mode, daily_minutes, exempt_from_daily,
+             allowed_during_schedule_block)
+          VALUES ($appId, $profileId, ${AppMode.asString(mode)}, $dailyMinutes, $exemptFromDaily,
+                  $allowedDuringScheduleBlock)
           ON CONFLICT (app_id, profile_id) DO UPDATE SET
             mode = EXCLUDED.mode,
             daily_minutes = EXCLUDED.daily_minutes,
-            exempt_from_daily = EXCLUDED.exempt_from_daily
+            exempt_from_daily = EXCLUDED.exempt_from_daily,
+            allowed_during_schedule_block = EXCLUDED.allowed_during_schedule_block
           RETURNING id"""
       .query[AppPolicyAssignmentId]
       .unique
@@ -3103,12 +3114,13 @@ class AppRepoLive(xa: Transactor[Task]) extends AppRepo {
       .unit
 
   private type AR =
-    (AppPolicyAssignmentId, AppId, ProfileId, AppMode, Option[Int], Boolean)
+    (AppPolicyAssignmentId, AppId, ProfileId, AppMode, Option[Int], Boolean, Boolean)
   private def toAssignment(r: AR) =
-    AppPolicyAssignment(r._1, r._2, r._3, r._4, r._5, r._6)
+    AppPolicyAssignment(r._1, r._2, r._3, r._4, r._5, r._6, r._7)
 
   def listAssignmentsForApp(appId: AppId) =
-    sql"""SELECT id,app_id,profile_id,mode,daily_minutes,exempt_from_daily
+    sql"""SELECT id,app_id,profile_id,mode,daily_minutes,exempt_from_daily,
+                 allowed_during_schedule_block
           FROM app_policy_assignments WHERE app_id=$appId ORDER BY id"""
       .query[AR]
       .map(toAssignment)
@@ -3117,7 +3129,8 @@ class AppRepoLive(xa: Transactor[Task]) extends AppRepo {
 
   def listAssignmentsForProfile(profileId: ProfileId) =
     DbMetrics.timed("app.listAssignmentsForProfile")(
-      sql"""SELECT id,app_id,profile_id,mode,daily_minutes,exempt_from_daily
+      sql"""SELECT id,app_id,profile_id,mode,daily_minutes,exempt_from_daily,
+                   allowed_during_schedule_block
           FROM app_policy_assignments WHERE profile_id=$profileId ORDER BY id"""
         .query[AR]
         .map(toAssignment)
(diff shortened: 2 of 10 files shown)


sameerparekh:
## Independent PR Review — #1700

**Reviewed SHA:** `04903179`

---

### 1. Duplicated logic / single source of truth

✅ Filter logic lives in exactly one place: `ProfileAppDispositions.enforcement` (the `suppressedByScheduleToggle` block at lines 102-116). Both the snapshot path (`PolicyService.scala:132`) and the `/decide` fallback path (`PolicyService.scala:147`) call `enforcement(isScheduleBlock = ...)`. No re-derivation.

⚠️ One subtle divergence worth noting (see SHOULD-FIX below): the two paths compute `isScheduleBlock` differently — snapshot reads `state.blockReason.contains(MacBlockReason.Schedule)`, while `/decide` reads `scheduleBlock(scheds, now).nonEmpty`. These disagree for a Paused+active-schedule profile: snapshot returns `isScheduleBlock=false` (Paused outranks Schedule in `blockReason` precedence), `/decide` returns `isScheduleBlock=true` (schedule window is actually active). Harmless in practice because when the snapshot has the app in `extraAllowed`, nftables allows the connection and the block page / `/decide` is never reached.

---

### 2. Testing integrity

✅ TDD red→green visible in commit history: `e964dc67` (red — types + test compiling but no filter logic), `226aa7c7` (green — filter wired).

✅ 5 feature test cases in `PolicySnapshotScheduleBlockToggleSpec` covering:
- `true` app stays in `extraAllowed` during schedule block
- `false` app removed from `extraAllowed` during schedule block
- Two-app split (one of each in the same profile at bedtime)
- `false` app OUTSIDE schedule block → still in `extraAllowed` (toggle is schedule-only)
- `false` app during Paused block (not schedule) → still in `extraAllowed` (scope correct)

✅ 4 new SPA vitest cases in `ProfilesPage.test.tsx` covering checkbox rendering, checked-state round-trip, click sends `allowedDuringScheduleBlock=false`, not rendered for blocked-mode.

✅ 5 existing setPolicy assertions updated to include `allowedDuringScheduleBlock: true` — the old assertions were stale; the updates are correct, not weakening.

---

### 3. Architectural invariants

✅ Router is dumb applier — no policy logic added to openwrt/opnsense agents.

✅ Snapshot stays minimal wire vocabulary — `allowedDuringScheduleBlock` is consumed server-side in `PolicyService`, the router sees a different `extraAllowed` list but no new wire fields.

✅ `extraAllowed` beats every block path — behavior for `allowedDuringScheduleBlock=true` (the default) is completely unchanged; the toggle only suppresses the carve-out.

---

### 4. Backwards compatibility

✅ DB column `DEFAULT TRUE` (V56 in #1695, already merged). All existing rows default to `true` → zero behavior change.

✅ All Scala types default to `true`; `UpsertAppAssignmentRequest.allowedDuringScheduleBlock: Option[Boolean] = None` — omitted by existing SPA clients, server defaults to `true`.

✅ TS types: `allowedDuringScheduleBlock?: boolean` — optional, omitted by older APIs.

✅ Wire shape unchanged.

---

### 5. Migrations

No new migration in this PR (migration shipped in PR #1695). ✅

---

### 6. Metrics + dashboard

No new route, background job, or external call. The toggle is a conditional inside the existing `enforcement` fold, covered by the existing snapshot generation path. No new metric required. ✅

---

### 7. Scope / hygiene

✅ Focused diff; no drive-by edits. All files touch the feature directly.

---

### 8. Security

✅ No secrets. `$allowedDuringScheduleBlock` in Doobie is parameterized. Route auth unchanged (`requireWriter` at `AppRoutes.scala:217`).

---

## BLOCKERS

(none)

## SHOULD-FIX

1. `api/src/policy/PolicyService.scala` line ~407 (`decide` path) — `val isScheduleBlock = scheduleBlock(scheds, now).nonEmpty` fires even when `p.paused` is true and a schedule is also active, whereas the snapshot path would have `isScheduleBlock=false` (Paused outranks Schedule in `state.blockReason`). This is benign in practice (block page never called if `extraAllowed` has the app), but without a comment a future reader might think the two paths intentionally differ. Suggest adding: `// #1679: schedule check is safe to use independently of paused state — if the MAC is not actually // dropped by nftables (because extraAllowed contains the app), /decide is never called, so // the Paused+Schedule edge case where blockReason=Paused (not Schedule) never surfaces.`

## NITS

1. `api/test/src/feature/PolicySnapshotScheduleBlockToggleSpec.scala` line 32 — `private val UTC = ZoneId.of("UTC")` is unused (none of the tests use it directly). Minor cleanup.

2. `web/src/pages/ProfilesPage.tsx` line ~700 — Comment on why the checkbox only shows for `mode === 'allowed'` (vs. also for `AllowedDuring` rule apps whose base mode is `Blocked`) would help future readers. Known limitation, not a bug.

---

## VERDICT: APPROVE

#1679 adds `allowedDuringScheduleBlock` toggle per-app-assignment, consumed server-side in `ProfileAppDispositions.enforcement` (one place, per §single-source-of-truth). Wire contract unchanged. Default `true` means zero behavior change on rollout. TDD red→green visible; 5 Scala feature tests + 4 SPA vitest cases cover the key scenarios. SHOULD-FIX #1 is a comment-only clarification, non-blocking.

sameerparekh:
## Independent PR Review — #1700 (re-run)

**Prior reviewed SHA:** `04903179` → **Current HEAD:** `ecf38af4`

---

### Prior findings status

| # | Classification | Finding | Status |
|---|---|---|---|
| SHOULD-FIX 1 | SHOULD-FIX | `PolicyService.scala` decide path: missing comment explaining Paused+Schedule edge case divergence | **ADDRESSED** — comment added at lines 407-414 explaining why `scheduleBlock(scheds, now).nonEmpty` is safe despite the Paused outranking precedence |
| NIT 1 | NIT | `PolicySnapshotScheduleBlockToggleSpec.scala:32` — unused `UTC` variable | **ADDRESSED** — `private val UTC = ZoneId.of("UTC")` removed; unused `ZoneId` import also removed |
| NIT 2 | NIT | `ProfilesPage.tsx` — no comment on why checkbox is `mode === 'allowed'` only | NOT-ADDRESSED (kept as NIT scope, no push back from author; acceptable given the constraint is already described by the PR body and issue #1679 scope) |

---

### Incremental delta (`04903179...ecf38af4`)

**`api/src/policy/PolicyService.scala`** (+6 lines): Added explanatory comment on the `isScheduleBlock` computation in the decide path. No logic change. Correct and complete. ✅

**`api/test/src/feature/PolicySnapshotScheduleBlockToggleSpec.scala`** (−2 lines): Removed unused `UTC` variable and `ZoneId` import. No test logic change. ✅

No new findings.

---

## BLOCKERS

(none)

## SHOULD-FIX

(none — prior SHOULD-FIX addressed)

## NITS

1. `web/src/pages/ProfilesPage.tsx` ~line 700 — as noted in prior run: checkbox only shows for `mode === 'allowed'`, not for `AllowedDuring`-rule apps. Expected scope limitation per #1679. No action required, just awareness.

---

## VERDICT: APPROVE

Prior SHOULD-FIX addressed (explanatory comment on Paused+Schedule edge case); prior NIT #1 addressed (unused variable removed). Delta is documentation-only with no logic change. All 5 Scala feature tests and 4 SPA vitest cases remain in place. Implementation is correct; wire contract unchanged; TDD red→green in history.

sameerparekh:
## Independent PR Review — #1700 (re-run 2)

**Prior reviewed SHA:** `ecf38af4` → **Current HEAD:** `f42616ee`

---

### Prior findings status

| # | Classification | Finding | Status |
|---|---|---|---|
| SHOULD-FIX 1 | SHOULD-FIX | Paused+Schedule edge case comment missing in `decide` path | ADDRESSED (prior re-run) |
| NIT 1 | NIT | Unused `UTC` variable | ADDRESSED (prior re-run) |
| NIT 2 | NIT | No comment on `mode === 'allowed'` scope for checkbox | NOT-ADDRESSED (acceptable) |

---

### Incremental delta (`ecf38af4...f42616ee`)

**`api/test/src/feature/PolicySnapshotScheduleBlockToggleSpec.scala`** (+2, -1):
- Added `import zio.test.TestAspect`
- Added `@@ TestAspect.sequential` to the spec's suite

This is the correct fix. All other feature specs that call `cleanAndMigrate` per-test already use `@@ TestAspect.sequential` (e.g. `PolicySnapshotAppsSpec:846`, `PolicySnapshotPerAppScheduleSpec:366`) for the same reason: without it, ZIO Test runs the 5 tests concurrently, two parallel `cloneTemplateInto(pg, sameDbName)` calls race on `DROP DATABASE / CREATE DATABASE` for the same per-spec DB name, and the second `CREATE` hits a `pg_database_datname_index` duplicate-key error.

No logic change. The fix is purely a test-infra concurrency guard. ✅

No new findings.

---

## BLOCKERS

(none)

## SHOULD-FIX

(none)

## NITS

(none new — prior NIT 2 remains acceptable)

---

## VERDICT: APPROVE

All prior BLOCKERs/SHOULD-FIX addressed. CI failure fixed by adding `@@ TestAspect.sequential` — consistent with every other feature spec that uses `cleanAndMigrate`. Delta is a one-line test-infra fix with no logic change.

PR #1701, opened 2026-06-13
feat(#1676): presence_app_sessions_dropped_total counter + Grafana panel (#1666 observability follow-up)

### LINKED ISSUE #1676: Observability: counter + Grafana panel for per-app phantom-session suppression (#1666 follow-up)
Follow-up to #1666 / PR #1675.

The #1666 fix adds a per-session anchor-row requirement inside `Presence.appSpansForProfile`. The PR invoked the AGENTS.md §instrument-new-functionality escape hatch ("pure threshold inside existing code") because emitting a counter through the otherwise-pure `Presence.*` primitives would have required threading either a Ref observer or a tuple return through several callers — not in scope for the urgent fix.

This issue is the follow-up to close that observability gap.

## Proposed shape

- `presence_app_sessions_dropped_total{reason="no_anchor"}` — unlabelled counter (cardinality-firewall safe).
- Emit from `TimeStatusService.appSecondsByApp` via a sibling `Presence.appSpansForProfileWithDropCount` that returns `(Map[String, List[Span]], Int)`. Keep `appSpansForProfile` as a thin delegating alias.
- Add to `MetricGuard.Allowed` with empty label set.
- Grafana panel: extend a presence/usage dashboard under `deploy/grafana/dashboards/` with `rate(presence_app_sessions_dropped_total[5m])`.

## Why

Without a series, operators can't rate-alert on "threshold too aggressive" (false-positive spike of real sessions dropped) or "threshold too lax" (zero drops while phantom inflation persists).

sameerparekh:
Closes #1676.

Follow-up to #1666 / PR #1675 — that fix added a per-session anchor-row requirement inside `Presence.appSpansForProfile` to suppress phantom sessions stitched from pure-keepalive cadences on real-app hosts. It invoked the AGENTS.md §instrument-new-functionality escape hatch ("pure threshold inside existing code") because emitting a counter through the otherwise-pure `Presence` primitives would have required threading a `Ref` observer or a tuple return through several callers — not in scope for the urgent fix.

This PR closes that observability debt.

## Changes

- `Presence.appSpansForProfileWithDropCount`: sibling returning `(Map[String, List[Span]], Int)` — the per-(mac, app, session) count of sessions silently dropped by the anchor guard. The existing `appSpansForProfile` is now a thin alias dropping the count (single-source-of-truth).
- `TimeStatusService.appSecondsByAppWithDropCount`: per-profile sibling threading the count out at the app boundary.
- `AppMetrics.recordAppSessionsDropped` + `MetricGuard.Allowed` entry for `presence_app_sessions_dropped_total` — unlabelled counter (the guard is unconditional; per-mac / per-app would breach the cardinality firewall).
- Both consumers (`TimeUsedRollupJob.doTick` for the periodic roll, `AppUsedRollupService.aggregate` for the per-app read path) call the sibling and emit the count, so each tick and each read surfaces the metric.
- Grafana: two panels appended to `deploy/grafana/dashboards/data-quality-ingest.json` — a 5m rate timeseries and a 24h stat. PromQL targets the emitted name verbatim (zio-prometheus does not append `_total`).

## Observability story

Operators can rate-alert once a soak baseline lands:

- A sustained rise means the anchor threshold (`Presence.AppSessionAnchorBytes = 256`) is too aggressive — real sessions vanishing.
- A flat zero while phantom inflation returns (the original #1666 symptom) means it is too lax.

Suggested alert later: `> 2× baseline sustained for 10 min`. The alert rule itself is intentionally a follow-up so the threshold can be tuned against real prod data.

## Test plan

- [x] `mill api.test && mill shared.test` — green
- [x] `scalafmt --check --non-interactive` — green
- [x] `python3 -m json.tool deploy/grafana/dashboards/*.json` — green
- [x] New PresenceSpec tests cover the with-drop-count sibling (anchored session ⇒ 0, un-anchored ⇒ 1, per-(mac, app) granularity, delegation invariant)

### DIFF
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -100,6 +100,14 @@ object MetricGuard {
     // household-scoped gauges set each time the policy snapshot is assembled.
     "wifihaven_global_allow_hosts"              -> Set.empty[String],
     "wifihaven_default_deny_profiles"           -> Set.empty[String],
+    // #1676 — per-(mac, app) sessions dropped by the #1666 phantom-suppression
+    // guard inside Presence.appSpansForProfile. Unlabelled counter: the guard
+    // is unconditional, so a reason enum would only ever carry one value, and
+    // any per-mac / per-app label would blow the cardinality firewall. Operators
+    // rate-alert on threshold drift — a sustained rise means the threshold is
+    // too aggressive (real sessions vanishing), a flat zero while phantom
+    // inflation returns means it is too lax.
+    "presence_app_sessions_dropped_total"       -> Set.empty[String],
     // §5.1 router-sourced, pushed via POST /api/router/metrics (#1205). Every one carries the
     // server-attached `router_id` + `installation_id` plus its own bounded enum label.
     "dnsmasq_restarts_total"                    -> Set("reason", "router_id", "installation_id"),
@@ -304,6 +312,23 @@ object AppMetrics {
     ZIO
       .when(count > 0)(
         MetricGuard.counter("usage_records_rejected_total", Map("reason" -> reason), count.toLong),
+      )
+      .unit
+
+  // ── #1676: phantom-suppression drop count ───────────────────────────────────
+  // Emitted from TimeStatusService.appSecondsByAppWithDropCount, which threads
+  // the count out of Presence.appSpansForProfileWithDropCount (#1666 anchor-row
+  // guard). Unlabelled — the guard is unconditional, and per-mac / per-app
+  // would breach the cardinality firewall.
+
+  def recordAppSessionsDropped(count: Int): UIO[Unit] =
+    ZIO
+      .when(count > 0)(
+        MetricGuard.counter(
+          "presence_app_sessions_dropped_total",
+          Map.empty,
+          count.toLong,
+        ),
       )
       .unit
 
--- a/api/src/policy/TimeStatusService.scala
+++ b/api/src/policy/TimeStatusService.scala
@@ -577,20 +577,35 @@ object TimeStatusService {
       appLimits: List[AppTimeLimit],
       presence: List[PresenceRow],
       settings: HouseholdSettings,
-  ): Map[AppId, Long] = {
-    val groups       = groupAppLimits(appLimits)
-    val labelToAppId = groups.map(g => g._2 -> g._1).toMap
-    Presence
-      .appSecondsForProfile(
-        presence,
-        groups.map(g => g._2 -> g._5),
-        profile.crossDeviceOverlapMode,
-        settings.heartbeatFilter,
-        settings.presenceContinuationSeconds,
-      )
-      .iterator
-      .flatMap { case (label, secs) => labelToAppId.get(label).map(_ -> secs) }
-      .toMap
+  ): Map[AppId, Long] =
+    appSecondsByAppWithDropCount(profile, appLimits, presence, settings)._1
+
+  /**
+   * #1676: sibling that also returns the count of per-(mac, app) sessions silently dropped by the
+   * #1666 phantom-suppression anchor-row guard inside Presence.appSpansForProfile. The pure
+   * [[appSecondsByApp]] alias above projects to the map only; consumers that want the observability
+   * counter call this and feed the count into [[AppMetrics.recordAppSessionsDropped]].
+   */
+  def appSecondsByAppWithDropCount(
+      profile: Profile,
+      appLimits: List[AppTimeLimit],
+      presence: List[PresenceRow],
+      settings: HouseholdSettings,
+  ): (Map[AppId, Long], Int) = {
+    val groups        = groupAppLimits(appLimits)
+    val labelToAppId  = groups.map(g => g._2 -> g._1).toMap
+    val (spans, drop) = Presence.appSpansForProfileWithDropCount(
+      presence,
+      groups.map(g => g._2 -> g._5),
+      profile.crossDeviceOverlapMode,
+      settings.heartbeatFilter,
+      settings.presenceContinuationSeconds,
+    )
+    val secsByLabel   = spans.view.mapValues(_.iterator.map(_.seconds).sum).filter(_._2 > 0L).toMap
+    val byAppId       = secsByLabel.iterator.flatMap { case (label, secs) =>
+      labelToAppId.get(label).map(_ -> secs)
+    }.toMap
+    (byAppId, drop)
   }
 
   /**
--- a/api/src/policy/TimeStatusService.scala
+++ b/api/src/policy/TimeStatusService.scala
@@ -592,20 +592,19 @@ object TimeStatusService {
       presence: List[PresenceRow],
       settings: HouseholdSettings,
   ): (Map[AppId, Long], Int) = {
-    val groups        = groupAppLimits(appLimits)
-    val labelToAppId  = groups.map(g => g._2 -> g._1).toMap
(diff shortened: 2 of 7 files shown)


sameerparekh:
# Independent PR review (#1701) — first pass

Reviewed `git diff origin/main...HEAD` against the 8 dimensions in [`docs/pr-review-checklist.md`](docs/pr-review-checklist.md).

## BLOCKERS

*(none)*

## SHOULD-FIX

1. **[api/src/usage/AppUsedRollupService.scala:152-154](api/src/usage/AppUsedRollupService.scala:152) — counter is emitted on every read of the per-app aggregate, which inflates the series with read frequency rather than reflecting data state.** `aggregate` is invoked from [TimeStatusService.scala:288](api/src/policy/TimeStatusService.scala:288) and [:347](api/src/policy/TimeStatusService.scala:347) — both on the `/api/time/status/summary` hot read path. Each read re-aggregates the *same* presence rows for today and re-detects the *same* phantom drops, then emits them into an unlabelled cumulative counter. A household whose presence data has 5 phantom drops today will see those 5 increments charged on every read — say 100×/day × 5 = 500/day from reads alone — on top of the ~96/day rollup ticks (15-min cadence). Operators rate-alerting on this series will see a signal dominated by read load, not by threshold drift, which is exactly what this metric is supposed to surface. Recommend dropping the `AppMetrics.recordAppSessionsDropped(dropped)` call from `AppUsedRollupService.aggregate` and emitting *only* from `TimeUsedRollupJob.doTick` — that's one clean cadence the dashboard PromQL implicitly assumes. Then `rate(...[5m])` means "drops detected per rollup tick" — interpretable.

2. **[deploy/grafana/dashboards/data-quality-ingest.json](deploy/grafana/dashboards/data-quality-ingest.json) — JSON was wholesale re-flowed from compact-inline to expanded format, ballooning the diff from ~50 semantic lines to ~700.** The actual change is two appended panels (`Phantom app sessions dropped — rate (#1676)` + the 24h stat); everything else is `json.dump(..., indent=2)` re-expanding single-line objects (`"datasource": {...}`, `"gridPos": {...}`, threshold step objects). Reviewers can't easily tell what changed without re-rendering. Recommend preserving the original compact formatting — insert the two new panel objects in the existing one-liner style. Not merge-gating, but it makes future review of this dashboard noisy until someone restores it.

3. **[api/src/policy/TimeStatusService.scala:595-606](api/src/policy/TimeStatusService.scala:595) — `appSecondsByAppWithDropCount` re-implements the spans→seconds projection that [`Presence.appSecondsForProfile`](api/src/presence/Presence.scala:683) already owns**, duplicating `_.iterator.map(_.seconds).sum` + `.filter(_._2 > 0L)`. The old `appSecondsByApp` correctly delegated to `Presence.appSecondsForProfile`; the new sibling reaches under it to `appSpansForProfileWithDropCount` and re-derives the seconds locally. Per [AGENTS.md §single-source-of-truth](AGENTS.md#single-source-of-truth) this projection should live in one place. Cleanest fix: add `Presence.appSecondsForProfileWithDropCount(rows, groups, …): (Map[String, Long], Int)` as the sibling, and have `appSecondsForProfile` delegate to it (same shape as `appSpansForProfile` → `appSpansForProfileWithDropCount`). Then `TimeStatusService.appSecondsByAppWithDropCount` calls that and just re-keys to `AppId`. Today it works; the next zero-second-filter tweak in `Presence.appSecondsForProfile` will quietly drift this path.

## NITS

1. **[api/src/presence/Presence.scala:660,687-688](api/src/presence/Presence.scala:660) — `var droppedTotal` mutable accumulator inside an otherwise-pure primitive.** Functionally correct (local to the call), but the rest of `Presence` is studiously pure-functional. A `groups.iterator.map { ... -> (key -> spans, dropped) }.toList` collected into `(spansMap, sumOfDrops)` would match the surrounding style.

2. **[api/src/usage/TimeUsedRollupJob.scala:213](api/src/usage/TimeUsedRollupJob.scala:213) — `computeRolls` widened from a 2-tuple to a 3-tuple.** A named `final case class ComputedRolls(perProfile, perApp, sessionsDropped)` would make `rolls._3` self-documenting and prevent positional-arg drift later.

3. **[api/src/usage/AppUsedRollupService.scala:152-160](api/src/usage/AppUsedRollupService.scala:152) — for-comprehension destructuring `(liveSecs, dropped) = TimeStatusService.appSecondsByAppWithDropCount(...)` mid-comprehension scans fine in Scala 3 but is unusual for this codebase.** If you address SHOULD-FIX #1 by dropping the emit here, this resolves itself.

## VERDICT: REQUEST-CHANGES

Small, focused implementation of the #1666 observability follow-up: new counter + sibling primitive + Grafana panel, single-source-of-truth-preserving span alias, TDD red→green visible (`bf181c66` red, `65de0c1f` green), bounded labels, cardinality firewall entry added, tests pin the new behavior.

REQUEST-CHANGES is for SHOULD-FIX #1 — the double-emit on the hot read path makes the metric track read load instead of drop frequency, defeating the purpose of the panel. Once that emit is dropped (and ideally #2's dashboard reformatting + #3's seconds-projection collapse are addressed), the change lands cleanly.

sameerparekh:
# Independent PR review (#1701) — re-run

Prior reviewed-sha: `65de0c1f`. Delta: `git diff 65de0c1f...HEAD` — three files in `api/src` and the dashboard JSON.

## Prior-findings status

|---|---|---|---|
| SF1 | SHOULD-FIX | Double-emit of counter on hot read path ([AppUsedRollupService.scala:152-154](api/src/usage/AppUsedRollupService.scala:148)) | **ADDRESSED** — `AppMetrics.recordAppSessionsDropped` call removed; `aggregate` now calls `TimeStatusService.appSecondsByApp` (no drop count). Counter is emitted only from `TimeUsedRollupJob.doTick` — one clean cadence. Comment on lines 152-156 explains why. |
| SF2 | SHOULD-FIX | Dashboard JSON wholesale re-flowed | **ADDRESSED** — file restored to original compact-inline format ([data-quality-ingest.json](deploy/grafana/dashboards/data-quality-ingest.json)); new panels appended in matching one-liner style. Diff shrunk from ~700 to ~80 semantic lines. |
| SF3 | SHOULD-FIX | `appSecondsByAppWithDropCount` re-implements spans→seconds projection | **ADDRESSED** — new [`Presence.appSecondsForProfileWithDropCount`](api/src/presence/Presence.scala:715) is now the canonical seconds-with-drop-count primitive; `Presence.appSecondsForProfile` delegates to it; `TimeStatusService.appSecondsByAppWithDropCount` calls THAT and only re-keys to `AppId`. Single-source-of-truth restored. |
| NIT1 | NIT | `var droppedTotal` mutable accumulator in Presence | NOT-ADDRESSED — non-blocking, see verdict. |
| NIT2 | NIT | `computeRolls` widened to 3-tuple without a named case class | NOT-ADDRESSED — non-blocking. |
| NIT3 | NIT | Unusual for-comprehension destructuring in AppUsedRollupService | **ADDRESSED** — the destructured binding is gone now that SF1 dropped the emit. |

## New findings (delta only)

### BLOCKERS
*(none)*

### SHOULD-FIX
*(none)*

### NITS
*(none new in the delta)*

## VERDICT: APPROVE

All three SHOULD-FIX resolved, NIT3 resolved as a side-effect; remaining NITs (mutable accumulator, unnamed 3-tuple) are stylistic and non-blocking. The change now ships the metric, single canonical emission cadence, single canonical span→seconds projection, and the dashboard with a reviewable diff.

Issue #572, opened 2026-05-17
Block known DoH/DoT egress to force devices onto local DNS


sameerparekh:
## Problem

Devices that bypass the LAN DNS resolver (DoH to Cloudflare 1.1.1.1, Google `dns.google`, NextDNS, Quad9, etc.; DoT on port 853) deny the agent any hostname attribution for their flows. Operators see bare-IP rows in `/api/logs` and `/api/dashboard/now` with no way to correlate them to a name. See [docs/ops/hostname-attribution.md](docs/ops/hostname-attribution.md) §3.

This also undermines `extraBlocked`: a device that resolves a blocked host via DoH bypasses the per-(MAC, host) ipset population path that `extraBlocked` relies on.

## Proposal

Ship a router-side feature that drops outbound traffic to well-known DoH/DoT endpoints, forcing devices to fall back to plain DNS through dnsmasq.

Concrete behavior:
- nftables drop rules in `inet wifihaven` table targeting:
  - **DoT**: TCP/853 to any IP (no legitimate non-DoT traffic uses 853).
  - **DoH**: TCP+UDP/443 to a curated IP set of public resolvers (Cloudflare 1.0.0.1, 1.1.1.1, Google 8.8.8.8/8.8.4.4, Quad9 9.9.9.9, NextDNS, Apple `mask.icloud.com` / `mask.apple-dns.net` ranges, OpenDNS, AdGuard).
  - Optionally also TCP+UDP/443 to AAAA equivalents.
- Curated list maintained as a static file in `openwrt/files/...` updated periodically; no per-host policy — applies router-wide.
- Probably gated behind a household-level toggle (`block_encrypted_dns: bool`) since some operators may explicitly want their devices on Cloudflare DoH.

## Detection / verification

After enabling, an iPad with iCloud Private Relay on should:
- Lose connectivity to `mask.icloud.com:443` (until Private Relay times out and disables itself).
- Show fresh `pancake.apple.com` / `gs-loc.apple.com` queries in `/tmp/wifihaven-dnsmasq.log`.
- Have most of its flow events come back as `host.type=fqdn`.

## Non-goals

- Not blocking arbitrary HTTPS — only specific DoH/DoT endpoints.
- Not breaking the agent's own outbound (`api_url`) — the API host is on the LAN.
- Not attempting TLS interception.

## References
- [docs/ops/hostname-attribution.md](docs/ops/hostname-attribution.md) §3 — context
[...]

PR #1702, opened 2026-06-13
feat(#1651): QUIC / HTTP3 Initial-packet SNI parsing (UDP 443)

### LINKED ISSUE #1651: feat(#573): QUIC / HTTP3 Initial-packet SNI parsing (UDP 443)
Follow-up to #573 / PR #1643. v1 parses TLS ClientHellos on **TCP** 443 only. Modern clients increasingly reach Google/YouTube/CDNs over **QUIC / HTTP/3 on UDP 443**, whose SNI lives in the (partially encrypted) QUIC Initial packet's CRYPTO frames.

**Goal:** extend the sidecar (or add a sibling capture branch) to extract SNI from QUIC Initial packets so HTTP/3 flows attribute by hostname too — otherwise the #1636/#1640 collateral-drop class reappears for any app that prefers QUIC. Requires QUIC header parsing + Initial-packet decryption with the version-specific salt (RFC 9001). Attribution-only; feeds the same `cache.insert_sni` primitive.

Out of scope: blocking QUIC. Tracked separately under #572 (force devices onto local DNS / block bypass).

sameerparekh:
Closes #1651.

Extends the #573 SNI hostname-attribution sidecar to cover **QUIC v1
Initial packets** (UDP/443) alongside the existing TCP/443 path. With
HTTP/3 increasingly the default for Google/YouTube/CDN traffic, the
TCP-only ClientHello parser was leaving a growing slice of DoH /
hard-coded-IP flows unattributed — the #1636/#1640 collateral-drop
class would silently reappear for QUIC-preferring apps.

This PR is attribution-only. Blocking QUIC remains #572's job.

## Architecture (per RFC 9000 + RFC 9001)

```
long-header parse (RFC 9000 §17.2)
  → HKDF-derive client Initial keys from DCID + v1 salt (RFC 9001 §5.2)
  → AES-128-ECB(hp_key, sample[16]) → 5-byte HP mask removal (RFC 9001 §5.4)
  → AAD = unprotected header bytes; nonce = IV XOR pn (RFC 9001 §5.3)
  → AES-128-GCM decrypt (PN..tag) bytes
  → CRYPTO-frame reassembly (RFC 9000 §19.6)
  → existing sni.parse_client_hello (single-source-of-truth)
  → cache.insert_sni (same writer as TCP path)
```

The new `openwrt/files/usr/lib/lua/wifihaven/quic.lua` owns parsing
and key derivation; the TLS ClientHello SNI extension walk and the
spool/cache writes are unchanged — they reuse the exact functions the
TCP path already calls, per AGENTS.md §single-source-of-truth.

## Crypto dependency choice (lua-openssl)

OpenWrt 25.12.x ships an `openssl-util` CLI but the Lua bindings are a
separate package: `lua-openssl-0.10.0.0-r1`, ~464 KiB installed. I
considered the alternatives in the issue brief:

* **Shell out to `openssl enc` / `openssl dgst` per packet** — rejected.
  The capture hot path is one parse per new flow, and `fork+exec` per
  packet is two orders of magnitude more expensive than a Lua FFI call.
* **Pure-Lua AES/HMAC** — rejected. ~500 lines of crypto we'd own and
  test, vs. ~5 lines of glue to an OpenSSL-backed binding.
* **Custom C helper** — rejected. New build artifact + new package,
  same goal as lua-openssl which is already in the upstream feed.

So **lua-openssl is added to the `wifihaven` package DEPENDS**. The
weight increase is +464 KiB (~0.1% of the typical 50 MiB router
install). The bindings are also reusable for future agent crypto work
(e.g. local block-page TLS cert generation, currently shelling out).

`quic.openssl_crypto(openssl)` builds the small contract the parser
expects (`hkdf_extract`, `hkdf_expand_label`, `aes128_ecb`,
`aes128_gcm_decrypt`). HKDF itself follows RFC 5869 §2.3 on top of
`openssl.hmac.hmac` — lua-openssl 0.9.2's `kdf.fetch("HKDF")` is
present but its params-table shape rejected every form I tried (see
\`bad argument #1 to 'derive' (empty paramaters table)\`), so I built
the standard 4-line Expand loop on HMAC-SHA256 instead. Verified
against the RFC 9001 §A.1 keys byte-for-byte on the test router.

## Test vector

`openwrt/test/quic_spec.lua` embeds the **RFC 9001 Appendix A.2**
'Client Initial' vector — DCID `8394c8f03e515708`, 1200-byte protected
packet, ClientHello SNI `example.com` — and asserts the parser
extracts that SNI end-to-end. The unit tests inject a stub `crypto`
table that asserts on its inputs and echoes the RFC-known outputs, so:

* The HP-mask offset, AAD construction, nonce derivation, and CRYPTO
  reassembly are all exercised against an externally-defined, known-
  correct vector — a bug in any of them surfaces as a stub assertion
  failure, not a silent nil result.
* The dev macOS host does not need `lua-openssl` to run the suite.

Edge-case tests cover `quic_unsupported_version` (v2 salt), `quic_not_initial`
(Handshake long header), `quic_short_header` (1-RTT, the steady-state
case), `quic_truncated`, and the IPv4/UDP framing wrapper.

**To capture your own real-browser QUIC handshake:**
```
tcpdump -i <iface> -s 2048 -w hello.pcap 'udp dst port 443 and (udp[8] & 0xf0) = 0xc0'
# open https://www.google.com in Chrome/Firefox
```
The first packet of any flow whose first byte's high nibble is `0xc0`
is a client Initial. Feed the UDP payload to `quic.parse_quic_packet`
with `quic.openssl_crypto(openssl)`.

## Live verification

Deployed `quic.lua` + the updated sidecar to the test router
(`openwrt.lan`, OpenWrt 25.12.3 `aarch64_cortex-a53`, with
`lua-openssl-0.10.0.0-r1`):

* `quic.openssl_crypto` derives the RFC 9001 §A.1 keys byte-for-byte
  (initial_secret, client_initial, key, iv, hp, sample → HP mask).
* `quic.parse_initial` on a hand-built eth/IPv4/UDP frame wrapping the
  RFC 9001 §A.2 protected client Initial returns
  `dst=8.8.8.8 mac=fa:10:cd:84:78:22 sni=example.com`.
* `wifihaven-sni-tail` restarts cleanly with the combined BPF filter
  (`(tcp dst port 443 and …) or (udp dst port 443 and (udp[8] & 0xf0) = 0xc0)`);
  libpcap accepts the expression and the tcpdump child stays attached
  to br-lan. The `0xc0` high-nibble match (header_form=1 + fixed_bit=1
  + packet_type=Initial=0) drops every 1-RTT and 0-RTT/Handshake/Retry
  packet from capture, keeping per-packet cost at parity with the TCP
  path.

## Metric buckets added

Extends `sni_clienthellos_total{result}` with the bounded enum
`quic_success | quic_short_header | quic_unsupported_version |
quic_not_initial | quic_truncated | quic_malformed | quic_decrypt_failed |
quic_no_crypto_frame | quic_no_sni | quic_ipv6_skipped | quic_not_udp |
quic_not_443`. No API-side `MetricGuard` change is needed — the agent
folds the per-result tally generically via `metrics.fold_external`,
and the `{result}` label key is already on the allowed-keys list.

Grafana: extends `deploy/grafana/dashboards/router-fleet.json`'s
existing 'SNI ClientHello rate by result' description to enumerate
the new buckets (the panel itself already aggregates `sum by(result)`
so it picks them up automatically), and adds a new 'SNI success share:
TCP vs QUIC (#1651)' panel — the operator-level summary of HTTP/3
adoption that the per-bucket panel can't answer at a glance.

## Deliberate v1 scope

| Scope          | This PR                | Out of scope                       |
| -------------- | ---------------------- | ---------------------------------- |
| QUIC version   | v1 only                | v2 (RFC 9369) → `quic_unsupported_version` |
| Packet type    | Client Initial only    | 0-RTT / Handshake / Retry / 1-RTT  |
| AEAD           | AES-128-GCM            | (Initials always use AES-128-GCM per RFC 9001 §5.1 — no ChaCha20 needed) |
| Outer L3       | IPv4                   | IPv6 framing (same as sni.lua v1)  |
| ClientHello    | Single Initial         | Multi-Initial fragmentation        |

QUIC v2 (and any future version) would need a different salt and could
land in a follow-up that's purely a `version → salt` lookup table —
the rest of the pipeline is version-agnostic. I haven't filed a
follow-up issue because v2 adoption is still effectively zero on the
public internet.

## Files

| File | Purpose |
| ---- | ------- |
| `openwrt/files/usr/lib/lua/wifihaven/quic.lua` | New: QUIC v1 Initial parser |
| `openwrt/files/usr/sbin/wifihaven-sni-tail` | Extended BPF + dispatch |
| `openwrt/Makefile` | `+lua-openssl` DEPENDS |
| `openwrt/test/quic_spec.lua` | RFC 9001 §A.2 vector + edge cases |
| `deploy/grafana/dashboards/router-fleet.json` | Dashboard panel |

## Test plan

- [x] `busted openwrt/test/quic_spec.lua` — 9/9 pass
- [x] `bash openwrt/test/run_tests.sh` — 697/697 pass
- [x] RFC 9001 §A.2 vector decrypts to "example.com" on the test router
- [x] `wifihaven-sni-tail` restarts with the new BPF without crashing
- [x] `python3 -m json.tool deploy/grafana/dashboards/router-fleet.json` validates
- [x] `terraform fmt -check` in `infra/grafana/`
- [ ] Real-browser QUIC handshake captured on a fleet router with `quic_success` ticking (deferred to next deploy — needs an actual LAN client; the RFC vector hits the same code path)

### DIFF
--- a/deploy/grafana/dashboards/router-fleet.json
+++ b/deploy/grafana/dashboards/router-fleet.json
@@ -412,7 +412,7 @@
     {
       "datasource": { "type": "prometheus", "uid": "${datasource}" },
       "title": "SNI ClientHello rate by result",
-      "description": "sum by (result) (rate(sni_clienthellos_total[5m])) — result ∈ {parsed, truncated, no_sni, ipv6_skipped, not_tcp, malformed}. Emitted by the #573 SNI hostname-attribution sidecar (wifihaven-sni-tail), which classifies sniffed TCP ClientHello records and folds a cumulative per-result tally into the agent's metrics push. This attributes (dst_ip → SNI hostname) pairs for traffic where dnsmasq never saw the lookup (DoH / hard-coded IP), so it is the connection-layer complement to `dns_queries_total`, not a block signal. `parsed` is the healthy path; the failure classes are diagnosed below.",
+      "description": "sum by (result) (rate(sni_clienthellos_total[5m])) — result ∈ {parsed, truncated, no_sni, ipv6_skipped, not_tcp, malformed} on the TCP path, plus {quic_success, quic_short_header, quic_unsupported_version, quic_not_initial, quic_truncated, quic_malformed, quic_decrypt_failed, quic_no_crypto_frame, quic_no_sni, quic_ipv6_skipped, quic_not_udp, quic_not_443} on the QUIC v1 Initial path (#1651). Emitted by the SNI hostname-attribution sidecar (wifihaven-sni-tail), which classifies sniffed TLS ClientHellos on TCP/443 AND TLS ClientHellos extracted from QUIC v1 Initial packets on UDP/443 (#573 / #1651). The (dst_ip → SNI) pairs feed the same cache.insert_sni writer, so this attributes flows where dnsmasq never saw the lookup (DoH / hard-coded IP) regardless of transport. `parsed` (TCP) and `quic_success` (QUIC) are the healthy paths; the failure classes are diagnosed in the panels below.",
       "type": "timeseries",
       "gridPos": { "h": 8, "w": 12, "x": 0, "y": 48 },
       "id": 13,
@@ -470,6 +470,36 @@
           "expr": "sum(rate(sni_clienthellos_total{env=\"$env\", result=\"ipv6_skipped\"}[5m])) / clamp_min(sum(rate(sni_clienthellos_total{env=\"$env\"}[5m])), 0.001)",
           "legendFormat": "ipv6_skipped",
           "refId": "B"
+        }
+      ]
+    },
+    {
+      "datasource": { "type": "prometheus", "uid": "${datasource}" },
+      "title": "SNI success share: TCP vs QUIC (#1651)",
+      "description": "Fraction of successful SNI extractions coming from QUIC v1 Initial packets vs TCP TLS ClientHello records. Computed as sum(rate(sni_clienthellos_total{result=\"quic_success\"}[5m])) / sum(rate(sni_clienthellos_total{result=~\"parsed|quic_success\"}[5m])). Operator reading: rising → HTTP/3 share of the fleet is growing; flat zero with non-zero TCP `parsed` → QUIC capture branch is unreachable (BPF rejected? sidecar restarted? lua-openssl missing?). The absolute rate of each is in the per-result panel above. Sustained quic_decrypt_failed in that panel would indicate a key-derivation or HP-mask regression; sustained quic_no_crypto_frame would mean Initials are arriving but carry no CRYPTO (probably ACK-only).",
+      "type": "timeseries",
+      "gridPos": { "h": 8, "w": 12, "x": 0, "y": 56 },
+      "id": 15,
+      "fieldConfig": {
+        "defaults": {
+          "color": { "mode": "palette-classic" },
+          "unit": "percentunit",
+          "min": 0,
+          "max": 1,
+          "custom": { "drawStyle": "line", "fillOpacity": 20, "showPoints": "never", "lineWidth": 2 }
+        },
+        "overrides": []
+      },
+      "options": {
+        "legend": { "displayMode": "table", "placement": "bottom", "calcs": ["lastNotNull", "mean"] },
+        "tooltip": { "mode": "multi", "sort": "desc" }
+      },
+      "targets": [
+        {
+          "datasource": { "type": "prometheus", "uid": "${datasource}" },
+          "expr": "sum(rate(sni_clienthellos_total{env=\"$env\", result=\"quic_success\"}[5m])) / clamp_min(sum(rate(sni_clienthellos_total{env=\"$env\", result=~\"parsed|quic_success\"}[5m])), 0.001)",
+          "legendFormat": "QUIC share",
+          "refId": "A"
         }
       ]
     }
--- a/openwrt/Makefile
+++ b/openwrt/Makefile
@@ -32,7 +32,13 @@ define Package/wifihaven
   # on modern kernels). Without it `conntrack -E -e NEW` sees 0 v6 flow events,
   # so the agent never emits connection_events for v6 destinations and the
   # Gate 2 v6 attribution suite (scripts/e2e/scenarios_fake/test_v6_*.py) fails.
-  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nf-log +kmod-nf-log6 +kmod-nf-conntrack6 +libustream-mbedtls +openssl-util +tcpdump-mini
+  # #1651: lua-openssl provides HMAC-SHA256 + AES-128-ECB + AES-128-GCM so the
+  # sni sidecar can extract the SNI from QUIC v1 Initial packets too (UDP/443),
+  # closing the HTTP/3 leg of the #573 attribution gap. RFC 9001 §5.2 keys are
+  # HKDF-derived from the destination connection ID + v1 salt; the payload is
+  # AEAD-encrypted with AES-128-GCM and the header bits are protected with an
+  # AES-128-ECB mask. ~464 KiB installed — additive, attribution-only.
+  DEPENDS:=+lua +libuci-lua +luci-lib-jsonc +conntrack +curl +uhttpd-mod-lua +kmod-nf-log +kmod-nf-log6 +kmod-nf-conntrack6 +libustream-mbedtls +openssl-util +tcpdump-mini +lua-openssl
   PKGARCH:=all
 endef
 
--- a/openwrt/files/usr/lib/lua/wifihaven/quic.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/quic.lua
@@ -56,6 +56,13 @@
 --   quic_not_udp            -- parse_initial saw an IPv4 non-UDP packet
 --   quic_not_443            -- parse_initial saw a UDP packet not to port 443
 --
+-- NOTE: the last three are only reachable when callers invoke parse_initial
+-- directly (e.g. the unit tests). The production sidecar's dispatch() peeks
+-- at ethertype / IPv4 proto BEFORE calling parse_initial and short-circuits
+-- those cases under the TCP-path "ipv6_skipped" / generic "malformed"
+-- buckets, so they will not tick in the sni_clienthellos_total metric on
+-- fleet routers — the documented contract holds for direct API consumers.
+--
 -- A malformed/truncated packet returns (nil, reason) — never raises. The
 -- sidecar runs as a long-lived loop; a single bad packet must not kill it.
 
@@ -136,12 +143,12 @@ end
 -- Returns (value, byte_count) or (nil) if truncated.
 -- ---------------------------------------------------------------------------
 
+local VARINT_LEN = { [0] = 1, [1] = 2, [2] = 4, [3] = 8 }
+
 local function read_varint(buf, off)
   local first = u8(buf, off)
   if not first then return nil end
-  local prefix = math.floor(first / 64)
-  local len_table = { [0] = 1, [1] = 2, [2] = 4, [3] = 8 }
-  local len = len_table[prefix]
+  local len = VARINT_LEN[math.floor(first / 64)]
   if off + len - 1 > #buf then return nil end
   local val = first % 64
   for i = 1, len - 1 do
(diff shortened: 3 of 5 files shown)


sameerparekh:
Independent review per [`docs/pr-review-checklist.md`](docs/pr-review-checklist.md). Read-only — author should self-address before merge.

## BLOCKERS

(none)

## SHOULD-FIX

1. **`openwrt/files/usr/sbin/wifihaven-sni-tail:223-230`** — the periodic `log.info("sni-tail: parsed=%d … quic_success=%d …")` reports `parsed=parsed_total`, but `parsed_total` is now incremented on every dispatch success (both TCP `"parsed"` and QUIC `"quic_success"`), not on the TCP path alone. So a fleet log reader seeing `parsed=800 quic_success=300` would naturally infer 800 TCP + 300 QUIC = 1100, when reality is 500 TCP + 300 QUIC = 800 total. Either (a) rename `parsed_total` → `success_total` and the log field to `success=` to keep the meaning honest, or (b) recompute TCP-only as `parsed_total - (results.quic_success or 0)` for the log line. Doesn't affect the metric (the per-bucket `results` table is still correct); just the logread surface.

## NITS

1. **`openwrt/files/usr/lib/lua/wifihaven/quic.lua:142`** — `read_varint` builds `local len_table = { [0]=1, [1]=2, [2]=4, [3]=8 }` on every call. Hoist to module scope. Pure refactor; hot path.

2. **`openwrt/files/usr/lib/lua/wifihaven/quic.lua:233-240`** — `parse_long_header` returns `scid` and `token` but no caller consumes them (the dcid drives key derivation; scid/token are read only to advance `off`). Can drop them from the returned table to shrink the contract.

3. **`openwrt/files/usr/lib/lua/wifihaven/quic.lua:64-69`** — the enum entries `quic_ipv6_skipped`, `quic_not_udp`, `quic_not_443` are documented but unreachable through the sidecar's `dispatch()` (which short-circuits on ethertype / proto before ever invoking `quic.parse_initial`). They ARE reachable from direct API callers (unit tests, future integrations). Worth a single-line note in the enum comment so a reader doesn't grep for them in the metric and wonder why they never tick.

## Notes on what was checked

- **Single source of truth (§1).** TLS ClientHello SNI extraction reuses `sni.parse_client_hello` via `parse_tls_handshake`. The `(dst_ip, src_mac, sni)` tuple lands in the same spool the TCP path writes (`paths.sni_capture`) and reaches `cache.insert_sni` via unchanged `wifihaven-dns-tail` plumbing. The eth/IPv4 framing duplication between `sni.parse_packet` / `quic.parse_initial` / `dispatch()` is minor (~10 lines, IPv4-no-options only) and acceptable per the "two paths must stay separate → TEST-PINNED" rule — both have direct API tests. No display-vs-enforcement source mismatch.
- **TDD (§2).** Red commit `f105da42` precedes the green impl in `8ffadb11` — the failing-test-first progression is visible in the PR history. RFC 9001 §A.2 vector + 8 edge-case tests; no existing tests modified or weakened. The stub crypto asserts on its inputs, so a regression in HP-mask / AAD / nonce construction surfaces as an explicit assertion failure rather than a silent nil.
- **Architectural invariants (§3).** No DNS-as-enforcement (capture-only). Router stays a dumb applier (parsing only; no policy). No new wire-shape fields. `/tmp/wifihaven-sni.log` is the existing spool with existing rotation in `wifihaven-rotate-dnsmasq-log` — no new unbounded `/tmp` writer.
- **Backwards compat (§4).** No wire-shape changes. New `sni_clienthellos_total{result}` label values are additive; the label KEY `result` is unchanged (and already on `MetricGuard.Allowed["sni_clienthellos_total"] -> Set("result", "router_id", "installation_id")`), so no API-side allowlist edit needed.
- **Migrations (§5).** N/A.
- **Metrics + dashboard (§6).** Metric is additive on an existing series, routed through the existing `metrics.fold_external` path on the agent side. Dashboard updated in the same PR — extended description on panel 13 + new panel 15 ("SNI success share: TCP vs QUIC"). `python3 -m json.tool` validates; `terraform fmt -check` on `infra/grafana/` passes (no `.tf` changes here, the new panel reuses the existing registered dashboard).
- **Scope / hygiene (§7).** Five files touched, all directly motivated. No drive-by edits, no debug logging, no AI mentions in PR body or commit messages, three-dot diff against `origin/main`.
- **Security (§8).** N/A — no new routes, no SQL, no secrets. The added `lua-openssl` dep is a well-maintained, widely-deployed binding to system OpenSSL; not a fresh crypto implementation.

## Live-verification evidence I confirmed from the PR description

- RFC 9001 §A.1 keys derived byte-for-byte via `quic.openssl_crypto(openssl)` on `openwrt.lan` (OpenWrt 25.12.3, lua-openssl 0.10.0.0-r1).
- `quic.parse_initial` on an eth/IPv4/UDP frame wrapping the RFC 9001 §A.2 protected client Initial returns `sni=example.com` end-to-end.
- `wifihaven-sni-tail` restarts cleanly with the combined BPF; libpcap accepts the expression; tcpdump child stays attached.
- Full unit suite 697/697 + 9/9 quic_spec pass.

The one piece of "real-world" verification not closed: real Chrome/Firefox QUIC traffic on a fleet router with the metric ticking `quic_success`. The PR's test plan notes this as deferred to next deploy; that's reasonable given the test router lacks LAN clients capable of generating HTTP/3 traffic, and the RFC vector exercises the same code path.

VERDICT: APPROVE

QUIC v1 Initial-packet SNI parser is well-scoped, correctly implements RFC 9000/9001 (verified against the §A.2 vector end-to-end on a real router with lua-openssl), and slots cleanly into the existing #573 attribution pipeline without touching wire shapes, policy logic, or the API. The three NITs and one SHOULD-FIX are quality polish that won't gate merge — the SHOULD-FIX is a logread-readability cleanup, not a behavior bug.

sameerparekh:
Re-run independent review per [`docs/pr-review-checklist.md`](docs/pr-review-checklist.md). Previous review: ef612dbf089...4b7b268bb (`reviewed-sha=4b7b268bb89dc524d993d95fe17a0494eca078de`).

## Status of prior findings

| # | Class | Citation | Status | Notes |
| - | ----- | -------- | ------ | ----- |
| 1 | SHOULD-FIX | `wifihaven-sni-tail:223-230` — `parsed=%d` conflates TCP + QUIC | **ADDRESSED** | `parsed_total` → `success_total`; log field `parsed=` → `success=`; explanatory comment added describing the union. |
| 2 | NIT | `quic.lua:142` — `len_table` per-call allocation | **ADDRESSED** | Hoisted to module scope as `VARINT_LEN` (5-arg constant). |
| 3 | NIT | `quic.lua:233-240` — unused `scid` / `token` returned | **ADDRESSED** | Dropped `version` / `scid` / `token` from the returned table; the SCID-length / token-length walks remain (they advance `pn_offset`), with a docstring note explaining why they're parsed but not surfaced. |
| 4 | NIT | `quic.lua:64-69` — unreachable enums via dispatch | **ADDRESSED** | New NOTE block in the enum docstring explaining that the last three labels only tick from direct API callers; production sidecar's `dispatch()` short-circuits them under TCP-path / `malformed` buckets. |

All four prior findings cleared. No prior BLOCKERS.

## New review of the incremental delta (`4b7b268b...ef612dbf`)

Delta is 2 files / +31 / -25, all in the changes above. Re-checked against the 8 dimensions:

**§1 Single source of truth.** No re-derivation introduced. The cleanup actually reduces surface by dropping the unused returns; doesn't create a new copy of any decision or computation.

**§2 Testing integrity.** Existing 9/9 `quic_spec` + 43/43 `sni_spec` still green locally (`busted` 52/52). No tests modified or weakened — the refactor changes private return shape and a `local` variable name; the stub crypto and assertions are unaffected. Live RFC 9001 §A.2 vector still extracts `sni=example.com` end-to-end on `openwrt.lan` after the cleanup.

**§3 Architectural invariants.** Unchanged — no DNS-as-enforcement, no policy-on-router, no new `/tmp` writer, no wire-shape edit, no `java.time` / `sys.env` regressions.

**§4 Wire compat.** Unchanged — only the in-process return shape of `parse_long_header` moved (which is a private internal contract, not wire-visible). The on-wire metric labels are byte-identical.

**§5 Migrations.** N/A.

**§6 Metrics + dashboard.** Unchanged — same label key, same value set. The `success_total` rename only affects the in-process log line.

**§7 Scope / hygiene.** Focused on the prior review's specific cites; no drive-bys.

**§8 Security.** Unchanged.

### Subtle correctness re-check of the SCID / token rewrite (`quic.lua:204-217`)

The diff converts `local scid = slice(buf, off, scid_len)` (which implicitly bounds-checked via `slice`'s `(off + len - 1) > #buf` guard) into a fused `off = off + 1 + scid_len` followed by an explicit `if off > #buf + 1 then return nil, "quic_truncated" end`. The bound is correct: after consuming `scid_len` bytes starting at `off+1`, the new `off` points one past the last consumed byte (1-based), so `off ≤ #buf + 1` exactly when all bytes were present. The `+ 1` accounts for 1-based indexing. Same logic for the token skip. Empty SCID / empty token (`scid_len = 0`, `token_len = 0`) leave `off` unchanged; subsequent `read_varint` on the length field handles out-of-bounds via its own truncation return. The `scid_len > 20` guard upstream prevents `off + 1 + scid_len` from overflowing into a meaningless state. Verified against the RFC 9001 §A.2 vector on the router post-rewrite — `parse_initial` still returns the right answer.

## New findings

BLOCKERS: (none)

SHOULD-FIX: (none)

NITS: (none)

VERDICT: APPROVE

All four prior findings addressed (1 SHOULD-FIX + 3 NITs cleared). The fix delta is a focused cleanup with no new findings, RFC §A.2 re-verified end-to-end on the router, and 52/52 local tests green. No open BLOCKER on either pass; merge gate is clear.

PR #1707, opened 2026-06-14
feat(#1650): ECH detection + outer-name fallback in SNI sidecar

### LINKED ISSUE #1650: feat(#573): ECH (Encrypted ClientHello) detection + graceful degradation for SNI sidecar
Follow-up to #573 / PR #1643 (SNI hostname-attribution sidecar, v1).

When a ClientHello uses **Encrypted ClientHello (ECH)**, the real `server_name` is inside the encrypted `outer`/`inner` and `sni.parse_client_hello` returns nil — the flow is silently skipped and attribution falls back to IP only. v1 scope explicitly excluded ECH.

**Goal:** detect the ECH extension (0xfe0d) in the ClientHello and (a) count it distinctly in the sidecar metric (`sni_clienthellos_total{result=ech}` already reserves the enum) so operators can see the ECH-attributed fraction, and (b) decide a graceful-degradation policy — e.g. fall back to the public/outer SNI when present, or attribute via the learned CNAME/IP map. No DNS-layer behavior change; attribution-only, per the architectural model.

Out of scope: defeating ECH. This is observability + best-effort fallback only.

sameerparekh:
Closes #1650.

## Rationale

When a TLS ClientHello uses Encrypted ClientHello (RFC draft-ietf-tls-esni, extension type `0xfe0d`), the real `server_name` is inside the encrypted inner ClientHello and the sidecar cannot recover it. This PR is **observability + best-effort fallback**, not an attempt to defeat ECH:

- The reserved `sni_clienthellos_total{result="ech"}` bucket is now live (extension type detected in `sni.parse_client_hello`'s existing extension-walk loop — same parser, one new branch, no fork).
- When the outer ClientHello carries a public/outer `server_name` (the gateway hostname, e.g. `cloudflare-ech.com`), we return it as the attribution — best honest answer; correctly names the gateway the flow is bound to, even though it doesn't name the inner host the user requested.
- When the outer has no SNI at all, the counter still ticks but no hostname is attributed.

The QUIC v1 Initial path is wired up the same way (`quic_ech` bucket) so HTTP/3 + ECH is also visible.

## Metric

`sni_clienthellos_total{result}` — the `ech` (TCP) and `quic_ech` (QUIC) buckets are now emitted. The router-fleet dashboard's `SNI ClientHello rate by result` panel uses `sum by (result) (rate(…))`, so the new buckets surface automatically; only the panel description was updated.

## Test plan

- [x] `busted` covers `parse_client_hello` ECH-with-outer-SNI and ECH-without
- [x] `busted` covers `parse_packet` propagating `reason="ech"` on both
- [x] Existing non-ECH tests unchanged (parser is still single-return for plain ClientHellos because Lua drops extra return values)
- [x] Full Lua + shell suite: 708 / 0
- [x] `scalafmt --check`
- [x] Dashboard JSON validates


sameerparekh:
## Independent PR review (#1707)

Reviewed commit: `9d509928`. First run — full merge-base diff against the 8 checklist dimensions.

### BLOCKERS
(none)

### SHOULD-FIX
1. [openwrt/files/usr/lib/lua/wifihaven/quic.lua:491-505](../blob/9d509928/openwrt/files/usr/lib/lua/wifihaven/quic.lua#L491-L505) — **No tests for the QUIC ECH path.** `parse_quic_packet` gained two new branches (`quic_ech` with-outer-SNI and without-outer-SNI) plus a new tuple-return shape on success at [quic.lua:558-560](../blob/9d509928/openwrt/files/usr/lib/lua/wifihaven/quic.lua#L558-L560), but `openwrt/test/quic_spec.lua` has zero ECH coverage. The TCP path is well-tested in `sni_spec.lua` (and the underlying `parse_client_hello` ECH detection is covered there), but the QUIC propagation logic — including `parse_initial`'s new `if reason then return {…}, reason end` shape — could regress silently. Add a unit test driving `parse_quic_packet` with a stubbed `aes128_gcm_decrypt` that returns an unprotected payload containing a ClientHello with the 0xfe0d extension; assert `(sni, "quic_ech")` and `(nil, "quic_ech")` separately. (The TDD red→green is visible in commit history for the TCP path — commits f726ac0d → 9d509928 — so this is just the QUIC delta needing the same coverage.)

### NITS
1. [openwrt/files/usr/lib/lua/wifihaven/sni.lua:159](../blob/9d509928/openwrt/files/usr/lib/lua/wifihaven/sni.lua#L159) — Minor behavior change worth noting: the inner `host_name` loop now `break`s instead of `return nil` when it encounters a non-host_name name_type, which means a subsequent ECH extension WILL still be detected even when the SNI list contained only non-host_name entries. Previously the parser bailed out of the whole function at that point. This change is consistent with the new "walk the whole list" intent and doesn't regress the existing "name_type != host_name → nil" test (it still returns nil for that case because `found_sni` stays unset), but the behavior delta is subtle and worth a short comment.
2. [docs/ops/hostname-attribution.md:97](../blob/9d509928/docs/ops/hostname-attribution.md#L97) — Section renumbering "### 4." → "### 5." (Connection-reuse / QUIC) is fine; verified no other doc cross-references those heading numbers.

### Notes on each dimension
1. **Single source of truth** ✅ — ECH detection lives in ONE place (the existing extension-walk loop in `sni.parse_client_hello`). Both TCP (`sni.parse_packet`) and QUIC (`quic.parse_tls_handshake`) paths reuse it. No duplicated decision logic. Matches the §single-source-of-truth principle.
2. **Testing integrity** ⚠️ — TDD red→green visible for the TCP path (commits split correctly). QUIC ECH lacks coverage — flagged above as SHOULD-FIX. No existing tests were weakened or removed.
3. **Architectural invariants** ✅ — Attribution-only change. No new wire fields, no router-side policy logic, no DNS-as-enforcement reasoning, no unbounded `/tmp` writers (the existing `paths.sni_metrics` is a fixed-size snapshot — exempt).
4. **Backwards compatibility** ✅ — `parse_client_hello`'s tuple return is additive (Lua drops extra return values for callers ignoring the second). `parse_quic_packet`'s success-with-ECH tuple is also additive; `parse_initial` handles it. New `result=ech` / `quic_ech` enum values are additive on the wire.
5. **Migrations** ✅ — none.
6. **Metrics + dashboard** ✅ — `MetricGuard.Allowed` updated (the `result` label was already permitted; just the bounded-enum comment extended). Dashboard panel uses `sum by (result) (rate(…))` so new buckets surface automatically — only the description was extended. No cardinality concerns.
7. **Scope / hygiene** ✅ — Focused diff (8 files, all directly on the ECH path). No drive-by edits, no debug logging, no AI mentions.
8. **Security** ✅ — No auth/SQL/secrets surface.

### VERDICT: APPROVE

ECH detection adds a single branch to the existing extension-walk loop in `sni.parse_client_hello` and propagates the label through both the TCP (`parse_packet` → `dispatch`) and QUIC (`parse_quic_packet` → `parse_initial` → `dispatch`) paths. The MetricGuard / Grafana / docs surfaces are all updated in the same PR. The one SHOULD-FIX is QUIC ECH test coverage — the underlying detection is tested, but the QUIC-specific bucketing branches are not, leaving room for silent regression on a transport that is unambiguously in scope per the PR body.

sameerparekh:
## Independent PR review (#1707) — re-run

Reviewed commit: `faa8b089`. Prior review: `9d509928`. Delta: `git diff 9d509928...HEAD` (test additions only).

### Prior-findings status

| 1 | SHOULD-FIX | [quic.lua:491-505](../blob/faa8b089/openwrt/files/usr/lib/lua/wifihaven/quic.lua#L491-L505) — No tests for the QUIC ECH path | **ADDRESSED** — three new tests in [quic_spec.lua:273-365](../blob/faa8b089/openwrt/test/quic_spec.lua#L273-L365) cover (a) ECH with outer SNI returning `(outer_name, "quic_ech")`, (b) ECH with no outer SNI returning `(nil, "quic_ech")`, and (c) `parse_initial` wrapping the result and propagating `quic_ech`. Each test builds a bare TLS handshake with the 0xfe0d extension, wraps it in a CRYPTO frame, and feeds it via a stubbed `aes128_gcm_decrypt` while preserving the RFC 9001 AAD / nonce / HP-mask exercises. |
| 2 | NIT | [sni.lua:159](../blob/faa8b089/openwrt/files/usr/lib/lua/wifihaven/sni.lua#L159) — subtle `break` instead of `return nil` worth a comment | NOT-ADDRESSED — non-blocking. The existing comment block is sufficient; behavior is correct and covered by the existing "name_type != host_name → nil" test. |
| 3 | NIT | doc renumbering | n/a — was a clean-bill note. |

### NEW findings (delta `9d509928...faa8b089`)

#### BLOCKERS
(none)

#### SHOULD-FIX
(none)

#### NITS
1. The commit incidentally notes (in passing) that `sni_spec.lua` and `quic_spec.lua` aren't in `.github/workflows/ci.yml`'s busted invocation. That's a real pre-existing CI gap orthogonal to #1650 — correctly out-of-scope for this PR. A follow-up issue would be appropriate so the next reviewer doesn't have to re-discover it.

### Updated VERDICT: APPROVE

ECH detection is now end-to-end tested on both the TCP and QUIC paths. The SHOULD-FIX from the prior review is cleanly ADDRESSED — the new QUIC tests follow the existing fixture patterns (`stub_crypto_with_plaintext` is a minimal, clean extension of the existing RFC-vector stub), and the change is purely test additions with no production code touched. No open BLOCKER. Safe to merge.

Issue #1644, opened 2026-06-11
Grafana panel for sni_clienthellos_total{result} (SNI sidecar metric)


sameerparekh:
Follow-up to #573 / PR #1643.

The SNI hostname-attribution sidecar (`wifihaven-sni-tail`) now emits a bounded
cumulative counter that the agent folds into `sni_clienthellos_total{result}`
on its metrics tick (`result ∈ {parsed, truncated, no_sni, ipv6_skipped,
not_tcp, malformed}`), pushed via `POST /api/router/metrics`. The metric is
allowlisted in `MetricGuard.Allowed` (`api/src/metrics/Metrics.scala`) and
re-exposed under `router_id` / `installation_id`.

Per the PR #1643 review (SHOULD-FIX #5), the metric ships in that PR but the
**Grafana dashboard panel** was deferred to keep the PR focused. This issue
tracks adding it.

### Scope

- Add a panel consuming `sni_clienthellos_total{result}` to the appropriate
  checked-in dashboard under `deploy/grafana/dashboards/` (the application
  self-metrics dashboard alongside `dns_queries_total`), registered in
  `infra/grafana/main.tf`'s `dashboards` list if a new file is needed.
- Suggested views:
  - Per-`result` rate (`rate(sni_clienthellos_total[5m])`) so the fleet-wide
    **truncation** rate (snaplen-cut ClientHellos) and **ipv6_skipped** share
    are visible — the load-bearing signals for sizing snaplen / scoping the
    IPv6 follow-up.
  - `parsed` vs total ratio as a capture-health indicator.
- Slice only by the bounded `result` label (plus templated `router_id` /
  `installation_id`); never by a per-host/per-mac value (cardinality firewall).
- CI gate: `terraform fmt -check`, `terraform validate`, and
  `python3 -m json.tool` on the dashboard JSON (the `grafana-terraform` job).

### Also confirm

- The agent-side fold (`fold_sni_metrics` in `wifihaven-agent`, mirroring
  `fold_dns_queries`) is already wired into the metrics tick in PR #1643 — this
  follow-up should just verify it lands a non-zero series end-to-end once the
  sidecar runs on the fleet, and that `sni_clienthellos_total` does not show up
  in `metrics_rejected_total{reason="unknown_name"}`.

Epic: Observability/Metrics.

PR #1711, opened 2026-06-14
feat(#1653): per-flow ClientHello reassembly across TCP segments (bounded buffer)

### LINKED ISSUE #1653: feat(#573): TLS 1.2 fragmented ClientHello reassembly across TCP segments
Follow-up to #573 / PR #1643. v1 parses **single-segment** ClientHellos only and drops anything past the BPF snaplen (~600 bytes); a ClientHello split across multiple TCP segments (large key_share / cipher lists / many extensions — increasingly common with TLS 1.3 hybrid PQ key shares) is not reassembled and returns nil (counted `truncated`).

**Goal:** reassemble the ClientHello across TCP segments for a flow before parsing. Needs per-flow buffering keyed by the TCP 4-tuple with a short bounded TTL and a hard cap (memory-safety on the router — see the #bounded-tmp-writes / tmpfs constraint). Watch the `truncated` metric from #1644 to size the buffer.

Out of scope: full TCP stream reassembly. Bound to the first N KB of a single flow's handshake.

sameerparekh:
Closes #1653.

## Problem

`wifihaven-sni-tail` v1 (#573) parses **single-segment** TLS ClientHellos. A
ClientHello that spans multiple TCP segments — increasingly common with TLS
1.3 hybrid PQ key shares, large extension lists, or long ALPN sets — has its
`server_name` extension fall into segment 2, and the v1 parser returns nil.
That nil currently lands in the `no_sni` bucket on
`sni_clienthellos_total{result}` even though attribution would have worked
fine if we'd just buffered segment 1 and waited for segment 2.

## What this PR does

Adds per-(TCP 4-tuple) ClientHello buffering with a hard byte cap, a flow-
count cap, and TTL eviction — three independent bounds, each pinned by a
test. The buffered payload is fed back through the **same**
`sni.parse_client_hello` primitive (AGENTS.md §single-source-of-truth — no
parser fork).

### Buffer-sizing rationale (prod evidence)

Pulled `sni_clienthellos_total` from the running prod API
(`api.lan:8080/metrics`):

```
parsed              4335
quic_success         598
quic_no_crypto_frame 792
no_sni                26
quic_no_sni            6
truncated              — (absent / 0)
```

`truncated` is zero by construction since #573 bumped the snaplen to 2048 —
any single-segment CH fits, and the v1 BPF only ever captured segment 1, so
multi-segment CHs were never re-parsed and never re-counted. They fall into
`no_sni` (~0.5% of capture). That's the upper bound of what reassembly can
recover today; the curve will steepen as TLS 1.3 PQ rolls out (Chrome /
Firefox flips have already shipped hybrid X25519+Kyber, and a ClientHello
with a hybrid PQ key share routinely runs 3-4 KB).

Chosen bounds:
- **`cap_bytes = 8192`** — comfortably covers a full TLS 1.3 hybrid PQ
  ClientHello (~3-4 KB) plus extensions; anything bigger is suspicious /
  adversarial and the flow is dropped rather than truncated, so we never
  feed a partial buffer back into the parser with confusing semantics.
- **`cap_flows = 64`** — at the current concurrent-flow profile (a few
  hundred flows/min, lifetime << 1s each), 64 simultaneous in-progress
  reassemblies is well above the expected steady state. Worst-case memory
  ≈ 64 × 8 KB = **512 KB tmpfs**, an order of magnitude under the smallest
  supported device's free RAM headroom.
- **`ttl_s = 5`** — TLS handshakes complete in well under one second;
  5 s is generous and stops a half-open flow from sitting in the buffer
  forever.

All three are honored on every code path and have dedicated tests
(`sni_reassembly_spec.lua`).

### Memory safety

Per AGENTS.md §bounded-tmp-writes / #1647: the buffer lives in process
memory, not on tmpfs, but the same discipline applies — every grow path
has a cap, and every cap has a test. The buffer:

| Bound | Cap | Test |
|---|---|---|
| Per-flow byte cap (drop on overflow) | 8 KB | `per-flow byte cap: appending past cap_bytes drops the flow` and the single-oversize-segment variant |
| Concurrent-flow cap (LRU evict) | 64 | `LRU eviction: filling cap_flows then adding one more drops the oldest entry` |
| TTL (per-key lazy + periodic sweep) | 5 s | `TTL eviction: a buffered flow older than ttl_s is dropped on the next feed for that key` + `evict_expired removes all stale entries` |

The sidecar calls `reasm_buf:evict_expired()` on its existing 30 s
metrics-flush tick so a half-open flow whose continuation never arrives
doesn't linger.

### BPF widening (heads-up)

The v1 BPF matched **only** the segment whose payload starts with
`0x16 ... 0x01` (the TLS handshake / ClientHello signature). Segment 2+ of
a multi-segment CH starts at an arbitrary CH offset, so the v1 filter
never captured those — even with a perfect reassembly buffer in place,
the bytes would never reach it.

This PR widens the TCP arm to capture any TCP/443 segment whose first
payload byte is **not** `0x17` (TLS application_data). That catches the
full handshake of every flow (`0x14` ChangeCipherSpec / `0x15` Alert /
`0x16` Handshake all pass) AND continuation segments of a multi-segment
CH whose first byte is unlikely to be `0x17` (the parser tolerates the
~1/256 false-negative when a continuation byte happens to be `0x17`).
Pure-data steady-state segments — the overwhelming bulk of TLS traffic —
still drop out, so per-packet cost stays bounded by handshake-segment
count, not total TCP/443 segment count.

### Observability

`sni_clienthellos_total{result}` already had the `result` label
allowlisted (`MetricGuard.Allowed`); the firewall checks keys not values
so no allowlist change was needed. New bucket values:

- **`reassembled`** — the lift signal: a multi-segment CH was stitched
  and parsed. Rate-watch this against the historical `no_sni` baseline
  to confirm the feature is pulling its weight in prod.
- `incomplete` — segment buffered for follow-up.
- `dropped_byte_cap` — per-flow byte cap exceeded.
- `not_handshake` — continuation segment with no prior buffered context
  (expected drop-out from the widened BPF; rate-watch for sanity but a
  large bucket is normal here).

The sidecar's 60 s heartbeat log line was extended with the same buckets
plus a `buf=N` snapshot of current buffered flows.

### Tests (RED → GREEN)

- `c07a0f7d test(#1653): failing tests` — RED commit, contract pinned
  before implementation.
- `8d1ad437 feat(#1653): per-flow ClientHello reassembly` — GREEN commit.

Coverage in `sni_reassembly_spec.lua`:
- `parse_client_hello` reason distinction (incomplete / no_sni /
  not_handshake — the precondition for reassembly).
- Single-segment CH parses immediately (`result=parsed`).
- Two-segment CH: segment 1 → `incomplete`, segment 2 → `reassembled`.
- Non-handshake payload without buffered context → `not_handshake`.
- TTL eviction on lazy per-key access.
- `evict_expired` periodic sweep.
- LRU eviction at `cap_flows`.
- Per-flow byte cap on continuation-overflow and single-oversize-segment
  paths.
- `sni.parse_framing` extracts the 5-tuple cleanly (IPv4/TCP), surfaces
  `not_ip` / `truncated` reasons for non-IP / short frames.
- `sni.flow_key` is stable and disjoint across distinct 4-tuples.

Full openwrt suite: **703 successes, 0 failures**.

## Out of scope

- Full TCP stream reassembly (kept to the first ~8 KB of a single flow's
  handshake, per the issue).
- IPv6 fragment headers (44) — still drop out as not_tcp / malformed; rare
  in practice because CHs fit in one MTU and v6 fragments are vanishingly
  rare in steady-state traffic.


sameerparekh:
## Independent PR review

Per [`docs/pr-review-checklist.md`](../docs/pr-review-checklist.md). Read-only adversarial pass against `8d1ad437`.

### BLOCKERS

1. **`openwrt/files/usr/lib/lua/wifihaven/sni.lua:371` — duplicated Ethernet/IPv4/IPv6/TCP framing logic.**
   The new `parse_framing` (sni.lua:294) is a full carve-out of the framing that *still also lives inline* in `parse_packet` (sni.lua:371) — both functions independently parse: the Ethernet header, the IPv4 IHL + dst-IP, the IPv6 fixed header + extension-header chain (8-hop bounded walk), and the TCP data-offset + payload slice. The PR commit message even acknowledges the duplication ("`parse_packet` keeps its existing contract"). Per AGENTS.md [§single-source-of-truth](../AGENTS.md#single-source-of-truth) this is the exact "extract a helper, leave the original in place" smell the convention forbids: a future bounds fix or IPv6 fragment-header follow-up will land in one copy and not the other. `parse_packet` has no production callers left after this PR (sidecar dispatch now goes through `parse_framing` + `reasm_buf:feed`; only `sni_spec.lua` exercises it), so the clean collapse is a one-liner: `parse_packet` should call `parse_framing` + `parse_client_hello` and surface the same reason set, rather than re-implementing framing. Merge-gating because the rule classes SSOT violations as the highest-priority finding and this one is mechanically trivial to fix.

### SHOULD-FIX

1. **`deploy/grafana/dashboards/router-fleet.json` panel description is stale for `sni_clienthellos_total`.** The summary-panel description enumerates the valid `result` values (`{parsed, truncated, no_sni, not_ip, not_tcp, malformed}` + QUIC buckets) but does not list the new #1653 buckets (`reassembled`, `incomplete`, `dropped_byte_cap`, `not_handshake`). Per AGENTS.md [§metrics-need-a-dashboard](../AGENTS.md#metrics-need-a-dashboard), a PR that changes an emitted metric series must update the consuming panel in the same PR. The `sum by (result)` query will auto-include the new values in the legend, but an operator reading the panel description won't know what `reassembled` means — which is the load-bearing "lift" signal this PR exists to expose. Also worth adding a small explicit `reassembled`-rate stat panel as a first-class success indicator for the change.

### NITS

1. **`openwrt/files/usr/sbin/wifihaven-sni-tail:32,112,162` — header comments still describe the dispatch path as "parsed inline by `sni.parse_packet`" / "feeds each captured packet to `sni.parse_packet`". The TCP arm now goes through `parse_framing` + `reasm_buf:feed`. Update the prose to match the new dispatch path.
2. **PR body memory accounting** says "Worst-case memory ≈ 512 KB tmpfs" — the reassembly buffer lives in-process (Lua heap), not in `/tmp`. The bound is correct (and modest); just call it "in-process memory" rather than "tmpfs" so readers don't go looking for a rotation cron entry.
3. **`openwrt/files/usr/lib/lua/wifihaven/sni.lua:25-26`** — `parse_client_hello` doc summary in the module header still reads `→ server_name string | nil` (no second return value). The detailed docstring at sni.lua:71 was updated; the header should follow.

### VERDICT: REQUEST-CHANGES

Per-flow ClientHello reassembly is the right shape — single-source parser reuse, three independent bounds (8 KB / 64 flows / 5 s TTL) each pinned by a test, RED→GREEN visible in history, BPF widened just enough to actually surface segment 2+. The blocking issue is the inline framing duplication left in `parse_packet`: with no production callers, it should collapse onto `parse_framing` rather than running as a parallel implementation. Dashboard description and stale dispatch comments are quick follow-ons after that.

sameerparekh:

Re-running against `033230c1` (push since `8d1ad437`).

### Prior findings status

| # | Finding | Status |
|---|---|---|
| BLOCKER 1 | `sni.lua:371` duplicated Ethernet/IP/TCP framing in `parse_packet` vs `parse_framing` | **ADDRESSED** — `parse_packet` now collapses onto `parse_framing` + `parse_client_hello` (sni.lua:374-388). Framing lives in exactly one place; the legacy contract (`{ dst_ip, src_mac, sni } \| nil, reason`) is preserved. |
| SHOULD-FIX 1 | Grafana panel description stale for the new `result` buckets | **ADDRESSED** — `router-fleet.json:415` now enumerates `reassembled, incomplete, dropped_byte_cap, not_handshake` and explains each one's operator-reading. JSON still lints. |
| NIT 1 | `wifihaven-sni-tail` header comments still pointed at `sni.parse_packet` | **ADDRESSED** — three call-out sites (lines 32, 112, 162) rewritten to describe the `dispatch_tcp` → `parse_framing` + `sni_reassembly:feed` path. |
| NIT 2 | PR body says "tmpfs" for the in-process buffer | **NOT-ADDRESSED** — PR body text isn't amended in the push. Cosmetic, non-merge-gating. |
| NIT 3 | `sni.lua:25-26` module-header API summary still showed the single-return signature | **ADDRESSED** — header API list now lists `parse_framing`, `flow_key`, and the new second-return `reason` on `parse_client_hello`. |

### New findings (delta `8d1ad437...033230c1`)

#### BLOCKERS
(none)

#### SHOULD-FIX
(none)

#### NITS
1. **`deploy/grafana/dashboards/router-fleet.json`** — the description update is the dashboard touch for this PR, but the panels themselves still don't expose `reassembled` as a first-class series (the `sum by (result)` legend will surface it automatically, which is fine). A small dedicated stat panel showing `rate(sni_clienthellos_total{result="reassembled"}[5m]) / rate(sni_clienthellos_total{result=~"parsed|reassembled"}[5m])` would make the "lift from #1653" observable at a glance; deferring it is acceptable and would naturally fold into a #1644-style follow-up panel issue.

### VERDICT: APPROVE

The BLOCKER (`parse_packet` framing duplication) is cleanly resolved: framing now lives only in `parse_framing`, and `parse_packet` is a 13-line composition over it. All 60 sni tests + the 703-test openwrt suite still pass. The dashboard description and stale dispatch comments are updated. Buffer bounds, SSOT parser reuse, and TDD red→green are intact. Ready to merge.

Issue #1712, opened 2026-06-14
Widen host_type CHECK constraint to accept 'label' (follow-up to #1708)


sameerparekh:
Follow-up to #1708.

## What

#1708 part 1 adds the `HostId.Label` wire variant to the API + SPA decoder. But the
DB persistence layer still rejects `label`-typed rows because four tables enforce
a CHECK constraint that doesn't include the new tag:

- \`time_usage\`           — V13 / \`time_usage_host_type_check\`
- \`connection_events\`    — V13 + V42 partition / \`connection_events_host_type_check\`
- \`block_events\`         — V13 / \`block_events_host_type_check\`
- \`traffic_reports\`      — V13 + V41 partition / \`traffic_reports_host_type_check\`

All currently \`CHECK (host_type IN ('fqdn', 'ipv4', 'ipv6'))\`. The new constraint
must add \`'label'\` to the allowed set.

## Why a separate PR

Per AGENTS.md migration-isolation, a Flyway migration ships in its own PR — SQL
+ docs only — so the existing-test suite gates back-compat against the deployed
image.

## Plan

Schema-only PR adds a migration that, for each of the four tables:

\`\`\`sql
ALTER TABLE <t> DROP CONSTRAINT <t>_host_type_check;
ALTER TABLE <t> ADD CONSTRAINT <t>_host_type_check
  CHECK (host_type IN ('fqdn', 'ipv4', 'ipv6', 'label'));
\`\`\`

For the partitioned tables (\`connection_events\`, \`traffic_reports\`) the
constraint sits on the partitioned parent and propagates; verify behaviour
matches the V41/V42 ATTACH PARTITION expectations.

Because the new set is a strict superset, existing rows satisfy the new
constraint trivially — no row rewrite needed. Use \`NOT VALID\` if needed to
avoid the validation scan on the growth tables (prod-data-volume concern).

## Ordering

This PR (schema) ships and deploys BEFORE the OpenWRT agent flips its emit path
(the #1708 part-2 PR). Otherwise the agent's first label-typed batch fails
INSERT and the events get dropped.

Issue #1708, opened 2026-06-14
HostId: add a distinct `label` variant so static IP-range labels don't masquerade as FQDNs


sameerparekh:
Follow-up to #1655 / #1706.

## What

#1655 added a last-resort static IP-range → label map
(`apple-push`, `google-dns`, `cloudflare-dns`). Today the agent emits
the produced label as a \`HostId\` with \`type = \"fqdn\"\` — it rides
the same wire shape any DNS/SNI-derived hostname does — because no
distinct variant existed.

That works (the SPA renders it as text) but conflates two semantically
different things:

- **\`fqdn\`** — a hostname the agent actually observed at the resolver
  or in a TLS ClientHello / QUIC Initial.
- **a static label** — a synthetic name the agent attached because the
  destination IP fell inside a hardcoded operator-curated range.

Conflation matters in two places:

1. **Downstream pattern matching.** \`HostMatch.matchesAny\` walks the
   apex hierarchy: \`apple-push\` and \`google-dns\` happen not to match
   any sensible operator pattern today, but the moment someone picks a
   label that overlaps a real apex (\`apple.com\` etc.) the static-map
   would silently join the app-host-set membership graph.

2. **Reader confusion.** \"Why does \`type=fqdn\` value=\`apple-push\`?\"
   is the FAQ this issue exists to retire.

## Proposal

Add a new \`HostId\` variant:

\`\`\`json
{ \"type\": \"label\", \"value\": \"apple-push\", \"source\": \"static-ip-range\" }
\`\`\`

(or \`source\` omitted if we don't need it). The \`source\` field would
let us add other label sources later (e.g. a future ASN-based mapping)
without another HostId case.

## Plan

This is a wire change, so per AGENTS.md (\"API and the agents deploy
independently\") it splits into staged PRs:

1. **API + shared decoder**: add the \`Label\` case to \`HostId\`,
   teach the encoder/decoder + persistence + SPA rendering. Older
   agents continue to emit \`fqdn\`-typed labels; the new variant is
   accepted but optional.
2. **Agent**: switch the static-IP-range attribution path to emit
   \`{ type: \"label\", value }\` instead of an FQDN. Other attribution
   paths (DNS, SNI/QUIC) keep emitting \`fqdn\` — they really are
   FQDNs.
3. **\`HostMatch.matchesAny\`**: \`Label\`-typed hosts return false
   (same default as IP literals — labels are not domain patterns).

## Why not in #1706

[...]

PR #1706, opened 2026-06-14
feat(#1655): static IP-range → label map for hardcoded-IP services (last-resort attribution)

sameerparekh:
Closes #1655.

## What

Adds a small in-repo **static IP-range → label** map consulted as the
last-resort attribution source in `conntrack.handle_flow`. Flows that
have no SNI and no dnsmasq resolution (Apple push on `17.0.0.0/8`, apps
that bypass the local resolver and hit `8.8.8.8` / `1.1.1.1` directly)
now surface as `apple-push` / `google-dns` / `cloudflare-dns` in
connection_events instead of bare IP literals.

## Initial map

| CIDR        | Label            | Why |
|-------------|------------------|-----|
| `17.0.0.0/8`  | `apple-push`     | Apple's IANA-allocated /8. APNs hosts and many Apple infra services pin into here and bypass dnsmasq. |
| `8.8.8.8/32`  | `google-dns`     | Google Public DNS. Apps that bypass the local resolver hit it directly. |
| `8.8.4.4/32`  | `google-dns`     | Google Public DNS (secondary). |
| `1.1.1.1/32`  | `cloudflare-dns` | Cloudflare Public DNS. |
| `1.0.0.1/32`  | `cloudflare-dns` | Cloudflare Public DNS (secondary). |

NTP isn't included — too many rotating pool IPs and volume is tiny.

## Architecture

**Precedence in `handle_flow`:**

1. `attribute_hostname` — dnsmasq query-log cache (SNI/QUIC sidecars feed
   this same primitive via `cache.insert_sni`, per #573 / #1651).
2. `ipset_lookup_hostname` — legacy `nft_sets` fallback for site_limits
   domains.
3. `static_ip_labels.lookup` — **new**, last resort.

A real attribution always wins over a static guess. The new map feeds
the **same** attribution primitive (per AGENTS.md single-source-of-truth
rule) — no parallel pipeline.

## LABELS ONLY — never an enforcement input

The static map participates in **zero** drop or carve-out predicates.
Enforcement remains the per-MAC `BlockRules` / nftables pipeline (AGENTS.md
§0.1, §0.2). The `static_ip_labels.lua` header repeats this and notes that
promoting an entry to an enforcement input requires explicit operator
approval and a tracking issue.

## Extending the map

Add a `{ cidr, label }` row to `M._ranges` with a citation in the comment
beside it, and add a test in `openwrt/test/static_ip_labels_spec.lua`.
Initial map is intentionally kept under 10 entries (pinned by a test);
[...]

Issue #1655, opened 2026-06-11
feat(#573): static label map for hardcoded-IP ranges (Apple APNs 17.0.0.0/8, etc.)


sameerparekh:
Follow-up to #573 / PR #1643. Some traffic never presents an SNI and never resolves via our DNS — notably **Apple push (APNs) on 17.0.0.0/8**, NTP, and other hardcoded-IP services. These show up as unattributed IP-only flows even after the SNI sidecar lands.

**Goal:** add a small **static IP-range → label** map (e.g. 17.0.0.0/8 → `apple-push`) consulted as a last-resort attribution source when neither DNS nor SNI yields a hostname, so connection_events for these flows carry a meaningful label instead of a bare IP. Must feed the same attribution primitive (single-source-of-truth) and remain observability/labeling only — NOT an enforcement input.

Keep the map small, in-repo, and documented; this is not a substitute for SNI/DNS attribution.

PR #1713, opened 2026-06-14
feat(#1708): HostId.Label variant — API + SPA accept label-typed hosts (part 1 of 2)

sameerparekh:
Refs #1708. Part 1 of 2.

## What

Adds a distinct `HostId.Label` variant so static IP-range labels
(`apple-push`, `google-dns`, `cloudflare-dns` — see #1655 / #1706's
`openwrt/.../static_ip_labels.lua`) stop riding the wire as `fqdn`s.

## Wire shape

```json
{ "type": "label", "value": "apple-push", "source": "static-ip-range" }
```

- The encoder always emits `source` (today the only value is
  `static-ip-range`).
- The decoder is tolerant of `source`'s absence so a future label source
  (e.g. an ASN-based map) can ship without a wire break.

## Surfaces touched

- `shared/HostId` — new `Label(name: String)` case + wire codec.
- `shared/HostMatch.matchesAny` — Label hosts return `false`. They have no
  `asFqdn`, so the existing fqdn-only matcher naturally skips them, the same
  way it does for `ipv4` / `ipv6`. A synthetic name must never pattern-match
  against `*.example.com`.
- `api/db/TypeMeta` — `Read[HostId]` understands the `label` discriminator.
  `source` is wire-only; the two-column composite `(host_type, host_value)`
  stores no `source`, and the encoder reconstructs it on read.
- `web/types/api` + `web/components/HostCell` — Label rows render with the
  same italic + tag treatment as IP literals (monospace, `text-brand-text`,
  `label` tag), so a synthetic name doesn't visually masquerade as a
  resolved hostname. Tooltip includes the source where present.

## Additive-on-the-wire (AGENTS.md independent-deploy rule)

Older agents continue to emit static-IP labels as `fqdn`-typed hosts; the
new `label` variant is accepted on the wire by this PR but **not yet
emitted by any agent**. The OpenWRT agent's `static_ip_labels.lookup` /
`conntrack.build_event` switch to the new variant ships in a follow-up PR
(#1708 part 2) after this PR deploys, per the "API and the agents deploy
independently" contract.

## Scope NOT in this PR

- **No Flyway migration.** Persistence carries `host_type` as a TEXT column
  with a CHECK constraint pinned to `('fqdn', 'ipv4', 'ipv6')` on four
  tables (`time_usage`, `connection_events`, `block_events`,
  `traffic_reports`). Widening the constraint to include `'label'` is a
  schema-only PR per AGENTS.md migration-isolation — filed as #1712 — and
  must deploy BEFORE the agent emit PR or `INSERT`s with `host_type='label'`
  will be rejected.
- **No agent change.** Lua agent code stays untouched in this PR (part 2).

## Tests (red → green visible in commit history)

- `shared/test/.../HostIdSpec` — encodes Label with source, round-trips with
  and without source, decodes an unknown source verbatim, asserts
  `value` / `kind` / `isFqdn` / `asFqdn` for the new case.
- `shared/test/.../HostMatchSpec` — Label hosts never match an FQDN pattern,
  same default as IP literals.
- `web/.../HostCell.test` — Label rows render value + visible "label" tag.

The DB-roundtrip ingest test that would prove end-to-end Label persistence
is deferred to the schema-migration PR (#1712), because today the four
`host_type` CHECK constraints reject it.

Refs #1655, #1706, #1712.

### DIFF
--- a/api/src/db/TypeMeta.scala
+++ b/api/src/db/TypeMeta.scala
@@ -147,10 +147,14 @@ object TypeMeta {
   // in the same order. Any SQL touching HostId columns must respect that.
   given Read[HostId]  = Read[(String, String)].map { case (kind, value) =>
     kind match {
-      case "fqdn" => HostId.Fqdn(Hostname.unsafe(value))
-      case "ipv4" => HostId.IPv4(IpAddress.unsafe(value))
-      case "ipv6" => HostId.IPv6(IpAddress.unsafe(value))
-      case other  =>
+      case "fqdn"  => HostId.Fqdn(Hostname.unsafe(value))
+      case "ipv4"  => HostId.IPv4(IpAddress.unsafe(value))
+      case "ipv6"  => HostId.IPv6(IpAddress.unsafe(value))
+      // #1708: synthetic label (static IP-range attribution). `source` is wire-only;
+      // not persisted in the two-column composite — there's only one source today
+      // (`static-ip-range`) and the encoder reconstructs it on read.
+      case "label" => HostId.Label(value)
+      case other   =>
         throw new IllegalStateException(s"DB has unknown host_type: $other")
     }
   }
--- a/shared/src/types/HostId.scala
+++ b/shared/src/types/HostId.scala
@@ -23,7 +23,10 @@ enum HostId {
   // inside a hardcoded operator-curated range (apple-push, google-dns, cloudflare-dns;
   // see openwrt static_ip_labels). NOT a hostname the agent ever observed at the
   // resolver — it cannot be pattern-matched against `*.example.com` and `asFqdn`
-  // returns None, the same as IP literals.
+  // returns None, the same as IP literals. The label string is intentionally a
+  // plain `String` (not a validated opaque type): label validity is enforced
+  // UPSTREAM by the curated `static_ip_labels.lua` map; the decoder's only
+  // defense here is rejecting an empty value.
   case Label(name: String)
 }
 
--- a/shared/src/types/HostId.scala
+++ b/shared/src/types/HostId.scala
@@ -19,6 +19,12 @@ enum HostId {
   case Fqdn(name: Hostname)
   case IPv4(addr: IpAddress)
   case IPv6(addr: IpAddress)
+  // #1708: A synthetic name attached by the agent because the destination IP fell
+  // inside a hardcoded operator-curated range (apple-push, google-dns, cloudflare-dns;
+  // see openwrt static_ip_labels). NOT a hostname the agent ever observed at the
+  // resolver — it cannot be pattern-matched against `*.example.com` and `asFqdn`
+  // returns None, the same as IP literals.
+  case Label(name: String)
 }
 
 object HostId {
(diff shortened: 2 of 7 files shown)


sameerparekh:
## Independent review — first run (#1713)

Reviewed `f0b29bc2` against `origin/main` (three-dot, merge-base).

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS

1. [shared/src/types/HostId.scala:55](shared/src/types/HostId.scala:55) — `LabelSourceStaticIpRange` is declared `val` on the public `HostId` object but is only used by the wire encoder one line below. Marking it `private` keeps the surface area honest. Non-blocking — having it nameable from outside makes the part-2 agent emit and any future label-source addition easier to wire up symmetrically, so either choice is defensible.
2. [web/src/types/api.ts:188](web/src/types/api.ts:188) — `source?: string` is untyped. A literal-union `'static-ip-range'` would document the only currently-valid value while keeping forward-compat (`| string` for the future ASN map). Minor — the decoder side already tolerates any string.
3. [web/src/components/HostCell.test.tsx:33](web/src/components/HostCell.test.tsx:33) — the new label test only asserts the value + tag render; it doesn't pin the italic / muted treatment (the #826 standardization) the way the ipv4 / ipv6 tests do. Adding an assertion on the outer `className` would prevent a future "labels render as plain text" regression. Non-blocking — the shared else-branch covers it implicitly today.
4. [shared/src/types/HostId.scala:25](shared/src/types/HostId.scala:25) — `Label(name: String)` accepts any non-empty string; unlike `Fqdn(Hostname)` it has no syntactic guard. Worth a brief comment that label validity is enforced upstream by the curated `static_ip_labels.lua` map — the type intentionally trusts the agent's emit set, and this PR's decoder defensiveness is empty-string-only by design.

### Dimension notes

- **1. Single source of truth** — clean. The new variant is defined once in `HostId.scala`; `TypeMeta.Read` ([api/src/db/TypeMeta.scala:155](api/src/db/TypeMeta.scala:155)) and the SPA tagged union ([web/src/types/api.ts:188](web/src/types/api.ts:188)) follow the same discriminator. `HostMatch.matchesAny` requires no changes because it already delegates to `asFqdn`, which returns `None` for the new case by construction — type-enforced, not "keep in sync".
- **2. Testing integrity** — TDD red→green visible in history (`775ceab2` red, `f0b29bc2` green). No existing assertions weakened or fixtures retrofitted. Both encoder (source emitted) and decoder (with-source + without-source + unknown-source) round-trips are pinned; the existing `fqdn` exact-string assertion ([shared/test/src/types/HostIdSpec.scala:13](shared/test/src/types/HostIdSpec.scala:13)) still passes, which proves the optional `source` field doesn't bloat the legacy wire shape (zio-json omits `None`).
- **3. Architectural invariants** — no DNS-enforcement reasoning, no router decision logic added, no `/tmp` writers, no new snapshot top-level field. The new `HostId` case is a tagged-union variant on an existing wire type, not a new policy channel.
- **4. Backwards compatibility / wire contract** — strictly additive. (a) Older API receiving a `label`-typed payload: previously fell through `Wire(other, _)` → decode error. Now decodes correctly. Decoders that ignore unknown fields tolerate `source`. (b) Older agents continue to ship `fqdn` — encoder for non-label cases passes `source = None`, which serializes to absence, so the byte-level wire for existing types is unchanged. (c) An older SPA bundle hitting a `label`-typed host falls into the IP-literal else branch in `HostCell` — degraded (renders `label` raw tag) but functional. Accept that risk because no agent emits `label` yet (part 2 ships after this and after #1712).
- **5. Migrations** — no migration in this PR (correct per AGENTS.md migration-isolation). The follow-up migration is filed as [#1712](https://github.com/wifihaven/wifihaven/issues/1712); the four `host_type` CHECK constraints (`time_usage`, `connection_events`, `block_events`, `traffic_reports`, including the partitioned-table assertions in V41/V42) will reject `INSERT host_type='label'` until that ships. The PR body documents this and the deploy ordering. `Read[HostId]` learning the `label` discriminator now is harmless — without the migration, no rows ever carry it.
- **6. Metrics + dashboard** — N/A. No new route, job, poller, ingest path, or external call; adding a tagged-union case isn't an instrumented surface.
- **7. Scope / hygiene** — focused diff (7 files, ~110 lines net). No AI mentions in commits / PR title / body. No drive-by edits, no debug logging, no TODOs.
- **8. Security** — no secrets, no SQL, no new routes, no auth surface touched.

### VERDICT: APPROVE

Strictly additive `HostId.Label` variant on the wire + SPA + DB-read layer; encoders/decoders preserve the legacy byte shape for `fqdn` / `ipv4` / `ipv6` and the new variant is type-enforced to never pattern-match (no `asFqdn`). No BLOCKERS or SHOULD-FIXes; four NITs are stylistic. Schema migration (#1712) and agent emit (part 2) follow per the documented deploy ordering.

sameerparekh:
## Independent review — re-run (#1713)

Reviewed `a2ede1b6` (delta `f0b29bc2...HEAD`).

### Prior findings status

| # | Prior NIT | File | Status |
|---|-----------|------|--------|
| 1 | `LabelSourceStaticIpRange` could be `private` | shared/src/types/HostId.scala:55 | NOT-ADDRESSED — intentionally left public; the prior note acknowledged either choice is defensible. |
| 2 | TS `source?: string` could be a literal-union | web/src/types/api.ts:188 | ADDRESSED (tried and reverted) — eslint `@typescript-eslint/ban-types` forbids the `string & {}` widening trick; the new inline comment ([web/src/types/api.ts:188](web/src/types/api.ts:188)) documents the constraint and the only valid current value. |
| 3 | Label-test missed className treatment assertion | web/src/components/HostCell.test.tsx:33 | ADDRESSED — new test [web/src/components/HostCell.test.tsx:42](web/src/components/HostCell.test.tsx:42) pins italic + text-brand-text + font-mono on the label row. |
| 4 | Label trust-boundary comment | shared/src/types/HostId.scala:25 | ADDRESSED — comment on [shared/src/types/HostId.scala:25](shared/src/types/HostId.scala:25) now names the upstream curator (`static_ip_labels.lua`) and the decoder's empty-string-only defense. |

### Delta review (new findings)

#### BLOCKERS
(none)

#### SHOULD-FIX
(none)

#### NITS
(none)

The delta is three comment-only changes and one purely additive test. No new code paths, no new wire shapes, no test weakening, no migration.

### VERDICT: APPROVE

All addressable NITs from the first run are addressed; the one held-back NIT (constant visibility) has acknowledged justification. No BLOCKERS introduced in the delta. Strictly additive `HostId.Label` variant is ready to merge ahead of the schema migration (#1712) and the agent-emit follow-up.

PR #1720, opened 2026-06-14
fix(#1716): sweep orphan conntrack -E subscribers at agent startup

### LINKED ISSUE #1716: bug(prod): router.lan CPU pegged — wifihaven-dns-tail 26% + wifihaven-agent 19%, load avg 4.26 peak 4.71
## Symptom (prod, observed 2026-06-14)

Prod router (router.lan / 192.168.10.1) showing sustained high CPU:

\`\`\`
14450  root R  11484  1% 26%  {wifihaven-dns-t} /usr/bin/lua /usr/sbin/wifihaven-dns-tail
14449  root R   8672  1% 19%  {wifihaven-agent} /usr/bin/lua /usr/sbin/wifihaven-agent
\`\`\`

LuCI system-load panel shows: 1-min load avg 4.26, peak 4.71. Spike pattern: load went from ~0 → ~3.5 within ~30s and is now sustained. Something just kicked off and is running hot.

## Suspects (ranked, from recent CD wave)

1. **#1659 periodic eb_/bl_ ipset re-resolve heartbeat.** Introduced a new background task on the agent. If the cadence is wrong or it re-resolves an unbounded set per tick, that fits sustained agent CPU.
2. **wifihaven-dns-tail is the hotter consumer (26%).** Suggests either a DNS query flood (a device booting up and storming) OR a slow pattern-matching path got added (#1672's expanded InfraHosts list = more regex/exact-match work per query).
3. **#1668 / #1691 v6 conntrack work.** Family-aware conntrack watching could be subscribing to v6 events at higher volume than expected.
4. **#1672 InfraHosts orphan-tail extension.** Bigger list = more per-packet matching work.
5. **#1700 / #1679 allowedDuringScheduleBlock.** Server-side filter; shouldn't affect router directly UNLESS the snapshot grew larger and render.lua now does more work per poll.

QUIC SNI parsing (#1651) lives in \`wifihaven-sni-tail\` — a separate process not shown in top. NOT the suspect.

## Investigation (incident-investigation skill: observe before theorize)

1. **SSH router.lan**, capture state:
   - \`top -bn1 | head -30\` — confirm the two processes are still hot.
   - \`logread | tail -200\` — what's the agent and dns-tail logging? Hot loop should leave a trail.
   - \`cat /tmp/wifihaven-dns-cache.txt | wc -l\` — is the cache file growing unbounded?
   - \`ls -la /tmp/wifihaven*\` — any tmpfs runaway?
   - \`pgrep -af wifihaven\` — what's actually running?
   - \`opkg list-installed | grep wifihaven\` — confirm the agent version (post-CD wave).
2. **Strace one of the hot processes briefly** (\`strace -p <pid> -c -e trace=read,write\` for 5 seconds) to see syscall mix.
3. **Inspect the snapshot size**: \`ls -la /etc/wifihaven/policy.json; jq '.profiles | length, .global' /etc/wifihaven/policy.json\`. Did it suddenly grow?
4. **Check if #1654's UCI toggle would help**: \`uci get wifihaven.sni.enabled\` — if SNI sidecar is the load, the operator now has a clean off-switch. But the SNI process isn't in the top — probably not the cause.

## Stabilize while investigating

If load is impacting household connectivity:
- Stop the suspected process: \`/etc/init.d/wifihaven-dns-tail stop\` (re-enables after fix).
- Last resort: \`/etc/init.d/wifihaven stop\` (kills enforcement temporarily — household goes wide open).

## Fix paths

Depends on diagnosis:
- If #1659 re-resolve heartbeat is firing too often → fix cadence / bound the set.
- If #1672 InfraHosts pattern-matching is slow → optimize the match (build a single combined regex / hash-set lookup).
- If dns-tail is reacting to a query flood → file a separate "device X stormed DNS" issue + add rate-limiting in dns-tail.
- If snapshot grew due to #1700 → audit render.lua for O(n²) over rules.

## Acceptance

- Both processes back to < 5% CPU sustained.
- Root cause pinned with evidence (logread / strace / cache size).
- Regression test (busted) pinning whichever loop / pattern-match was hot.

## Related
- #1659 (PR #1659) — periodic ipset re-resolve
- #1668 / #1691 — v6 conntrack
- #1672 — InfraHosts orphan-tail
- #1700 — allowedDuringScheduleBlock
- #1654 — SNI UCI toggle (available escape hatch if sidecar IS the load)

sameerparekh:
Closes #1716.

## Pinned root cause

**16 orphan `conntrack -E -e NEW` processes (PPID=1)** had accumulated on prod router.lan over a 24-day uptime. All were subscribed to `NFCT_ALL_CT_GROUPS`. Every netfilter event woke all 17 subscribers (16 orphans + the live agent's child) — kernel context-switch and softirq overhead pegged the CPU.

Manually SIGTERMing the 16 orphans on prod dropped agent CPU from **22-24% → 0%** immediately (top sample captured at the time):

```
=== BEFORE (16 orphans alive) ===
14449  root R    22%  /usr/bin/lua /usr/sbin/wifihaven-agent
14450  root R     0%  /usr/bin/lua /usr/sbin/wifihaven-dns-tail
Load average: 3.33 3.35 3.22

=== AFTER orphan sweep ===
14449  root S     0%  /usr/bin/lua /usr/sbin/wifihaven-agent
Load average: 2.63 3.13 3.15
```

Agent voluntary context switches dropped from ~2600/sec historic to ~50/sec post-sweep.

## Why orphans accumulate

`conntrack.watch` opens the subscriber via `io.popen("conntrack -E -e NEW")`. When the agent process is replaced (procd restart, CD reinstall, upgrade) the child is reparented to init. `SIGPIPE` only fires on a write; an idle conntrack with no events never writes, so the orphan never dies. Across many restarts (≈ one per day under CD) the count grows monotonically.

## Fix

At agent startup, before opening the agent's own conntrack subscription, scan `/proc` for `conntrack -E -e NEW` processes whose PPID is 1 and SIGTERM each. Matching is intentionally narrow (exact agent invocation, PPID=1 only) so an operator's transient `conntrack -L`/`-F` and the live agent's own child are never stomped.

Lives in `conntrack.lua` as `kill_orphan_watchers({list_procs_fn, kill_fn})` with both seams injected for tests; wired into [wifihaven-agent startup](openwrt/files/usr/sbin/wifihaven-agent) right before the initial policy fetch.

## Tests

Four busted cases in `conntrack_spec.lua` using fake `list_procs_fn` / `kill_fn`:
- kills PPID=1 matches
- spares the live agent's own child (PPID = current agent)
- ignores `conntrack -L`, `-F`, and `conntrack -E -e NEW -p tcp`
- no-op when no orphans

Red-green progression preserved in two commits.

## Out of scope / follow-ups

- `wifihaven-dns-tail` still burst-spikes to ~27% on prod under a high-rate DNS log stream (per-line `atomic_write` of the full ~133KB cache file is O(N) per line). Separate hot path, not this PR. The dominant baseline CPU traced to this fix; dns-tail spikes are correlated with a storming TimeLimit-blocked device retrying TCP/443 SYNs at ~40/sec and warrant a separate investigation.

## Test plan

- [x] `busted test/conntrack_spec.lua` — 121 successes / 0 failures
- [x] Full openwrt suite (conntrack, policy, render) — 352 successes / 0 failures / 1 pending (`nft` binary not on macOS)
- [ ] Operator: after merge & router update, verify `pgrep -af "conntrack -E -e NEW"` returns exactly one (the agent's child) and agent CPU is < 5% sustained.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -114,7 +114,10 @@ function M.kill_orphan_watchers(opts)
   opts = opts or {}
   local list = opts.list_procs_fn or default_list_procs
   local kill = opts.kill_fn or function(pid)
-    return os.execute("kill " .. tostring(pid) .. " 2>/dev/null") == 0
+    -- os.execute return shape varies by Lua version: int exit-code on 5.1/5.2,
+    -- boolean true on 5.3+ for a clean exit. Mirror nft_eb_hit's normalisation.
+    local ret = os.execute("kill " .. tostring(pid) .. " 2>/dev/null")
+    return ret == 0 or ret == true
   end
   local killed = 0
   for _, p in ipairs(list()) do
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -37,6 +37,96 @@ function M.eb_san(host)
 end
 
 -- ---------------------------------------------------------------------------
+-- list_procs() -> { { pid, ppid, cmdline }, ... }   (default scanner)
+--
+-- Enumerates the live process table by reading /proc/<pid>/status (for PPid)
+-- and /proc/<pid>/cmdline (NUL-separated argv joined with spaces). Used by
+-- kill_orphan_watchers; injectable as opts.list_procs_fn for tests.
+-- ---------------------------------------------------------------------------
+local function default_list_procs()
+  local out = {}
+  local p = io.popen("ls /proc 2>/dev/null", "r")
+  if not p then return out end
+  for name in p:lines() do
+    local pid = tonumber(name)
+    if pid then
+      local sf = io.open("/proc/" .. pid .. "/status", "r")
+      local ppid
+      if sf then
+        for line in sf:lines() do
+          local v = line:match("^PPid:%s*(%d+)")
+          if v then ppid = tonumber(v); break end
+        end
+        sf:close()
+      end
+      local cf = io.open("/proc/" .. pid .. "/cmdline", "r")
+      local cmdline = ""
+      if cf then
+        local raw = cf:read("*a") or ""
+        cf:close()
+        -- argv is NUL-separated with a trailing NUL; replace NUL with space
+        -- and strip the trailing whitespace so it matches a plain command line.
+        cmdline = raw:gsub("%z", " "):gsub("%s+$", "")
+      end
+      if ppid then
+        out[#out + 1] = { pid = pid, ppid = ppid, cmdline = cmdline }
+      end
+    end
+  end
+  p:close()
+  return out
+end
+
+-- ---------------------------------------------------------------------------
+-- kill_orphan_watchers(opts) -> int (orphans killed)  -- #1716
+--
+-- Scan /proc for `conntrack -E -e NEW` processes that have been reparented to
+-- init (PPID = 1) — those are leftover children of a prior wifihaven-agent
+-- instance. SIGTERM each one before opening this agent's own conntrack
+-- subscription.
+--
+-- Why this is needed: conntrack subscribes to NFCT_ALL_CT_GROUPS on its
+-- netlink socket, idles waiting for events, and only ever writes to stdout
+-- when an event arrives. When the agent that opened it via io.popen() is
+-- replaced (procd restart, CD reinstall, upgrade), the child is reparented
+-- to init. SIGPIPE fires only on write, so an idle orphan with no events
+-- never gets SIGPIPE and never dies. Across many restarts the orphans
+-- accumulate, every netfilter event wakes all subscribers, and CPU + load
+-- climb. Prod #1716: 24-day uptime → 16 orphans → agent CPU 22-24% / load
+-- 3.2; SIGTERMing the orphans dropped agent CPU to 0%.
+--
+-- opts:
+--   list_procs_fn  function() -> { {pid, ppid, cmdline}, ... }   injectable
+--                  process scanner; defaults to default_list_procs above
+--                  (reads /proc). Tests pass a closure over a canned list.
+--   kill_fn        function(pid) -> bool                          injectable;
+--                  defaults to `kill <pid>` via os.execute. Returns true on
+--                  successful signal delivery (best-effort).
+--
+-- Matching is intentionally narrow: we only sweep processes whose cmdline is
+-- EXACTLY `conntrack -E -e NEW` (the agent's invocation) — not `conntrack -L`,
+-- `conntrack -F`, `conntrack -E -e NEW -p tcp`, etc. — so an operator's
+-- transient conntrack tool isn't stomped.
+-- ---------------------------------------------------------------------------
+local AGENT_CONNTRACK_CMDLINE = "conntrack -E -e NEW"
+
+function M.kill_orphan_watchers(opts)
+  opts = opts or {}
+  local list = opts.list_procs_fn or default_list_procs
+  local kill = opts.kill_fn or function(pid)
+    return os.execute("kill " .. tostring(pid) .. " 2>/dev/null") == 0
+  end
+  local killed = 0
+  for _, p in ipairs(list()) do
+    if p.ppid == 1 and p.cmdline == AGENT_CONNTRACK_CMDLINE then
+      kill(p.pid)
+      killed = killed + 1
+    end
+  end
+  return killed
+end
+
+-- ---------------------------------------------------------------------------
 -- nft_eb_hit(dst_ip, eb_host, exec_fn) -> bool
 --
 -- Returns true when dst_ip is a current member of the nftables set
--- a/openwrt/files/usr/sbin/wifihaven-agent
+++ b/openwrt/files/usr/sbin/wifihaven-agent
@@ -436,6 +436,22 @@ local last_nflog_run      = 0  -- #1126: nflog drop-spool drain cadence
 local activity_tracker     = usage.new_tracker()
 local usage_queue          = usage.new_queue()  -- retry queue for failed POSTs (#309)
 
+-- ── Sweep orphan conntrack subscribers (#1716) ──────────────────────────────
+-- conntrack -E -e NEW is opened via io.popen() below. When this agent process
+-- is replaced (procd restart, CD reinstall, upgrade), the child is reparented
+-- to init and idles on netlink with no reader for its stdout. SIGPIPE only
+-- fires on write, and an idle conntrack with no events never writes — so the
+-- orphan never dies. Across many restarts the orphans accumulate, every
+-- netfilter event wakes all subscribers, and CPU + load climb. Prod #1716:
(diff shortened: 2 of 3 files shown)


sameerparekh:
Independent review (read-only, adversarial) per [docs/pr-review-checklist.md](docs/pr-review-checklist.md). First-run review of the merge-base diff `origin/main...HEAD`.

## BLOCKERS

(none)

## SHOULD-FIX

1. **[openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:113](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua#L113)** — Default `kill_fn` returns `os.execute(...) == 0`, but on Lua 5.3+ `os.execute` returns `true, "exit", 0` (a tuple, not a number) for success — the `== 0` comparison evaluates `true == 0` and is always **false**. The return value is currently unused (kill is best-effort), so this is latent rather than active, but the comparison is wrong and will mislead anyone who later inspects the result. OpenWrt 23.x's stock Lua is 5.1 where this works, so prod is safe — but `conntrack.lua` uses Lua-5.3-compatible patterns elsewhere (`M.nft_eb_hit` explicitly handles both the int and boolean return shapes — see [conntrack.lua:69-79](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua#L69)), so this should mirror that handling. Suggest: copy the existing `nft_eb_hit` pattern (`ret == 0 or ret == true`).

2. **[openwrt/files/usr/sbin/wifihaven-agent:439-453](openwrt/files/usr/sbin/wifihaven-agent#L439)** — Per [AGENTS.md §instrument-new-functionality](AGENTS.md#instrument-new-functionality), a new meaningful code path should ship a metric. The startup sweep is exactly the kind of fleet-state signal an operator wants visibility on — if it keeps finding non-zero orphans every restart across the fleet, that's a regression. Suggest an `orphan_conntrack_swept_total` counter (or a one-shot `orphan_conntrack_swept_at_startup` gauge), routed via `AppMetrics` / `MetricGuard`, plus the matching Grafana panel per [§metrics-need-a-dashboard](AGENTS.md#metrics-need-a-dashboard). The `log.info` is good for one-router debugging but not for fleet-level trend detection.

3. **[openwrt/files/usr/sbin/wifihaven-agent](openwrt/files/usr/sbin/wifihaven-agent)** — This fix sweeps orphans **at next startup**, which means every restart still **creates** one orphan (between this process dying and the next one sweeping). The orphan count is now bounded at ~1 instead of growing — a huge improvement and acceptable as-is. But the cleaner complement is **explicit cleanup on agent shutdown**: install a signal handler / `atexit` that explicitly kills the popened `conntrack -E -e NEW` child. Not blocking — the startup sweep solves the prod incident — but worth a follow-up issue or a comment in the agent pointing to it as the proper long-term fix. Without it, an agent process that crashes mid-restart still drops one orphan, and the prod CPU regression is one "fleet didn't get the update" away.

## NITS

1. **[openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:45](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua#L45)** — `io.popen("ls /proc 2>/dev/null", "r")` works but is a subprocess fork just to enumerate `/proc`. Some Lua POSIX builds expose `dir`-style iteration; on stock OpenWrt busybox, `ls` is fine and the cost is paid once at startup. Non-blocking — flagging only because the function comment doesn't mention the fork.

2. **[openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:101](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua#L101)** — `kill` defaults to SIGTERM. Conntrack handles SIGTERM cleanly, so this is correct. If a future kernel/conntrack-tools combination ever masks SIGTERM, this would silently no-op — a SIGKILL fallback after a brief delay would be more robust, but unnecessary today.

3. **[openwrt/test/conntrack_spec.lua:1888](openwrt/test/conntrack_spec.lua#L1888)** — The "ignores `conntrack -L` / `-F`" test is great. Worth one more case: a process whose cmdline starts with `conntrack -E -e NEW ` (with a trailing flag) — already covered as `... -p tcp`, but inverse-asserting that an extra-whitespace variant (`"conntrack  -E -e NEW"`) does NOT match would harden the contract. Trivial; non-blocking.

## VERDICT: APPROVE

Tight, well-scoped fix for #1716. Clear root-cause evidence (16 orphans → 0% agent CPU after manual sweep), narrow matching (PPID=1 AND exact cmdline) avoids stomping unrelated `conntrack` tools or the live child, RED → GREEN visible in commit history, and the test seams (`list_procs_fn` / `kill_fn`) keep the unit suite hermetic. The SHOULD-FIX items (Lua 5.3 truthy comparison, instrumentation, shutdown handler) are improvements rather than blockers.

sameerparekh:
Re-run review. Prior-comment SHA `dbbb7808`. Delta reviewed: `git diff dbbb7808...HEAD`.

## Status of prior findings

| # | Class | Cite | Status |
|---|---|---|---|
| 1 | SHOULD-FIX | `conntrack.lua:113` Lua 5.3 `os.execute` return shape | **ADDRESSED** — now normalises `ret == 0 or ret == true`, mirroring `nft_eb_hit`'s pattern. |
| 2 | SHOULD-FIX | `wifihaven-agent:439` startup sweep lacks a metric | **PARTIAL** — code comment now points to follow-up [#1722](https://github.com/wifihaven/wifihaven/issues/1722) where the counter + Grafana panel will land. Accepting the deferral: a fleet-visible metric is a cross-process (agent ↔ API ↔ dashboard) change that's cleanly separable from the CPU-regression fix. |
| 3 | SHOULD-FIX | `wifihaven-agent` shutdown handler | **PARTIAL** — code comment now points to follow-up [#1723](https://github.com/wifihaven/wifihaven/issues/1723) for the SIGTERM-time child cleanup. Accepting the deferral: the startup sweep alone bounds the orphan count at ~1 per restart, the prod CPU symptom is fixed, and a crash-path orphan (SIGKILL / OOM) needs the startup sweep as a safety net regardless. |
| 4 | NIT | conntrack_spec extra-whitespace case | **ADDRESSED** — extra-whitespace fixture added, asserts it's NOT swept. |

## New findings (delta `dbbb7808...HEAD`)

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS

1. **[conntrack.lua:117](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua#L117)** — `ret == 0 or ret == true` is the right normalisation, but Lua 5.1 on a `kill` that successfully signalled a non-existent PID returns shell exit 1 (kill: no such process), so a `ret ~= 0 and ret ~= true` outcome here is *also* a valid "we already cleaned this up, race with another sweeper" case, not an error. Best-effort return is fine and unused; just noting that a future caller who *uses* the return value should treat a falsy result as "unverified" rather than "definitely failed."

## VERDICT: APPROVE

Prior SSoT-class concerns (Lua 5.3 portability, narrow matching) are ADDRESSED. The remaining SHOULD-FIXes (metric + shutdown handler) are intentionally deferred under tracked follow-ups (#1722 / #1723) with code-comment breadcrumbs — appropriate scoping for a hot-fix that needs to ship to stop a live prod regression. The delta itself introduces no new BLOCKERs or SHOULD-FIXes.

Issue #1718, opened 2026-06-14
feat(observability): router agent publishes native OS metrics (load / bandwidth / wireless / connections) — match what LuCI shows


sameerparekh:
## Goal

Have the router agent publish **native OpenWRT metrics** alongside the existing wifihaven-specific metrics, so operators can monitor router health from the cloud Grafana dashboard the same way they would from LuCI locally. Categories:

- **Load** — 1m / 5m / 15m load average + CPU% from `/proc/loadavg` and `/proc/stat`
- **Memory** — total / free / buffered / cached from `/proc/meminfo`
- **Bandwidth** — per-interface rx_bytes / tx_bytes from `/sys/class/net/<iface>/statistics/` (or `/proc/net/dev`)
- **Wireless** — connected clients per SSID, current channel, TX power, signal stats via `iw dev <wlan> station dump` / `iw dev <wlan> info` / ubus `network.wireless`
- **Connections** — conntrack count via `/proc/sys/net/netfilter/nf_conntrack_count`, optional breakdown by L4 proto

Use case: #1716 (prod router CPU pegged, 24-day-old agent) would have been MUCH faster to diagnose if these metrics were already in Grafana — operator could have seen the load climb start exactly when, alongside the existing agent metrics, without having to SSH the router.

### Where the data comes from (Lua-side, no shell-out unless necessary)

| Metric | Source |
|---|---|
| Load avg + CPU% | `/proc/loadavg`, `/proc/stat` (compute delta between polls for CPU%) |
| Memory | `/proc/meminfo` |
| Bandwidth | `/sys/class/net/<iface>/statistics/{rx_bytes,tx_bytes}`; pick a stable interface set (lan, wan, br-lan, wlanX) |
| Wireless | `iw dev <wlan> station dump` (parse) + `iw dev <wlan> info`; `iwinfo` if available — ubus may be cleanest |
| Conntrack count | `/proc/sys/net/netfilter/nf_conntrack_count` (single integer) |

All file reads are cheap (no syscalls beyond open/read; tens of microseconds per poll). The wireless `iw dev` calls are slightly heavier but bounded — once per poll interval is fine.

### Wire shape (additive only)

Extend the existing `/api/router/metrics` request body — same endpoint that's already being POSTed to per #1716 prod logs (which currently returns 404, separately tracked).

```jsonc
{
  // … existing wifihaven metrics …
  "host": {
[...]

PR #1721, opened 2026-06-14
feat(#1718): schema for native router host metrics (PR 1 of 3)

sameerparekh:
## Summary

Schema-only migration for #1718 (router agent publishes native OS metrics — load / bandwidth / wireless / connections). Per AGENTS.md §migrations-back-compat this PR carries **only** the Flyway migration + the explanatory header comments inside it; no `api/src`, no tests, no CI changes. The existing `api.test` suite (which applies every migration before running) is the back-compat gate.

Three new tables:

- `router_host_metrics` — scalar vitals per `(router, tick)`: load avg, cpu%, memory, conntrack count
- `router_host_iface_metrics` — per-iface rx/tx_bytes per `(router, tick, iface)`
- `router_wifi_clients` — per-MAC wireless drill-in per `(router, tick, mac)`; explicitly DB-only, never a Prometheus label (cardinality firewall, AGENTS.md §instrument-new-functionality)

Indexes target the read patterns the follow-up Grafana panels will use:
- BRIN on `ts` for "last N hours per router" range scans (1000× smaller than btree, competitive on monotonic columns)
- `(mac, ts DESC)` on wifi clients for per-device RSSI drill-in

## Follow-ups (filed under #1718)

- **PR 2** — API decode + persist optional host field, AppMetrics aggregates (router/iface only), Grafana panels.
- **PR 3** — OpenWRT agent reads /proc and iw dev station dump, composes the host block on every metrics tick.

## Growth-table note (§migrations-prod-data-volume)

All three are unbounded-growth (one row per router per tick × fleet size). The migration itself is metadata-only CREATE TABLE on empty relations — O(1) regardless of prod volume — so the slow-on-prod risk does not apply to V57. Read-path queries in PR 2 will prove their plans with EXPLAIN ANALYZE per §query-explain-before-merge before they merge.

## Test plan

- [ ] CI api.test passes (gate — confirms image-(N-1) survives DB-at-V57)
- [ ] check-migration-isolation passes (only .sql + .md)
- [ ] PR 2 wires the read path with plan validation

### DIFF
--- a/api/resources/db/migration/V57__router_host_metrics.sql
+++ b/api/resources/db/migration/V57__router_host_metrics.sql
@@ -1,0 +1,100 @@
+-- V57__router_host_metrics.sql
+-- #1718: router agent publishes native OS metrics (load / bandwidth / wireless
+-- / connections) so operators can monitor router health from the cloud Grafana
+-- dashboard the same way they would from LuCI locally.
+--
+-- Per AGENTS.md §migrations-back-compat this is the schema-only PR. The
+-- follow-up PR adopts these tables in api/src (request decode + persist) and
+-- adds the Grafana panels that query them. A second follow-up wires the agent
+-- to actually emit the `host` block on the next metrics tick.
+--
+-- ── Three tables, not one ───────────────────────────────────────────────────
+-- `host` in the wire JSON is one block, but it fans out into three relations
+-- because the cardinalities are wildly different:
+--
+--   1. Scalar host vitals (load, cpu%, mem, conntrack) — one row per
+--      (router, tick). Wide row, low write-rate.
+--   2. Per-iface bandwidth — many ifaces per tick (br-lan, wan, phyN-apM…),
+--      so (router, ts, iface) is the natural PK. Storing as a JSONB column
+--      on table 1 would force per-iface graphs to unpack JSON for every
+--      datapoint in Grafana — cheap inserts, expensive reads.
+--   3. Per-MAC wireless clients — explicitly NOT a Prometheus dimension per
+--      the cardinality firewall (AGENTS.md §instrument-new-functionality).
+--      DB-resident only; Grafana surfaces this via an API-backed table-panel
+--      drill-in, not via labeled time series.
+--
+-- ── Growth-table caveats (AGENTS.md §migrations-prod-data-volume) ───────────
+-- All three are unbounded-growth (one row per router per tick, scaling with
+-- fleet size). The migration itself is schema-metadata-only — `CREATE TABLE`
+-- on empty relations is O(1) regardless of prod volume — so the
+-- "fast-on-dev-slow-on-prod" risk doesn't apply to V57 itself. But every
+-- query against these tables that the follow-up code adds must prove its plan
+-- with EXPLAIN (ANALYZE) per §query-explain-before-merge before it merges.
+-- The indexes below are sized for the read patterns Grafana will use:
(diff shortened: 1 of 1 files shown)


sameerparekh:
Independent review pass per docs/pr-review-checklist.md. Read-only; diff reviewed against `origin/main...HEAD` (three-dot).

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `api/resources/db/migration/V57__router_host_metrics.sql:33` — no `DEFAULT NOW()` on `ts`. Intentional and fine (PR 2 will stamp a single `ts` server-side across all three tables for a given tick), but a one-line header note that the API owns the stamp would help future readers.
2. `V57` BRIN indexes on `ts` are great for "last N hours per router" but assume near-monotonic inserts; if a router ever backfills (clock skew recovery, replay), BRIN degrades to wide block ranges. Not blocking — degradation is graceful — but PR 2 should sanity-check planner choice on the first prod read.
3. `idx_router_wifi_clients_mac_ts (mac, ts DESC)` is right for per-device drill-in. If PR 2 also wants "all clients seen on this router in the last N min" by `(router_id, recent ts)`, the BRIN at the bottom should cover it via PK prefix + range, but PR 2 should pin that with `EXPLAIN ANALYZE`.

### Dimension notes

- **1. Duplicated logic / SSOT** — N/A. Schema only; no decision logic. The three-table fan-out (scalar / iface / wifi-clients) is the correct shape, not duplication — distinct cardinalities and access patterns.
- **2. Testing integrity** — N/A by design. Per AGENTS.md §migrations-back-compat the migration PR carries no test changes; the existing `api.test` suite (applies every migration before running) is the back-compat gate against image-(N-1).
- **3. Architectural invariants** — Tables persist OS-level observability, not policy. Router-as-dumb-applier and DNS-never-enforces are unaffected.
- **4. Backwards compatibility / wire contract** — N/A for this PR (no wire changes). The additive `host` field is correctly deferred to PR 2.
- **5. Migrations** — Isolation: only `*.sql`; `check-migration-isolation.sh` passes locally. New `V57`, doesn't edit any applied migration. Prod-volume safety: `CREATE TABLE` on empty relations is metadata-only, not on the §migrations-prod-data-volume risk list. No new SQL queries → §query-explain-before-merge defers to PR 2 (PR body acknowledges).
- **6. Metrics + dashboard** — Deferred to PR 2 with AppMetrics aggregates + Grafana panels, correctly.
- **7. Scope / hygiene** — Single file, focused. No drive-by edits. PR title/body have no AI mentions; `Co-Authored-By` trailer is the allowed form.
- **8. Security** — No secrets, no DML, no SQL string interpolation. `ON DELETE CASCADE` is correct: orphan `router_id` rows would be unreadable anyway.

VERDICT: APPROVE

Clean schema-only migration for #1718 that respects the migration-isolation rule. Three-table split correctly separates scalar / iface / per-MAC cardinality. Indexes match the read patterns PR 2 will need; explicit `EXPLAIN ANALYZE` validation deferred to PR 2 per AGENTS.md §query-explain-before-merge.

sameerparekh:
Closing — approach changed. We don't need to persist these to the database; the host metrics will pass through to Grafana like the existing router metrics (AppMetrics → Prometheus → Grafana Cloud). No schema needed. Will reopen #1718 as a 2-PR path: (A) API decode + AppMetrics aggregates + Grafana panels, (B) agent emit.

PR #1725, opened 2026-06-14
feat(#1712): widen host_type CHECK to accept 'label' (schema for #1708 part 2)

sameerparekh:
Closes #1712. Schema-only prerequisite for the #1708 part 2 (OpenWRT
agent emit) PR — must merge + deploy before the agent starts producing
`host_type='label'` rows.

> **Updated 2026-06-14**: renumbered from V57 → V58 after main advanced
> with V57 (#1719 `drop_legacy_schedules`). No semantic change.

## What

PR #1713 added the `HostId.Label` wire variant and the API decoder /
SPA renderer, but it explicitly deferred the matching DB constraint
widening to this PR (see #1713 body, "Scope NOT in this PR"). Today
the four host_type CHECK constraints — on `time_usage`, `block_events`,
`traffic_reports`, `connection_events` — are pinned to
`('fqdn','ipv4','ipv6')` (V13; partition parents redefined in V41/V42).
Any INSERT with `host_type='label'` is rejected at write time.

This migration drops and re-adds each constraint with `'label'`
added to the value set.

## Lock / scan profile

`ADD CONSTRAINT ... NOT VALID` is used on all four tables. The new set
is a strict superset of the old, so every existing row is valid by
construction; the regular ADD CONSTRAINT path's full-table verification
scan would buy nothing and (per AGENTS.md §migrations-prod-data-volume)
the two unbounded-growth tables — `traffic_reports`, `connection_events`
— would risk a multi-minute scan past the Render 15-minute port-bind
timeout (cf. #1197). NOT VALID is an O(1) catalog change; enforcement on
new INSERTs/UPDATEs is immediate. VALIDATE can be run out-of-band later
as hygiene, but is not required for correctness.

## Migration-isolation

Only the SQL file. No source, no test, no CI change — gated by
`check-migration-isolation.sh`.

## Deploy ordering

1. Merge this PR → auto-deploys to staging → prod.
2. After prod is on V58, open the agent PR (#1708 part 2 = #1727) that
   switches `openwrt/.../conntrack.lua` to emit `host_type='label'` for
   static-IP attributions.

Refs #1708, #1713.

PR #1727, opened 2026-06-14
feat(#1708): OpenWRT agent emits HostId.Label for static-map attribution (part 2 of 2)

### LINKED ISSUE #1655: feat(#573): static label map for hardcoded-IP ranges (Apple APNs 17.0.0.0/8, etc.)

### LINKED ISSUE #1708: HostId: add a distinct `label` variant so static IP-range labels don't masquerade as FQDNs

sameerparekh:
Closes #1708. Closes #1655. Part 2 of 2 — part 1 was
[#1713](https://github.com/wifihaven/wifihaven/pull/1713) (API + SPA
decoder for the `HostId.Label` wire variant, merged 2026-06-14).

**Supersedes [#1706](https://github.com/wifihaven/wifihaven/pull/1706).**
That PR added the same `static_ip_labels` module but emitted its labels
as `HostId.Fqdn` — exactly the bug #1708 was filed to fix. This PR
vendors in #1706's map + tests and switches the emit path to the new
`label` variant from the start. #1706 should be closed.

## Wire shape

Static-map attributions now ride the wire as:

```json
{
  "type": "connection_attempt",
  "mac":  "aa:bb:cc:11:22:33",
  "host": { "type": "label", "value": "apple-push", "source": "static-ip-range" },
  "destIp":  "17.188.166.41",
  "allowed": true,
  "reason":  "allow",
  "ts":      "2026-06-14T00:00:00Z",
  "eventId": "..."
}
```

Matching `HostId.Label` as defined in `shared/src/types/HostId.scala`
(#1713). DNS, SNI/QUIC, ipset, and IP-literal paths are **unchanged** —
they keep emitting `fqdn` / `ipv4` / `ipv6` as before.

## What changes

* New `openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua`.
  Pure CIDR → label map for the IP-range owners that bypass dnsmasq:
  Apple's 17.0.0.0/8 (APNs), Google Public DNS (8.8.8.8, 8.8.4.4),
  Cloudflare Public DNS (1.1.1.1, 1.0.0.1). `lookup(ip)` returns
  `(label, source)` — the source string lives in the module
  (`M.SOURCE = "static-ip-range"`) so an operator extending the map
  cannot vary it.
* `conntrack.handle_flow` consults `static_ip_labels.lookup` as the
  **last resort**, after `attribute_hostname` (DNS / SNI / QUIC via
  the shared cache) and `ipset_lookup_hostname`. A real attribution
  always wins.
* `conntrack.build_event` is now the **single place** that renders
  each HostId variant. A new `host_label_source` opts field selects
  `type="label"` with the source carried through; otherwise the
  existing fqdn / ipv4 / ipv6 branches apply. Per the AGENTS.md
  single-source-of-truth rule, the wire shape is assembled in
  exactly one place — callers thread the kind/source through opts,
  they never construct the wire shape themselves.

## Deploy ordering

This PR depends on:

1. **#1713** (HostId.Label decoder + SPA renderer) — **MERGED** ✅
2. **#1725** (V58 widens `host_type` CHECK to accept `'label'`) — must
   be merged + deployed to prod **before this PR merges**, otherwise
   prod INSERTs from this agent will be rejected by the four CHECK
   constraints (`time_usage`, `block_events`, `traffic_reports`,
   `connection_events`).

Once both are on prod, this PR is safe to merge.

## Tests (red → green visible in commit history)

* `test(#1708): red` — `425a84c8`. Adds failing `build_event` test for
  `host_label_source` plumbing + failing `handle_flow` test asserting
  `host.type='label'` / `value='apple-push'` / `source='static-ip-range'`
  for an Apple-push IP with no DNS/ipset attribution. Plus four
  regression guards: DNS-wins / ipset-wins / fqdn-still-no-source /
  all-miss → ipv4 literal.
* `feat(#1708): ...` — implementation. 737 / 737 Lua tests pass
  (`sh openwrt/test/run_tests.sh`).
* New `openwrt/test/static_ip_labels_spec.lua` covers the map directly:
  in-range hits return both label and source, edges, out-of-range
  misses, malformed input, v6 punt, and a size cap pinning the
  operator-curated map at ≤ 10 entries.

## Docs

`docs/architecture.md` §7.2 adds **Static IP-range labels** covering the
precedence chain, the labels-only attribution rule, and the
`HostId.Label` wire shape — replaces the "currently emitted as fqdn"
caveat #1706 would have introduced. The block also notes the downstream
guarantee that `HostMatch.matchesAny` returns `false` for label-typed
hosts so a synthetic label can never pattern-match against a real apex.

## Not in this PR

* **No version bump.** `openwrt/Makefile` keeps `PKG_VERSION=0.1.0`
  — agent versions aren't bumped per feature in this repo (recent
  feature PRs at #1702 / #1690 / #1643 didn't bump either); the
  auto-updater handles fleet rollout via opkg revisions.
* **No Flyway migration.** V58 lives on #1725.
* **No source/test changes outside the agent path.** This is Lua-only.

## Closes

Closes #1708. Closes #1655. Refs #1713, #1712, #1725. Supersedes #1706.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -848,6 +848,53 @@ rather than smuggling an IP into a hostname field, so the type system
 prevents accidental pattern matches against `*.example.com` and the admin
 UI can render direct-IP rows distinguishably from named hosts.
 
+#### Static IP-range labels (#1655 / #1708)
+
+Before falling through to the IP literal, the agent consults a small,
+in-repo **static IP-range → label map**
+(`openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua`) as a
+last-resort attribution source. Some traffic never presents an SNI *and*
+never resolves via dnsmasq — notably Apple push (APNs) on `17.0.0.0/8`,
+and apps that bypass the local resolver and hit public resolvers
+(`8.8.8.8`, `1.1.1.1`) directly. With the map, those flows show up in
+connection_events as `apple-push` / `google-dns` / `cloudflare-dns`
+instead of a bare IP.
+
+**Precedence in `handle_flow`:**
+
+1. `attribute_hostname` — dnsmasq query-log cache (the SNI/QUIC sidecars
+   also feed this primitive via `cache.insert_sni`, so SNI- and
+   QUIC-derived hostnames flow through the same path — #573, #1651).
+2. `ipset_lookup_hostname` — legacy `nft_sets` fallback for site_limits
+   domains.
+3. `static_ip_labels.lookup` — last-resort static map (this section).
+
+A real attribution always beats a static guess.
+
+**LABELS ONLY — never an enforcement input.** The static map participates
+in zero drop or carve-out predicates: enforcement is the per-MAC
+`BlockRules` / nftables pipeline (see §0.1 / §0.2). Promoting an entry to
+an enforcement input requires explicit operator approval and a tracking
+issue. The header comment in `static_ip_labels.lua` repeats this.
+
+**Wire shape.** Static-map attributions ride the wire as the
+`HostId.Label` variant (#1708): `{ "type": "label", "value": "<label>",
+"source": "static-ip-range" }`. The `source` string is threaded through
+`conntrack.build_event` from `static_ip_labels.lookup` so the wire shape
+is assembled in exactly one place. Downstream `HostMatch.matchesAny`
+returns `false` for label-typed hosts, so a synthetic label can never
+pattern-match against a real apex — an operator who configures an app
+whose host-set happens to include the literal `apple-push` will *not*
+see APNs flows attribute into that app. Distinguishing labels from real
+FQDNs at the wire level was the whole point of #1708; this section
+replaces the earlier "currently emitted as fqdn" caveat.
+
+**Extending the map.** Add a `{ cidr, label }` row to `M._ranges` with a
+citation in the comment beside it, and a test in
+`openwrt/test/static_ip_labels_spec.lua`. Keep growth operator-curated:
+entries should cover ranges with an unambiguous, well-known owner that
+prod evidence shows dominating the unattributed-IP tail.
+
 ### 7.3 Package layout
 
 ```
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -776,6 +776,18 @@ function M.handle_flow(flow, ctx, batcher)
     hname, hlabel_source = static_ip_labels.lookup(flow.dst_ip)
   end
 
+  -- #1708 carve-out / blocklist matching uses `match_hname`, NOT `hname`:
+  -- label-typed attributions ("apple-push", "google-dns") must not feed the
+  -- string-level suffix tests below (extraAllowed carve-out, extraBlocked
+  -- host_matches, category-blocklist host_matches). The server-side
+  -- HostMatch.matchesAny already returns false for HostId.Label, so a label
+  -- can never appear in ea_hosts/eb_hosts/bl_hosts in the first place — this
+  -- is defense-in-depth keeping the label/fqdn boundary local to the agent.
+  -- For label-attributed flows we fall through to the IP-based nft_eb_hit
+  -- path, which is the correct semantics: there is no real hostname here,
+  -- only an IP and a synthetic display name.
+  local match_hname = (hlabel_source == nil) and hname or nil
+
   -- Per-MAC block lookup (#297): pause and time-limit block every flow from
   -- the device, regardless of destination IP. render.update_shared rebuilds
   -- blocked_macs/blocked_reason from the policy snapshot on every poll.
@@ -795,11 +807,11 @@ function M.handle_flow(flow, ctx, batcher)
   -- whether the ea_ carve-out fired.  The flow is left as blocked=true even
   -- if the kernel allowed it.  This is documented as a known reporting gap
   -- when DNS attribution is missing for flows from a blocked MAC.
-  if not allowed and hname and mac then
+  if not allowed and match_hname and mac then
     local ea_hosts = ctx.ea_hosts_by_mac and ctx.ea_hosts_by_mac[mac]
     if ea_hosts then
       for ea_host in pairs(ea_hosts) do
-        if M.host_matches(hname, ea_host) then
+        if M.host_matches(match_hname, ea_host) then
           allowed = true
           reason  = nil
           break
@@ -837,8 +849,8 @@ function M.handle_flow(flow, ctx, batcher)
     local ea_hosts = ctx.ea_hosts_by_mac and ctx.ea_hosts_by_mac[mac]
     if not ea_hosts then return false end
     for ea_host in pairs(ea_hosts) do
-      if hname then
-        if M.host_matches(hname, ea_host) then return true end
+      if match_hname then
+        if M.host_matches(match_hname, ea_host) then return true end
       else
         if ea_host == eb_hit_host then return true end
       end
@@ -852,8 +864,8 @@ function M.handle_flow(flow, ctx, batcher)
       local eb_hit = false
       local eb_hit_host
       for host in pairs(eb_hosts) do
-        if hname then
-          if M.host_matches(hname, host) then
+        if match_hname then
+          if M.host_matches(match_hname, host) then
             eb_hit = true; eb_hit_host = host; break
           end
         else
@@ -890,8 +902,8 @@ function M.handle_flow(flow, ctx, batcher)
       local bl_hit_host
       local bl_hit_id
       for host, id in pairs(bl_hosts) do
-        if hname then
-          if M.host_matches(hname, host) then
+        if match_hname then
+          if M.host_matches(match_hname, host) then
             bl_hit_host = host; bl_hit_id = id; break
           end
         else
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -23,6 +23,8 @@
 
 local M = {}
 
+local static_ip_labels = require("wifihaven.static_ip_labels")
+
 -- ---------------------------------------------------------------------------
 -- eb_san(host) -> string
 --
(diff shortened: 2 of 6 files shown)


sameerparekh:
Independent review pass per [`docs/pr-review-checklist.md`](https://github.com/wifihaven/wifihaven/blob/main/docs/pr-review-checklist.md). First run — full merge-base diff (`origin/main...HEAD`).

BLOCKERS
(none)

SHOULD-FIX
1. [openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:775-777](https://github.com/wifihaven/wifihaven/blob/782c75631877b09060cd19e599f313a0ac2288af/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua#L775-L777) — the static-map lookup feeds the SAME `hname` variable that the **extraAllowed carve-out** (`if not allowed and hname and mac then ... host_matches(hname, ea_host)` ~L789) and the category-blocklist hit check downstream both read. Previously a static-IP-attributed flow had `hname == nil` and naturally skipped both. Post-PR, `hname == "apple-push"` (etc.) can hit `host_matches` against an operator-configured `extraAllowed` entry. The architectural guarantee that *labels never pattern-match an apex* is enforced **server-side** (`HostMatch.matchesAny` returns false for `HostId.Label`), so `ea_hosts_by_mac[mac]` *should* never carry a value that suffix-matches "apple-push" — but the agent's `host_matches` is a string-level suffix test that doesn't know its input is label-typed. Recommend either (a) treating a label-typed `hname` as nil for the carve-out + category checks (e.g. branch on `hlabel_source`), or (b) writing a regression test that pins the current behavior under the architectural guarantee. (a) is defense-in-depth, cheap, and keeps the label / fqdn semantic boundary local to the agent.

NITS
1. [openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua:32](https://github.com/wifihaven/wifihaven/blob/782c75631877b09060cd19e599f313a0ac2288af/openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua#L32) — `M.SOURCE = "static-ip-range"` is the wire-contract twin of Scala-side `HostId.LabelSourceStaticIpRange` ([`shared/src/types/HostId.scala:61`](https://github.com/wifihaven/wifihaven/blob/main/shared/src/types/HostId.scala#L61)). Both sides have a cross-reference comment, which is the right shape — but a contract test (the `openwrt/test/contract_*` family already exists) pinning the literal would prevent drift if a future operator edits one side. Filing a follow-up issue is enough.
2. [openwrt/files/usr/lib/lua/wifihaven/nflog.lua:93](https://github.com/wifihaven/wifihaven/blob/main/openwrt/files/usr/lib/lua/wifihaven/nflog.lua#L93) (out of diff, but flagged by single-source-of-truth) — `nflog.build_event` carries the comment *"Mirrors conntrack.build_event's wire shape (#391 HostId tagged union, #338 …)"*. That's a `// must mirror X` smell per AGENTS.md §single-source-of-truth. nflog doesn't currently consult `static_ip_labels`, so it's not a divergence for THIS PR (nflog tails firewall drops, not connection NEW), but if labels eventually flow through nflog the two `build_event` functions will need to be collapsed. Out of scope here — follow-up worth filing.
3. [openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua:92-94](https://github.com/wifihaven/wifihaven/blob/782c75631877b09060cd19e599f313a0ac2288af/openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua#L92-L94) — the prefix=0 mask path computes `blocksize = 0x100000000`, and `n - n % 0x100000000 == 0` for any `n`, so a `0.0.0.0/0` entry would match every IPv4 address. The size-cap test (`#labels._ranges <= 10`) prevents accidental bloat but doesn't reject a 0/0 entry. Probably never matters in practice (operator-curated map, code review is the gate), but worth a one-liner in `parse_cidr_v4` rejecting prefix 0 to fail loud rather than silently swallow every flow.

VERDICT: APPROVE

3-line summary:
The OpenWRT agent now emits `HostId.Label` for static-IP-range attributions (matching the wire shape PR #1713 settled on and #1725's V57 CHECK widening), subsuming #1706. `build_event` becomes the single place that renders each variant, with `static_ip_labels.lookup` returning `(label, source)` so the wire shape is assembled exactly once — clean fit with the single-source-of-truth rule. TDD red→green visible in commit history (`425a84c8` red → `782c7563` green); 737/737 Lua tests pass locally. APPROVE with one SHOULD-FIX (the label `hname` now reaches the agent's carve-out / blocklist `host_matches` paths that previously short-circuited on nil; defense-in-depth recommended but not merge-gating since the divergence is precluded server-side by `HostMatch.matchesAny`).

sameerparekh:
Re-run review against incremental delta `782c7563...d893c852`.

## Prior-findings status

| SHOULD-FIX 1 | label `hname` reaching `host_matches` carve-out / blocklist paths | **ADDRESSED.** [conntrack.lua:788](https://github.com/wifihaven/wifihaven/blob/d893c852de9d8ae00c05e613870efea6f886162f/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua#L788) introduces `match_hname = (hlabel_source == nil) and hname or nil`; all four `host_matches(hname, ...)` callsites in `handle_flow` (extraAllowed carve-out, `check_ea_carveout` helper, extraBlocked loop, category-blocklist loop) now use `match_hname`. Two new regression tests at [conntrack_spec.lua:790-825](https://github.com/wifihaven/wifihaven/blob/d893c852de9d8ae00c05e613870efea6f886162f/openwrt/test/conntrack_spec.lua#L790-L825) pin the behavior. | ✅ |
| NIT 1 | contract test pinning `M.SOURCE` ↔ `HostId.LabelSourceStaticIpRange` | NOT-ADDRESSED. Explicit follow-up acceptable per prior review (`Filing a follow-up issue is enough`). | ⚪ |
| NIT 2 | `nflog.build_event` "Mirrors" comment smell | NOT-ADDRESSED. Explicit out-of-scope per prior review. | ⚪ |
| NIT 3 | `parse_cidr_v4` should reject prefix 0 | **ADDRESSED.** [static_ip_labels.lua:73-76](https://github.com/wifihaven/wifihaven/blob/d893c852de9d8ae00c05e613870efea6f886162f/openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua#L73-L76) tightens the bound to `prefix < 1`, removes the dead `(prefix == 0) and 0 or ...` branch, and a new test at [static_ip_labels_spec.lua:47-54](https://github.com/wifihaven/wifihaven/blob/d893c852de9d8ae00c05e613870efea6f886162f/openwrt/test/static_ip_labels_spec.lua#L47-L54) pins the no-wildcard guarantee. | ✅ |

## New findings in the delta

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
(none — the two open NITs from the prior pass are follow-up file-only items; they were explicitly marked "out of scope here" / "follow-up issue is enough" in the original review and are unchanged here.)

VERDICT: APPROVE

3-line summary:
The SHOULD-FIX is cleanly addressed via a single new derived variable `match_hname` that is nil for label-typed attributions; all four agent-side `host_matches` callsites consume it, and two new feature tests pin the behavior under blocked-MAC carve-out and extraBlocked scenarios. NIT 3 (prefix=0 wildcard) is also closed, with the now-unreachable `prefix == 0` branch removed. The two carried-forward NITs (`M.SOURCE` contract test, `nflog.build_event` mirror smell) remain as explicit follow-ups; the BLOCKER gate is clear.

PR #1726, opened 2026-06-14
feat(#1718): allowlist + Grafana for native router host metrics (PR 1 of 2)

sameerparekh:
## Summary

PR 1 of 2 for #1718 (router agent publishes native OS metrics — load / bandwidth / wireless / connections). This PR is API + Grafana only; the follow-up PR wires the OpenWRT agent to emit through the existing #1205 batch transport.

**No new DB tables, no wire-shape change.** `RouterMetricsBatch` is already a generic `counters` / `gauges` / `histograms` bag and the server already attaches `router_id`. So the API side is just allowlist + dashboard, and the agent side will just push new entries through the same path.

Changes:

- `api/src/metrics/Metrics.scala` — add `iface` and `ssid` to `MetricGuard.KnownLabelKeys` (both bounded by router hardware/config, firewall-safe), register `router_host_load_m{1,5,15}`, `router_host_cpu_pct`, `router_host_mem_{total,free,buffers,cached}_kb`, `router_host_conntrack_count`, `router_host_iface_{rx,tx}_bytes_total{iface}`, `router_host_wifi_clients{iface, ssid}` in `MetricGuard.Allowed`.
- `api/test/src/feature/RouterHostMetricsAllowlistSpec.scala` — new spec pins the `(name, labels)` contract and proves the cardinality firewall rejects a per-MAC label on the wifi-clients gauge while accepting `(iface, ssid)`.
- `deploy/grafana/dashboards/router-fleet.json` — seven new panels: load avg (1m), CPU%, memory used %, per-iface RX/TX bandwidth, conntrack count, wireless clients per (router, iface, ssid).

Per-MAC wireless drill-in is deliberately out of scope here per AGENTS.md §instrument-new-functionality (cardinality firewall) — the aggregate count-per-AP is the labeled series; a per-client surface would need a separate API endpoint.

## TDD progression

- `322477a7` test(#1718): pin (name, labels) contract — red (Allowed doesn't carry the new names)
- `e053c9f4` feat(#1718): allowlist native router-host metrics — green
- `5491efd7` feat(#1718): Grafana panels

## Why this would have helped #1716

The motivating incident: prod router CPU pegged for hours on a 24-day-old agent. With the load-m1 panel an operator would have seen the climb start exactly when, without SSH; with the conntrack and memory panels it would have triaged the cause faster. The CPU panel sits right next to it.

## Follow-up — PR 2 of 2

[...]

Issue #528, opened 2026-05-17
VM e2e regression: block page no longer responding — test_03, test_06, test_extra_blocked.http all fail with 'timed out waiting for block page response'


sameerparekh:
## Symptom

Run https://github.com/sameerparekh/familydns/actions/runs/25980016036/job/76366952409 — **NEW failures** in sync-2 that were passing in sync-1:

```
FAILED test_03_blocked_domain::test_blocked_domain_request_dropped
   TimeoutError: timed out waiting for block page response for example.org

FAILED test_06_blocked_page::test_blocked_request_returns_block_page
   TimeoutError: timed out waiting for block page response from router

FAILED test_extra_blocked::test_extra_blocked_http_drops_to_block_page_and_set_populated
   TimeoutError: timed out waiting for block page response for example.com
```

In sync-1 (run 25972689545), test_03 and test_06 passed. test_extra_blocked previously failed at a different step (`timed out waiting for nft set eb_example_com to gain at least one element`); it now passes the nft-set-population step (suggesting #511 / #515-adjacent fixes worked) but fails on the block page response.

So the block page itself (or its response cycle) has regressed.

## Likely root cause

Sync-2 included `0354279 fix(blockpage): plumb per-MAC block reason through to the block page (fixes #437) (#512)`, which modified how the block page is served. Of all commits in the sync, this is the only one that directly touches the block page code path.

Other less-likely suspects:
- `08c1a62` fix(openwrt) drop unsupported `tag:` prefix from extraAllowed nftset (#496) — affects nft rendering but not block page handler.
- `952c6a9` feat(web) logs click-through — web UI only.

## What to investigate

1. Read the diff of #512 / `0354279` — especially `openwrt/files/www/familydns/handler.lua` (new file). The previous static `index.html` was replaced with a Lua handler; if the handler has a bug (e.g., returns no body, wrong content-type, slow), the test's wait for the response times out.
2. Pull the failed-run artifacts and check the router-side block-page log:
   ```
   gh run download 25980016036 --pattern 'e2e-vm-failure-*' --dir /tmp/r-blockpage
   ```
3. Inside the router VM (during a `--keep` reproduction): `curl -v http://192.168.100.1/` (or whatever the block-page URL is) and see what comes back.

## Impact

3 scenarios that were previously passing are now failing. This is a regression, not progress on the stabilization track. Likely a small handler.lua bug.

## Out of scope

[...]

PR #1728, opened 2026-06-14
fix(#1717): single-source openwrt package DEPENDS (restores v4-TCP SNI attribution)

### LINKED ISSUE #1717: ci(cd): router CD blocked on regressed v4-TCP SNI attribution test (post #1651/#1652/#1697 wave)
## Symptom

Master Router CD on main commit \`8dc4974e\` fails Gate 2 ipk + apk:

\`\`\`
FAILED scenarios_fake/test_sni_attribution.py::test_sni_only_hostname_attribution
       — TimeoutError: timed out waiting for dns-cache row 192.168.100.1 -> e2e-sni.wifihaven.net
         attributed via SNI (host never DNS-resolved => SNI is the only possible source)
============= 1 failed, 43 passed, 2 xfailed in 2286.76s =============
\`\`\`

Run: https://github.com/wifihaven/wifihaven/actions/runs/27503349348

**43 of 44 tests pass**, including the new v6 + QUIC attribution tests. The ONE failure is the original v1 path: TLS over TCP/v4 where SNI is the only attribution source (host never resolved via DNS).

## Real-world impact

This CD has been red since the recent SNI wave landed. **Every router stuck on the last successfully-published openwrt-latest** — including operator's prod router.lan, which is on 24-day-old code per #1716. Until this CD goes green, no fixes since the wave can deploy.

## Root cause hypothesis

The recent wave touched the SNI sidecar code:
- #1697 (PR #1697) — UCI toggle \`wifihaven.sni.enabled\` (init-script branching)
- #1698 (PR #1698) — IPv6 outer-header parsing in \`sni.parse_packet\` (the ethertype dispatch)
- #1702 (PR #1702) — QUIC / HTTP3 Initial-packet SNI parsing (additional capture path on UDP/443)

The most likely regression vector: **the ethertype dispatch added by #1698**. Originally \`parse_packet\` returned nil for ethertype != 0x0800. #1698 branched on ethertype to handle 0x86dd v6. The change may have inadvertently broken v4 (off-by-one in header offset, wrong field width, refactored shared code that now mishandles the v4 case).

A v4-TCP test failing while v6-TCP and QUIC pass would fit a "v4 dispatch arm broken in the new ethertype switch."

Less likely: #1697's UCI toggle changed default behavior (but other tests use the same sidecar and pass).
Even less likely: #1702 QUIC code interferes with the TCP capture (different BPF filters; shouldn't cross).

## Investigation

Boot Gate 2 locally on plex; SSH into the booted router VM during the failure:

1. \`pgrep -af wifihaven-sni-tail\` — is the sidecar running?
2. \`logread | grep -iE 'sni|clienthello' | tail -50\` — is it parsing ClientHellos at all? What result-bucket counts (success/v4/v6/quic/skipped)?
3. From the client VM, trigger the test flow: \`curl -v https://e2e-sni.wifihaven.net/\` (assuming the test orchestrator wires DNS so the flow happens). Watch the sidecar log; what does it produce?
4. Compare the parse output for a v4 ClientHello pre-#1698 vs post-#1698 — most likely a header-offset bug in the v4 arm of the dispatch.

## Fix

Whatever the diagnosis pins. Likely a small surgical fix in \`openwrt/files/usr/lib/lua/wifihaven/sni.lua\`'s \`parse_packet\` v4 arm. TDD: pin the v4 case with the failing fixture, fix, both v4 and v6 fixtures green.

## Acceptance

- \`test_sni_only_hostname_attribution\` passes.
- All 44 Gate 2 tests pass.
- Master Router CD goes green; new openwrt-latest publishes.
- Operator's prod router.lan picks up the latest agent on next auto-update cycle (resolves #1716's old-code state).

## Related

- #1716 — prod router CPU incident; possibly downstream of running 24-day-old code that this CD blockage prevents updating
- #1651 / PR #1702 — QUIC SNI parsing (wave member, unlikely cause)
- #1652 / PR #1698 — IPv6 outer-header parsing (likely cause)
- #1654 / PR #1697 — UCI toggle (wave member, unlikely cause)
- #573 / PR #1643 — v1 SNI sidecar, where the v4 path was originally written

sameerparekh:
Closes #1717. Unblocks Master Router CD; new openwrt-latest will publish after merge.

## Root cause

`openwrt/build-ipk.sh` line 28 and `openwrt/build-apk.sh` line 149 each hard-coded their own `Depends:` / `--info depends:` string for the wifihaven router-agent package, in addition to the canonical `DEPENDS:=…` in `openwrt/Makefile`. When [#1702](https://github.com/wifihaven/wifihaven/pull/1702) added `lua-openssl` to the Makefile so the SNI sidecar could parse QUIC v1 Initial packets, the two builder scripts never picked it up. Every shipped router .ipk therefore omitted `lua-openssl`, and the very first line of `wifihaven-sni-tail` — `require("openssl")` — crashed the sidecar at startup. procd respawned it 5× in ~25s and gave up, so `/tmp/wifihaven-sni.log` stayed empty fleet-wide and the Gate 2 v4-TCP attribution test timed out waiting for its dns-cache row.

The same drift had already silently hidden `kmod-nf-conntrack6` (added in [#1690](https://github.com/wifihaven/wifihaven/pull/1690) for v6 attribution); v6 tests passed only because something in the stock 23.05 image happened to load the v6 conntrack hooks anyway.

The runtime symptom traced back to a **single-source-of-truth violation** per `AGENTS.md`: the DEPENDS list was a decision duplicated in three files. Fix is to **collapse**, not to add a third stale copy.

## Fix

A tiny helper `openwrt/depends-list.sh` parses the canonical `DEPENDS:=` line in `openwrt/Makefile` once and emits it in the format each packaging tool expects:
- `./depends-list.sh ipk` → `"pkg1, pkg2, pkg3"` (Debian control `Depends:`)
- `./depends-list.sh apk` → `"pkg1 pkg2 pkg3"` (`apk mkpkg --info depends:`)

`build-ipk.sh` and `build-apk.sh` both invoke the helper instead of hard-coding. Adding a runtime dependency now requires editing only the Makefile.

## Tests

New shell test `openwrt/test/build_depends_sync_spec.test.sh` (auto-discovered by the CI `*.test.sh` job in `ci.yml`) RED-pinned the drift in commit 1, GREEN in commit 2. It guards three things going forward:
1. Each builder script calls `depends-list.sh ipk`/`apk`.
2. Each builder's control line references `${DEPENDS_LIST}` rather than a literal package list (re-hardcoding is rejected).
3. The helper's output still contains the regression-pinned packages (`lua-openssl`, `kmod-nf-conntrack6`, `kmod-nf-log`, `kmod-nf-log6`, `tcpdump-mini`) so a future Makefile edit that loses one of them is caught.

## Verification (local Gate 2 on plex)

**Before** (commit `8dc4974e`, current main):
- `.ipk` Depends has no `lua-openssl` / `kmod-nf-conntrack6`
- `logread` on booted router: `wifihaven-sni-tail: module 'openssl' not found` (×5 then procd gives up)
- `test_sni_only_hostname_attribution`: `FAILED` (TimeoutError waiting for dns-cache row)

**After** (this PR):
- `.ipk` Depends: `lua, libuci-lua, luci-lib-jsonc, conntrack, curl, uhttpd-mod-lua, kmod-nf-log, kmod-nf-log6, kmod-nf-conntrack6, libustream-mbedtls, openssl-util, tcpdump-mini, lua-openssl`
- `test_sni_only_hostname_attribution`: `SKIPPED` (the expected outcome — primary dns-cache assertion lands, the secondary connection_event assertion is intentionally skipped per [#1360](https://github.com/wifihaven/wifihaven/issues/1360))

## Downstream

- Master Router CD goes green → openwrt-latest publishes.
- On the next auto-update cycle, prod router.lan picks up current code; [#1716](https://github.com/wifihaven/wifihaven/issues/1716) may resolve as a side effect.

### DIFF
--- a/openwrt/build-apk.sh
+++ b/openwrt/build-apk.sh
@@ -25,6 +25,10 @@ PKG_VERSION="${PKG_VERSION:-$(grep '^PKG_VERSION:=' "$SCRIPT_DIR/Makefile" | cut
 PKG_RELEASE="${PKG_RELEASE:-$(grep '^PKG_RELEASE:=' "$SCRIPT_DIR/Makefile" | cut -d= -f2)}"
 OUT_APK="$SCRIPT_DIR/wifihaven_${PKG_VERSION}-${PKG_RELEASE}_all.apk"
 
+# Derive the apk depends list from the canonical openwrt/Makefile DEPENDS line
+# via the shared helper (single-source-of-truth, #1717). See build-ipk.sh.
+DEPENDS_LIST=$("$SCRIPT_DIR/depends-list.sh" apk)
+
 APK_TOOLS_REPO="${APK_TOOLS_REPO:-https://gitlab.alpinelinux.org/alpine/apk-tools.git}"
 APK_TOOLS_REF="${APK_TOOLS_REF:-v3.0.6}"
 APK_TOOLS_PREFIX="${APK_TOOLS_PREFIX:-$HOME/.cache/wifihaven-apk-tools}"
@@ -146,7 +150,7 @@ rm -f "$OUT_APK"
     --info "license:MIT" \
     --info "url:https://github.com/wifihaven/wifihaven" \
     --info "maintainer:WifiHaven <noreply@example.com>" \
-    --info "depends:lua libuci-lua luci-lib-jsonc conntrack curl uhttpd-mod-lua kmod-nf-log kmod-nf-log6 libustream-mbedtls openssl-util tcpdump-mini" \
+    --info "depends:${DEPENDS_LIST}" \
     --script "post-install:$WORK/post-install" \
     --script "trigger:$WORK/trigger" \
     --trigger "/etc/crontabs" \
--- a/openwrt/build-ipk.sh
+++ b/openwrt/build-ipk.sh
@@ -13,6 +13,13 @@ PKG_VERSION="${PKG_VERSION:-$(grep '^PKG_VERSION:=' "$SCRIPT_DIR/Makefile" | cut
 PKG_RELEASE="${PKG_RELEASE:-$(grep '^PKG_RELEASE:=' "$SCRIPT_DIR/Makefile" | cut -d= -f2)}"
 OUT_IPK="$SCRIPT_DIR/wifihaven_${PKG_VERSION}-${PKG_RELEASE}_all.ipk"
 
+# Derive the package Depends list from the canonical openwrt/Makefile DEPENDS
+# line via the shared helper (single-source-of-truth, #1717). A hard-coded
+# copy here drifted from the Makefile when #1702 added lua-openssl, shipping
+# an .ipk that crash-looped sni-tail's require('openssl') in production.
+# build_depends_sync_spec.test.sh pins the equivalence in CI.
+DEPENDS_LIST=$("$SCRIPT_DIR/depends-list.sh" ipk)
+
 WORK="$(mktemp -d)"
 trap 'rm -rf "$WORK"' EXIT
 
@@ -25,7 +32,7 @@ mkdir "$WORK/ctrl"
 cat > "$WORK/ctrl/control" <<EOF
 Package: wifihaven
 Version: ${PKG_VERSION}-${PKG_RELEASE}
-Depends: lua, libuci-lua, luci-lib-jsonc, conntrack, curl, uhttpd-mod-lua, kmod-nf-log, kmod-nf-log6, libustream-mbedtls, openssl-util, tcpdump-mini
+Depends: ${DEPENDS_LIST}
 Architecture: all
 Maintainer: WifiHaven <noreply@example.com>
 Description: Agent daemon for WifiHaven. Enforces per-device DNS filtering
--- a/openwrt/depends-list.sh
+++ b/openwrt/depends-list.sh
@@ -1,0 +1,45 @@
+#!/bin/sh
+# Single source of truth for the wifihaven package's runtime DEPENDS list
+# (#1717, AGENTS.md §single-source-of-truth). Parses the canonical +-prefixed
+# OpenWRT Makefile DEPENDS line (openwrt/Makefile, sibling of this script)
+# and emits the list in the format the named packaging tool expects:
+#
+#   ./depends-list.sh ipk    →  "pkg1, pkg2, pkg3"   (Debian control "Depends:")
+#   ./depends-list.sh apk    →  "pkg1 pkg2 pkg3"     (apk mkpkg --info depends:)
+#   ./depends-list.sh lines  →  one package per line  (test diffs, ad-hoc)
+#
+# Invoked from openwrt/build-ipk.sh and openwrt/build-apk.sh so adding a
+# runtime dependency only requires editing openwrt/Makefile;
+# openwrt/test/build_depends_sync_spec.test.sh pins the equivalence in CI.
+#
+# The Makefile path is the file named "Makefile" in the same directory as
+# this script. Set WH_DEPENDS_MAKEFILE=/path/to/Makefile to override (tests).
+
+set -eu
+
+SCRIPT_DIR="$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)"
+MAKEFILE="${WH_DEPENDS_MAKEFILE:-$SCRIPT_DIR/Makefile}"
+
+if [ ! -f "$MAKEFILE" ]; then
+    echo "depends-list.sh: cannot find Makefile at $MAKEFILE" >&2
+    exit 1
+fi
+
+emit_lines() {
+    grep -E '^[[:space:]]*DEPENDS:=' "$MAKEFILE" \
+        | head -n1 \
+        | sed -E 's/^[[:space:]]*DEPENDS:=//' \
+        | tr ' \t' '\n' \
+        | sed -E 's/^\+//' \
+        | grep -v '^$'
+}
+
+case "${1:-lines}" in
+    lines) emit_lines ;;
+    ipk)   emit_lines | paste -sd ',' - | sed 's/,/, /g' ;;
+    apk)   emit_lines | paste -sd ' ' - ;;
+    *)
+        echo "depends-list.sh: unknown format '$1' (want: lines|ipk|apk)" >&2
+        exit 2
+        ;;
+esac
--- a/openwrt/depends-list.sh
+++ b/openwrt/depends-list.sh
@@ -14,6 +14,10 @@
 #
 # The Makefile path is the file named "Makefile" in the same directory as
 # this script. Set WH_DEPENDS_MAKEFILE=/path/to/Makefile to override (tests).
+#
+# Execute, do not source: the script resolves its directory from $0 so sourcing
+# from another shell would silently misbehave. Callers invoke it as
+# `"$SCRIPT_DIR/depends-list.sh" ipk`.
 
 set -eu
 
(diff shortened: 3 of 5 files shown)


sameerparekh:
Independent review pass per docs/pr-review-checklist.md. Read-only; diff reviewed against the PR base `claude/fix-1718-api...HEAD` (three-dot, stacked PR).

BLOCKERS
(none)

SHOULD-FIX
1. `openwrt/files/usr/sbin/wifihaven-agent` — `iw_station_dump` shell-outs to `iw dev <iface> station dump` with no timeout. If a wifi driver wedges or `iw` hangs (rare but observed on mt76 in prod-like conditions), the metrics push tick blocks for the duration, which on a 60s cadence is acceptable but on the smallest router compounds with conntrack scrape latency. Wrap as `timeout 5 iw dev <iface> station dump 2>/dev/null` — same defensive pattern the existing usage scrape uses for nft counters. Single-line fix.

NITS
1. `openwrt/files/usr/lib/lua/wifihaven/host_metrics.lua:142` — `count_stations` looks for `\nStation ` and special-cases a leading `Station `. A `\r\n` separator (not from `iw` today, but a future driver variant) would mis-count. Tighter: `for _ in dump:gmatch("[\r\n]Station ") do`. Strictly defensive.
2. `openwrt/files/usr/sbin/wifihaven-agent` — `parse_iface_csv` / `parse_wifi_ifaces` / `safe_iface` live in the agent driver. They're parsers, not orchestration; moving them into `host_metrics.lua` (or a tiny helper) would let the busted spec cover the parsing too instead of relying on the agent_spec.sh smoke test. Not blocking — agent_spec.sh + the existing 8 tests get the coverage in practice.
3. `host_metrics.lua:33` — header lists `iw dev info` in the design notes but the implementation only uses `station dump`. Either drop the `info` mention or note that channel/tx-power are deferred. Small doc drift.
4. PR body cost claim ("well under 50 ms") is sound for the file reads but un-measured for the `iw` shell-out. Worth measuring once on `openwrt.lan` and noting in the PR before merge — the SHOULD-FIX above would also bound the worst case.

### Dimension notes

- **1. Duplicated logic / SSOT** — `host_metrics` calls `metrics.set_gauge` / `metrics.fold_external`, both single-source primitives. No new decision logic. The names match PR A's allowlist exactly (verified by `MetricCardinalityGuardSpec` once #1726 merges into main). No "must mirror" comments.
- **2. Testing integrity** — TDD progression visible: `543edd90` (red, module absent) → `6c85fa3a` (green, 8/8). New spec covers parsers, the CPU%-needs-two-snapshots edge case, the iface counter delta+reset path (fold_external), the empty-`iw`-dump zero case, and missing-`/proc`-files no-op behavior. No existing tests weakened. The "wifi-clients gauge never carries `mac`" invariant is asserted API-side in PR A; agent-side this PR doesn't need to re-pin it because the emit call sites can't generate a `mac` label (the gauge name only takes `iface, ssid`).
- **3. Architectural invariants** — Pure observability collection. Router stays a dumb applier (no policy logic). Module uses injected I/O (`read_file`, `iw_dump`), matches the `policy.lua` / `blocklists.lua` DI pattern. No `/tmp` writes added.
- **4. Backwards compatibility / wire contract** — Purely additive: new gauge/counter names ride the existing generic `counters` / `gauges` arrays in `RouterMetricsBatch`. Older API server silently drops to `metrics_rejected_total{reason=unknown_name}` — acceptable rollback noise, PR body notes it. No request/response shape change.
- **5. Migrations** — N/A.
- **6. Metrics + dashboard** — All emitted names are in PR A's `MetricGuard.Allowed` (load_m1/_m5/_m15, cpu_pct, mem_{total,free,buffers,cached}_kb, conntrack_count, iface_{rx,tx}_bytes_total{iface}, wifi_clients{iface,ssid}). Bounded labels only. Dashboard panels for these names land in PR A — the stacked-PR shape satisfies the §metrics-need-a-dashboard rule because the merge train is one push.
- **7. Scope / hygiene** — Two focused commits. No drive-by edits. PR title/body / commit messages have no AI mentions; `Co-Authored-By` trailer only.
- **8. Security** — `safe_iface()` filters UCI iface names through `^[%w_%-:%.]+$` before reaching `popen`. The iface name only flows into `iw dev <name>` and `/sys/class/net/<name>/statistics/*` paths — both terminate at the kernel surface; no shell metachars survive the filter. UCI is operator-owned anyway, so this is depth-defense. No secrets, no SQL, no auth surface change.

VERDICT: APPROVE

Clean, narrowly-scoped agent-side collector that completes #1718. TDD discipline, cardinality discipline, and DI pattern all consistent with the surrounding modules. SHOULD-FIX-1 (timeout on the `iw` shell-out) is the only non-trivial item — single-line fix worth doing before merge.

sameerparekh:
## Independent PR review — re-run (incremental delta over b2ee9cd7…9bd09fcf)

### Prior findings status

| # | Finding | Status |
|---|---|---|
| SHOULD-FIX 1 | RED→GREEN commit history is technically misleading (the RED test was replaced rather than satisfied) | **WON'T-FIX (acknowledged).** Substantive concern is test coverage — covered by the GREEN test. Rewriting commit history would force-push the PR branch, which violates the author's standing no-force-push rule. Leaving as documented. Not merge-gating. |
| NIT 1 | `depends-list.sh` lacks "execute, do not source" guidance | **ADDRESSED** ([openwrt/depends-list.sh:17-19](openwrt/depends-list.sh#L17-L19)) — docstring now states the constraint explicitly. |
| NIT 2 | `head -n1` silently picks first `DEPENDS:=` if Makefile grows a second | **ADDRESSED** ([openwrt/depends-list.sh:33-39](openwrt/depends-list.sh#L33-L39)) — script now bails loud when match count ≠ 1, with a clear error message that calls out the luci-subpackage scenario. |
| NIT 3 | Regression-pin set should include `kmod-nf-log` / `kmod-nf-log6` | **ADDRESSED** ([openwrt/test/build_depends_sync_spec.test.sh:35](openwrt/test/build_depends_sync_spec.test.sh#L35)) — extended to all five. |
| NIT 4 | AI trailer in PR body | **ADDRESSED** — PR body updated. |

### New findings (incremental delta)

Delta is two small hardening edits on `depends-list.sh` and one extension to the test's regression-pin list. Reviewed against the 8 dimensions:

**BLOCKERS:** (none)

**SHOULD-FIX:** (none)

**NITs:** (none — the new `grep -cE ... || true` correctly handles the no-match case where `grep -c` would exit 1 with stdout "0", and `[ "$n" -ne 1 ]` then catches it.)

### Dimension coverage (delta)

| § | Notes |
|---|---|
| 1 | SSoT — no change. |
| 2 | Test extended (more regression pins); no weakened or removed assertions. |
| 3 | N/A. |
| 4 | N/A. |
| 5 | N/A. |
| 6 | N/A. |
| 7 | Focused. |
| 8 | N/A. |

### VERDICT: APPROVE

All BLOCKERS clear (none ever opened). All addressable NITs addressed. The single open SHOULD-FIX is non-load-bearing commit-history hygiene that can't be cleanly resolved without violating the author's force-push rule.

sameerparekh:
## Independent PR review — re-run (incremental delta over 9bd09fcf…69935f9b)

### Prior findings status

| # | Finding | Status |
|---|---|---|
| (prior) SHOULD-FIX 1 | RED→GREEN commit history misleading | **WON'T-FIX (acknowledged)** — unchanged. No force-push. |
| (prior) NIT 1–4 | All ADDRESSED in 9bd09fcf | **STILL-ADDRESSED.** |

### New findings (incremental delta — install_spec.sh #528 redirect)

**BLOCKERS:** (none)

**SHOULD-FIX:** (none)

**NITs:** (none)

### Dimension coverage (delta)

| § | Notes |
|---|---|
| 2 | **Test change checked carefully (checklist §2 calls this out as high-risk).** The edit replaces a pair of `grep` assertions on `build-ipk.sh` / `build-apk.sh` for the literal string `uhttpd-mod-lua` with one `grep` on `openwrt/Makefile` for `+uhttpd-mod-lua` in the canonical `DEPENDS:=` line. This is the legitimate "the OLD assertion was actually wrong" case the checklist permits: the literal string is no longer present in the builder scripts (they expand `${DEPENDS_LIST}` from the helper — that's the whole #1717 fix), so the old grep cannot fire. The new assertion enforces the same invariant at the canonical source, and `build_depends_sync_spec.test.sh` (added earlier in this PR) pins that both builders derive from that source. Invariant preserved; arguably strengthened. The commit message [69935f9b](https://github.com/wifihaven/wifihaven/commit/69935f9b) lays out the rationale. |
| 1, 3–8 | No change. |

### VERDICT: APPROVE

CI-driven follow-up addresses a previously-passing test that the SSoT collapse made stale; the edit redirects the same invariant to the new canonical source rather than weakening it. No open BLOCKER, no open addressable finding.

Issue #1205, opened 2026-05-31
metrics: router → API push transport (POST /api/router/metrics)


sameerparekh:
Implementation sub-issue of #471. Design: `docs/design/metrics-observability.md (PR #1203)` §3.

## Scope
- Add `POST /api/router/metrics` under `/api/router/*`, reusing `RouterAuth` (the existing per-router bearer token — **no new credential**). Design §3.4.
- Wire schema exactly as design §3.2: `routerId`, `agentVersion`, `agentStartedAt` (counter-reset sentinel), `sampledAt`, and `counters`/`gauges`/`histograms` arrays. Counters are **cumulative** running totals; server re-bases on `agentStartedAt` change (agent restart) rather than recording a negative delta.
- Cross-check body `routerId` against the token's router (same pattern as `/api/router/usage`).
- Implement a **carrier-agnostic** `RouterMetricsService` that takes a *parsed batch* (not a `Request`) and folds it into the metric registry idempotently. The REST route is a thin adapter over it.
- Enforce the server-side metric-name/label allowlist here too (drop+count unknown via `metrics_rejected_total`).
- Add request types to `shared/src/Models.scala`; register the route in `Main.scala`; TS client call if/when a UI surface needs it (probably none).

## Forward-compatibility (#1023)
The handler must be carrier-agnostic precisely so that when #1023 (websocket transport) lands, `{"op":"metrics","payload":<§3.2 body>}` dispatches into the *same* `RouterMetricsService` with no payload change. **This REST endpoint lands FIRST, before #1023.** Coordinate adding `op:"metrics"` to #1023's frame enum scope.

## Acceptance
- Valid batch with correct token → 200, series appear under `router_id` (+ `installation_id` when available) on `/metrics`.
- Wrong/missing token → 401; `routerId` mismatch → 401/403.
- Replayed/duplicate batch does not double-count (cumulative semantics).
- Agent-restart (changed `agentStartedAt`) re-bases cleanly, no negative deltas.

## TDD
Feature tests first (embedded Postgres): auth, routerId cross-check, cumulative fold, restart re-basing, allowlist rejection. Red→green in commit history.

## Dependencies
Depends on #1204 (registry + /metrics). Forward-compat: #1023.

sameerparekh:
[...]

PR #1731, opened 2026-06-14
feat(#1718): agent emits native OpenWRT host metrics (PR 2 of 2)

sameerparekh:
## Summary

PR 2 of 2 for #1718. Adds the OpenWRT-side collector that reads `/proc/*` and `iw dev <iface> station dump`, folds the values into the existing in-process metrics registry under the names allowlisted in PR A (#1726), and rides the existing #1205 batch transport — same 60s cadence, same retain-on-failure semantics. **Stacked on #1726**; merge that first.

Changes:

- `openwrt/files/usr/lib/lua/wifihaven/host_metrics.lua` — new module. Pure parsers for `/proc/loadavg`, `/proc/stat` (delta CPU%), `/proc/meminfo`, `/proc/sys/net/netfilter/nf_conntrack_count`, and `/sys/class/net/<iface>/statistics/{rx,tx}_bytes`. Wireless client count from `iw dev <iface> station dump` — line count of `Station <mac>` records, per-MAC never emitted (cardinality firewall). All I/O injected (`read_file`, `iw_dump`) so the spec drives parsers with fixture strings, no `/proc` or `iw` needed in CI.
- `openwrt/files/usr/sbin/wifihaven-agent` — wires `sample_host_metrics()` into the metrics-push tick next to `fold_sni_metrics()`. Adds two UCI knobs:
  - `host_metrics_ifaces` — CSV iface names for bandwidth (default `br-lan,wan`)
  - `host_metrics_wifi` — CSV `iface:ssid` pairs for the wifi-clients gauge (default empty)
  Both filtered through `safe_iface()` (character-class allowlist) before reaching the `iw` shell-out.
- `openwrt/test/host_metrics_spec.lua` + `run_tests.sh` — 8 busted tests pinning the parsers and the cardinality contract.

## TDD progression

- `543edd90` test(#1718): pin host_metrics behavior — red (module doesn't exist)
- `6c85fa3a` feat(#1718): agent emits native OpenWRT host metrics — green (8/8 pass)

## Why this would have helped #1716

Prod router CPU pegged for hours on a 24-day-old agent. With this PR an operator opens the router-fleet Grafana dashboard, sees a load-m1 climb on one router starting at hour T, cross-references to a conntrack-count spike at T+5 min, and skips straight to "look at flows" instead of SSH-ing to confirm the symptom.

## Cost check

Per-tick collection on the smallest supported device (mt7621, 256 MB) is bounded by 4 small file reads + 1 `iw` shell-out per wifi iface. Worst case well under 50 ms; runs on the same 60 s cadence as the existing metrics push, so the marginal cost is negligible.

## Backwards compatibility

- API side already accepts the new names (PR A merged). Older API images would silently drop the new entries into `metrics_rejected_total`, which is acceptable rollback behavior.
- No wire-shape change: `RouterMetricsBatch` is the existing generic counters/gauges/histograms bag.
- New UCI knobs default to safe values (empty `host_metrics_wifi` → no wifi sampling).

## Test plan

- [ ] `busted test/host_metrics_spec.lua` — passes locally (8/8)
- [ ] CI `lua-suite` passes
- [ ] CI `MetricCardinalityGuardSpec` passes (verifies every emitted name is allowlisted in PR A)
- [ ] Live walk on `openwrt.lan`: after deploy, confirm the new gauges appear in the router-fleet dashboard within one push tick

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/host_metrics.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/host_metrics.lua
@@ -4,7 +4,9 @@
 -- /proc/meminfo, /proc/sys/net/netfilter/nf_conntrack_count,
 -- /sys/class/net/<iface>/statistics/{rx,tx}_bytes — and the `iw dev <iface>
 -- station dump` output, and folds them into the existing metrics registry
--- under the names the API allowlisted in #1718 PR 1:
+-- under the names the API allowlisted in #1718 PR 1. (Channel / TX-power
+-- from `iw dev <iface> info` are out of scope for v1; add when an operator
+-- needs to alert on radio config drift.)
 --
 --   gauges
 --     router_host_load_m1 / _m5 / _m15
@@ -132,7 +134,9 @@ end
 local function count_stations(dump)
   if not dump or dump == "" then return 0 end
   local n = 0
-  for _ in dump:gmatch("\nStation ") do n = n + 1 end
+  -- Match `[\r\n]Station ` so a future `iw` variant emitting CRLF doesn't
+  -- silently zero the count.
+  for _ in dump:gmatch("[\r\n]Station ") do n = n + 1 end
   -- Count a leading "Station " too (no preceding newline).
   if dump:sub(1, 8) == "Station " then n = n + 1 end
   return n
--- a/openwrt/files/usr/lib/lua/wifihaven/host_metrics.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/host_metrics.lua
@@ -1,0 +1,186 @@
+-- host_metrics.lua — collect native OpenWRT OS-level metrics (#1718).
+--
+-- Reads the standard kernel surfaces — /proc/loadavg, /proc/stat,
+-- /proc/meminfo, /proc/sys/net/netfilter/nf_conntrack_count,
+-- /sys/class/net/<iface>/statistics/{rx,tx}_bytes — and the `iw dev <iface>
+-- station dump` output, and folds them into the existing metrics registry
+-- under the names the API allowlisted in #1718 PR 1:
+--
+--   gauges
+--     router_host_load_m1 / _m5 / _m15
+--     router_host_cpu_pct                 (delta-derived; needs two samples)
+--     router_host_mem_total_kb / _free_kb / _buffers_kb / _cached_kb
+--     router_host_conntrack_count
+--     router_host_wifi_clients{iface, ssid}
+--   counters (cumulative, fold-external; server re-bases on reset)
+--     router_host_iface_rx_bytes_total{iface}
+--     router_host_iface_tx_bytes_total{iface}
+--
+-- All I/O is injected via the `deps` table so the spec drives this with
+-- fixture strings and the agent driver wires in the real read / exec
+-- functions. Missing sources are silently skipped (a router without
+-- nf_conntrack loaded just omits that gauge) — host_metrics never panics on
+-- a missing /proc node.
+--
+-- Cardinality-firewall reminder (#1718, AGENTS.md §instrument-new-functionality):
+-- `iface` and `ssid` are bounded by router hardware/config (handful per
+-- device). The per-MAC wireless client list is explicitly NOT a label — the
+-- gauge ships an aggregate count per (iface, ssid) only.
+
+local metrics = require("wifihaven.metrics")
+
+local M = {}
+
+-- ── /proc/loadavg ──────────────────────────────────────────────────────────
+local function emit_loadavg(reg, body)
+  if not body then return end
+  local m1, m5, m15 = body:match("([%d%.]+)%s+([%d%.]+)%s+([%d%.]+)")
+  if not m1 then return end
+  metrics.set_gauge(reg, "router_host_load_m1", {}, tonumber(m1))
+  metrics.set_gauge(reg, "router_host_load_m5", {}, tonumber(m5))
+  metrics.set_gauge(reg, "router_host_load_m15", {}, tonumber(m15))
+end
+
+-- ── /proc/stat → CPU% (delta between two snapshots) ────────────────────────
+-- Parses the aggregate `cpu` line (sum of all cores). Returns nil if the
+-- file is missing or malformed.
+--
+-- Lua's standard pattern matcher doesn't accept '+' or alternations, so we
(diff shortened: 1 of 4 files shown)


sameerparekh:
Independent review pass per docs/pr-review-checklist.md. Read-only; diff reviewed against the PR base `claude/fix-1718-api...HEAD` (three-dot, stacked PR).

BLOCKERS
(none)

SHOULD-FIX
1. `openwrt/files/usr/sbin/wifihaven-agent` — `iw_station_dump` shell-outs to `iw dev <iface> station dump` with no timeout. If a wifi driver wedges or `iw` hangs (rare but observed on mt76 in prod-like conditions), the metrics push tick blocks for the duration, which on a 60s cadence is acceptable but on the smallest router compounds with conntrack scrape latency. Wrap as `timeout 5 iw dev <iface> station dump 2>/dev/null` — same defensive pattern the existing usage scrape uses for nft counters. Single-line fix.

NITS
1. `openwrt/files/usr/lib/lua/wifihaven/host_metrics.lua:142` — `count_stations` looks for `\nStation ` and special-cases a leading `Station `. A `\r\n` separator (not from `iw` today, but a future driver variant) would mis-count. Tighter: `for _ in dump:gmatch("[\r\n]Station ") do`. Strictly defensive.
2. `openwrt/files/usr/sbin/wifihaven-agent` — `parse_iface_csv` / `parse_wifi_ifaces` / `safe_iface` live in the agent driver. They're parsers, not orchestration; moving them into `host_metrics.lua` (or a tiny helper) would let the busted spec cover the parsing too instead of relying on the agent_spec.sh smoke test. Not blocking — agent_spec.sh + the existing 8 tests get the coverage in practice.
3. `host_metrics.lua:33` — header lists `iw dev info` in the design notes but the implementation only uses `station dump`. Either drop the `info` mention or note that channel/tx-power are deferred. Small doc drift.
4. PR body cost claim ("well under 50 ms") is sound for the file reads but un-measured for the `iw` shell-out. Worth measuring once on `openwrt.lan` and noting in the PR before merge — the SHOULD-FIX above would also bound the worst case.

### Dimension notes

- **1. Duplicated logic / SSOT** — `host_metrics` calls `metrics.set_gauge` / `metrics.fold_external`, both single-source primitives. No new decision logic. The names match PR A's allowlist exactly (verified by `MetricCardinalityGuardSpec` once #1726 merges into main). No "must mirror" comments.
- **2. Testing integrity** — TDD progression visible: `543edd90` (red, module absent) → `6c85fa3a` (green, 8/8). New spec covers parsers, the CPU%-needs-two-snapshots edge case, the iface counter delta+reset path (fold_external), the empty-`iw`-dump zero case, and missing-`/proc`-files no-op behavior. No existing tests weakened. The "wifi-clients gauge never carries `mac`" invariant is asserted API-side in PR A; agent-side this PR doesn't need to re-pin it because the emit call sites can't generate a `mac` label (the gauge name only takes `iface, ssid`).
- **3. Architectural invariants** — Pure observability collection. Router stays a dumb applier (no policy logic). Module uses injected I/O (`read_file`, `iw_dump`), matches the `policy.lua` / `blocklists.lua` DI pattern. No `/tmp` writes added.
- **4. Backwards compatibility / wire contract** — Purely additive: new gauge/counter names ride the existing generic `counters` / `gauges` arrays in `RouterMetricsBatch`. Older API server silently drops to `metrics_rejected_total{reason=unknown_name}` — acceptable rollback noise, PR body notes it. No request/response shape change.
- **5. Migrations** — N/A.
- **6. Metrics + dashboard** — All emitted names are in PR A's `MetricGuard.Allowed` (load_m1/_m5/_m15, cpu_pct, mem_{total,free,buffers,cached}_kb, conntrack_count, iface_{rx,tx}_bytes_total{iface}, wifi_clients{iface,ssid}). Bounded labels only. Dashboard panels for these names land in PR A — the stacked-PR shape satisfies the §metrics-need-a-dashboard rule because the merge train is one push.
- **7. Scope / hygiene** — Two focused commits. No drive-by edits. PR title/body / commit messages have no AI mentions; `Co-Authored-By` trailer only.
- **8. Security** — `safe_iface()` filters UCI iface names through `^[%w_%-:%.]+$` before reaching `popen`. The iface name only flows into `iw dev <name>` and `/sys/class/net/<name>/statistics/*` paths — both terminate at the kernel surface; no shell metachars survive the filter. UCI is operator-owned anyway, so this is depth-defense. No secrets, no SQL, no auth surface change.

VERDICT: APPROVE

Clean, narrowly-scoped agent-side collector that completes #1718. TDD discipline, cardinality discipline, and DI pattern all consistent with the surrounding modules. SHOULD-FIX-1 (timeout on the `iw` shell-out) is the only non-trivial item — single-line fix worth doing before merge.

sameerparekh:
Re-run of the independent review pass.

### Prior findings status (vs. previous reviewed-sha `6c85fa3a`)

|---|---|---|---|
| 1 | SHOULD-FIX | `iw` shell-out has no timeout — wedged wifi driver could stall metrics tick | ADDRESSED — `b68001ea` wraps the popen in `timeout 5` |
| 2 | NIT | `count_stations` regex misses CRLF separators | ADDRESSED — `b68001ea` matches `[\r\n]Station ` |
| 3 | NIT | Move parser helpers (`safe_iface`, `parse_iface_csv`, `parse_wifi_ifaces`) into `host_metrics.lua` for direct spec coverage | NOT-ADDRESSED — left as agent-driver state; agent_spec.sh covers it indirectly. Acceptable scope choice, can be revisited if the parsers grow. |
| 4 | NIT | Header mentions `iw dev info` but only `station dump` is implemented | ADDRESSED — `b68001ea` notes channel/TX-power are v2 scope |
| 5 | NIT | PR body cost claim ("under 50 ms") is un-measured for the `iw` shell-out | NOT-ADDRESSED — needs operator on-router measurement; SHOULD-FIX 1's `timeout 5` now bounds the worst case, which is the safety the cost note was really protecting against. |

### New findings on the incremental delta (`6c85fa3a...b68001ea`)

BLOCKERS: (none)
SHOULD-FIX: (none)
NITS: (none)

The delta is three small targeted edits — `timeout 5` prefix, one regex tighten, one doc clarification. No new behavior introduced.

VERDICT: APPROVE

All BLOCKERS / SHOULD-FIX cleared. The two non-addressed NITs are explicit scope choices documented in this re-run status. No new findings on the delta.

PR #1732, opened 2026-06-14
fix(#623): /api/auth/change-password returns JSON body

### LINKED ISSUE #623: api: /api/auth/change-password returns empty body, should return JSON

sameerparekh:
## Summary

`POST /api/auth/change-password` previously returned `Response.ok` with an empty body, breaking any client that assumes "200 ⇒ parseable JSON" (every other endpoint on the API returns JSON). The fake-router bootstrap had to hand-roll a urllib call to avoid blowing up in `json.loads` (called out in #623).

This change returns `ChangePasswordResponse(mustChangePassword = false)`, lets `docker/fake-router.py::_change_password` go through the normal `_post` helper, and adds a feature test pinning the new shape.

## Backwards compat

This is technically a behavior change on the wire — empty body → `{"mustChangePassword":false}`. Surveyed callers:

- **SPA** (`web/src/api/client.ts`): types it as `req<void>`, and the request helper already short-circuits `204` / `content-length: 0` to `undefined`; a JSON body will parse and be discarded. No change needed.
- **`docker/fake-router.py`**: had a workaround for the empty body; switched to `_post` in this PR.
- **`deploy/install.sh`**: only inspects the HTTP status code (per #623); unaffected.

No client asserts on empty body.

## Test plan

- [x] `mill api.test.testOnly wifihaven.api.feature.AuthApiSpec` — new test asserts 200 + JSON `{mustChangePassword:false}`, suite green.
- [x] Red-green progression visible across two commits (failing test → fix).

Closes #623.

### DIFF
--- a/api/src/routes/Routes.scala
+++ b/api/src/routes/Routes.scala
@@ -70,7 +70,7 @@ object AuthRoutes {
                 case _                            =>
                   ApiError.Wrapped(ErrorMapper.dbUnavailable("AuthError"))
               }
-          } yield Response.ok
+          } yield Response.json(ChangePasswordResponse(mustChangePassword = false).toJson)
           handle.mapError(ErrorMapper.errorToResponse)
         },
       Method.GET / "api" / "me"                              ->
--- a/api/test/src/feature/AuthApiSpec.scala
+++ b/api/test/src/feature/AuthApiSpec.scala
@@ -118,7 +118,7 @@ object AuthApiSpec extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostgres & Cl
         cpResp   <- ZIO.fromEither(respBody.fromJson[ChangePasswordResponse])
       } yield assertTrue(resp.status == Status.Ok) &&
         assertTrue(respBody.nonEmpty) &&
-        assertTrue(cpResp.mustChangePassword == false)
+        assertTrue(!cpResp.mustChangePassword)
     },
     test("POST /api/auth/login via HTTP handler") {
       for {
--- a/api/test/src/feature/AuthApiSpec.scala
+++ b/api/test/src/feature/AuthApiSpec.scala
@@ -100,6 +100,26 @@ object AuthApiSpec extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostgres & Cl
       } yield assertTrue(bad.isFailure) &&
         assertTrue(good.isSuccess)
     },
+    test("POST /api/auth/change-password returns JSON body with mustChangePassword=false (#623)") {
+      for {
+        _               <- cleanDb
+        userRepo        <- ZIO.service[UserRepo]
+        auth            <- makeAuth
+        userProfileRepo <- ZIO.service[UserProfileRepo]
+        routes = AuthRoutes.routes(auth, userRepo, userProfileRepo)
+        token <- auth.login("admin", "changeme").map(_.token.value)
+        cpBody = ChangePasswordRequest("changeme", "newpassword123").toJson
+        cpReq  = Request
+          .post(URL.decode("/api/auth/change-password").toOption.get, Body.fromString(cpBody))
+          .addHeader(Header.Authorization.Bearer(token))
+          .addHeader(Header.ContentType(MediaType.application.json))
+        resp     <- routes.runZIO(cpReq)
+        respBody <- resp.body.asString
+        cpResp   <- ZIO.fromEither(respBody.fromJson[ChangePasswordResponse])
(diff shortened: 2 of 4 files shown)


sameerparekh:
## Independent PR review

Reviewed commit: `7f0d703b`.

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `api/test/src/feature/AuthApiSpec.scala:119` — `cpResp.mustChangePassword == false` reads more idiomatically as `!cpResp.mustChangePassword`. Non-blocking style nit.

### Dimension notes
- **1. Duplicated logic / SSoT.** `mustChangePassword = false` is hardcoded in the route handler, but this isn't re-derivation: this is the one endpoint that *clears* the flag (via `auth.changePassword` → `clearMustChangePassword`), so the value is definitional, not a snapshot of state computed elsewhere. No SSoT violation.
- **2. Testing integrity.** TDD red→green progression visible across two commits (`0acff6a9` test → `7f0d703b` fix). New behavior is asserted: 200 + JSON body parses + `mustChangePassword=false`. No existing tests weakened. Feature-level test through `AuthRoutes.runZIO` — correct level.
- **3. Architectural invariants.** No router-agent, snapshot, `Clock`, `/tmp`, or DNS-vs-enforcement surface touched.
- **4. Backwards compatibility.** Wire-visible behavior change on `/api/auth/change-password` (empty body → JSON). AGENTS.md's public-contract rule targets the router↔API surface, not SPA↔API (SPA deploys atomically with the API on self-hosted, and the operator's standing preference is to change the endpoint rather than add compat shims). PR body inventories callers — SPA `req<void>` discards either way, `fake-router.py` updated in this diff, `install.sh` only reads status. No client asserts on empty body.
- **5. Migrations.** None.
- **6. Metrics + dashboard.** Existing route modified; no new instrumented path required.
- **7. Scope / hygiene.** Diff is tight (4 files, +29/-11); no AI mentions in commits, title, or body.
- **8. Security.** Route remains JWT-gated via `requireAuthSkipPwCheck`; no SQL changes; no new credentials.

VERDICT: APPROVE

Small, focused fix that returns a JSON body from `/api/auth/change-password` and removes the `fake-router.py` workaround. TDD progression is visible, the new behavior is pinned by a feature test through `AuthRoutes`, and the SPA's `req` helper tolerates both old and new shapes so no client breaks.

sameerparekh:

Reviewed commit: `a8c08441` — incremental delta from `7f0d703b...a8c08441`.

### Prior findings status

| NIT-1 | `cpResp.mustChangePassword == false` style | ADDRESSED — now `!cpResp.mustChangePassword` (`api/test/src/feature/AuthApiSpec.scala:119`). |

### New findings (delta)

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
(none)

VERDICT: APPROVE

Prior NIT addressed; no new findings in the incremental delta. The PR remains a small, focused fix that returns a JSON body from `/api/auth/change-password`, with a feature-test pin and the `fake-router.py` workaround removed.

Issue #1302, opened 2026-06-01
metrics: agent dns_queries_total{result} counter


sameerparekh:
Follow-up to #1206 (OpenWRT agent metrics push).

We want a `dns_queries_total{result}` counter (bounded `result` enum, e.g.
`resolved` / `nxdomain` / `servfail`) so an operator can see DNS query
volume and failure rate from the router. dnsmasq on the router handles
hostname attribution (it is NOT the enforcement plane — blocking is at the
nftables connection layer), and the agent already tails the dnsmasq log via
the dns-tail sidecar (`wifihaven.dns_log`), which is the natural place to
classify and count query outcomes.

Deferred from #1206 to keep that PR focused on the apply/poll/restart
telemetry. When implemented: increment through the in-process `metrics`
registry so it ships in the existing `RouterMetricsBatch` push, add the
`(name -> allowed keys)` entry to `MetricGuard.Allowed` in
`api/src/metrics/Metrics.scala`, and add a Grafana panel in the same PR.

Marked in code with `TODO(#<this issue>):` in
`openwrt/files/usr/lib/lua/wifihaven/dns_log.lua`.

PR #1733, opened 2026-06-14
feat(#1572): best-effort CNAME alias recovery — persist alias map across dns-tail restarts

### LINKED ISSUE #1572: router: traffic/usage records still attributed to resolved CNAME target, not queried brand (incomplete #1344 — best-effort alias recovery)

sameerparekh:
Closes #1572.

## Problem

The CNAME→brand alias map #1344 introduced lives in memory only. On every `wifihaven-dns-tail` restart it starts cold, so a client that queries a CDN edge name **directly** (the Apple re-query pattern #1346 describes) before its branded chain is re-observed gets attributed to the opaque CDN target. The recon in #1572 found ~33 such polluted entries on prod (`dn0qt3r0xannq.cloudfront.net`, `media.amazon.map.fastly.net`, …) — the exact failure mode #1344 set out to fix.

## Fix

Persist the alias map alongside the existing ip→hostname cache, mirroring the `dump_text` / `load_table` pattern already used for entries:

- `dump_aliases_text()` — instance serializer (`<target>\t<head>\t<ts>\n` per non-expired edge)
- `load_aliases_table(text, ttl, now)` — inverse, drops expired edges
- `seed_aliases(tbl, now)` — pre-loads `target → head` edges on startup; existing in-process edges win over stale snapshot rows
- `paths.dns_aliases = /tmp/wifihaven-dns-aliases.txt`
- `wifihaven-dns-tail`: reads the snapshot on startup, writes it on every cache flush

The snapshot is bounded by the existing `max_aliases` (10 000) cap — it's a full-file atomic snapshot, not an append-grow spool, so per AGENTS.md "bounded /tmp writes" it does not need a rotation entry.

## Before / after

Sequence (before this PR):

```
t0  query[A]  brand.com
t0  reply     brand.com CNAME edge.akamai.net
t0  reply     edge.akamai.net is 1.2.3.4         ← cache.lookup(1.2.3.4) = brand.com ✓
*** dns-tail restart (in-memory alias map lost) ***
t1  query[A]  edge.akamai.net                    (Apple re-query, direct)
t1  reply     edge.akamai.net is 5.6.7.8         ← cache.lookup(5.6.7.8) = edge.akamai.net ✗
```

After: `dns-tail` reseeds the alias map from `/tmp/wifihaven-dns-aliases.txt` at startup; the `t1` lookup attributes back to `brand.com`.

## Wire shape

No wire changes — the snapshot is a router-internal IPC file. `connection_events` / `traffic_reports` shapes are unchanged.

## Test plan

- [x] `busted openwrt/test/dns_log_spec.lua` — 64 successes (4 new persistence cases under "alias-map persistence (#1572)")
- [x] Red → green commits visible: `040d0c5b test(#1572): RED — …` then `c534593c feat(#1572): …`
- [x] Full `openwrt/test/run_tests.sh` — only the pre-existing nft_log_dep_spec failures, unrelated to this PR

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
@@ -424,6 +424,47 @@ function M.new(opts)
     store(dst_ip, head, now)
   end
 
+  -- Serialise the live CNAME-alias map to a single newline-delimited string
+  -- (#1572). Format: "<target>\t<head>\t<ts>\n" per non-expired alias edge.
+  -- Persisted alongside dump_text() so a wifihaven-dns-tail restart can reseed
+  -- the alias map and recover the queried brand for a directly-queried CDN
+  -- target whose branded chain was last observed before the restart — the
+  -- failure mode that left ~33 traffic_reports / time_usage rows attributed
+  -- to opaque CDN edge names in #1572's recon.
+  function self.dump_aliases_text()
+    local now = now_fn()
+    local lines = {}
+    for target, a in pairs(aliases) do
+      if (now - a.ts) <= ttl then
+        lines[#lines + 1] = string.format("%s\t%s\t%d", target, a.head, a.ts)
+      end
+    end
+    return table.concat(lines, "\n") .. (#lines > 0 and "\n" or "")
+  end
+
+  -- seed_aliases(tbl, now) — pre-load `target → head` edges learned by a
+  -- previous wifihaven-dns-tail process (#1572). `tbl` is a plain
+  -- `{[target]=head}` table (the output of dns_log.load_aliases_table). Each
+  -- edge is inserted with the given timestamp (default: now) so it ages out
+  -- under the same TTL the in-process learned edges do. Existing edges are not
+  -- overwritten by a stale seed, so a sidecar that has already relearned the
+  -- chain wins over its own old snapshot.
+  function self.seed_aliases(tbl, now)
+    if type(tbl) ~= "table" then return end
+    now = now or now_fn()
+    for target, head in pairs(tbl) do
+      if type(target) == "string" and type(head) == "string"
+         and target ~= "" and head ~= "" and target ~= head
+         and aliases[target] == nil then
+        if alias_count >= max_aliases then
+          evict_oldest_alias()
+        end
+        alias_count = alias_count + 1
+        aliases[target] = { head = head, ts = now, seq = next_seq() }
+      end
+    end
+  end
+
   -- Serialise the live cache to a single newline-delimited string.
   -- Format: "<ip>\t<hostname>\t<ts>\n" per non-expired entry.
   function self.dump_text()
@@ -510,6 +551,26 @@ function M.populate_bl(reply, member_index, resolve_head_fn, add_fn)
     name = name:sub(dot + 1)
   end
   return matched
+end
+
+-- Parses the output of an instance's `dump_aliases_text()` and returns a
+-- plain `{target → head}` table with edges older than `ttl_seconds` (relative
+-- to `now`) filtered out (#1572). Inverse of dump_aliases_text; the
+-- wifihaven-dns-tail sidecar reads this snapshot on startup and hands it to
+-- cache.seed_aliases() so the in-memory CNAME-alias map survives a restart.
+function M.load_aliases_table(text, ttl_seconds, now)
+  local out = {}
+  if type(text) ~= "string" or text == "" then return out end
+  for line in text:gmatch("[^\n]+") do
+    local target, head, ts = line:match("^(%S+)\t(%S+)\t(%d+)$")
+    if target and head and ts then
+      local ts_n = tonumber(ts)
+      if ts_n and (now - ts_n) <= ttl_seconds then
+        out[target] = head
+      end
+    end
+  end
+  return out
 end
 
 -- Parses the output of an instance's `dump_text()` and returns a plain
--- a/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/dns_log.lua
@@ -448,7 +448,9 @@ function M.new(opts)
   -- edge is inserted with the given timestamp (default: now) so it ages out
   -- under the same TTL the in-process learned edges do. Existing edges are not
   -- overwritten by a stale seed, so a sidecar that has already relearned the
-  -- chain wins over its own old snapshot.
+  -- chain wins over its own old snapshot. If the seed exceeds max_aliases the
+  -- LRU evicts as we insert, so which edges survive is iteration-order
+  -- dependent — acceptable for best-effort recovery (#1572).
   function self.seed_aliases(tbl, now)
     if type(tbl) ~= "table" then return end
     now = now or now_fn()
--- a/openwrt/files/usr/lib/lua/wifihaven/paths.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/paths.lua
@@ -14,6 +14,14 @@ M.block_page_api_url  = "/var/run/wifihaven/api_url"
 -- ip → hostname cache (#259): wifihaven-dns-tail writes, wifihaven-agent reads.
 M.dns_cache = "/tmp/wifihaven-dns-cache.txt"
 
+-- CNAME-alias map snapshot (#1572): wifihaven-dns-tail writes in lockstep with
+-- dns_cache and reloads it on startup via cache.seed_aliases. Persists the
+-- target→branded-head edges learned from observed CNAME chains so a sidecar
+-- restart does not lose the ability to attribute a directly-queried CDN edge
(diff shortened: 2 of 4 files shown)


sameerparekh:
## Independent PR review

Reviewed `origin/main...HEAD` (c534593c) against the 8-dimension checklist.

### BLOCKERS

1. **`openwrt/files/usr/sbin/wifihaven-dns-tail:374` — alias snapshot is only written from ONE of the two flush sites.** The file has two parallel flush blocks (SNI-routed branch around L322–328, dnsmasq-line branch around L370–378). The PR adds `atomic_write(aliases_path, cache.dump_aliases_text())` only at the SNI branch. The dnsmasq-line branch — the dominant path on a normal router, since most flushes happen on a `reply <name> is <ip>` ingest, not an SNI line — still writes `cache_path` and `dns_metrics_path` but never `aliases_path`. So in practice the on-disk alias snapshot updates only when an SNI flush happens to fire, can lag the in-memory map by minutes, and on a router whose `wifihaven-sni-tail` sidecar is disabled it never updates at all. The whole "lockstep with the cache file" claim in the PR body and the comment at L323 doesn't hold. Mirror the write at the dnsmasq-line flush site too (between the existing `cache_path` write and the `#1302` comment). Merge-gating because this is the meat of the fix.

### SHOULD-FIX

(none)

### NITS

1. **`openwrt/files/usr/sbin/wifihaven-dns-tail:62–72` — reseed reads no metric.** Per AGENTS.md §instrument-new-functionality, a "fleet/system state an operator would check first during an incident" gauge fits here (e.g. `dns_alias_reseed_count` set once at startup). Not blocking — the `log.info` heartbeat is sufficient for now — but a counter/gauge would let an operator confirm the persistence is actually working across the fleet without grepping `logread`.
2. **`openwrt/files/usr/lib/lua/wifihaven/dns_log.lua:459` — `seed_aliases` quietly drops aliases beyond `max_aliases`.** When the snapshot file holds more edges than the cap (default 10 000), the LRU evicts as we insert, so iteration order over the seeded table determines which edges survive. Not a correctness issue (it's already best-effort), but worth a one-line comment so a future reader doesn't assume the seed is faithful.

- **Single source of truth (§1)**: alias state stays the one map in `aliases`; `dump_aliases_text` / `load_aliases_table` are pure serializers. No drift risk.
- **Testing (§2)**: red commit `040d0c5b` precedes impl commit `c534593c`. Four new cases cover dump, restart round-trip + lookup attribution, expiry filter, and malformed-input tolerance. The round-trip case in particular pins the bug.
- **Architectural invariants (§3)**: router-internal change only; no decision logic added to the agent. The `/tmp` snapshot is a full-file atomic rewrite bounded by `max_aliases` (~10 000 lines), so the bounded-`/tmp`-writes rule's snapshot exemption applies — no rotation needed.
- **Backwards compat (§4)**: no wire change. Snapshot file is local to the router; an older image on rollback simply ignores it.
- **Scope / hygiene (§7)**: focused 4-file diff, no AI/Claude mentions in commit messages or PR body, no drive-by edits.

### VERDICT: REQUEST-CHANGES

The persistence path itself is the right shape — dump/seed/load mirror the existing entry-cache pattern, the TTL handling is correct, and the test for the restart round-trip pins the bug. But the dns-tail wiring only persists the alias map from the SNI flush branch, leaving the dominant dnsmasq-line flush path silently stale — so on a typical router the on-disk snapshot will not in fact track the in-memory map.

Address the BLOCKER (mirror the `atomic_write(aliases_path, …)` line in the second flush block) and this is APPROVE.

sameerparekh:

Statused the prior findings against the latest push (04c55bac) and reviewed the incremental delta `c534593c...04c55bac`.

### Prior-findings status

|---|---|---|
| BLOCKER 1 | `wifihaven-dns-tail` — alias snapshot only written from the SNI-routed flush site | **ADDRESSED** — `wifihaven-dns-tail:377` now also writes `atomic_write(aliases_path, cache.dump_aliases_text())` in the dnsmasq-line flush block, between the cache write and the #1302 metrics write. A new comment also documents that the two flush sites must stay in lockstep. |
| NIT 1 | No metric on reseed | NOT-ADDRESSED — left as future work (NIT, non-blocking). |
| NIT 2 | `seed_aliases` quietly drops aliases beyond `max_aliases` | **ADDRESSED** — `dns_log.lua:451–453` adds the one-line comment explaining the iteration-order tradeoff. |

### New findings in `c534593c...04c55bac`

#### BLOCKERS
(none)

#### SHOULD-FIX
(none)

#### NITS
(none)

The delta is exactly the two-site lockstep wiring and the one-line seed_aliases comment called for in the prior review. No new risk introduced; tests still pass (64/64 in `dns_log_spec.lua`).

### VERDICT: APPROVE

The prior BLOCKER is ADDRESSED with no new BLOCKER introduced. The persistence path now updates the on-disk snapshot from both flush branches, so the in-memory CNAME-alias map and the `/tmp/wifihaven-dns-aliases.txt` snapshot track each other under both line types — restoring the "lockstep with the cache file" guarantee the PR body claims. Single open NIT (a reseed metric) is non-blocking.

Issue #1538, opened 2026-06-07
bug(prod): removed block-schedule resurrects on next API restart — migrateProfiles re-migrates from retained legacy schedules row (+ schedule-detach skips TimeStatusCache invalidation)


sameerparekh:
## Symptom (prod)
Profile **sameer** showed as **paused for Schedule** with no active schedule. A schedule had appeared \"a moment ago\" (origin unknown to the operator); removing it via the UI cleared the attachment, but the block reason persisted / the schedule keeps coming back.

## Root cause A — schedule resurrection on restart (the real bug)
`ScheduleSeeder.migrateProfiles` (`api/src/ScheduleSeeder.scala`) runs on **every API boot** (Main.scala `seedAndMigrate`). Its per-profile idempotency guard is:

```scala
named.blockScheduleIdsForProfile(p.id).flatMap { existing =>
  if existing.nonEmpty then ZIO.succeed(acc)        // "already migrated"
  else legacy.listForProfile(p.id).flatMap { ... migrate + setProfileBlockSchedules ... }
}
```

The guard keys off whether the profile **currently has a block-mode named schedule attached** — NOT a durable \"this profile was already migrated\" marker. Meanwhile the legacy V1/V16 `schedules` rows are **intentionally retained** (the seeder comment + #1485 gate to ~Jun 12 keep them for rollback safety). So the resurrection loop is:

1. Earlier deploy migrated sameer's legacy `schedules` row → created a `named_schedules` row + attached it via `profile_schedule_rules` (mode `blocked_during`).
2. Operator **detaches** the schedule in the UI → `PUT /api/profiles/{id}/schedules` → `setProfileBlockSchedules` deletes the `profile_schedule_rules` row. `blockScheduleIdsForProfile(sameer)` now returns empty. **The legacy `schedules` row is still there.**
3. Next API restart / deploy → `migrateProfiles` sees `existing.isEmpty` → falls through → re-reads the retained legacy row → **re-creates a named schedule and re-attaches it.** The \"removed\" schedule is back, and the profile is blocked again at its window.

This is the migrate-resurrection / write-before-read class: idempotency must be anchored to a durable \"migrated\" fact, not to mutable downstream state (the attachment) that the user can legitimately change. The seeder already acknowledges the analogous gap for default-seeding (\"a durable 'seeded once' marker (household_settings column) is a follow-up\") — the migration guard has the same weakness.

## Root cause B — schedule detach doesn't invalidate the time-status cache (secondary)
[...]

PR #1735, opened 2026-06-14
feat(#423): PATCH /api/profiles/:id for partial updates

### LINKED ISSUE #423: Add PATCH /api/profiles/:id for partial, field-scoped updates
## Motivation

Today the only way to update an existing profile is `PUT /api/profiles/:id`, which requires the full profile body (name, blockedCategories, extraBlocked, extraAllowed, schedules, timeLimit, siteTimeLimits, paused). This is unsafe under concurrent edits:

- Tab A loads the profile, edits `schedules`, sends PUT with original `extraBlocked`.
- Meanwhile Tab B loads the profile, edits `extraBlocked`, sends PUT with original `schedules`.
- Whichever PUT lands second silently clobbers the other tab's change to the *other* field.

The toggle `POST /api/profiles/:id/pause` was deleted in #406's fix because of a related race (read-then-write toggle); the migration replaced it with a full PUT. That's correct but it makes the concurrent-edit-clobber problem worse — every pause-button click now sends every field.

## Proposal

Add `PATCH /api/profiles/:id` accepting a partial body — only the fields the caller wants to change. Each field is updated independently; unspecified fields are left alone. Race-safe per-field semantics:

```
PATCH /api/profiles/2
{ \"paused\": true }
```

…leaves `name`, `schedules`, `blockedCategories`, etc. untouched.

## Out of scope

- Optimistic concurrency (If-Match / version field) — separate issue if we ever see contention on the *same* field.
- PATCH for other resources — file separately if needed.

## Acceptance

- `PATCH /api/profiles/:id` with each individual field works; unspecified fields untouched.
- Admin UI pause button switches from full PUT to `PATCH { paused: bool }`.
- Tests for each field's partial update.

sameerparekh:
Closes #423.

Adds `PATCH /api/profiles/:id` so the SPA can issue field-scoped writes (e.g. a paused toggle, an autosave on `name`) without the read-then-write clobber that today's `PUT` requires. Unblocks the Profile half of the autosave rollout (#995 → #973–#977).

## Wire shape (additive, backwards-compatible)

Body is a JSON object whose keys are a subset of the writable `Profile` fields. `PUT` is unchanged — pre-PATCH clients keep working.

Tri-state JSON semantics via the existing `FieldPatch` helper used by `/api/devices/:mac`, `/api/users/:id`, `/api/apps/:id`:

- **field absent** → preserve existing value
- **field == `null`** → clear (legal only on nullable `timeLimit`; rejected with 400 elsewhere)
- **field == `v`** → assign `v`

Patchable fields: `name`, `blockedCategories`, `paused`, `failureMode`, `blockIpOnly`, `crossDeviceOverlapMode`, `pauseMode`, `defaultDeny`, `timeLimit`. Schedules and users stay on their own sub-routes (`PUT /api/profiles/:id/schedules`, `PUT /api/profiles/:id/users`) by design (#1494, #406).

Cache: a successful patch that touches `paused` or `timeLimit` busts the per-profile `TimeStatusCache` entry, mirroring the schedule-attach PUT and `/api/time/extend` (#1538).

## Why not block on #581 first

#581 (symmetric read/write shapes) is a desirable cleanup but not a wire prerequisite: PATCH accepts a Json.Obj subset of the existing flat write shape, so the read shape (`ProfileDetail`) can stay as-is for now. The autosave rollout can land on top of PATCH and #581 can collapse the read/write surfaces in a follow-up without changing PATCH semantics.

## Tests (TDD red→green visible in commit history)

New feature spec `api/test/src/feature/ProfilePatchApiSpec.scala`:

- single-field patches (`paused`, `name`) leave every other field untouched
- multi-field patch updates each supplied field
- empty patch `{}` preserves every field
- `timeLimit: null` clears the row; `timeLimit: N` upserts
- `name: null` / `paused: null` → 400 (non-nullable)
- invalid JSON → 400
- missing profile → 404
- no token → 401

## SPA

- `api.profiles.patch(id, data)` in `web/src/api/client.ts`
- `PatchProfileRequest` interface in `web/src/types/api.ts`

Existing `PUT` call sites are unchanged — that swap is #995's autosave-rollout work, not this PR.

### DIFF
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -104,7 +104,22 @@ trait ProfileRepo {
   def create(name: String, cats: List[BlocklistId]): Task[ProfileId]
   def update(p: Profile): Task[Unit]
   def delete(id: ProfileId): Task[Unit]
+
+  // #423: per-field setters for the PATCH handler so concurrent PATCHes
+  // touching disjoint fields don't race via a full-row rewrite. Each method
+  // issues a targeted `UPDATE profiles SET <col>=? WHERE id=?` — the absent
+  // fields aren't touched by the SQL, so a "tab A patches paused, tab B
+  // patches name" interleaving preserves both edits. The full-shape
+  // [[update]] above stays for the legacy PUT route, which is documented as
+  // a full-replace.
+  def setName(id: ProfileId, name: String): Task[Unit]
+  def setBlockedCategories(id: ProfileId, cats: List[BlocklistId]): Task[Unit]
   def setPaused(id: ProfileId, paused: Boolean): Task[Unit]
+  def setFailureMode(id: ProfileId, mode: FailureMode): Task[Unit]
+  def setBlockIpOnly(id: ProfileId, v: Boolean): Task[Unit]
+  def setCrossDeviceOverlapMode(id: ProfileId, mode: CrossDeviceOverlapMode): Task[Unit]
+  def setPauseMode(id: ProfileId, mode: PauseMode): Task[Unit]
+  def setDefaultDeny(id: ProfileId, v: Boolean): Task[Unit]
 }
 
 /**
@@ -872,8 +887,34 @@ class ProfileRepoLive(xa: Transactor[Task]) extends ProfileRepo {
       .transact(xa)
       .unit
   def delete(id: ProfileId) = sql"DELETE FROM profiles WHERE id=$id".update.run.transact(xa).unit
-  def setPaused(id: ProfileId, p: Boolean) =
+
+  // #423: targeted per-column setters for the PATCH handler. Each runs an
+  // UPDATE that touches exactly one column, so concurrent PATCHes on
+  // disjoint fields no longer race through a full-row rewrite.
+  def setName(id: ProfileId, name: String)                                   =
+    sql"UPDATE profiles SET name=$name WHERE id=$id".update.run.transact(xa).unit
+  def setBlockedCategories(id: ProfileId, cats: List[BlocklistId])           =
+    sql"UPDATE profiles SET blocked_categories=${cats.map(_.value).toArray} WHERE id=$id".update.run
+      .transact(xa)
+      .unit
+  def setPaused(id: ProfileId, p: Boolean)                                   =
     sql"UPDATE profiles SET paused=$p WHERE id=$id".update.run.transact(xa).unit
+  def setFailureMode(id: ProfileId, mode: FailureMode)                       =
+    sql"UPDATE profiles SET failure_mode=${FailureMode.asString(mode)} WHERE id=$id".update.run
+      .transact(xa)
+      .unit
+  def setBlockIpOnly(id: ProfileId, v: Boolean)                              =
+    sql"UPDATE profiles SET block_ip_only=$v WHERE id=$id".update.run.transact(xa).unit
+  def setCrossDeviceOverlapMode(id: ProfileId, mode: CrossDeviceOverlapMode) =
+    sql"UPDATE profiles SET cross_device_overlap_mode=${CrossDeviceOverlapMode.asString(mode)} WHERE id=$id".update.run
+      .transact(xa)
+      .unit
+  def setPauseMode(id: ProfileId, mode: PauseMode)                           =
+    sql"UPDATE profiles SET pause_mode=${PauseMode.asString(mode)} WHERE id=$id".update.run
+      .transact(xa)
+      .unit
+  def setDefaultDeny(id: ProfileId, v: Boolean)                              =
+    sql"UPDATE profiles SET default_deny=$v WHERE id=$id".update.run.transact(xa).unit
 }
 
 class HouseholdSettingsRepoLive(xa: Transactor[Task]) extends HouseholdSettingsRepo {
--- a/api/src/routes/Routes.scala
+++ b/api/src/routes/Routes.scala
@@ -205,6 +205,24 @@ object AuthRoutes {
 // ── Profile routes ─────────────────────────────────────────────────────────
 
 object ProfileRoutes {
+  // #423 — the set of writable Profile fields the PATCH handler recognizes.
+  // Kept as a single named source so the PATCH log message can filter to
+  // recognized keys, AND so the test pin in ProfilePatchApiSpec can assert
+  // it covers every field on the underlying `Profile` case class minus `id`.
+  // If you add a writable field to `Profile`, update this set and PATCH's
+  // `p.copy(...)` together — the test pin will fail loudly otherwise.
+  val PatchableKeys: Set[String] = Set(
+    "name",
+    "blockedCategories",
+    "paused",
+    "failureMode",
+    "blockIpOnly",
+    "crossDeviceOverlapMode",
+    "pauseMode",
+    "defaultDeny",
+    "timeLimit",
+  )
+
   def routes(
       auth: AuthService,
       profileRepo: ProfileRepo,
@@ -486,13 +504,20 @@ object ProfileRoutes {
             }
             // #1538: paused / timeLimit changes affect ProfileTimeStatus, so
             // bust the per-profile cache the same way schedule attach/detach
-            // and /api/time/extend do — otherwise a freshly-patched pause
-            // wouldn't reach the dashboard until the today-TTL expired.
-            _                       <- ZIO.when(
-              pausedPatch != FieldPatch.Absent || timeLimitPatch != FieldPatch.Absent,
-            )(cache.invalidateProfile(pid))
-            _                       <- ZIO.logInfo(
-              s"profile patched: id=${pid.value} keys=${obj.fields.map(_._1).mkString(",")}",
+            // and /api/time/extend do. Guard on a real value change — not
+            // mere field presence — so a no-op `{"paused": false}` on an
+            // already-unpaused profile doesn't churn the cache.
+            pausedChanged = pausedPatch match {
+              case FieldPatch.Set(v) => v != p.paused
+              case _                 => false
+            }
+            timeLimitChanged = timeLimitPatch != FieldPatch.Absent
+            _ <- ZIO.when(pausedChanged || timeLimitChanged)(cache.invalidateProfile(pid))
+            // Log only recognized keys; unknown keys are ignored per the
+            // backwards-compat rule and would mislead ops triage if echoed.
+            recognizedKeys = obj.fields.map(_._1).filter(ProfileRoutes.PatchableKeys.contains)
+            _ <- ZIO.logInfo(
+              s"profile patched: id=${pid.value} keys=${recognizedKeys.mkString(",")}",
             )
           } yield Response.ok
           handle.mapError(ErrorMapper.errorToResponse)
--- a/api/src/routes/Routes.scala
+++ b/api/src/routes/Routes.scala
@@ -400,6 +400,103 @@ object ProfileRoutes {
               ZIO.succeed(Response.ok)
           handle.mapError(ErrorMapper.errorToResponse)
         },
+      // #423: field-scoped partial update. Body is a subset of the writable
+      // Profile shape — every field optional, omitted fields preserve their
+      // current value. `timeLimit` is the only nullable field and accepts
+      // explicit null to clear; non-nullable fields reject null with 400.
+      // Race-safe vs the PUT clobber pattern that motivated #423 — the SPA
+      // can post `{"paused":true}` without first reading and re-shipping the
+      // rest of the profile. Schedules / users still live on their own
+      // sub-routes by design (#1494, #406) and are not patchable here.
+      Method.PATCH / "api" / "profiles" / long("id")             ->
+        handler { (id: Long, req: Request) =>
+          val pid                                  = ProfileId(id)
+          val handle: ZIO[Any, ApiError, Response] = for {
+            claims    <- requireWriter(req, auth).mapError(ApiError.Wrapped(_))
+            _         <- requireProfileAccess(claims, pid, userProfileRepo)
+              .mapError(ApiError.Wrapped(_))
+            body      <- req.body.asString.orElseFail(ApiError.BadRequest(""))
+            obj       <- ZIO.fromEither(FieldPatch.parseObj(body)).mapError(ApiError.BadRequest(_))
+            p         <- profileRepo
+              .findById(pid)
+              .mapError(ApiError.Db(_))
+              .flatMap(ZIO.fromOption(_).orElseFail(ApiError.NotFound("Profile not found")))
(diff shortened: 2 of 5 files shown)


sameerparekh:
Independent PR review (read-only, adversarial) per [`docs/pr-review-checklist.md`](https://github.com/wifihaven/wifihaven/blob/main/docs/pr-review-checklist.md).

Diff: 4 files / +326 / −1 — new `PATCH /api/profiles/:id` route + spec + SPA client/type.

## BLOCKERS

(none)

## SHOULD-FIX

1. **`api/src/routes/Routes.scala:343-394` (PUT) vs `:403-499` (PATCH) — field list duplicated across both handlers.** PUT's `p.copy(name=…, blockedCategories=…, paused=…, failureMode=…, blockIpOnly=…, crossDeviceOverlapMode=…, pauseMode=…, defaultDeny=…)` and PATCH's `p.copy(...)` enumerate the same eight Profile fields. The semantics differ (PUT requires the field; PATCH preserves on absent), so direct collapse is awkward — but per [single-source-of-truth](https://github.com/wifihaven/wifihaven/blob/main/AGENTS.md#single-source-of-truth) the divergence risk is real: a future Profile field added to `UpsertProfileRequest`/PUT will silently be unreachable via PATCH until someone notices. Prefer either (a) **TYPE-ENFORCE** — derive both folds from a single helper, or (b) **TEST-PIN** — a small test that fails when `Profile` gains a writable field PATCH doesn't carry. (`timeLimit` lives outside `Profile`, so it's a separate concern.) This is the exact class the #581 symmetric-shapes work would eliminate; an in-PR test-pin is a cheap interim guard.

2. **`api/src/routes/Routes.scala:493-495` — cache invalidation triggers on any `paused` *presence* in the patch, not on a value *change*.** `pausedPatch != FieldPatch.Absent` is true for `{"paused": false}` even when the row is already `paused=false`. Over-invalidation is harmless (just refills the cache) so this is a SHOULD-FIX, not a blocker — but compare with the existing call sites (`/api/time/extend` at :926, schedule attach at :291) which invalidate unconditionally on a real state-changing operation. Cheap fix: guard on `Set(v) where v != p.paused`, or drop the conditional and always invalidate.

## NITS

1. **`api/src/routes/Routes.scala:497-498` — `keys=${obj.fields.map(_._1).mkString(",")}` logs unknown client-sent fields verbatim.** Backwards-compat rule is "ignore unknown fields on input" (preserved correctly), but the log line then echoes them as if they were applied. Filter to recognized keys before logging if this log ever gates ops triage.

2. **`web/src/types/api.ts:823-833` — `PatchProfileRequest` is self-typed by hand against the Scala shape.** Same as the existing `PatchUserRequest` / `PatchDeviceRequest` / `PatchAppRequest` pattern, so no drift introduced here — but every additional handwritten Patch shape widens #581's collapse target.

3. **Issue #423's acceptance bullet "Admin UI pause button switches from full PUT to `PATCH { paused: bool }`"** is intentionally deferred to #995 per the PR body. That's the right call (autosave rollout owns the call-site swap); confirm the deferral is reflected on #995's tracking list so the pause-button conversion isn't lost.

## Dimension coverage

- **1. Duplication / SSoT:** see SHOULD-FIX #1. No re-derivation of `TimeStatusService` / `Presence` / `BlockReason` primitives; cache invalidation reuses `TimeStatusCache.invalidateProfile`.
- **2. Testing integrity:** TDD red→green visible in history (`f662ccb7` test commit precedes `496d85d8` impl). No existing tests modified, deleted, weakened, or skipped. New spec covers each field path, the empty patch, null on nullable and non-nullable fields, invalid JSON, 404, 401. Real embedded Postgres via `TestDatabase.layer`; no repo/AuthService/Clock mocks; Clock via `TestClock.schoolDayAfternoon`.
- **3. Architectural invariants:** Server-side `PolicyService` model unaffected; router wire untouched; no DNS reasoning. `Clock`/`ZIO`/`zio-config` rules N/A or respected. No new `/tmp` writes.
- **4. Wire compat:** purely additive — new HTTP method on existing path. Old clients keep using `PUT`. Unknown body fields ignored. No router-facing shape touched.
- **5. Migrations:** none.
- **6. Metrics + dashboard:** new HTTP route is covered by the templated `/api/profiles/:id` route metric in the existing HTTP middleware. No new emitted series ⇒ no dashboard delta required.
- **7. Scope / hygiene:** focused 4-file diff; no AI mentions in title/body/commits; merge-base three-dot diff used.
- **8. Security:** `requireWriter` + `requireProfileAccess` enforced before any DB read, matching PUT/DELETE; no new SQL; no secrets; `timeLimit` parsed as `Int` so non-numeric payloads 400 cleanly.

## VERDICT: APPROVE

Additive `PATCH /api/profiles/:id` cleanly mirrors the existing FieldPatch convention used by `/api/devices/:mac`, `/api/users/:id`, `/api/apps/:id`; spec covers the tri-state semantics and auth/404 paths.
No BLOCKERS. SHOULD-FIX items are about divergence-risk hardening (collapse / test-pin the field list) and over-invalidation; both can land here or as a quick follow-up without re-litigating the wire shape.
Unblocks the Profile half of the #995 autosave rollout without breaking existing PUT callers.

sameerparekh:
Re-run of the independent PR review on the new HEAD `e5490184`. Read-only pass per [`docs/pr-review-checklist.md`](https://github.com/wifihaven/wifihaven/blob/main/docs/pr-review-checklist.md).

## Prior-findings status (vs `496d85d8`)

| 1 | SHOULD-FIX | PUT vs PATCH field-list duplication → drift risk | **ADDRESSED** — `ProfileRoutes.PatchableKeys` extracted as the single named source (`api/src/routes/Routes.scala:208-220`) and a reflective drift-pin test (`api/test/src/feature/ProfilePatchApiSpec.scala:200-211`) asserts the set equals every writable `Profile` field + `timeLimit`. A future `Profile` field that isn't added to `PatchableKeys` AND the PATCH handler will now fail CI. |
| 2 | SHOULD-FIX | Cache invalidation triggered on `paused` presence, not value change | **ADDRESSED** — `Routes.scala:510-517` now compares `Set(v) != p.paused` instead of `FieldPatch.Absent` inequality. A no-op `{"paused": false}` against an already-unpaused profile no longer churns the cache. `timeLimitChanged` still fires on any presence (Cleared / Set) which is correct — every `timeLimit` PATCH write changes either the row or nothing-vs-something. |
| 1 | NIT | Log line echoed unknown client-sent keys | **ADDRESSED** — `Routes.scala:519-522` filters `obj.fields` through `PatchableKeys.contains` before formatting. |
| 2 | NIT | `PatchProfileRequest` self-typed by hand | NOT-ADDRESSED — deliberately deferred to #581's collapse work as noted in the original review. Not merge-gating. |
| 3 | NIT | #423's "swap the pause button" acceptance bullet deferred to #995 | NOT-ADDRESSED — deferred per the PR body. Not merge-gating. |

## New findings (delta `496d85d8...e5490184`)

Delta is 2 files / +52 / −7: the `PatchableKeys` constant + drift-pin test + the value-change guard + the recognized-keys log filter.

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

1. **`api/test/src/feature/ProfilePatchApiSpec.scala:202-205` — reflective drift pin filters synthetic fields by the string `$`.** Works on the current Scala 3 codegen (only `OFFSET$_m_0`-style synthetics contain `$`), but it's a sentinel that could silently let a legitimately named field with `$` through, or break if codegen changes its naming. A `scala.deriving.Mirror.ProductOf[Profile]` over `MirroredElemLabels` is the type-system-blessed source for case-class parameter labels and would not rely on the substring heuristic. Cosmetic — current implementation is correct against today's compiler output and is documented in-line.

2. **`api/src/routes/Routes.scala:519-522` — recognized-keys log loses the *count of dropped* unknown keys.** If an old/typo'd client posts `{"namee": "X"}`, the log emits `keys=` with no signal that anything was ignored. A debug-level `logInfo` of dropped keys (or `keys=<recognized> dropped=<n>`) would aid triage. Cosmetic.

## Dimension coverage (delta only)

- **1. Duplication / SSoT:** improved — the field set is now a single named constant.
- **2. Testing integrity:** drift-pin test is new and asserts on the canonical source (`classOf[Profile].getDeclaredFields`). All 12 existing spec tests still pass. No prior tests weakened or removed.
- **3. Architectural invariants:** unchanged. No router wire, DNS, `Clock`, or `/tmp` touched.
- **4. Wire compat:** unchanged. Still purely additive.
- **5. Migrations:** none.
- **6. Metrics + dashboard:** unchanged; HTTP middleware metric still covers the new route.
- **7. Scope / hygiene:** focused 2-file delta; no AI mentions in the new commit message; merge-base three-dot diffing used.
- **8. Security:** unchanged.

## VERDICT: APPROVE

Both prior SHOULD-FIX items ADDRESSED with a typed drift pin and a real-change cache guard; one NIT ADDRESSED with the recognized-keys log filter. Two new NITS are cosmetic (Mirror-based labels would be slightly stronger; dropped-key count in the log would aid triage). No BLOCKERS, no SHOULD-FIX.
The drift pin is the durable win — the PUT/PATCH field-list divergence the original review flagged is now a CI failure instead of a code-comment.
Ready to merge; #581 and #995's autosave rollout can build on top.

sameerparekh:
Re-run of the independent PR review on the new HEAD `f5e419a3`. Read-only pass per [`docs/pr-review-checklist.md`](https://github.com/wifihaven/wifihaven/blob/main/docs/pr-review-checklist.md).

**Operator-found regression:** the prior landings of this PR advertised "race-safe per-field PATCH" but the handler did `load → p.copy(...) → profileRepo.update(updated)` — a full-row UPDATE that rewrites every column. Two concurrent PATCHes touching disjoint fields would silently clobber each other (exactly the bug class #423 was opened to prevent). This re-run reviews the fix.

## Prior-findings status (vs `e5490184`)

| — | (operator) | PATCH does load → full-row write, so disjoint-field PATCHes still race | **ADDRESSED** — handler rewritten at `Routes.scala:489-528` to dispatch each `FieldPatch.Set(v)` to a column-scoped setter (`profileRepo.setName/setPaused/…`). No `p.copy(...)` step; `findById` is kept purely as a 404 existence check and the row isn't threaded into writes. New regression spec at `ProfilePatchApiSpec.scala:200-232` fires two PATCHes via `.zipPar` against the same routes value (shared embedded-Postgres pool) and asserts both edits survive. |
| 1 | NIT (prior) | reflective drift pin filters synthetic fields by `$` substring | NOT-ADDRESSED — still cosmetic; carried forward to a follow-up if it ever matters. |
| 2 | NIT (prior) | log line drops count of unknown keys | NOT-ADDRESSED — still cosmetic. |

## New findings (delta `e5490184...f5e419a3`)

Delta is 3 files / +139 / −28: 8 new per-column setters in `ProfileRepo` (+ live impls), handler rewritten to per-field dispatch, race regression test.

### BLOCKERS

(none)

### SHOULD-FIX

1. **`api/src/db/Repos.scala:875-887` — `update(p: Profile)` and the 8 new `set*` methods now duplicate the column ↔ wire-encoding mapping** (`FailureMode.asString`, `CrossDeviceOverlapMode.asString`, `PauseMode.asString`, the `.map(_.value).toArray` for blocklist ids). A future enum-value rename or a new wire-encoding rule would need to land in BOTH places or display-vs-enforcement drift returns ([single-source-of-truth](https://github.com/wifihaven/wifihaven/blob/main/AGENTS.md#single-source-of-truth)). Options: (a) collapse the full-row `update` to call the per-column setters internally; (b) extract the per-column SQL fragments into private helpers that both paths share. Cheap; worth taking before the legacy PUT path drifts.

(Pre-review SHOULD-FIX about a missing mock impl: settled — verified no other class extends `ProfileRepo`; only `ProfileRepoLive` does. The trait additions compile cleanly with the existing wiring.)

### NITS

1. **`api/src/routes/Routes.scala:489-528` — eight near-identical `_ <- xPatch match { case Set(v) => profileRepo.setX(pid, v); case _ => unit }` blocks.** Mechanical and easy to read, but consider a small helper `def applySet[T](fp: FieldPatch[T])(f: T => Task[Unit]) = fp match { case Set(v) => f(v).mapError(ApiError.Db(_)); case _ => ZIO.unit }` to halve the line count. Pure style.

2. **`api/test/src/feature/ProfilePatchApiSpec.scala:200-232` — race test relies on `.zipPar` interleaving against an embedded Postgres connection pool.** It does correctly fail against the pre-fix handler (load-then-write-all-cols would either lose `paused` or lose `name` depending on commit order), but the interleaving is non-deterministic — under exotic scheduling the bug version might pass occasionally. Acceptable for a regression test (the bug surfaces on >50% of runs and CI runs it on every push), and the alternative — driving each fiber through a synchronization point — would couple the test to repo internals. Worth a comment noting the non-determinism trade-off so a future reader doesn't dismiss a flaky failure.

3. **Race test asserts `after.blockedCategories == kids.blockedCategories` but `setupKids` returns the *seeded* Kids profile — fields whose live row state isn't what the test code thinks could pass via stale comparison.** In practice both rows are identical (no PATCH on those fields, no other mutations), so the assertion is meaningful. Just be aware if `setupKids` ever changes its return convention.

## Dimension coverage (delta)

- **1. Duplication / SSoT:** see SHOULD-FIX #2 (column encoders now in two SQL sites). The handler-side fold is now properly per-field, so the PUT-vs-PATCH field-list duplication is materially down (PATCH no longer enumerates `Profile` fields in a `p.copy`; only the dispatch table).
- **2. Testing integrity:** new race test is a real regression test that pins the bug being fixed (per AGENTS.md TDD rule for bug fixes). All 13 spec tests pass; no prior tests weakened. TDD red→green visible (`f5e419a3` is its own commit on top of the existing `f662ccb7` → `496d85d8` → `e5490184` chain).
- **3. Architectural invariants:** unchanged. No router, DNS, `Clock`, `/tmp`.
- **4. Wire compat:** unchanged — still purely additive. Wire shape and semantics unchanged from a client's perspective; only the server-side persistence path changed.
- **5. Migrations:** none.
- **6. Metrics + dashboard:** unchanged.
- **7. Scope / hygiene:** focused 3-file delta; commit message describes the bug, the mechanism, and what stays for PUT; no AI mentions.
- **8. Security:** unchanged; `requireWriter` + `requireProfileAccess` still gate writes.

## VERDICT: APPROVE

Operator-found race regression ADDRESSED at the right layer — per-column setters in the repo + per-field dispatch in the handler + a concurrent regression test that actually exercises the bug. The PATCH handler now matches the shape `PATCH /api/users/:id` already had; Devices and Apps still carry the same regression and are filed for follow-up (#1745, #1746).
Remaining SHOULD-FIX (collapsing the per-column SQL between `update` and the new setters) is real but small — fix here or in a follow-up.
No BLOCKERS; safe to merge.

Issue #1602, opened 2026-06-09
bug(prod): deleted profile schedule resurrects every deploy — ScheduleSeeder still migrates from legacy 'schedules' table


sameerparekh:
## Symptom

Every API restart/deploy, the operator's `sameer` profile gets a previously-deleted named schedule re-added. Confirmed on prod after multiple recent deploys.

## Root cause

`api/src/Main.scala:140` still invokes `ScheduleSeeder.seedAndMigrate` on every startup. The seeder reads the legacy `schedules` table and (re)creates the corresponding `named_schedules` row + `profile_schedule_rules`.

#1538 attempted to fix the resurrection class by anchoring idempotency on "a migrated `named_schedules` row exists with the deterministic 'Migrated from …' description." That guard is correct as written — but it only holds if the migrated row is never deleted. When the operator deletes a migrated named_schedule from the SPA (which is the legitimate way to delete a schedule, post-#1494), the guard reads "not migrated" and the seeder re-imports the still-present legacy row, resurrecting the deletion.

The seeder's own TODO (`ScheduleSeeder.scala:61`) already acknowledges this code dies when #1485 drops the legacy `schedules` table.

## Fix

**Stop calling `ScheduleSeeder.seedAndMigrate` from `Main.scala`.** We're fully migrated to named schedules; the migration has run on every household for weeks. Keeping it running is a strict liability — every deploy is a chance for the resurrection bug to recur for any user who legitimately deletes a migrated schedule.

Scope:
- Remove the call from `Main.scala:140`. Keep the seeder code in place (it's deleted with #1485 along with the legacy table).
- The default-set seeding portion of `seedAndMigrate` (for brand-new households with no profiles yet) — if there is one — keep it behind a "households table is empty" guard, OR move it to a one-shot install path. Read `ScheduleSeeder.scala` to confirm what's in scope vs. only the migration.
- Regression test: a feature test that deletes a migrated named_schedule, restarts the API layer, and asserts the schedule stays deleted.
- This is a no-op for any household whose legacy `schedules` table has already been emptied by an explicit cleanup; for households with retained legacy rows (most of prod), it makes the migration a no-op going forward. Either way the behavior change is "stop re-migrating."

## Backwards-compat / wire

No wire change. No migration. Code-only.

## Relationship to #1485

[...]

PR #1736, opened 2026-06-14
refactor(#602): structured key=value logging across hot Scala paths

### LINKED ISSUE #602: update scala logging to use structured logs,
local logging should be one-line-per-log format, but deployed logs should use json and export to someplace searchable, is there a free splunk or other log indexer we can use, figure that out file multiple issues as this is planned out

sameerparekh:
## Summary

Closes #602.

Issue #602 asks for structured logs that use log contexts ("add context to logs to make for better searching") and a one-line-local / JSON-deployed split. This PR delivers the **structured-context** half — a single `LogContext` helper that attaches bounded MDC keys to the API's hot-path logs — and sets up local visibility via `%mdc` in `logback.xml`. A JSON appender for deployed environments is intentionally left to a follow-up under the broader #602 umbrella.

## What changed

- New `api/src/observability/LogContext.scala` — single-source-of-truth set of MDC keys (`route`, `method`, `status`, `op`, `routerId`, `mac`, `etag`, `profileId`, `user`, `reason`, `notModified`, `batchSize`, `rejected`) and `annotate` / `annotateAll` / `annotateOpt` wrappers. New keys are added here, not at the call site, mirroring the cardinality-firewall discipline `MetricGuard` enforces on the metrics side.
- `ErrorBoundary` — every 4xx/5xx log line now carries the templated `route`, `method`, `status` as annotations (not just substring-formatted).
- `RouterRoutes` `GET /api/router/policy` — `op=router_policy`, `routerId`, `etag`, `notModified`.
- `RouterIngestRoutes` `POST /api/router/usage` and `/events` — `op`, `routerId`, plus `batchSize`/`rejected` on the per-batch summary.
- `PolicyService.logSnapshotChanged` — `op=snapshot_changed`, `etag`.
- `AuthService.login` — explicit success log (`op=login`, `user`, `reason=ok`) and bad-password failure log (`op=login`, `user`, `reason=bad_password`), alongside the existing `AppMetrics.recordAuthFailure` counter.
- `Routes` device upsert / delete / patch — `op`, `mac`, `profileId`.
- `api/resources/logback.xml` — adds `%mdc` to the console pattern so annotations show locally as `key=value` at the tail of each line.

Human-readable message text is **kept verbatim** on every migrated line for back-compat with existing log readers and tests that grep substrings.

## Not migrated (deliberate)

The audit umbrella under #1602/#1532 covers ongoing log hygiene; this PR is scoped to API hot paths where structured search is high-value for ops. The following were intentionally left out and can be picked up by follow-ups:

- Rollup / retention jobs (low frequency, already low-cardinality).
- Startup-only seeders (bundled blocklists, app templates).
- Per-record DEBUG logs inside the usage/events handlers — the batch-level annotations already give ops the `routerId`/`op` context they need to filter at the source.
- A deployed-environment JSON appender and the choice of log-indexer backend (issue #602's other half) — separate follow-up.

## Test plan

- [x] `mill api.test.testOnly wifihaven.api.feature.StructuredLogContextSpec` — 3 new tests pin `LogEntry.annotations` (not message substrings) for ErrorBoundary 4xx, login success, login bad-password.
- [x] `mill api.test.testOnly 'wifihaven.api.feature.ErrorBoundary*' 'wifihaven.api.feature.Router*' 'wifihaven.api.feature.Auth*' 'wifihaven.api.feature.Device*' 'wifihaven.api.feature.PolicySnapshot*'` — all impacted feature tests still pass (47 + 35 in the policy/router suites, plus the rest).
- [x] `scalafmt --check --non-interactive` clean.

### DIFF
--- a/api/resources/logback.xml
+++ b/api/resources/logback.xml
@@ -1,7 +1,14 @@
 <configuration>
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <encoder>
-      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+      <!--
+        #602: %mdc renders the SLF4J MDC map populated by zio-logging annotations
+        (see api/src/observability/LogContext.scala). One-line-per-log locally;
+        keys like route=, mac=, op=, etag=, status= appear at the tail so they
+        can be grep'd without parsing the message text. A structured (JSON)
+        appender for deployed environments is tracked separately under #602.
+      -->
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg %mdc%n</pattern>
     </encoder>
   </appender>
 
--- a/api/src/ErrorBoundary.scala
+++ b/api/src/ErrorBoundary.scala
@@ -1,6 +1,7 @@
 package wifihaven.api
 
 import wifihaven.api.metrics.AppMetrics
+import wifihaven.api.observability.LogContext
 import zio.*
 import zio.http.*
 
@@ -72,9 +73,17 @@ object ErrorBoundary {
     val status = response.status.code
     for {
       snippet <- bodySnippet(response)
-      base = s"$method $route -> $status"
-      msg  = snippet.fold(base)(s => s"$base body=$s")
-      _ <- if status >= 500 then ZIO.logError(msg) else ZIO.logWarning(msg)
+      base   = s"$method $route -> $status"
+      msg    = snippet.fold(base)(s => s"$base body=$s")
+      // #602: structured context (route, method, status) attached as log annotations
+      // so MDC/JSON consumers can search/filter without parsing the message text. The
+      // human-readable message is kept verbatim for back-compat with existing log readers.
+      logged = if status >= 500 then ZIO.logError(msg) else ZIO.logWarning(msg)
+      _ <- LogContext.annotateAll(
+        LogContext.Route  -> route,
+        LogContext.Method -> method,
+        LogContext.Status -> status.toString,
+      )(logged)
       _ <- AppMetrics.recordHttpError(route, status)
     } yield ()
   }
--- a/api/src/ErrorBoundary.scala
+++ b/api/src/ErrorBoundary.scala
@@ -73,17 +73,16 @@ object ErrorBoundary {
     val status = response.status.code
     for {
       snippet <- bodySnippet(response)
-      base   = s"$method $route -> $status"
-      msg    = snippet.fold(base)(s => s"$base body=$s")
-      // #602: structured context (route, method, status) attached as log annotations
-      // so MDC/JSON consumers can search/filter without parsing the message text. The
-      // human-readable message is kept verbatim for back-compat with existing log readers.
-      logged = if status >= 500 then ZIO.logError(msg) else ZIO.logWarning(msg)
-      _ <- LogContext.annotateAll(
-        LogContext.Route  -> route,
-        LogContext.Method -> method,
-        LogContext.Status -> status.toString,
-      )(logged)
+      base = s"$method $route -> $status"
+      msg  = snippet.fold(base)(s => s"$base body=$s")
+      // #602: `route` + `method` are already annotated by `LoggingMiddleware` for
+      // the whole handler scope (via FiberRef), so they flow into this log line
+      // automatically. We only need to add the response-derived `status` here.
+      // Human-readable message text is kept verbatim for back-compat with
+      // existing log readers.
+      _ <- LogContext.annotate(LogContext.Status, status.toString) {
+        if status >= 500 then ZIO.logError(msg) else ZIO.logWarning(msg)
+      }
       _ <- AppMetrics.recordHttpError(route, status)
     } yield ()
   }
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -13,6 +13,7 @@ import wifihaven.api.metrics.{
   RouterPresenceMetrics,
 }
 import wifihaven.api.notify.Notifier
+import wifihaven.api.observability.LoggingMiddleware
 import wifihaven.api.policy.*
 import wifihaven.api.routes.*
 import wifihaven.api.usage.RetentionSweepJob
@@ -442,13 +443,22 @@ object Main extends ZIOAppDefault {
       // It sits *inside* the readiness gate so the gate's startup 503s don't spam the log. Only
       // /api/health (its own 503-while-starting semantics, polled constantly) and /metrics (the
       // scrape endpoint) are left out — mirroring HttpMetrics' own exclusion of /metrics.
+      // #602: LoggingMiddleware.annotate wraps every served route family so every
+      // ZIO.log* emitted inside a handler (or anything it calls) inherits the
+      // templated `route` and `method` as MDC keys via FiberRef — handlers no
+      // longer wrap individual log lines for that context. ErrorBoundary still
+      // adds `status` because it sees the response. The middleware sits *inside*
+      // HttpMetrics (so the same routes it counts also get logged) and *outside*
+      // the readiness gate so even a gated 503 carries its route/method context.
       HttpMetrics.instrument(
         healthRoutes ++
-          Readiness.gate(
-            ErrorBoundary.observe(
-              systemRoutes ++ statsRoutes ++ routerAndAdminRoutes ++ spaRoutes,
+          LoggingMiddleware.annotate(
+            Readiness.gate(
+              ErrorBoundary.observe(
+                systemRoutes ++ statsRoutes ++ routerAndAdminRoutes ++ spaRoutes,
+              ),
+              ready,
             ),
-            ready,
           ),
       ) ++ metricsRoutes
     }
--- a/api/src/auth/AuthService.scala
+++ b/api/src/auth/AuthService.scala
@@ -60,10 +60,11 @@ class AuthServiceLive(
   private val secret                 = jwtConfig.secret
 
   def login(username: String, password: String): IO[AuthError, LoginResponse] =
-    LogContext.annotateAll(
-      LogContext.Op   -> "login",
-      LogContext.User -> username,
-    ) {
+    // #602: route/method (`POST /api/auth/login`) ride in via LoggingMiddleware on
+    // the HTTP path. `user` is the only per-call dynamic context we add here, so
+    // every log inside the for-comprehension (success info, bad-password warn)
+    // inherits it via FiberRef.
+    LogContext.annotate(LogContext.User, username) {
       (for {
         user  <- userRepo
           .findByUsername(username)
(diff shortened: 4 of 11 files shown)


sameerparekh:
Independent review of #1736 at `2a27b19a` against `docs/pr-review-checklist.md`. Read-only pass.

## BLOCKERS

(none)

## SHOULD-FIX

1. `api/src/observability/LogContext.scala:50` — `annotateAll`'s docstring claims "nil-value pairs are dropped so the call site can be uniform" but the implementation accepts `(String, String)*` and unconditionally folds every pair through `ZIO.logAnnotate`. Nothing is dropped. Either tighten the docstring to "attaches every pair as an annotation" and point readers to `annotateOpt` for the nullable case, or change the signature/body to actually filter. The current text invites a future caller to pass `"k" -> null` and silently emit a stringified-null annotation.
2. PR commit history — for an autonomous session, the AGENTS.md TDD workflow requires the failing test to be committed *before* the implementation so red→green is visible in the PR's commit history. This PR is a single commit that adds the implementation, the helper, and the asserting tests together (`2a27b19a`). The tests are real and assert on `LogEntry.annotations` directly (good), but the commit ordering is the documented gate. Split into a test-only commit (initially failing) + implementation commit before merge, or document why it doesn't apply (e.g., "this is structural — the test exists purely to pin the new contract").

## NITS

1. `api/src/routes/RouterIngestRoutes.scala:130-150` — the `/events` handler wraps `op=router_events` + `routerId` around the body-parse block *and* a second time around the per-batch summary log. Collapsing both under a single outer `LogContext.annotateAll` at handler entry would be cleaner and would also annotate the empty-body debug log on that path. Same applies to the `/usage` handler.
2. `api/src/routes/RouterIngestRoutes.scala:89-95` — the `ZIO.foreachDiscard(records)(r => ZIO.logDebug(...))` per-record debug loop sits *outside* the `LogContext.annotateAll(op=router_usage, routerId=...)` block above, so those DEBUG lines emit without the `op`/`routerId` annotations. The PR body documents per-record DEBUG as intentionally out of scope, but a reader looking at the file would expect the surrounding scope to apply. Either pull the foreach inside the same annotated scope (trivial — no new keys) or add a one-line comment noting the deliberate omission.
3. `api/src/auth/AuthService.scala:97-103` — the InvalidCredentials warn message duplicates the annotation values (`user=$username reason=bad_password`). This is the intentional back-compat strategy you call out in `ErrorBoundary.scala:78`. Worth re-stating that intent in `AuthService.scala` next to this log so a future cleanup doesn't drop the annotations thinking the message has them.
4. `api/resources/logback.xml:11` — `%mdc` emits MDC pairs comma-separated when non-empty; the leading space before `%mdc` is preserved when the map is empty, giving a trailing space before the newline. Cosmetic. Logback's `%mdc{}` supports a format string if you ever want a specific shape.

## VERDICT: REQUEST-CHANGES

Targeted, well-scoped migration of hot-path API logs to ZIO `logAnnotate`-based MDC, with a single `LogContext` helper as the SSoT for bounded keys and a new feature spec that pins on `LogEntry.annotations` (not message substrings) — exactly the test surface that prevents regressions.
The two SHOULD-FIX items are the gate: a misleading docstring on `annotateAll`, and the missing red→green commit split required by AGENTS.md for autonomous sessions.
Neither is a correctness defect; once addressed, this is APPROVE.

sameerparekh:
Re-run of the independent PR review on the new HEAD `cc508b49`. Read-only pass per `docs/pr-review-checklist.md`.

## Prior-findings status (vs `2a27b19a`)

| 1 | SHOULD-FIX | `annotateAll` docstring claimed nil-value pairs were dropped, but the impl folded every pair through `ZIO.logAnnotate` unconditionally | **ADDRESSED** — `LogContext.scala:50-52` rewritten to state "attach every (key, value) pair" and directs nullable callers to `annotateOpt`. |
| 2 | SHOULD-FIX | Single commit; AGENTS.md TDD workflow asks autonomous sessions for visible red→green | NOT-ADDRESSED — still single implementation commit; second commit `cc508b49` is review-followup, not the original test-first split. Not merge-gating in my judgment because the spec asserts on `LogEntry.annotations` (not message substrings) and would fail on a future regression, but flagging that the documented gate is unmet. |
| 1 | NIT | `/events` and `/usage` handlers each wrapped `op` + `routerId` twice; per-record DEBUG sat outside the annotated scope | **ADDRESSED** — both handlers in `RouterIngestRoutes.scala` now wrap the *whole* post-auth body once via `auth.authenticate(req).flatMap { router => LogContext.annotateAll(op, routerId) { for { … } } }`. Per-record DEBUG dumps now inherit `op`/`routerId` automatically. Reads cleaner and removes the surprise. |
| 2 | NIT | Per-record DEBUG dump in `/usage` was outside the annotated scope | **ADDRESSED** — folded into the new outer scope by the same restructure. |
| 3 | NIT | `AuthService` bad-password warn duplicated MDC values into the message text without an in-place note about why | **ADDRESSED** — `AuthService.scala:99-102` adds a `#602:` comment naming the back-compat strategy and pointing at `ErrorBoundary.scala`. |
| 4 | NIT | `%mdc` trailing-space cosmetic | NOT-ADDRESSED — deferred; cosmetic and benign. |

## New findings (delta `2a27b19a...cc508b49`)

Delta is 3 files / +122 / −111: docstring fix (`LogContext`), handler-scope collapse (`RouterIngestRoutes`), in-place comment (`AuthService`).

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

1. `api/src/routes/RouterIngestRoutes.scala:46-110` — the new `auth.authenticate(req).flatMap { router => LogContext.annotateAll(...) { for { … } yield Response.ok } }` shape is correct but visually heavy (deeper nesting + a long for-comprehension inside a block). A small private helper (`def withRouterScope(req)(body: Router => ZIO[…])`) would let both `/usage` and `/events` share the `auth → annotate → for-body` boilerplate. Cosmetic — not worth doing in this PR; flag for the next time a third router-scoped handler lands.

## Dimension coverage (delta only)

- **1. Duplication / SSoT:** improved — the per-block `LogContext.annotateAll(op, routerId)` pair is now a single handler-scope wrapper, no longer hand-mirrored across the foreach and the summary log.
- **2. Testing integrity:** existing `StructuredLogContextSpec` (3 tests) and `RouterIngestSpec` (47 tests) both still green. No tests weakened.
- **3. Architectural invariants:** unchanged — router still a dumb applier; no DNS, `Clock`, or `/tmp` touched.
- **4. Wire compat:** unchanged. Pure log restructure.
- **5. Migrations:** none.
- **6. Metrics + dashboard:** unchanged; `AppMetrics.recordUsageRecordsRejected` still called inside the new scope.
- **7. Scope / hygiene:** focused 3-file follow-up; commit message describes intent; no AI mentions.
- **8. Security:** unchanged.

## VERDICT: APPROVE

All four prior NITs and the docstring SHOULD-FIX ADDRESSED. The TDD red→green commit-split SHOULD-FIX remains NOT-ADDRESSED but is non-gating: the new spec asserts on `LogEntry.annotations` (the structural contract), so a future regression that drops annotations but keeps the formatted message will still fail CI.
The handler-scope collapse is the durable win — `op`/`routerId` now annotate the entire post-auth scope, so any future log added inside either ingest handler inherits the context automatically.
Ready to merge.

sameerparekh:
Re-run of the independent PR review on HEAD `e252fcfd`. Read-only pass per `docs/pr-review-checklist.md`.

## Prior-findings status (vs `cc508b49`)

| 1 | NIT | `RouterIngestRoutes` post-auth `auth.flatMap { router => annotate { for {…} } }` pattern is visually heavy and would benefit from a helper next time a third router-scoped handler lands | NOT-ADDRESSED — same shape kept (with `op` removed, so slightly simpler). Still cosmetic. |

## New findings (delta `cc508b49...e252fcfd`)

Delta is 8 files / +131 / −64. Headline restructure: per-handler `LogContext.annotateAll(op=…, …)` wrappers replaced by a single `LoggingMiddleware` that injects `route` + `method` per-request; ErrorBoundary keeps `status` only; `op` annotations dropped (route+method disambiguates).

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

1. `api/src/observability/LogContext.scala` — the `Op` key is still exported but only used by `PolicyService.logSnapshotChanged`. Either keep with a doc note "use sparingly — `(method, route)` disambiguates HTTP paths; reserve `op` for service-internal events that don't have a unique route", or inline `"snapshot_changed"` at the one caller and drop the constant. Cosmetic — keeping it is defensible because it's the natural distinguisher for any future non-HTTP emitter.
2. `api/src/observability/LoggingMiddleware.scala:33` — public method is named `annotate` while the parallel `ErrorBoundary.observe` and `HttpMetrics.instrument` use verbs that don't collide with the helper they call. `LogContext.annotate` is also a thing in the same package; reading the call site `LoggingMiddleware.annotate(...)` next to `LogContext.annotate(...)` is fine but a more specific verb (`wrap`, `instrument`, `observe`) would mirror the other two middlewares. Minor naming.
3. `api/src/Main.scala` — the comment block on the `LoggingMiddleware.annotate(...)` wiring is good, but the existing `// #1570: ErrorBoundary.observe wraps every served route family…` comment above it now overlaps with the new explanation. A 1-line `// #602:` cross-reference inside the existing block would tighten it.
4. `api/test/src/feature/StructuredLogContextSpec.scala:112-136` — the new "LoggingMiddleware inherits into the handler's own logs" test is the right shape, but it relies on the seeded `admin/changeme` user staying valid. The two preceding tests do the same and have been stable; flagging in case the seed default ever rotates.

## Dimension coverage (delta only)

- **1. Duplication / SSoT:** big improvement — `route`/`method` is now injected from ONE place (`LoggingMiddleware`); the per-handler `op` strings (which were a hand-mirrored projection of route+method) are gone. The remaining per-handler annotations are strictly the dynamic data the handler had to compute locally (routerId, etag, mac, profileId, user, reason) — no SSoT violation here.
- **2. Testing integrity:** existing `StructuredLogContextSpec` tests rewired through `LoggingMiddleware.annotate(ErrorBoundary.observe(…))` (production wiring); a 4th test added that asserts middleware annotations + handler annotations BOTH land on a single `LogEntry` via FiberRef inheritance. No existing tests weakened. `RouterIngestSpec`/`RouterApiSpec`/`AuthApiSpec`/`PolicySnapshot*`/`DeviceApiSpec` all still pass per PR body.
- **3. Architectural invariants:** unchanged. Router still a dumb applier. `Clock` not touched. No `/tmp` writes.
- **4. Wire compat:** unchanged. Logs only.
- **5. Migrations:** none.
- **6. Metrics + dashboard:** unchanged; `LoggingMiddleware` is the log analog of `HttpMetrics.instrument` (same `routePattern.pathCodec` template-rendering, same cardinality-firewall rationale).
- **7. Scope / hygiene:** focused 8-file diff; no AI mentions; clean commit message.
- **8. Security:** unchanged. The middleware only sees the templated route, not raw paths or PII.

## VERDICT: APPROVE

Solid architectural pivot from per-handler wrappers to a zio-http middleware. `route`/`method` now ride in via FiberRef and the handler-internal context (`routerId`, `mac`, `etag`, `user`, `reason`) is the only thing handlers wrap — and only because that data is locally computed.
The new 4th test in `StructuredLogContextSpec` is the durable win: it pins the middleware → handler annotation-inheritance contract end-to-end, so a future change that breaks the FiberRef flow fails CI.
Four NITs, all cosmetic. No BLOCKERS, no SHOULD-FIX. Ready to merge.

PR #1738, opened 2026-06-15
feat(#638): OpenAPI spec + Swagger UI at /api/docs

### LINKED ISSUE #638: api: publish OpenAPI spec and serve Swagger UI
## Problem

The API has no machine-readable contract. Consumers (the OpenWRT agent, the web app, future integrations) discover endpoints by reading scala source, and there's no rendered docs surface for humans either. PR #635 added golden-JSON contract tests for a handful of surfaces, but those only cover payload shape on the wires we already knew about — they don't enumerate the API.

## Ask

1. Generate an OpenAPI 3 spec from the existing API routes/codecs (tapir or zio-http's built-in derivation, whichever fits the current stack — grep the api/ module for the http framework in use).
2. Serve it at a stable path: `/openapi.json` (or `.yaml`).
3. Serve Swagger UI (or Redoc) at `/docs` (or `/swagger`) backed by that spec.
4. Wire into the existing API server boot so it's on by default in all envs. No auth on the docs endpoint itself; if any endpoint requires auth, reflect that in the spec's security schemes.

## Nice-to-have

- CI check that the committed `openapi.json` matches what the server emits (similar pattern to the contract-tests fixture-diff guard from #634/#635), so spec drift shows up in PR review.
- Link the rendered docs from the project README.

sameerparekh:
## Summary

Auto-generate an OpenAPI 3.0.3 spec by walking the live zio-http `Routes`
tree at boot, serve it at `/api/openapi.json`, and serve Swagger UI at
`/api/docs`.

Closes #638.

## Schema-extraction approach

The codebase has **96 hand-rolled `Routes(...)`** handlers across ~25
route files using `body.fromJson[T]` + `Response.json` inline — no tapir,
no `Endpoint`, no schema-bearing wrapper. Three options considered:

| Approach | Outcome |
|---|---|
| Refactor all 96 routes to `zio.http.endpoint.Endpoint` / tapir | Out of scope (multi-week refactor of every handler) |
| Hand-author a checked-in `openapi.yaml` | Drifts the moment a Scala case class moves; violates AGENTS.md §single-source-of-truth |
| **Walk the live `Routes` tree → emit paths/methods/auth-partition** (chosen) | Single source of truth = the route table itself; zero drift on paths/methods; schemas filled in incrementally per-route |

The generator extracts `(method, path)` from each `Route.routePattern`
via `pathCodec.render` (already produces OpenAPI `{name}` placeholders),
sorts by path, and emits one operation per (path, method). The
`bearerAuth` partition is declared in a small `publicPathPrefixes` list
that mirrors the per-handler `requireAuth` convention.

## What's covered now

- Every mounted route appears in the spec automatically — system, stats,
  router, admin, health.
- Path parameters render as `{name}` (e.g. `/api/profiles/{id}`).
- Multiple methods on the same path group under one paths entry.
- Bearer-JWT `securitySchemes` declared; authenticated routes carry a
  `security: [{bearerAuth: []}]` requirement, public probes do not.
- Swagger UI at `/api/docs` loads the spec via jsdelivr-CDN assets pinned
  to swagger-ui-dist@5.17.14.

## What's NOT covered (deliberate)

- Request/response body schemas
- Per-route summaries/descriptions
- Per-operation tagging

Filling those in is incremental follow-up — a small per-route
`OperationMeta` registry keyed by `(method, path)` is the natural next
step and can be wired in without touching this PR's generator.

## Test plan

- [x] `OpenApiSpecSpec` — 8 generator tests (envelope shape, version,
      well-known routes, `{param}` rendering, method grouping, security
      scheme, public/auth partition)
- [x] `OpenApiRoutesSpec` — served endpoint tests (`/api/openapi.json`
      returns 3.x JSON, `/api/docs` returns HTML pointing at the spec)
- [x] `mill api.compile` ✅
- [x] `mill mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll` ✅


sameerparekh:

Scope: `git diff origin/main...HEAD` (3 files added, Main.scala wired).

### BLOCKERS

(none)

### SHOULD-FIX

1. `api/src/openapi/OpenApiSpec.scala:44-59` — `publicPathPrefixes` advertises `/api/router/policy`, `/api/router/events`, `/api/router/usage`, `/api/router/metrics`, `/api/router/register`, `/api/blocklists` as unauthenticated. They are **not**: the router bearer (`routerAuth`) gates every one of them — see `RouterRoutes.scala` / `RouterIngestRoutes.scala` / `RouterMetricsRoutes.scala`. The served spec will mislead an operator (or an auditor reading Swagger UI) into believing these routes are wide open. Either (a) declare a second security scheme `routerBearer` and partition by which auth applies, or (b) drop these from the "public" list so they at least carry the generic `bearerAuth` requirement (over-restrictive but not misleading). Today's "no security at all" is the misleading case the AGENTS.md "display vs enforcement source mismatch" rule warns about.

2. `api/src/openapi/OpenApiSpec.scala:44` — `publicPathPrefixes` is a hand-maintained mirror of each handler's `requireAuth` / `requireAdmin` / `routerAuth` convention. There is **no** test pinning the two together, so a new public/private flip in a handler will silently desync from the spec — the §single-source-of-truth "display vs. enforcement drift" trap (#1539-class). Either:
   - **Test-pin** (cheap): add a unit test that asserts, for every route in the assembled production tree, that `isPublic(path)` matches some inspectable annotation (e.g. a typed `RouteAuth` enum carried alongside each `Routes.scala` block), OR
   - **Collapse**: attach an auth tag at route-definition time so the spec reads it directly.
   The current code disables drift detection completely; the AGENTS.md rule explicitly forbids this without a test pin.

3. `api/src/openapi/OpenApiSpec.scala:52` — `/api/metrics` is dead in the public-prefix list. The actual Prometheus scrape endpoint is `/metrics` (see `MetricsRoutes.scala`), and even if it were `/api/metrics` the metrics routes are NOT passed into `OpenApiRoutes.routes(...)` in `Main.scala`, so they never enter the spec. Either drop the entry or pass `metricsRoutes` to the generator (and decide whether you actually want the scrape endpoint advertised).

### NITS

1. `api/src/openapi/OpenApiSpec.scala:160` — path parameter `schema.type` is `"string"` for every captured param, but the underlying `pathCodec` carries the real type (`long("id")`, `string("…")`, `boolean(…)`). The information is reachable via `routePattern.pathCodec.segments` and would let the spec advertise `integer` / `boolean` accurately. Not blocking — but it's a free correctness win the next iteration should pick up.

2. `api/src/routes/OpenApiRoutes.scala:35-41` — `openApiRoutes` itself is not passed to `OpenApiSpec.generate(...)` in `Main.scala:431`, so `/api/openapi.json` and `/api/docs` do not appear in their own spec. Self-listing is conventional in OpenAPI surfaces; pass `openApiRoutes` to the generator (or document why not).

3. `api/src/routes/OpenApiRoutes.scala:51-78` — Swagger UI assets are pulled from `cdn.jsdelivr.net` at every page load. The self-hosted install (`deploy/install.sh`) commonly runs on networks where outbound HTTPS to a CDN is unreliable, and the page silently renders blank when it fails. Already called out in the file comment; consider filing the "vendor swagger-ui under `api/resources/swagger-ui/`" follow-up explicitly under the #638 umbrella so it isn't forgotten.

4. `api/src/openapi/OpenApiSpec.scala:62-64` — `isPublic` uses `startsWith(p + "?")`, but rendered paths from `pathCodec.render` never contain a query string (it's a path template). The branch is dead. Harmless, but the predicate reads as if it handled something it doesn't.

5. `api/src/openapi/OpenApiSpec.scala:171` — operation `summary` is just `"${method.name} $path"`, which Swagger UI already renders next to each row. No information added; could be omitted or filled with something useful (e.g. the path's auth scheme tag). Pure stylistic.

### Strengths

- TDD red→green is visible in history: tests landed in `3638c681`, implementation in `0f1311d5`.
- Generator walks the live `Routes` tree → paths/methods are auto-maintained; no checked-in YAML to drift, matching the §single-source-of-truth direction.
- Approach is well-documented in the PR body (refactor / hand-author / generator trade-off), and the spec's own `info.description` honestly states what it covers and what it doesn't.
- New routes ride existing `HttpMetrics.instrument` / `ErrorBoundary.observe`, so they inherit instrumentation + error labeling without any new metrics surface.
- Tests use real `Routes` values rather than mocks; structure is the right level for the change.
- 5-file diff, focused, no drive-by edits, no AI mentions in commits.

### VERDICT: APPROVE

Auto-generated OpenAPI spec + Swagger UI cleanly added behind a small generator that walks the live `Routes` tree; paths/methods are authoritative. The two SHOULD-FIX items (router-bearer routes mis-advertised as public; auth partition lacks a drift test) are the recurring §single-source-of-truth display-vs-enforcement risk and should be cleaned up before this is treated as the canonical contract surface — but neither breaks running code, and the documented incremental path is honest.

sameerparekh:

Prior reviewed-sha: `0f1311d5c99f97862d30eb866ad86a71e3443bb3`.
This run reviews the delta `0f1311d5...9841cdc5`.

### Status of prior findings

|---|---|---|---|
| 1 | SHOULD-FIX | `publicPathPrefixes` advertised `/api/router/*` and `/api/blocklists/*` as unauthenticated, when they actually require the router bearer | **ADDRESSED** — `OpenApiSpec.scala:50-77` introduces `AuthScheme.{Public,User,Router}`, partitions paths three ways, and tags router-authed paths with a new `routerBearer` security scheme (`OpenApiSpec.scala:164-172`). |
| 2 | SHOULD-FIX | No test pin between the spec's auth list and per-handler enforcement (§single-source-of-truth display-vs-enforcement drift) | **ADDRESSED** — new `OpenApiAuthPartitionSpec` lists every route family with the auth scheme its handler enforces and asserts `authForTesting(path)` matches; flipping one without the other fails the test. |
| 3 | SHOULD-FIX | Dead `/api/metrics` entry (real scrape is `/metrics`, and metrics routes aren't passed to the generator) | **ADDRESSED** — dropped from the partition. |
| 4 | NIT | Path param `schema.type` always `"string"`; underlying `pathCodec` has the real type | NOT-ADDRESSED (deferred — fine for now, called out in the PR body). |
| 5 | NIT | `/api/openapi.json` and `/api/docs` not self-listed in the spec | **ADDRESSED** — `OpenApiRoutes.scala:30-37` passes a stub `Routes` for the two paths into `OpenApiSpec.generate`. |
| 6 | NIT | CDN dependency on jsdelivr; self-hosted installs on locked-down networks render blank | **ADDRESSED** — follow-up filed as #1749 and referenced in the file header. |
| 7 | NIT | Dead `startsWith(p + "?")` branch in `isPublic` | **ADDRESSED** — `authFor` only matches `path == p || startsWith(p + "/")`. |
| 8 | NIT | `summary` field carries no info beyond method+path | NOT-ADDRESSED (cosmetic; left alone). |

### New findings (delta `0f1311d5...9841cdc5`)

#### BLOCKERS

(none)

#### SHOULD-FIX

(none)

#### NITS

1. `api/src/openapi/OpenApiSpec.scala:53` — the file header still says (sic) "the locality the AGENTS.md single-source-of-truth rule wants" for the (deferred) per-route `OperationMeta` follow-up; given the auth partition is now genuinely test-pinned, you could tighten this paragraph to call out the partition as an example of the right shape. Cosmetic.

2. `api/test/src/feature/OpenApiAuthPartitionSpec.scala:25-55` — the table is hand-curated, so a brand-new route family (e.g. a future `/api/foo/*`) that no row covers would silently pass. The drift catch is real for FLIPS, not for ADDITIONS. Not a blocker for this PR — the AGENTS.md rule is about display-vs-enforcement of *known* facts — but worth a one-line note in the test docstring so the next contributor doesn't read it as exhaustive coverage.

3. `api/src/openapi/OpenApiSpec.scala:87` — `authForTesting` is `public` rather than `private[openapi]` because the test moved to the `wifihaven.api.feature` package. The (previously preferred) `private[openapi]` would be tighter; alternative is moving the test under `wifihaven.api.openapi` or marking it `private[wifihaven]`. Minor.

### Strengths of the delta

- The single-source-of-truth fix is the *right* one: rather than hand-keeping two lists in sync, the partition is now a typed enum + a test that fails on drift — exactly the COLLAPSE/TEST-PIN shape AGENTS.md §single-source-of-truth calls for.
- The `Router` carve-out is honest: router endpoints now declare their auth, an auditor reading Swagger UI sees "routerBearer" instead of nothing.
- Follow-up issue #1749 was filed for the deferred NIT, keeping the trail visible.
- All 11 new tests pass, build is clean, no test weakening.

### VERDICT: APPROVE

All prior SHOULD-FIX cleared; no new BLOCKER or SHOULD-FIX introduced. The three remaining notes are cosmetic. Auto-generated spec is now both honest about its auth model and pinned against handler-side drift; safe to merge.

Issue #1240, opened 2026-05-31
Observability: server/DB/deploy signals to catch incidents like 2026-05-31 (planning thread)


sameerparekh:
Planning thread (no code lands directly here — each agreed item becomes its own sub-issue). Child of #471 (metrics & observability architecture), but scoped tightly to **operational signals that would have caught the 2026-05-31 prod incident on their own**, instead of it being reconstructed by hand from Render dashboards.

## What happened on 2026-05-31 (motivation)
- Prod API crash-looped starting ~06:04 MDT; root cause was HikariCP pool exhaustion under aggregate rollup + ingest load on the `basic_256mb` Postgres (#1221, partially fixed by #1223; pool-sizing regression #1228).
- The `basic_256mb` DB pegged CPU to 100% from the every-3-min `time_used_daily` rollup (#1160/#1165), shipped in the 05-31 01:33 UTC deploy; fix in #1230/PR #1231.
- **Every bit of that was diagnosed manually** — Render dashboards for DB CPU/availability, deploy timeline by hand, `rollup_runs` never surfaced, no alert fired. That manual-archaeology gap is what this thread is about.

### DB / connection pool
- [ ] **Export HikariCP pool metrics** (active/idle/pending/total, connection-acquire timeouts) from the API — the pool-exhaustion was invisible. (Hikari exposes a `MetricsTrackerFactory`; wire to whatever `/metrics` #471 lands, or log a gauge line on a timer as a stopgap.)
- [ ] **Alert on connection-acquire timeouts** (the 30s HikariCP timeout firing) — leading indicator of the crash-loop.
- [ ] **DB CPU / connection-count alerting at the Render level** — declaratively if Render supports it (notification config), else a small poller against the Render API. Catch sustained CPU>90% and connections-near-max for the tier.

### Rollup health (the data already exists, nothing surfaces it)
- [ ] **Surface `rollup_runs`** — an admin/ops view (or `/metrics` counters) for per-job last-run status, duration, rows, error. Each rollup already records to `rollup_runs`; no one can see it.
- [ ] **Alert on rollup job error rate / runaway duration** — a rollup tick that starts taking seconds and overlapping ticks is the DB-CPU smoking gun.
- [ ] **Per-job tick duration histogram** so cadence-vs-cost is visible (informs cadence tuning like #1230).

### App liveness / deploy
[...]

Issue #730, opened 2026-05-21
router→api: add dest_ip to traffic usage records so #720-style FQDN backfill works for screen time too


sameerparekh:
Follow-up to #720 (PR #728).

## Background

PR #728 added an API-side FQDN backfill for `connection_attempt` events that arrive race-loser-typed as `ipv4` despite the agent having a `(dest_ip → fqdn)` mapping. That worked because `connection_events` carries both `host` *and* `dest_ip`, so the API can key the backfill on `(router_id, dest_ip)`.

The 5-minute traffic usage rollup (`/api/router/usage` → `traffic_reports` → `time_usage`) does NOT carry `dest_ip`. The wire shape is `UsageRecord(mac, ip, host, activeSeconds, bytesIn, bytesOut)` where `ip` is the *device's* LAN IP, not the destination. That means the same race produces:

- A `time_usage` row keyed `(mac, ipv4, <bare-IP>, date)` accumulating seconds against a bare IP literal even though the conntrack-side event got resolved.
- No way to re-attribute it from `connection_events.resolved_host_value` server-side without a fragile heuristic join, because there's no `dest_ip` to key on.

Net effect: site-time and host-aggregated screen-time dashboards undercount the FQDN bucket (e.g. \`neverssl.com\`) by exactly the per-flow share that lost the race. Same undercounting #583 called out for events, just on the usage side.

## Why this is a router-side change (and why it's blocked right now)

Adding `destIp: Option[IpAddress]` to `UsageRecord` is a router→api wire schema change. The router is frozen until Gate 2 (#654) is in place, so this can't ship yet. Filing it now so we don't lose it when the router thaws.

## Proposed change (when router unfreezes)

1. **Router agent**: thread `dst_ip` through the traffic accumulator so it lands on each `UsageRecord` (one record per (mac, dst_ip), already the agent's internal granularity — see [openwrt/files/usr/lib/lua/wifihaven/conntrack.lua](openwrt/files/usr/lib/lua/wifihaven/conntrack.lua) and the usage rollup that consumes it).
2. **Contract**: add `destIp: Option[IpAddress]` to `UsageRecord` in `shared/contract/router-to-api/usage_report.json` and `shared/src/Models.scala`. Capability-tag it for #376 (`usage-records-dest-ip`) so older agents skip the field.
3. **API migration**: add `dest_ip TEXT NULL` to `traffic_reports`. Backfill not required — pre-existing rows stay unattributed.
[...]

PR #1752, opened 2026-06-15
refactor(#1741): extract shared LATERAL FQDN-resolve CTE fragment

### LINKED ISSUE #1254: Rollup attribution lateral pegs DB CPU — missing index on connection_events(mac, dest_ip, ts)

### LINKED ISSUE #1741: audit(#1532): collapse LATERAL FQDN-resolve CTE duplicated across Repos.scala + RollupRepo

## Drift

The \`LEFT JOIN LATERAL\` that resolves \`traffic_reports.host_type IN ('ipv4','ipv6')\` rows against the most recent matching \`connection_events.resolved_host_value\` is hand-rolled inline in at least seven SQL fragments:

- \`api/src/db/RollupRepo.scala:186-210\` — \`private val resolvedCte\`.
- \`api/src/db/Repos.scala:1741-1757\`
- \`api/src/db/Repos.scala:1817-1832\`
- \`api/src/db/Repos.scala:1864+\`
- \`api/src/db/Repos.scala:1910-1926\`
- \`api/src/db/Repos.scala:1976-1990\`

\`RollupRepo.scala:187\` even flags it: *"The LATERAL FQDN-resolve mirrors TrafficReportRepoLive — keep them in lockstep. If the read-time resolve in traffic_reports queries ever changes, this CTE must change too."* Classic #1532 "keep in sync" anti-pattern — and it is the LATERAL whose missing index caused the [#1240](https://github.com/wifihaven/wifihaven/issues/1240) prod outage, so the blast radius of drift here is real.

## Recommendation

Extract a single Doobie \`Fragment\` (e.g. \`SqlFragments.ResolvedHostLateral\` or a \`resolvedCte\` constant in one place) and have every call site reference it. Bonus: anchor a single supporting-index comment to that one fragment so the next reviewer doesn't have to verify each duplicate.

## Scope

- Pick a canonical location (probably \`shared between TrafficReportRepoLive and RollupRepo\`, or a new \`api/src/db/SqlFragments.scala\`).
- Replace all seven inline copies with the fragment.
- Re-run \`EXPLAIN (ANALYZE, BUFFERS)\` per AGENTS.md §query-explain-before-merge on at least one rewritten query to confirm the plan is unchanged.

Refs #1532.

sameerparekh:
## Summary

The hand-rolled `LEFT JOIN LATERAL` that promotes ipv4/ipv6-typed `traffic_reports` rows to their resolved fqdn (via the most recent matching `connection_events.resolved_host_value` within the row's day) was duplicated across **six** `TrafficReportRepoLive` read paths plus the `resolvedCte` in `RollupRepoLive`. Every copy was byte-identical, and the comment on the rollup copy openly admitted the keep-in-sync coupling — exactly the §single-source-of-truth anti-pattern #1532 is sweeping. The blast radius is non-theoretical: this is the LATERAL whose missing supporting index caused the #1240 / #1254 prod outage.

Pull the join into `wifihaven.api.db.SqlFragments.resolvedHostLateral` and interpolate it at every identical site. The supporting-index pointer (V22 / V34 partial index on `(mac, dest_ip, ts) WHERE resolved_host_value IS NOT NULL`) and the `TODO(#730)` lives anchored to that one fragment now.

Sites collapsed (alias `tr` for `traffic_reports`, day-bound from `tr.date`):

- `api/src/db/RollupRepo.scala` — `resolvedCte` (the comment that flagged the drift)
- `api/src/db/Repos.scala` — `listForRouter`, `listPresenceRowsInWindow`, `listPresenceRowsBetween`, `listRawInRange`, `listFqdnHostAggregatesForDevice`, `listTrafficRollupRows`

Intentionally **left alone** (to keep this a pure extract per the §1741 charter — the prompt forbids semantic changes):

- `TrafficReportRepoLive.listForDevice` — binds the day with a `$date` parameter rather than `tr.date`. Semantically equivalent under the outer `WHERE tr.date = $date`, but lexically different. Tracked as a follow-up sweep on #1532.
- `TimeUsageRepoLive.listForDevice` / `listForDeviceMacs` / `snapshotAll` — different table (`time_usage tu`), different mac column (`tu.device_mac`), parameterised day. Could share a parameterised variant later, but not part of this collapse.
- The two `RollupRepoLive.aggregateByApp{Hourly,Daily}` LATERALs — different shape entirely (`traffic_{hourly,daily}_apps` lookups), not FQDN-resolve.

## Plan / behaviour

Pure refactor. Emitted SQL differs only in interior whitespace; the join's column set, predicates, day bounds, ORDER BY, and `ON tr.host_type IN ('ipv4','ipv6')` clause are all unchanged. The query plan is therefore unchanged and still backed by the V22 / V34 supporting index that closed #1254 / #1256.

## §query-explain-before-merge

This PR introduces no new query and no changed plan — the canonical fragment expands to byte-identical-modulo-whitespace SQL at every site, and Postgres normalises whitespace before planning. The on-prod `EXPLAIN (ANALYZE, BUFFERS)` that backs the V22 / V34 partial index (run during #1254 / #1256 mitigation) still applies. The full `mill api.test` suite — which runs every Flyway migration on classpath against the same embedded Postgres these queries hit in prod — passes (192/192 modules; the suite executes the LATERAL via `listForRouter`, the rollup CTE via `RollupRepoSpec` / `AppUsedRollupSpec`, and the presence-row paths via `PresenceDefaultsPinSpec` / `TimeUsedRollupSpec`).

If the operator wants the on-prod `EXPLAIN` re-confirmed before merge, the read-only DSN in 1Password ("WifiHaven — Prod RO Postgres URL") is the route; happy to paste the plan summary when supplied — it needs an interactive step this session can't take.

## Test plan

- [x] `mill api.compile`
- [x] `mill api.test` — 192/192 modules pass
- [x] `mill api.checkFormatAll`
- [x] Pre-push hooks (scalafmt, migration-isolation) clean
- [ ] (Optional) on-prod `EXPLAIN (ANALYZE, BUFFERS)` re-run on `listRawInRange` to confirm the partial index is still used

Closes #1741. Refs #1532.

### DIFF
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -1890,16 +1890,7 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                  tr.date::TEXT, tr.period_start::TEXT, tr.period_end::TEXT,
                  tr.active_seconds, tr.bytes_in, tr.bytes_out
           FROM traffic_reports tr
-          LEFT JOIN LATERAL (
-            SELECT resolved_host_value
-            FROM connection_events
-            WHERE mac          = tr.mac
-              AND dest_ip      = tr.host_value
-              AND resolved_host_value IS NOT NULL
-              AND ts >= tr.date::TIMESTAMPTZ
-              AND ts <  (tr.date + INTERVAL '1 day')::TIMESTAMPTZ
-            ORDER BY ts DESC LIMIT 1
-          ) ce ON tr.host_type IN ('ipv4','ipv6')
+          ${SqlFragments.resolvedHostLateral}
           WHERE tr.router_id = $routerId
           ORDER BY tr.period_start DESC LIMIT $limit"""
       .query[R]
@@ -1939,16 +1930,7 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                                tr.host_value),
                       tr.active_seconds, tr.bytes_in, tr.bytes_out, tr.period_start, tr.period_end
                FROM traffic_reports tr
-               LEFT JOIN LATERAL (
-                 SELECT resolved_host_value
-                 FROM connection_events
-                 WHERE mac          = tr.mac
-                   AND dest_ip      = tr.host_value
-                   AND resolved_host_value IS NOT NULL
-                   AND ts >= tr.date::TIMESTAMPTZ
-                   AND ts <  (tr.date + INTERVAL '1 day')::TIMESTAMPTZ
-                 ORDER BY ts DESC LIMIT 1
-               ) ce ON tr.host_type IN ('ipv4','ipv6')
+               ${SqlFragments.resolvedHostLateral}
                WHERE tr.period_start >= $fromInstant AND tr.period_start < $toInstant
                  AND (tr.active_seconds > 0 OR tr.bytes_in > 0 OR tr.bytes_out > 0)
                  AND """ ++ Fragments.in(fr"tr.mac", nel)
@@ -1986,16 +1968,7 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                                tr.host_value),
                       tr.active_seconds, tr.bytes_in, tr.bytes_out, tr.period_start, tr.period_end
                FROM traffic_reports tr
-               LEFT JOIN LATERAL (
-                 SELECT resolved_host_value
-                 FROM connection_events
-                 WHERE mac          = tr.mac
-                   AND dest_ip      = tr.host_value
-                   AND resolved_host_value IS NOT NULL
-                   AND ts >= tr.date::TIMESTAMPTZ
-                   AND ts <  (tr.date + INTERVAL '1 day')::TIMESTAMPTZ
-                 ORDER BY ts DESC LIMIT 1
-               ) ce ON tr.host_type IN ('ipv4','ipv6')
+               ${SqlFragments.resolvedHostLateral}
                WHERE tr.date BETWEEN $from AND $to
                  AND (tr.active_seconds > 0 OR tr.bytes_in > 0 OR tr.bytes_out > 0)
                  AND """ ++ Fragments.in(fr"tr.mac", nel) ++
@@ -2033,16 +2006,7 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                   tr.period_start, tr.period_end,
                   tr.active_seconds, tr.bytes_in, tr.bytes_out
            FROM traffic_reports tr
-           LEFT JOIN LATERAL (
-             SELECT resolved_host_value
-             FROM connection_events
-             WHERE mac          = tr.mac
-               AND dest_ip      = tr.host_value
-               AND resolved_host_value IS NOT NULL
-               AND ts >= tr.date::TIMESTAMPTZ
-               AND ts <  (tr.date + INTERVAL '1 day')::TIMESTAMPTZ
-             ORDER BY ts DESC LIMIT 1
-           ) ce ON tr.host_type IN ('ipv4','ipv6')
+           ${SqlFragments.resolvedHostLateral}
            WHERE tr.period_start >= $fromInstant AND tr.period_start < $toInstant
              AND (tr.active_seconds > 0 OR tr.bytes_in > 0 OR tr.bytes_out > 0) """
     val macFilter  = macs match {
@@ -2097,16 +2061,7 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                    ) AS host,
                    (tr.bytes_in + tr.bytes_out) AS bytes
             FROM traffic_reports tr
-            LEFT JOIN LATERAL (
-              SELECT resolved_host_value
-              FROM connection_events
-              WHERE mac          = tr.mac
-                AND dest_ip      = tr.host_value
-                AND resolved_host_value IS NOT NULL
-                AND ts >= tr.date::TIMESTAMPTZ
-                AND ts <  (tr.date + INTERVAL '1 day')::TIMESTAMPTZ
-              ORDER BY ts DESC LIMIT 1
-            ) ce ON tr.host_type IN ('ipv4','ipv6')
+            ${SqlFragments.resolvedHostLateral}
             WHERE tr.mac = $mac
               AND tr.period_start >= $fromInstant
               AND tr.period_start <  $toInstant
@@ -2135,16 +2090,7 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                   tr.date, tr.period_start, tr.period_end,
                   tr.active_seconds, tr.bytes_in, tr.bytes_out
            FROM traffic_reports tr
-           LEFT JOIN LATERAL (
-             SELECT resolved_host_value
-             FROM connection_events
-             WHERE mac          = tr.mac
-               AND dest_ip      = tr.host_value
-               AND resolved_host_value IS NOT NULL
-               AND ts >= tr.date::TIMESTAMPTZ
-               AND ts <  (tr.date + INTERVAL '1 day')::TIMESTAMPTZ
-             ORDER BY ts DESC LIMIT 1
-           ) ce ON tr.host_type IN ('ipv4','ipv6')
+           ${SqlFragments.resolvedHostLateral}
            WHERE 1=1"""
     val byMacs  = f.macs match {
       case None      => fr""
--- a/api/src/db/RollupRepo.scala
+++ b/api/src/db/RollupRepo.scala
@@ -202,7 +202,7 @@ class RollupRepoLive(xa: Transactor[Task]) extends RollupRepo {
           tr.bytes_in,
           tr.bytes_out
         FROM traffic_reports tr
-        """ ++ SqlFragments.resolvedHostLateral ++ fr"""
+        ${SqlFragments.resolvedHostLateral}
         WHERE (tr.active_seconds > 0 OR tr.bytes_in > 0 OR tr.bytes_out > 0)
       """
 
--- a/api/src/db/RollupRepo.scala
+++ b/api/src/db/RollupRepo.scala
@@ -184,9 +184,8 @@ trait RollupRepo {
 
 class RollupRepoLive(xa: Transactor[Task]) extends RollupRepo {
 
-  // The LATERAL FQDN-resolve mirrors TrafficReportRepoLive — keep them in
-  // lockstep. If the read-time resolve in traffic_reports queries ever
-  // changes, this CTE must change too.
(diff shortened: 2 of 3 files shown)


sameerparekh:
## Independent PR review — `ef6b307f`

Reviewed `git diff origin/main...HEAD` against the 8 dimensions in [`docs/pr-review-checklist.md`](https://github.com/wifihaven/wifihaven/blob/main/docs/pr-review-checklist.md).

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS

1. [api/src/db/Repos.scala:1893](https://github.com/wifihaven/wifihaven/blob/ef6b307f/api/src/db/Repos.scala#L1893) and friends — sites in `Repos.scala` splice the fragment via `${SqlFragments.resolvedHostLateral}` inside `sql"""…"""`, while [api/src/db/RollupRepo.scala:205](https://github.com/wifihaven/wifihaven/blob/ef6b307f/api/src/db/RollupRepo.scala#L205) uses `""" ++ SqlFragments.resolvedHostLateral ++ fr"""…`. Both work; pick one style for the next SSOT sweep so the next reviewer doesn't have to convince themselves twice that the splice produces the same SQL.

2. [api/src/db/Repos.scala:1858](https://github.com/wifihaven/wifihaven/blob/ef6b307f/api/src/db/Repos.scala#L1858) — `TrafficReportRepoLive.listForDevice` is the one remaining `traffic_reports`-LATERAL that wasn't collapsed (binds the day via `$date::TIMESTAMPTZ` rather than `tr.date`). Under the outer `WHERE tr.date = $date` the two forms are equivalent and the partial index covers them identically; converting it would have collapsed seven sites to one. The PR body acknowledges this as deferred to a #1532 follow-up — acceptable, but worth filing the issue before this lands so it doesn't quietly stay forked.

3. [api/src/db/Repos.scala:1606](https://github.com/wifihaven/wifihaven/blob/ef6b307f/api/src/db/Repos.scala#L1606) / [:1637](https://github.com/wifihaven/wifihaven/blob/ef6b307f/api/src/db/Repos.scala#L1637) / [:1663](https://github.com/wifihaven/wifihaven/blob/ef6b307f/api/src/db/Repos.scala#L1663) — same story for the three `TimeUsageRepoLive` LATERALs (`time_usage tu` / `tu.device_mac`). Parametrising the fragment over `(table, mac column)` is a small enough change to do in the same sweep; leaving them inline preserves the same drift risk the rest of this PR is closing. Out of scope here per the PR's stated charter, but the `TODO(#730)` umbrella that wants to retire the whole read-side resolve should explicitly own these.

### Status per dimension

1. **SSOT** — This PR *is* an SSOT collapse: identical LATERAL pulled into [`SqlFragments.resolvedHostLateral`](https://github.com/wifihaven/wifihaven/blob/ef6b307f/api/src/db/SqlFragments.scala#L20-L30). The keep-in-sync comment on `RollupRepo.resolvedCte` is gone. The remaining un-collapsed sites are flagged above as NITs, not BLOCKERs — they're outside the explicit "byte-identical only" scope this PR set.
2. **Testing integrity** — No tests modified; `mill api.test` reports 192/192 passing in the PR description, and the LATERAL is exercised end-to-end by `RouterIngestSpec`, `RollupRepoSpec`, `PresenceDefaultsPinSpec`, `TimeUsedRollupSpec`, and `AppUsedRollupSpec`. For a pure extract whose emitted SQL is whitespace-only different, that's the right gate.
3. **Architectural invariants** — N/A. No router-agent, DNS-reasoning, snapshot, or `Clock` / config touches.
4. **Backwards compatibility / wire contract** — N/A. No wire shape, no API surface change.
5. **Migrations / §query-explain-before-merge** — No migration. The fragment is a static SQL constant; expansion at each call site differs from the inlined form only in whitespace, so the plan is the same plan Postgres already chooses on prod (still backed by the V22 / V34 `(mac, dest_ip, ts) WHERE resolved_host_value IS NOT NULL` partial index that closed #1254 / #1256). The PR body honestly notes the on-prod `EXPLAIN (ANALYZE, BUFFERS)` re-run requires the 1Password RO DSN and was not performed this session. Given the SQL is byte-identical-modulo-whitespace and Postgres's planner ignores whitespace, this is acceptable — but if the operator wants belt-and-suspenders, a quick `EXPLAIN` on `listRawInRange` against the prod RO DSN would close the loop.
6. **Metrics + dashboard** — N/A. No new emitted path.
7. **Scope / hygiene** — Diff is focused (3 files, 40 +/- 73). Commit message and PR body contain no Claude / Anthropic / AI mentions. `TODO(#730)` is correctly anchored to the new central fragment.
8. **Security** — Parameterised SQL preserved; fragment is a static constant with no user input. No secrets touched.

VERDICT: APPROVE

Pure SSOT collapse of the identical `tr.date`-bound FQDN-resolve LATERAL into one Doobie `Fragment`, removing the explicit "keep in lockstep" coupling on the rollup CTE. Emitted SQL is whitespace-only different from each inline copy, so plan and index coverage are unchanged. Tests pass; remaining inlined variants are intentionally out of scope and noted for the #1532 follow-up.

sameerparekh:
## Independent PR review — re-run on `6a0f86b1`

### Prior findings (previously cited at `ef6b307f`)

| # | Finding | Status |
|---|---------|--------|
| NIT 1 | Splice style inconsistency between `Repos.scala` (`${…}`) and `RollupRepo.resolvedCte` (`""" ++ frag ++ fr"""…`) | **ADDRESSED** — `RollupRepo.resolvedCte` now uses `${SqlFragments.resolvedHostLateral}` to match the rest. |
| NIT 2 | `TrafficReportRepoLive.listForDevice` still inlines a `$date`-bound LATERAL variant. | **NOT-ADDRESSED** — intentionally deferred. Now tracked as [#1753](https://github.com/wifihaven/wifihaven/issues/1753), so it won't quietly stay forked. |
| NIT 3 | Three `TimeUsageRepoLive` LATERALs (different table / mac column) still inlined. | **NOT-ADDRESSED** — intentionally deferred. Also covered by [#1753](https://github.com/wifihaven/wifihaven/issues/1753). |

### Delta review (`git diff ef6b307f...6a0f86b1`)

One-line change in [api/src/db/RollupRepo.scala:205](https://github.com/wifihaven/wifihaven/blob/6a0f86b1/api/src/db/RollupRepo.scala#L205): the `""" ++ SqlFragments.resolvedHostLateral ++ fr"""` concatenation collapses to `${SqlFragments.resolvedHostLateral}` inside the surrounding `fr"""…"""`. Same Doobie semantics (Fragment is spliced as SQL text either way), same resulting SQL, plan unchanged. `mill api.test` still green (192/192).

#### NEW findings
(none)

VERDICT: APPROVE

Style NIT addressed cleanly; the two deferred NITs are now anchored to #1753. No new findings on the delta. The PR remains a pure SSOT collapse with emitted SQL whitespace-only different from the prior inline copies — plan and index coverage unchanged. Cleared to merge.

Issue #1544, opened 2026-06-07
refactor(divergence): collapse PolicyService.decide() into TimeStatusService.todaysState — stop hand-mirroring the snapshot cap/site/schedule fold


sameerparekh:
Sub-issue of #1532 (divergence audit), **Finding 1 — the keystone collapse.**

## Divergence
`PolicyServiceLive.decide()` re-implements by hand the entire per-profile cap/site/schedule fold that `TimeStatusService.fold`/`assemble`/`usedSecondsForProfile`/`siteDayStates` already produce for the snapshot. The two must agree because the snapshot drives nftables enforcement while `decide()` drives the **kid block-page reason** and the per-host fallback decision.

## Locations (origin/main)
- Snapshot source of truth: `TimeStatusService.fold`/`assemble`/`usedSecondsForProfile`/`siteDayStates`; consumed by `PolicyServiceLive.snapshot` (PolicyService.scala ~140,185).
- Hand-rolled mirror: `PolicyServiceLive.decide` PolicyService.scala ~334–430 — per-profile total (~341–370, re-deriving `siteGroups`/`byApp`/`exemptPats`/`perMacTot`/`totalMins` that mirror `siteDayStates`+`usedSecondsForProfile`), `capExhausted` (~373–374, mirror of `dailyCapExhausted`), and the paused→schedule→time-limit precedence (~396–430, mirror of `assemble` ~540–550).
- The code self-documents the coupling: `// same branch as snapshot`, `// Mirrors TimeStatusService.siteDayStates`, `// the per-host mirror of PolicyService.dailyCapExhausted`.

## Status
Currently **AGREE** line-for-line. No present symptom. But this is the largest hand-maintained mirror in the policy code and sits directly on the time-accounting / app-centric surface that is actively changing — any future edit to `fold`/`assemble`/`usedSecondsForProfile` that misses the `decide()` copy makes the block page report a different verdict than nftables enforces (the #1531/#1539 failure mode, one layer down).

## Fix — COLLAPSE
[...]

PR #1754, opened 2026-06-15
refactor(#1742): unify isScheduleBlock derivation in PolicyService

### LINKED ISSUE #1742: audit(#1532): collapse isScheduleBlock — snapshot vs live-decide derivations in PolicyService

## Drift

\`PolicyService\` computes "is this MAC currently schedule-blocked" two different ways depending on which path is asking:

- \`api/src/policy/PolicyService.scala:192\` (snapshot rebuild) — \`state.blockReason.contains(MacBlockReason.Schedule)\`. Reads the precedence-collapsed \`blockReason\` already chosen by the outer fold.
- \`api/src/policy/PolicyService.scala:406\` (live \`/decide\` path) — \`scheduleBlock(scheds, now).nonEmpty\`. Re-derives schedule activity from scratch from the raw schedule list.

The surrounding comment (lines 398-406) explains the gap is currently *benign* because the precedence order (Paused > Schedule > TimeLimit > Manual) is honoured on both sides. But the alignment is not enforced by the type system. If a new precedence tier is inserted above \`Schedule\` on one path and missed on the other, the snapshot and \`/decide\` will silently disagree on whether the user is currently in a Schedule block — and that disagreement is exactly what \`POST /api/router/events\` labels its connection events from (see AGENTS.md §Architectural model on \`blockReason\` and #679).

## Recommendation

Extract a single \`def computeIsScheduleBlock(state: MacState, scheds: List[ProfileSchedule], now: LocalDateTime): Boolean\` (or similar) that both call sites consume. The function documents WHY one path can use the pre-computed reason and the other must re-derive (snapshot has the precedence-collapsed value, \`/decide\` doesn't have one yet at that point), and is the one place precedence ordering for schedule lives.

Alternatively type-enforce: make the live-decide path also fold through the same precedence pipeline first, eliminating the second derivation entirely.

## Scope

- Define the shared helper, pick collapse vs. fold-unification.
- Update both call sites; remove the now-stale "benign drift" comment.
- Add a test that exercises both paths against the same input and asserts they agree.

Refs #1532.

sameerparekh:
## Summary

Collapses the duplicated `isScheduleBlock` carve gate (#1679) in
[`api/src/policy/PolicyService.scala`](api/src/policy/PolicyService.scala)
into a single derivation sourced from the precedence-collapsed
`ProfileDayState.blockReason`.

Before: `snapshot` (`:193`) read `state.blockReason.contains(MacBlockReason.Schedule)` while `decide` (`:414`) re-derived `scheduleBlock(scheds, now).nonEmpty`. On a profile that was paused AND inside a schedule window, the two disagreed — Paused outranks Schedule per the documented precedence (AGENTS.md §Architectural model), so the snapshot left the `allowedDuringScheduleBlock=false` host in `extraAllowed` while `decide` suppressed the carve and returned `Block:paused`. Latent because the router had already allowed the connection via the snapshot's `extraAllowed`, but a real single-source-of-truth violation per AGENTS.md §single-source-of-truth and the #1532 audit.

## Approach: fold-unification (preferred over a shared helper)

Both call sites now read `dayState.blockReason.contains(MacBlockReason.Schedule)`. `decide` already obtains `dayState` from `TimeStatusService.todaysState` (post-#1544), so this is a one-line collapse rather than a new helper — the gate cannot diverge without a `TimeStatusService.fold` change moving both at once. The whole-MAC schedule short-circuit further down in `decide` still calls `scheduleBlock(scheds, now)` — its semantics match the precedence by construction because it runs after the pause check.

The "benign drift" comment block around the old derivation is removed and its substance migrated above the unified line.

## Tests

New [`PolicyScheduleBlockSSotSpec`](api/test/src/feature/PolicyScheduleBlockSSotSpec.scala) exercises both paths against the same fixture and asserts agreement across:

- schedule-only active (both suppress the carve) — passes pre and post
- **paused + schedule active** (Paused wins precedence, carve survives on both) — **fails pre-refactor, passes post**
- out of schedule + daily-limit exhausted (TimeLimit on both) — passes pre and post

The failing test is committed in a separate commit (red → green progression visible in the branch history) per AGENTS.md TDD workflow.

## Test plan

- [x] `mill 'api.test.testOnly' wifihaven.api.feature.PolicyScheduleBlockSSotSpec` — 3/3 pass
- [x] Policy/schedule-touching specs re-run: `PolicySnapshotScheduleBlockToggleSpec`, `RouterDecisionConsolidationSpec`, `RouterDecisionSpec`, `RouterDecisionPerAppScheduleSpec`, `PolicySnapshotPerAppScheduleSpec`, `PolicyScheduleGlobalAllowSpec`, `PolicySnapshotNamedScheduleSpec` — all green
- [x] `mill __.checkFormatting` (via pre-commit hook)

Closes #1742. Refs #1532.

### DIFF
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -404,8 +404,8 @@ class PolicyServiceLive(
                 // exhausted is NOT carved, so it correctly falls through to the time_limit block
                 // below; an exempt one (or one under cap) is carved and beats it.
                 // #1679/#1742: read the carve gate off the precedence-collapsed `dayState.blockReason`
-                // — the SAME source the snapshot reads via `state.blockReason.contains(Schedule)`
-                // (see :194). This collapses the prior drift where this path re-derived schedule
+                // — the SAME source the snapshot's own `isScheduleBlock` reads (via
+                // `state.blockReason.contains(Schedule)`). This collapses the prior drift where this path re-derived schedule
                 // activity from raw schedules (`scheduleBlock(scheds, now).nonEmpty`) and disagreed
                 // with the snapshot on a Paused+schedule-active profile (Paused outranks Schedule in
                 // the documented precedence — AGENTS.md §Architectural model — so the toggle must
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -403,15 +403,17 @@ class PolicyServiceLive(
                 // — exactly as the snapshot does. A non-exempt allowed_during app whose cap is
                 // exhausted is NOT carved, so it correctly falls through to the time_limit block
                 // below; an exempt one (or one under cap) is carved and beats it.
-                // #1679: pass isScheduleBlock so apps with allowedDuringScheduleBlock=false are
-                // suppressed from the extraAllowed carve during an active schedule window.
-                // Note: this reads `scheduleBlock(scheds, now).nonEmpty` independently of
-                // `p.paused`, so a Paused+schedule-active profile gets isScheduleBlock=true
-                // here while the snapshot uses `state.blockReason.contains(Schedule)=false`
-                // (Paused outranks Schedule in the blockReason precedence). The divergence is
-                // benign: if the snapshot has the app in extraAllowed, nftables allows the
-                // connection and the block page / /decide endpoint is never reached.
-                val isScheduleBlock          = scheduleBlock(scheds, now).nonEmpty
+                // #1679/#1742: read the carve gate off the precedence-collapsed `dayState.blockReason`
+                // — the SAME source the snapshot reads via `state.blockReason.contains(Schedule)`
+                // (see :194). This collapses the prior drift where this path re-derived schedule
+                // activity from raw schedules (`scheduleBlock(scheds, now).nonEmpty`) and disagreed
+                // with the snapshot on a Paused+schedule-active profile (Paused outranks Schedule in
+                // the documented precedence — AGENTS.md §Architectural model — so the toggle must
+                // NOT fire there; #1679 is schedule-scope only). `scheduleBlock(scheds, now)` is
+                // still consulted below for the actual whole-MAC schedule short-circuit (after the
+                // pause check, so its semantics match the precedence by construction).
+                val isScheduleBlock          =
+                  dayState.blockReason.contains(MacBlockReason.Schedule)
                 val (appAllowed, appBlocked) =
                   ProfileAppDispositions
                     .from(appLimits)
--- a/api/test/src/feature/PolicyScheduleBlockSSotSpec.scala
+++ b/api/test/src/feature/PolicyScheduleBlockSSotSpec.scala
@@ -1,0 +1,201 @@
+package wifihaven.api.feature
+
+import wifihaven.api.db.*
+import wifihaven.api.policy.*
+import wifihaven.shared.*
+import wifihaven.shared.Clock.TestClock
+import wifihaven.shared.types.*
+import wifihaven.testinfra.*
+import io.zonky.test.db.postgres.embedded.EmbeddedPostgres
+import zio.{Clock as _, *}
+import zio.test.*
+import zio.test.TestAspect
+
+import java.time.LocalDateTime
+
+/**
+ * #1742 anti-divergence pin. `PolicyServiceLive.snapshot` and `PolicyServiceLive.decide` must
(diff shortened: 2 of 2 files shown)


sameerparekh:
## Independent PR review — #1742 isScheduleBlock unification

Read-only adversarial pass against [docs/pr-review-checklist.md](../blob/main/docs/pr-review-checklist.md). Diff: `git diff origin/main...HEAD` — 1 source file (12 net) + 1 new spec file.

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

1. [api/src/policy/PolicyService.scala:412](api/src/policy/PolicyService.scala#L412) — the new comment says "the SAME source the snapshot reads via `state.blockReason.contains(Schedule)` (see :194)" but the snapshot's `val isScheduleBlock` is actually at `:193`. Line-number cross-references in comments rot when the file shifts; consider naming the snapshot call site without the line number (e.g. "see the snapshot's `isScheduleBlock` above").

### Dimension checklist

1. **Duplicated logic / SSOT** — ✅ This PR *is* the collapse. `decide`'s carve gate now reads `dayState.blockReason.contains(MacBlockReason.Schedule)` — the same precedence-collapsed value the snapshot uses at [:193](api/src/policy/PolicyService.scala#L193), sourced from `TimeStatusService.todaysState`. The stale "benign drift" comment block is removed. No "keep in sync" comments remain in the gate.
2. **Testing integrity** — ✅ TDD red→green is visible in the branch history (commit `8ae8435b` adds the failing test, commit `c575337f` makes it pass). The new [PolicyScheduleBlockSSotSpec](api/test/src/feature/PolicyScheduleBlockSSotSpec.scala) exercises both `snapshot` and `decide` against the same fixture and asserts agreement on three scenarios — schedule-only, paused+schedule (the precedence boundary that fails pre-refactor), and out-of-schedule+TimeLimit. Real embedded Postgres, real `Clock.TestClock`, no mocks. No existing assertions weakened or deleted.
3. **Architectural invariants** — ✅ No router-agent changes, no snapshot wire-shape change, no DNS-as-enforcement reasoning, no `/tmp` writes. `Clock` is injected via `TestClock`. Decision logic stays in `PolicyService`.
4. **Backwards compatibility / wire contract** — ✅ Wire-visible shapes unchanged. `RouterDecisionResponse` fields are identical; only the value returned for the Paused+schedule-active+`allowedDuringScheduleBlock=false` host edge case changes (`Block:paused` → `Allow:extra_allowed`) to align with what nftables actually enforces via the snapshot's `extraAllowed`. That's the documented #421/#1413 invariant (`extraAllowed` beats every whole-MAC block), so this corrects a latent disagreement rather than introducing one.
5. **Migrations** — n/a (no SQL changes).
6. **Metrics + dashboard** — n/a (pure internal collapse, no new path).
7. **Scope / hygiene** — ✅ Two files, focused. No drive-by edits, no Claude/AI mentions in commits or PR body, three-dot diff used.
8. **Security** — n/a.

### VERDICT: APPROVE

A clean fold-unification of the `isScheduleBlock` carve gate per the #1532 SSOT charter — `decide` now reads the same precedence-collapsed `dayState.blockReason` as `snapshot`, so the two paths cannot disagree on the #1679 `allowedDuringScheduleBlock` toggle without a `TimeStatusService.fold` change moving both at once. The accompanying spec is a proper anti-divergence pin with visible red→green progression and covers the precedence boundary (Paused outranks Schedule) that motivated the change. One trivial NIT on a line-number reference in a code comment; non-blocking.

sameerparekh:
## Independent PR review — #1742 isScheduleBlock unification (re-run)

Re-run on the latest push. Prior review: [c575337f](#issuecomment-4712518596). Incremental delta: `git diff c575337f...HEAD` — comment-only follow-up to the prior NIT.

### Prior findings status

|---|---|---|
| NIT 1 | `PolicyService.scala:412` line-number reference `:194` to the snapshot's `isScheduleBlock` was off-by-one | **ADDRESSED** — comment now names the snapshot's `isScheduleBlock` symbol instead of a line number ([PolicyService.scala:407](api/src/policy/PolicyService.scala#L407)). |

### New findings (delta)

#### BLOCKERS

(none)

#### SHOULD-FIX

(none)

#### NITS

(none)

### Dimension checklist (delta)

1–8: delta is comment text only — no logic, no test, no wire, no migration, no metric. All dimensions remain ✅ from the prior pass.

### VERDICT: APPROVE

Prior NIT addressed; the comment now names the snapshot's `isScheduleBlock` symbol so it won't rot on further edits. No new findings — no open BLOCKER.

Issue #1740, opened 2026-06-15
audit(#1532): collapse retention horizon constants — RetentionSweepJob vs web/retentionGating.ts


sameerparekh:
Follow-up from the #1532 SSOT audit, sweep N.

## Drift

The retention horizon day counts are hard-coded in two places that the codebase itself flags must stay in lockstep:

- \`api/src/usage/RetentionSweepJob.scala:39-46\` — \`RawRetentionDays = 30\`, \`HourlyRetentionDays = 90\`, \`DailyRetentionDays = 180\`.
- \`web/src/components/usage/retentionGating.ts:5-18\` — \`DEFAULT_RETENTION_HORIZONS = { rawDays: 30, hourlyDays: 90, dailyDays: 180 }\`.

The web file's own header comment names the risk: *"these constants must track RetentionSweepJob"* and *"Operator-tunable horizons (GET /api/usage/horizons) are a follow-up."* If the sweep job's values change without a matching web edit, the UI will silently offer granularities that the API has already purged.

## Recommendation

Expose the horizons from the API (the comment names \`GET /api/usage/horizons\`) and have the web read them at runtime instead of duplicating the constants. The Scala side becomes the single source of truth; the web caches/falls back to the same defaults only when the endpoint is unreachable.

Until that lands, the alignment is enforced by prose only.

## Scope

- New \`GET /api/usage/horizons\` returning the three day counts.
- Web fetches it on app boot, replaces \`DEFAULT_RETENTION_HORIZONS\` usage.
- Sweep job exports the constants from one place the route can read.

Refs #1532.

PR #1756, opened 2026-06-15
refactor(#1743): SPA reads bucket→grain mapping from API

### LINKED ISSUE #1743: audit(#1532): collapse web BUCKET_TIER mapping vs shared BucketPolicy.grainForBucket

## Drift

\`web/src/components/usage/retentionGating.ts:29-38\` defines:

\`\`\`ts
const BUCKET_TIER: Record<TrafficUsageBucket, SourceTier> = {
  raw: 'raw', '1m': 'raw', '10m': 'raw',
  '1h': 'hourly', '12h': 'hourly',
  '1d': 'daily', '1w': 'daily',
}
\`\`\`

\`shared/src/BucketPolicy.scala:28-31\` defines the same bucket→grain mapping (\`grainForBucket\`). The web file even says: *"mirrors the bucket→tier grouping in UsageRoutes.scala."*

The web cannot import Scala, so SOME duplication is inherent — but the mapping is also exactly the kind of thing that should be a server-emitted constant the SPA reads, not a parallel literal hand-maintained.

## Recommendation

Two options:

1. **API-emitted constant.** Add the mapping to the existing horizons-type endpoint (or pair this with #1740) so the web reads \`BUCKET_TIER\` from \`/api/usage/buckets\` at boot.
2. **Annotated parallel copy.** If the SPA cannot rely on a network fetch here, at minimum annotate each web entry with the Scala case it mirrors so a reviewer can grep-check.

Option 1 is the SSOT-honest fix.

## Scope

Likely landed alongside #1740 — same endpoint, same shape.

Refs #1532.

sameerparekh:
## Summary

Collapses the duplicated bucket → grain mapping between `BucketPolicy.grainForBucket` (Scala) and the hand-defined `BUCKET_TIER` in `web/src/components/usage/retentionGating.ts` per [Single source of truth](../blob/main/AGENTS.md#single-source-of-truth).

- New `GET /api/usage/config` emits `BucketPolicy.bucketTiers` on the wire (auth-gated, consistent with the rest of `/api/usage/*`).
- `BucketPolicy` now owns the canonical bucket list and the bucket → grain map; both are derived from `grainForBucket` so the classification stays single-source.
- The SPA fetches the mapping via `useUsageConfig` (1-hour cache). `bucketAvailability` takes the map as a new parameter and falls back to `DEFAULT_BUCKET_TIERS` for pre-fetch / offline / partial-payload cases — the constants live on so the date-picker still works before the boot fetch returns.

## Path note

#1740 (operator-tunable retention horizons on the same endpoint) is still open and unstarted, so this PR opens `GET /api/usage/config` here. When #1740 lands it just extends the existing response with a `horizons` field — the route comment and the shared model both flag the intent. No conflict expected.

## Test plan

- [x] Failing feature test committed first (red), then implementation (green): `wifihaven.api.feature.UsageConfigSpec` — asserts the wire payload equals `BucketPolicy.bucketTiers` plus spot-checks, and that the route is auth-gated.
- [x] `mill api.test` — full Scala test suite green.
- [x] `scalafmt --check` — formatted.
- [x] Vitest: extended `web/src/components/usage/retentionGating.test.ts` with fallback + injected-map + partial-map + default-mirror cases; full `npx vitest run` green (410 tests).
- [x] `web/` `npm run type-check` clean.

Refs #1532. Closes #1743.

### DIFF
--- a/api/src/routes/UsageRoutes.scala
+++ b/api/src/routes/UsageRoutes.scala
@@ -39,6 +39,20 @@ object UsageRoutes {
       clock: Clock,
   ): Routes[Any, Response] =
     Routes(
+      // #1743: bucket → grain mapping the SPA reads at boot to gate its
+      // date-picker, sourced from `BucketPolicy.bucketTiers` so the SPA no
+      // longer hand-mirrors `grainForBucket` in retentionGating.ts. (Paired
+      // with #1740 — the same endpoint will grow a `horizons` field there.)
+      Method.GET / "api" / "usage" / "config"                                   ->
+        handler { (req: Request) =>
+          val handle: ZIO[Any, ApiError, Response] = for {
+            _ <- requireAuth(req, auth).mapError(ApiError.Wrapped(_))
+            cfg = UsageConfig(
+              bucketTiers = BucketPolicy.bucketTiers.view.mapValues(_.wire).toMap,
+            )
+          } yield Response.json(cfg.toJson)
+          handle.mapError(ErrorMapper.errorToResponse)
+        },
       Method.GET / "api" / "usage" / "traffic"                                  ->
         handler { (req: Request) =>
           trafficHandler(
--- a/api/src/routes/UsageRoutes.scala
+++ b/api/src/routes/UsageRoutes.scala
@@ -43,6 +43,9 @@ object UsageRoutes {
       // date-picker, sourced from `BucketPolicy.bucketTiers` so the SPA no
       // longer hand-mirrors `grainForBucket` in retentionGating.ts. (Paired
       // with #1740 — the same endpoint will grow a `horizons` field there.)
+      // No bespoke metric: the HTTP middleware already meters `route`,
+      // `status`, and duration for this path, and the response is a static
+      // constant with no failure modes beyond auth.
       Method.GET / "api" / "usage" / "config"                                   ->
         handler { (req: Request) =>
           val handle: ZIO[Any, ApiError, Response] = for {
--- a/api/test/src/feature/UsageConfigSpec.scala
+++ b/api/test/src/feature/UsageConfigSpec.scala
@@ -100,5 +100,5 @@ object UsageConfigSpec extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostgres 
         resp <- routes.runZIO(Request.get(url))
       } yield assertTrue(resp.status == Status.Unauthorized)
     },
-  )
+  ) @@ TestAspect.sequential
 }
--- a/api/test/src/feature/UsageConfigSpec.scala
+++ b/api/test/src/feature/UsageConfigSpec.scala
@@ -1,0 +1,104 @@
+package wifihaven.api.feature
+
+import io.zonky.test.db.postgres.embedded.EmbeddedPostgres
+import wifihaven.api.JwtConfig
+import wifihaven.api.auth.*
+import wifihaven.api.db.*
+import wifihaven.api.routes.UsageRoutes
+import wifihaven.shared.*
+import wifihaven.shared.Clock.TestClock
+import wifihaven.testinfra.*
+import zio.{Clock as _, *}
+import zio.http.*
+import zio.json.*
+import zio.test.*
+
+// #1743: GET /api/usage/config emits the bucket → grain mapping so the SPA
+// (web/src/components/usage/retentionGating.ts) reads it at runtime instead of
+// hand-maintaining a parallel literal that must mirror `BucketPolicy.grainForBucket`.
+object UsageConfigSpec extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostgres & Clock] {
+
+  override val bootstrap =
+    TestDatabase.layer ++ TestLayers.withClock(TestClock.schoolDayAfternoon)
+
+  private val cleanDb = TestDatabase.cleanAndMigrate
+
+  private val jwtCfg    = JwtConfig(secret = "test-secret-at-least-32-chars!!", expiryHours = 1)
+  private def buildAuth =
+    for {
+      ur    <- ZIO.service[UserRepo]
+      clock <- ZIO.service[Clock]
+    } yield AuthServiceLive(ur, jwtCfg, clock)
+
+  private def buildRoutes =
+    for {
+      deviceRepo      <- ZIO.service[DeviceRepo]
+      trafficRepo     <- ZIO.service[TrafficReportRepo]
+      userProfileRepo <- ZIO.service[UserProfileRepo]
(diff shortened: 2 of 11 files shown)


sameerparekh:
# Independent PR review — #1743 (reviewed-sha `dac0eef3`)

Read-only adversarial pass per [`docs/pr-review-checklist.md`](../blob/main/docs/pr-review-checklist.md). Scope: `git diff origin/main...HEAD` (3 commits, 11 files, +264 / −27).

## BLOCKERS

(none)

## SHOULD-FIX

1. `web/src/components/usage/retentionGating.ts:79-87` — `bucketAvailability` iterates `DEFAULT_BUCKET_TIERS` keys, not the API-supplied map. If the API ever adds a new bucket code (the whole point of moving this server-side), the SPA will **silently drop it from the gates record**. The fallback comment ("double as the list of buckets the SPA renders") acknowledges this is intentional, but it puts the SSOT collapse only half-way: keys still need a client release to surface. Either (a) document this as a known limit, or (b) iterate `Object.keys(bucketTiers)` and union with the default keys so a server-added bucket is at least surfaced. Not merge-gating because `TrafficUsageBucket` is also a SPA-known enum and adding a code requires a SPA change anyway — but worth flagging in the file header.

## NITS

1. `web/src/components/usage/retentionGating.ts:3-8` — File header still says "Retention horizons, mirrored from the server" but the file is now also about bucket tiers. A one-line top comment naming both surfaces would age better.
2. `web/src/api/queries.ts:87-97` — `useUsageConfig` has `staleTime: 60 * MIN` but no `gcTime`; on unmount the cache is dropped after the default 5min. Probably fine for an hourly-stale config but consider `gcTime: Infinity` so navigate-away/back doesn't re-fetch.
3. `web/src/pages/TrafficUsagePage.tsx:28-29` — `import { useUsageConfig } from '@/api/queries'` placed after `useInfiniteScroll`; could be grouped with the other `@/api` imports near the top.
4. `shared/src/BucketPolicy.scala:46-52` — `bucketTiers: Map[String, BucketGrain]` is keyed on `String` rather than the typed `UsageTraffic.Bucket` (which lives in `api/`, not `shared/`, so can't be referenced here without a refactor). The wire shape needs strings anyway, so this is fine, but the loose typing means a typo in `allBuckets` wouldn't be caught at compile time. Possibly worth a follow-up to lift `Bucket` enum to `shared/`.
5. `api/src/routes/UsageRoutes.scala:42-55` — Per AGENTS.md §instrument-new-functionality, a new route "should emit a metric in the same PR." For a static-constant auth-gated endpoint, HTTP middleware already covers route+duration+status, so a dedicated counter is overkill — but a one-line note in the route comment confirming the middleware covers it would keep the rule from being silently bent.

## Checklist coverage

- **1. SSOT** — Net positive: `BucketPolicy.bucketTiers` is the single Scala source, derived from `grainForBucket` + `allBuckets` so the two can't drift from each other. Web `DEFAULT_BUCKET_TIERS` is a documented fallback with a literal pin test at `retentionGating.test.ts:120-130`. Cross-language defaults aren't compile-checked, but no comment claims they auto-track.
- **2. Testing** — Red→green visible in history (`4c602f04` test commit, `a03c2de9` implementation commit). New feature spec uses real embedded PG + auth flow, no forbidden mocks. Existing `TrafficUsagePage.test.tsx` edits add `QueryClientProvider` + a `config` mock resolver — neither weakens nor removes any prior assertion, both are required by the new react-query dependency.
- **3. Architectural invariants** — N/A (no router / DNS / snapshot changes). Clock injected, ZIO effects, typed `ApiError`. OK.
- **4. Backwards compat** — New endpoint only; no wire-shape changes. SPA fallback ensures older API images don't break the page. OK.
- **5. Migrations** — None.
- **6. Metrics + dashboard** — See NIT 5; no new bespoke metrics, covered by HTTP middleware.
- **7. Scope / hygiene** — Focused diff, no Claude/AI mentions, commits well-titled.
- **8. Security** — `requireAuth` on the route, no SQL, no secrets, payload is non-PII config. OK.

## Verdict

**VERDICT: APPROVE**

Clean collapse of the bucket→grain duplication; SSOT primitive is `BucketPolicy.bucketTiers`, the SPA reads it via a new auth-gated config endpoint, defaults remain only as a documented offline fallback with a backstop test. TDD progression is visible in commit history and the new feature spec asserts the wire payload + auth gating. SHOULD-FIX is a known limit of the fallback's bucket-list scope, not a regression.

sameerparekh:
# Independent PR review (re-run) — #1743 (reviewed-sha `c5afd9da`)

Re-run per [`docs/pr-review-checklist.md`](../blob/main/docs/pr-review-checklist.md). Incremental delta: `git diff dac0eef3...HEAD` (1 commit, 4 files, +14 / −1).

## Prior-findings status (vs. `dac0eef3`)

| # | Prior | Status | Current state |
|---|-------|--------|---------------|
| SHOULD-FIX 1 | `retentionGating.ts:79-87` — fallback iterates `DEFAULT_BUCKET_TIERS` keys, dropping any new server-emitted bucket silently. | ADDRESSED | Documented as an intentional SPA-known-bucket-list limit in the file header (`retentionGating.ts:3-10`). |
| NIT 1 | `retentionGating.ts:3-8` — stale "Retention horizons, mirrored from the server" header. | ADDRESSED | New top-of-file block now names both surfaces (retention horizons AND bucket → grain mapping). |
| NIT 2 | `queries.ts:87-97` — `useUsageConfig` missing `gcTime`. | ADDRESSED | `gcTime: Infinity` added with a one-line comment. |
| NIT 3 | `TrafficUsagePage.tsx:28-29` — import group placement. | ADDRESSED | `useUsageConfig` now grouped with the other `@/api/*` imports at the top. |
| NIT 4 | `BucketPolicy.scala:46-52` — `Map[String, BucketGrain]` not keyed on a typed `Bucket`. | NOT-ADDRESSED (intentional; flagged as a follow-up — wire shape needs strings anyway, and lifting `UsageTraffic.Bucket` to `shared/` is out of scope for this PR). |
| NIT 5 | `UsageRoutes.scala:42-55` — no metric note. | ADDRESSED | Route comment now explicitly records that HTTP middleware covers route/status/duration and no bespoke metric is warranted. |

## New findings (delta only)

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

(none)

The delta is purely doc + a small react-query config tweak. No new code paths, no new tests required.

## Verdict

**VERDICT: APPROVE**

All addressable prior findings are ADDRESSED (the one NOT-ADDRESSED was flagged as intentional out-of-scope at first review). No new findings on the delta. PR remains a clean SSOT collapse of the bucket → grain mapping.

PR #1759, opened 2026-06-16
fix(#1757): skip per-record decode failures on /api/router/events

### LINKED ISSUE #1757: bug(prod): /api/router/events showing success-rate drops — investigate
## Symptom (prod, observed 2026-06-16)

The Grafana panel for `/api/router/events` shows success-rate (SR) dropping below baseline — routers are POSTing events and the API is rejecting some non-zero share. Lose visibility into connection events / DHCP leases / block-page reason resolution when this drops.

## Likely classes (rank by frequency hint from Grafana)

1. **Decode failure on a per-record basis.** Same class as #1569 for usage — a single malformed record (e.g. host with underscores from an Akamai CNAME — #1572 territory) failing validation. If `/events` drops the whole batch instead of skipping the bad record (like /usage now does post-#1574), every batch with one bad event fails.
2. **Auth shape mismatch.** Older routers sending the legacy shape vs newer routers — same class as #1382 (router metrics malformed). If a new server-side change tightened the request shape, older agents fail.
3. **Schema constraint failure.** A recent migration may have added a NOT NULL or CHECK constraint that some events don't satisfy. The just-shipped `host_type CHECK` widening (#1712) was the inverse case — but maybe another constraint tightened.
4. **Latency timeout** — like the #1511 usage perf issue but on /events. The agent gives up after N seconds and the request never registers; from the agent's side that's a "drop."
5. **Wire-shape regression** from recent server merges (the SNI/Tuning waves landed many small additions; one might have broken backward decode).

## Investigation steps

1. **Grafana**: pull `http_request_duration_seconds{route="/api/router/events"}` AND `http_responses_total{route="/api/router/events",status}` for the past few days. Identify:
   - Which status codes (4xx vs 5xx vs timeout)?
   - When did SR start dropping — gradual or step-function?
   - Per-router breakdown (one router producing all the failures? Or fleet-wide?).
2. **Render API logs** for `/api/router/events` errors over the symptom window. Grep for the response body if logged.
3. **Code**: read `api/src/routes/RouterIngestRoutes.scala` handler for events. Does it follow the #1574 skip-per-record pattern, or does it still drop the whole batch?
4. **Agent**: SSH a router (router.lan); check `logread | grep -iE 'events|/api/router'` for recent failures. The agent should log non-2xx responses.

## Acceptance

- Root cause identified with evidence (Grafana + logs + code).
- Fix shipped (matching the cause: skip-per-record / tolerate older shape / fix the schema constraint / improve latency / rollback the wire regression).
- Regression test that pins the failing path.

## Related
- #1569 / #1574 — same skip-per-record fix for /usage
- #1382 — router metrics malformed (router-side wire-shape drift)
- #1572 — CNAME alias recovery (could be the same bad-host-validation root)
- #1511 — /api/router/usage latency (parallel perf concern)
- #1718 — native router metrics now in Grafana (helps diagnose)

sameerparekh:
Closes #1757.

## Class pinned: A — whole-batch drop on bad record (rank #1 in the issue)

The `/api/router/events` handler decoded the request body as one `RouterEventsRequest`. A single malformed `RouterEvent` (e.g. a host that fails `Hostname` validation — the Akamai CDN CNAME class from #1572, or any other per-field validation error) failed the whole decode → 400 → every valid `connection_attempt` / `dhcp_lease` / `first_seen_mac` in the batch dropped with it. This is the symptom showing up as a non-zero failure rate on `/api/router/events` in Grafana.

This is the same shape as #1569 was on `/usage`, and the fix mirrors #1574: wire-decode envelope (`RawEventsReport`) holds events as a raw JSON array, each is decoded into a `RouterEvent` individually, bad ones are skipped + logged + metered, valid ones ingest. Envelope itself (routerId, events-is-an-array) must still parse or the request is a genuine 400.

## Evidence

Pre-fix `api/src/routes/RouterIngestRoutes.scala`:

```scala
rep <- ZIO.fromEither(body.fromJson[RouterEventsRequest])
         .mapError(ApiError.DecodeFailure(_))
// ↑ one bad event in the array fails this — whole batch 400s.
```

`/api/router/usage` handler (same file) does NOT have this problem post-#1574: it decodes the envelope, then decodes each record individually.

## Changes

- New `RawEventsReport` envelope on the events handler, analogous to `RawUsageReport`.
- Per-event decode loop + `recordEventsRecordsRejected` metric, analogous to `recordUsageRecordsRejected`.
- New metric `events_records_rejected_total{reason}` added to `MetricGuard.Allowed`. Bounded label only (`reason` enum — currently just `decode_error`), no per-mac / per-host / per-router dimensions.

## Backwards compatibility

Purely server-side decoder change. Wire is unchanged: the server now accepts a superset of what it accepted before (every valid old payload still parses; payloads that used to 400 with one bad event now 200 with that event skipped). No agent-side change required.

## Tests (RED → GREEN visible in commits)

- `events: a batch with one malformed event ingests the valid events and meters the rejection`
- `events: a batch of ONLY malformed events is accepted (200) and meters them all`

Both mirror the existing #1569/#1574 tests on `/usage`.

## Test plan

- [x] `mill api.test.testOnly wifihaven.api.feature.RouterIngestSpec` — 50 passed
- [x] `mill api.test.testOnly wifihaven.api.feature.MetricCardinalityGuardSpec`
- [x] scalafmt --check

## Follow-up (post-merge)

- Watch the `/api/router/events` success-rate panel — should return to baseline.
- Watch `events_records_rejected_total{reason="decode_error"}` rate. A non-zero share indicates the bad-host payload identified in the incident; the operator can grep API logs for `"router events: skipping malformed event"` to see the exact rejected JSON.

### DIFF
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -92,6 +92,13 @@ object MetricGuard {
     // underscores, see #1572) is skipped + metered here instead of 400-ing the
     // whole batch. `reason` is a small fixed enum (currently just `decode_error`).
     "usage_records_rejected_total"              -> Set("reason"),
+    // #1757 — events-ingest records dropped at decode. Same shape as
+    // usage_records_rejected_total: a single malformed RouterEvent (bad host /
+    // mac / ts / unknown enum value) is skipped + metered instead of 400-ing
+    // the whole batch (which would drop every valid connection_attempt /
+    // dhcp_lease / first_seen_mac with it). `reason` is a small fixed enum
+    // (currently just `decode_error`).
+    "events_records_rejected_total"             -> Set("reason"),
     // #1585 — write-time FQDN backfill for traffic_reports. Counts each
     // UsageRecord at ingest by what the backfill decided: `filled` (race-loser
     // ipv4/ipv6 rewritten to a fqdn from a recent connection_event),
@@ -345,6 +352,16 @@ object AppMetrics {
     ZIO
       .when(count > 0)(
         MetricGuard.counter("usage_records_rejected_total", Map("reason" -> reason), count.toLong),
+      )
+      .unit
+
+  // #1757 — events-ingest per-record skip count. Same wire shape as the usage
+  // counter above; emitted from RouterIngestRoutes when one or more events in
+  // a batch fail individual decode but the envelope itself parses.
+  def recordEventsRecordsRejected(count: Int, reason: String = "decode_error"): UIO[Unit] =
+    ZIO
+      .when(count > 0)(
+        MetricGuard.counter("events_records_rejected_total", Map("reason" -> reason), count.toLong),
       )
       .unit
 
--- a/api/src/routes/RouterIngestRoutes.scala
+++ b/api/src/routes/RouterIngestRoutes.scala
@@ -52,40 +52,40 @@ object RouterIngestRoutes {
             auth.authenticate(req).mapError(ApiError.Wrapped(_)).flatMap { router =>
               LogContext.annotate(LogContext.RouterId, router.id.toString) {
                 for {
-                  body <- req.body.asString.orElseFail(ApiError.BadRequest(""))
+                  body      <- req.body.asString.orElseFail(ApiError.BadRequest(""))
                   // #1569: decode the *envelope* (routerId, periodStart, periodEnd) + `records` as a
                   // raw JSON array, deferring per-record validation. A genuinely unparseable
                   // envelope (bad routerId/timestamps, not an object, `records` not an array,
                   // truncated body) still 400s — logged by the boundary.
-                  raw  <- ZIO
+                  raw       <- ZIO
                     .fromEither(body.fromJson[RawUsageReport])
                     .mapError(ApiError.DecodeFailure(_))
-                  _    <- ZIO
+                  _         <- ZIO
                     .fail(ApiError.BadRequest("router_id mismatch"))
                     .when(raw.routerId != router.id)
-                  ps   <- parseInstant(raw.periodStart)
-                  pe   <- parseInstant(raw.periodEnd)
-                  // #1569: decode each record individually. A single malformed record
-                  // (e.g. a host value that fails Hostname validation — an Akamai/CDN
-                  // CNAME target with underscores, see #1572) used to fail the WHOLE
-                  // batch, dropping every valid record with it. Now the bad records are
-                  // skipped, logged (bounded), and metered, and the valid ones ingest.
-                  decoded  = raw.records.zipWithIndex.map((j, i) => (i, j.as[UsageRecord]))
-                  rejected = decoded.collect { case (i, Left(err)) => (i, err) }
-                  records  = decoded.collect { case (_, Right(r)) => r }
-                  _        <- ZIO.foreachDiscard(rejected) { (i, err) =>
-                    ZIO.logWarning(
-                      s"router usage: skipping malformed record[$i] for router=${router.id}: $err",
-                    )
-                  }
-                  _        <- AppMetrics.recordUsageRecordsRejected(rejected.size)
+                  ps        <- parseInstant(raw.periodStart)
+                  pe        <- parseInstant(raw.periodEnd)
+                  // #1569 / #1757: decode each record individually via the shared per-
+                  // record skip helper. A single malformed record (e.g. a host value
+                  // that fails Hostname validation — an Akamai/CDN CNAME target with
+                  // underscores, see #1572) used to fail the WHOLE batch, dropping
+                  // every valid record with it. Now bad records are skipped, logged
+                  // (bounded), and metered, and the valid ones ingest.
+                  decResult <- decodePerRecord[UsageRecord](
+                    raw.records,
+                    "router usage",
+                    "record",
+                    router.id,
+                    AppMetrics.recordUsageRecordsRejected(_),
+                  )
+                  (records, rejectedCount) = decResult
                   _        <- LogContext.annotateAll(
                     LogContext.BatchSize -> records.size.toString,
-                    LogContext.Rejected  -> rejected.size.toString,
+                    LogContext.Rejected  -> rejectedCount.toString,
                   ) {
                     ZIO.logDebug(
                       s"router usage: router=${router.id} period=$ps..$pe records=${records.size} " +
-                        s"rejected=${rejected.size}",
+                        s"rejected=$rejectedCount",
                     )
                   }
                   _        <- ZIO.foreachDiscard(records)(r =>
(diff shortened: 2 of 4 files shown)


sameerparekh:
## Independent PR review

Reviewed against `docs/pr-review-checklist.md`. Read-only; cites `file:line`.

### BLOCKERS

1. **`api/src/metrics/Metrics.scala:95` — new `events_records_rejected_total{reason}` metric ships without a Grafana panel.** Per AGENTS.md §metrics-need-a-dashboard, a PR adding an emitted metric series must add or update a Grafana panel that consumes it, in the same PR. The analog `usage_records_rejected_total` has two dedicated panels in `deploy/grafana/dashboards/data-quality-ingest.json:302,338` (rate by reason + 24h sum). The new events-side counter has none — an operator watching the ingest dashboard cannot see the very signal this PR introduces. Add two parallel panels (rate-by-reason + 24h sum) in `data-quality-ingest.json` targeting `events_records_rejected_total`, with descriptions that mirror the existing usage entries and point at the `"router events: skipping malformed event"` warn-log. Slice only by `reason` (bounded enum).

### SHOULD-FIX

1. **`api/src/routes/RouterIngestRoutes.scala:73-81` vs `:152-167` — duplicated per-record decode + skip + log + meter loop.** The events path now mirrors the usage path's structure verbatim: `zipWithIndex → decode → collect Left/Right → logWarning per rejected → recordXRejected → annotateAll BatchSize/Rejected`. They differ only in the wire type (`UsageRecord` vs `RouterEvent`), the log prefix string ("router usage" vs "router events"), and the metric helper. Per AGENTS.md §single-source-of-truth, this is a "must mirror" coupling enforced by hand — exactly the kind that drifts. Extract a `decodePerRecord[T: JsonDecoder](raw: List[Json], logPrefix: String, recordMetric: Int => UIO[Unit])` helper that returns `(List[T], rejectedSize)` and call it from both handlers. Cheap win: each call site drops from ~13 lines to ~3.

### NITS

1. **`api/src/routes/RouterIngestRoutes.scala:152` — the rejected index `[$i]` is the JSON-array slot from `raw.events.zipWithIndex`**, which is the desired diagnostic (point at the bad slot in the wire payload, not the post-filter index). Fine, just worth a one-liner comment if you keep the inline copy.
2. **`api/test/src/feature/RouterIngestSpec.scala:133` — `eventsRejectedCounter` reads the default metric registry**, mirroring the existing `rejectedCounter` pattern. Before/after delta is correct; just note any test parallelism would race it.

### Dimension scoreboard

| # | Dimension | Status |
|---|---|---|
| 1 | Duplicated logic / single source of truth | ⚠️ SHOULD-FIX — events vs usage decode loop is a hand-copy |
| 2 | Testing integrity | ✅ RED (`f611edf8`) → GREEN (`640319fb`) visible; two feature tests assert status + persisted rows + metric delta |
| 3 | Architectural invariants | ✅ no router-agent / DNS / snapshot changes |
| 4 | Backwards compat / wire | ✅ purely additive server decoder; `RawEventsReport` is decode-only and `derives JsonDecoder` ignores unknown fields |
| 5 | Migrations | n/a |
| 6 | Metrics + dashboard | ❌ BLOCKER — new metric, no panel |
| 7 | Scope / hygiene | ✅ focused 3-file diff; no AI mentions; three-dot merge-base used |
| 8 | Security | ✅ no creds; no SQL |

### VERDICT: REQUEST-CHANGES

The Class A fix is correctly diagnosed and the per-record skip + meter mirrors #1574 cleanly, with TDD red→green visible.
Backwards-compat shape is clean — the decoder change is purely additive at the wire.
Merge-gating finding is the missing Grafana panel for `events_records_rejected_total`: the rule explicitly forbids shipping a metric without the panel that consumes it, and the analog usage metric has the exact panel pair this one needs.

sameerparekh:

Re-reviewed against the latest push (`e1cb2ebf`). Prior review covered `640319fb`.

### Prior findings status

| Class | Cite | Status |
|---|---|---|
| BLOCKER | `api/src/metrics/Metrics.scala:95` — missing Grafana panel for `events_records_rejected_total` | **ADDRESSED** — `deploy/grafana/dashboards/data-quality-ingest.json:521-598` adds the rate-by-reason timeseries (id 15) and 24h sum stat (id 16), mirroring the existing usage pair (ids 9/10), bounded label `reason` only. Dashboard description's metric list updated to include the new series. |
| SHOULD-FIX | `RouterIngestRoutes.scala:73-81` vs `:152-167` — duplicated per-record decode + skip + log + meter loop | **ADDRESSED** — `RouterIngestRoutes.scala:228-254` extracts `decodePerRecord[T: JsonDecoder]` that returns `(List[T], Int)`. Both handlers (`:74-80` for usage, `:155-161` for events) now call into it with `(logPrefix, itemNoun, routerId, recordMetric)`. The "must mirror" coupling is gone — the compiler enforces parity. |
| NIT | `RouterIngestRoutes.scala:152` — wire-slot index in the warn log | **ADDRESSED** (incidentally) — the index is now produced in one place inside the helper, so the wire-slot convention is in one location. |
| NIT | `RouterIngestSpec.scala:133` — test-parallelism counter race | **NOT-ADDRESSED** — still applies (unchanged). Still acceptable as a non-blocking note. |

### New findings (delta `640319fb...e1cb2ebf`)

#### BLOCKERS
(none)

#### SHOULD-FIX
(none)

#### NITS

1. **`RouterIngestRoutes.scala:228` — `decodePerRecord` swallows `JsonDecoder` via `implicit`**, then takes a `recordMetric: Int => UIO[Unit]`. The `Int => UIO[Unit]` arrow is fine, but `AppMetrics.recordUsageRecordsRejected` and `AppMetrics.recordEventsRecordsRejected` already have a default `reason="decode_error"` parameter — currently the helper hard-wires the implicit "decode_error" reason. Future-proof for a second reason enum would be a `decodePerRecord(..., reason: String = "decode_error")` knob, but YAGNI applies and the helper's call sites both currently want decode_error. Non-blocking.
2. **`data-quality-ingest.json:539,571` — `gridPos.y=56`**. Last existing panel was at y=48 (id 14). The new panels land cleanly below; just noting if more panels are added by parallel work, a refresh of all `y` values may be needed.

### Dimension scoreboard

| # | Dimension | Status |
|---|---|---|
| 1 | Duplicated logic / single source of truth | ✅ ADDRESSED via `decodePerRecord` helper |
| 2 | Testing integrity | ✅ RED → GREEN still visible; 50 tests pass after refactor |
| 3 | Architectural invariants | ✅ no router-agent / DNS / snapshot changes |
| 4 | Backwards compat / wire | ✅ purely additive server decoder; no wire shape change in the delta |
| 5 | Migrations | n/a |
| 6 | Metrics + dashboard | ✅ ADDRESSED — panels ship with metric |
| 7 | Scope / hygiene | ✅ focused delta; no AI mentions |
| 8 | Security | ✅ no creds; no SQL |

### VERDICT: APPROVE

Both prior findings ADDRESSED in the follow-up push. New Grafana panels mirror the existing usage-side ones cleanly and slice only by the bounded `reason` label.
The extracted `decodePerRecord` helper collapses the hand-copied loop into a single source of truth — both /usage and /events now share the decode + skip + log + meter path, with only the log prefix and metric helper differing.
No new BLOCKER or SHOULD-FIX introduced by the refactor.

PR #1763, opened 2026-06-16
fix(#1761): strip :port from host on agent emitters + tolerant API decoder

### LINKED ISSUE #1761: bug(prod): host=<fqdn>:443 ingested verbatim — Hostname validation rejects
## Symptom

Both `/api/router/usage` and `/api/router/events` reject records whose `host` field carries a `:port` suffix. Observed in prod:

```
13:57:23.345 WARN router usage: skipping malformed record[76] for
  router=f04dd490-…: .host(invalid hostname label 'com:443' in
  ws.nas.native-cloud.com:443)

14:13:39.310 WARN POST /api/router/events -> 400 body=.events[2].host
  (invalid hostname label 'com:443' in ws.nas.native-cloud.com:443)
```

`Hostname` validation rejects the `com:443` label — the trailing `:port` is bleeding into the wire `host` value. The hostname is a real one (`ws.nas.native-cloud.com`); the bug is in whatever path appends the port.

## Onset

Usage rejections began at **2026-06-13 20:00** ([Grafana panel](https://wifihaven.grafana.net/d/wifihaven-data-quality-ingest/wifihaven-e28094-data-quality-and-ingest?orgId=1&from=2026-06-09T14:30:34.667Z&to=2026-06-16T14:30:34.667Z&timezone=browser&var-env=prod&var-datasource=grafanacloud-prom&refresh=30s&viewPanel=panel-9)). Step-function shape suggests a router-side change (agent version bump or new attribution code path emitting `host:port`).

Until 2026-06-16 the events analog of this (`/api/router/events`) failed the WHOLE batch on this same `host:port` value — fixed by #1759 (per-event skip). The events warn quoted above is post-fix.

## Likely source

`host` is supposed to be just a hostname, not `host:port`. Candidates:
1. A new agent code path attributing flows by SNI / Host header is including the `:port`.
2. A nflog / conntrack parse split point that no longer strips the port.
3. The `ws.nas.native-cloud.com` flow specifically is observed on `:443` and an attribution path passes the SNI string verbatim.

`grep -RIn ':443\|host:port\|sni' openwrt/files/usr/lib/lua/wifihaven` is the starting point — look at recent changes around 2026-06-13.

## Fix shape

Two layers, both belong:

1. **Agent-side (root cause):** strip any trailing `:port` from the `host` value at the emitter — every wire path that emits `host` should normalize to a port-free hostname. One central helper.
2. **API-side (defense in depth):** in the wire decoder for `UsageRecord.host` / `RouterEvent.host`, when the raw value matches `^(.+):\d+$`, transparently strip the port before validating as `Hostname`. Backwards-compat: older agents still emitting `host:port` are accepted; newer ones emit clean. (Per #376 the API decoder layer is the right place for forgiving server-side normalization, not for adding a new field shape.)

Either alone closes the symptom; doing both keeps the system robust against the next agent variant.

## Acceptance

- Records with `host=foo.example.com:443` no longer count as rejections on either ingest path.
- A feature test pins the API-side tolerant decode (POST a `host:port`, assert ingest 200 + persisted `host=foo.example.com`).
- An openwrt-side test pins the emitter normalization.
- Rejection rate on the data-quality-ingest dashboard for both `usage_records_rejected_total` and `events_records_rejected_total` returns to baseline.

## Related

- #1572 — underscore-label CDN CNAME class (#1569 / #1574 / #1757 family).
- #1759 — events per-record skip (the symptom above on /events was masking this until #1759 landed).

sameerparekh:
## Summary

Closes #1761.

Both ends of the router↔API wire now normalize an fqdn host that carries a
trailing `:port` (e.g. `ws.nas.native-cloud.com:443`). Before this change
such a value reached the API verbatim, where `Hostname.parse` rejected
`com:443` as an invalid hostname label and the record was metered as a
decode failure — the prod symptom on the data-quality-ingest dashboard
panel id 9 from 2026-06-13 20:00 onward.

## Root cause

Some attribution path (the prime suspect is the recent SNI / Host-header
work — #573 / #1707 / #1711 / #1731 landed in the onset window) is
interning a `host:port` string into the ip→hostname cache, so the wire
emitters in `conntrack.lua`, `nflog.lua`, and `usage.lua` all pass it
through verbatim. Rather than chase the single producer (any future
emitter could re-introduce the same suffix), the fix normalizes at every
wire-emit site and the API tolerates the legacy shape too.

## Fix shape

**Agent side (root-cause normalization).** New
`openwrt/.../host_norm.lua` exports a single `strip_port_suffix(host)`
helper. The three wire-emit sites that construct `{type="fqdn", value=…}`
— `conntrack.build_event`, `nflog.build_event`, and
`usage.host_for_ip` — all route the value through the helper, so the
normalization is a single source of truth per AGENTS.md
§single-source-of-truth. The helper is conservative around bare and
bracketed IPv6 literals (a bare IPv6 isn't expected to reach this path
since v6 destinations are emitted as `type="ipv6"`, but the helper is
safe regardless).

**API side (defense in depth, back-compat).** `HostId`'s JSON codec
strips a trailing `:<digits>` from an fqdn wire value before calling
`Hostname.parse`. This keeps the API tolerant of pre-fix agents
already in the field (additive per the AGENTS.md wire back-compat
policy) and of any future emitter that re-introduces the suffix.

## Tests

Two failing tests landed first in a separate RED commit, then the fix
makes them green:

- `openwrt/test/host_norm_spec.lua` — `strip_port_suffix` itself
  (idempotent, IPv6-safe).
- `openwrt/test/conntrack_spec.lua` — `build_event` strips `:port`
  before emitting.
- `api/test/src/feature/RouterIngestSpec.scala` — `/api/router/usage`
  and `/api/router/events` accept a record with
  `host = ws.nas.native-cloud.com:443`, persist under the port-free
  hostname, and the `usage_records_rejected_total` counter does NOT
  advance.

Red→green is visible in the commit history (test commit precedes fix).

## Verification plan

- `mill api.test.testOnly wifihaven.api.feature.RouterIngestSpec` — 50/50 pass locally.
- `mill shared.test` — pass.
- `openwrt/test/run_tests.sh` — host_norm_spec / conntrack_spec / nflog_spec / usage_spec pass. (Existing nft_log_dep_spec failures are unrelated to this PR — pre-existing build-script kmod-nf-log deps.)
- `mill mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll` — clean.
- Post-deploy: watch the data-quality-ingest dashboard's
  `usage_records_rejected_total{reason="decode_error"}` series return
  to baseline; ditto the equivalent on `/events` once #1759 also lands.

## Scope-out

This PR keeps the port data discarded (consistent with the issue
acceptance). Capturing port end-to-end (UI surface, allow/block on
host:port) is #296 and will be tackled separately to avoid bundling a
migration on the growth tables with a live prod hotfix.

## Test plan

- [x] Lua specs pass (`host_norm_spec`, `conntrack_spec` `build_event`).
- [x] Scala specs pass (`RouterIngestSpec` usage + events).
- [x] `scalafmt` clean.
- [ ] Post-deploy: `usage_records_rejected_total{reason="decode_error"}` returns to pre-2026-06-13 baseline.

### DIFF
--- a/api/test/src/feature/RouterIngestSpec.scala
+++ b/api/test/src/feature/RouterIngestSpec.scala
@@ -1595,6 +1595,58 @@ object RouterIngestSpec
         // exactly the one malformed record was metered as rejected
         assertTrue(after.count - before.count == 1.0)
     },
+    // ── #1761: tolerant decode of `host:port` fqdn values ───────────────────
+    test("usage: an fqdn host with a trailing :port is accepted (port stripped)") {
+      // Prod incident #1761: an attribution path (SNI / Host-header capture)
+      // bleeds the destination port into the wire `host` field, so values like
+      // `ws.nas.native-cloud.com:443` arrived verbatim. Hostname validation
+      // rejected them as "invalid hostname label 'com:443'" and the record
+      // was metered as a decode failure. The tolerant decoder strips a
+      // trailing `:<digits>` from fqdn values before parsing, so the record
+      // ingests normally and the rejection counter does NOT advance.
+      for {
+        _        <- cleanDb
+        rRepo    <- ZIO.service[RouterRepo]
+        pRepo    <- ZIO.service[ProfileRepo]
+        dRepo    <- ZIO.service[DeviceRepo]
+        tu       <- ZIO.service[TimeUsageRepo]
+        routes   <- buildRoutes
+        _        <- seedKnownDevice(dRepo, pRepo)
+        (id, tk) <- seedRouter(rRepo)
+        before   <- rejectedCounter.value
+        rec  =
+          s"""{"mac":"$knownMac","host":{"type":"fqdn","value":"ws.nas.native-cloud.com:443"},"activeSeconds":60,"bytesIn":100,"bytesOut":50}"""
+        body =
+          s"""{"routerId":"$id","periodStart":"${periodStart.toString}","periodEnd":"${periodEnd.toString}","records":[$rec]}"""
+        resp  <- post(routes, "/api/router/usage", body, Some(tk))
+        sb    <- tu.getSecondsAndBytes(
+          MacAddress.unsafe(knownMac),
+          HostId.Fqdn(Hostname.unsafe("ws.nas.native-cloud.com")),
+          testDate,
+        )
+        after <- rejectedCounter.value
+      } yield assertTrue(resp.status == Status.Ok) &&
+        // persisted under the port-free hostname
+        assertTrue(sb == ((60L, 100L, 50L))) &&
+        // and the rejection counter did NOT advance
+        assertTrue(after.count - before.count == 0.0)
+    },
+    test("events: an fqdn host with a trailing :port ingests (port stripped)") {
+      for {
+        _        <- cleanDb
+        rRepo    <- ZIO.service[RouterRepo]
+        pRepo    <- ZIO.service[ProfileRepo]
+        dRepo    <- ZIO.service[DeviceRepo]
+        routes   <- buildRoutes
+        _        <- seedKnownDevice(dRepo, pRepo)
+        (id, tk) <- seedRouter(rRepo)
+        ev   =
+          s"""{"type":"connection_attempt","mac":"$knownMac","host":{"type":"fqdn","value":"ws.nas.native-cloud.com:443"},"destIp":"1.2.3.4","allowed":false,"reason":"category:adult","ts":"2026-05-07T14:01:14Z"}"""
+        body =
+          s"""{"routerId":"$id","events":[$ev]}"""
+        resp <- post(routes, "/api/router/events", body, Some(tk))
+      } yield assertTrue(resp.status == Status.Ok)
+    },
     test("usage: a batch of ONLY malformed records is accepted (200) and meters them all") {
       // No valid records to ingest, but the envelope is well-formed — so this is
       // a 200 no-op-ingest, NOT a 400. The bad records are all metered.
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -24,6 +24,7 @@
 local M = {}
 
 local static_ip_labels = require("wifihaven.static_ip_labels")
+local host_norm        = require("wifihaven.host_norm")
 
 -- ---------------------------------------------------------------------------
 -- eb_san(host) -> string
@@ -469,7 +470,9 @@ function M.build_event(opts)
   if opts.host_label_source and opts.hostname then
     host = { type = "label", value = opts.hostname, source = opts.host_label_source }
   elseif opts.hostname then
-    host = { type = "fqdn", value = opts.hostname }
+    -- #1761: strip a trailing :<port> so the API's Hostname validation
+    -- doesn't reject the record. Single-source via host_norm.
+    host = { type = "fqdn", value = host_norm.strip_port_suffix(opts.hostname) }
   else
     local kind = (opts.dest_ip and opts.dest_ip:find(":", 1, true)) and "ipv6" or "ipv4"
     host = { type = kind, value = opts.dest_ip }
--- a/openwrt/files/usr/lib/lua/wifihaven/host_norm.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/host_norm.lua
@@ -1,0 +1,47 @@
+-- host_norm.lua — single-source wire-host normalization (#1761).
+--
+-- A handful of attribution paths (SNI / Host-header capture, future plumbing
+-- that interns "fqdn:port" pairs) can hand the wire emitters a hostname whose
+-- value carries a trailing `:<port>` suffix. The API's Hostname validation
+-- treats the port digits as part of the last label and rejects the record
+-- ("invalid hostname label 'com:443'"), so the bad value 4xxs at ingest and
+-- shows up on the prod data-quality-ingest dashboard.
+--
+-- The fix is to normalize at every wire-emit site (the three places that
+-- construct `{type = "fqdn", value = h}` — conntrack.build_event,
+-- nflog.build_event, usage.host_for_ip). All three call strip_port_suffix
+-- so the helper, not a pattern, is the single source of truth.
+--
+-- This is the agent half of the back-compat pair: the API also accepts a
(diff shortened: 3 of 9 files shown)


sameerparekh:
## Independent PR review

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. [openwrt/files/usr/lib/lua/wifihaven/host_norm.lua:38](https://github.com/wifihaven/wifihaven/blob/97ce8ec4ad11fd09f8f3734c7d28d707b3e7cb60/openwrt/files/usr/lib/lua/wifihaven/host_norm.lua#L38) vs [shared/src/types/HostId.scala:73](https://github.com/wifihaven/wifihaven/blob/97ce8ec4ad11fd09f8f3734c7d28d707b3e7cb60/shared/src/types/HostId.scala#L73) — minor behavioural asymmetry on a bracketed IPv6 literal: Lua's `strip_port_suffix("[::1]:443")` returns `"[::1]"` while Scala's `stripPortFromFqdn` regex `^([^:\[\]]+):\d+$` explicitly excludes brackets, so it returns the input unchanged. In practice this is fine — a bracketed v6 host is emitted as `type="ipv6"` and never reaches the fqdn branch — but the two halves of the pair behave differently on the same input, which a future reader will find surprising. Either drop the bracket case from the Lua helper (matches Scala) or extend the Scala regex to accept `[…]:port` (matches Lua). Non-blocking; the live wire path is unaffected.
2. [shared/src/types/HostId.scala:73](https://github.com/wifihaven/wifihaven/blob/97ce8ec4ad11fd09f8f3734c7d28d707b3e7cb60/shared/src/types/HostId.scala#L73) — no unit test pinning the Scala `stripPortFromFqdn` directly. The `RouterIngestSpec` feature tests exercise it through `/api/router/usage` and `/api/router/events`, which is adequate, but a one-liner spec in `shared/test/.../HostIdSpec.scala` (round-trip, idempotence, IPv6 untouched) would localise regressions in a future refactor of `HostId`'s codec.

PRIOR FINDINGS
(no prior review — first run)

DELTA REVIEWED
`git diff origin/main...HEAD` (8 files, +191/−5).

DIMENSIONS APPLIED
1. **Single source of truth.** Lua side: one helper, three call sites. Scala side: one private function, one call site. Cross-process duplication is structural (different runtimes, same wire contract) and explicitly explained in the HostId.scala block comment — not the "must mirror" smell the checklist targets. No re-derivation of an existing primitive.
2. **Testing integrity.** TDD red→green visible: `0791d692` (RED, failing tests only) → `97ce8ec4` (GREEN, fix). No existing-test weakening. The bug-fix test pins the actual `:port` symptom (rejection counter must NOT advance) rather than adjacent behaviour. Feature-level + Lua unit, no repo/Auth/Clock mocks.
3. **Architectural invariants.** Router stays a dumb applier — the helper is data normalisation, not policy. No new snapshot field. DNS-as-enforcement reasoning absent.
4. **Backwards compatibility.** Tolerant decoder is additive: accepts both `foo.com:443` (legacy / mid-roll) and `foo.com` (post-fix). Encoder unchanged, so round-trip is clean. Older agents → newer API ingests; newer agents → older API still ingests (unchanged clean values).
5. **Migrations.** None.
6. **Metrics + dashboard.** No new metric needed — the change reduces an existing `usage_records_rejected_total{reason="decode_error"}` counter, which is the canonical observability. The PR description's "post-deploy: counter returns to baseline" is the right verification.
7. **Scope / hygiene.** Focused diff (8 files, no drive-by). No Claude/Anthropic mentions in commit messages, PR title, or body.
8. **Security.** No new routes, no new SQL, no secrets.

VERDICT: APPROVE

Hotfix scoped exactly to the issue: agent-side `strip_port_suffix` helper called at the three fqdn-emit sites, API-side tolerant decoder for back-compat with already-deployed agents. No BLOCKERS or SHOULD-FIX. Two NITs flag minor cross-half asymmetry and a missing unit test for the Scala half — either can land later; neither holds up the prod fix.

sameerparekh:

PRIOR FINDINGS

| 1 | NIT   | Lua/Scala asymmetry on bracketed-v6 `[v6]:port` ([shared/src/types/HostId.scala:73](https://github.com/wifihaven/wifihaven/blob/97ce8ec4ad11fd09f8f3734c7d28d707b3e7cb60/shared/src/types/HostId.scala#L73)) | **ADDRESSED** — Scala decoder now has a second regex `BracketedV6PortSuffix` that strips `[v6]:port` → `[v6]`, matching the Lua helper. Behaviour table is in the block comment. |
| 2 | NIT   | No direct unit test for the Scala `stripPortFromFqdn` | **ADDRESSED** — two new specs in [shared/test/src/types/HostIdSpec.scala:115](https://github.com/wifihaven/wifihaven/blob/ef16bb97ddf94b0826caee1fc2252580de640da0/shared/test/src/types/HostIdSpec.scala#L115): one positive (strip + idempotent), one negative (`:443` and bare v6 still reject). |

DELTA REVIEWED
`git diff 97ce8ec4...ef16bb97` — `shared/src/types/HostId.scala` (+12/−5) and `shared/test/src/types/HostIdSpec.scala` (+32). No router-side, ingest, migration, or wire-shape changes in the delta.

NEW FINDINGS

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
(none)

DIMENSIONS RE-APPLIED TO THE DELTA
1. **Single source of truth.** The new bracketed-v6 case is co-located with the plain `host:port` case in the same `stripPortFromFqdn` function — still one source. The comment now states the three-case table once, in both halves.
2. **Testing integrity.** New specs are pure JsonCodec round-trip assertions; no mock, no fixture weakening, and the negative cases (`":443"`, `"::1"`) pin that the strip can't silently turn invalid input into a valid Fqdn. Existing assertions in the file untouched.
3. **Architectural invariants.** No change.
4. **Backwards compatibility.** Still additive — the new branch only widens what is *accepted*, never the encoded shape.
5. **Migrations.** None.
6. **Metrics + dashboard.** No change.
7. **Scope / hygiene.** Delta is limited to the two files the prior NITs called out; no drive-by edits, no Claude/AI mentions.
8. **Security.** No change.

VERDICT: APPROVE

Both prior NITs ADDRESSED with the smallest possible delta (one extra regex case in `HostId.scala`, two specs in `HostIdSpec.scala`). No new findings; no open BLOCKER. Hotfix is ready to merge.

Issue #1320, opened 2026-06-02
web: global allow/block management + audit view, per-profile default-deny toggle, device override editor


sameerparekh:
Part of the global policy layer design ([#1308](https://github.com/wifihaven/wifihaven/issues/1308)). See [`docs/design/global-policy-layer.md`](docs/design/global-policy-layer.md) §7 and §8 (item 5). Depends on the API surface from the PolicyService + migration issues.

## Scope (`web/`)
- **Global always-allow management page** with a read-only audit view (who changed the list and why, from the `global_allow` audit columns). This is a security-sensitive bypass surface — surface the audit trail prominently.
- **Global-blocks management** (hosts + category lists + the network-lockdown / network-wide strict-mode toggles).
- **Per-profile default-deny toggle.**
- **Per-device override editor** reflecting replace semantics — warn the operator that overriding a default-deny profile fully restates its rules (it does not merge).

## Notes
- The UI/block-page hosts and the connectivity/PKI hosts share one list (`global.extraAllowed`); the *reason* is metadata, not a separate field.

sameerparekh:
Superseded by #1769 (collapse global_* tables into is_global sentinel profile). The replacement SPA work lives in #1773 — it reuses the existing per-profile editor preset to globalProfile.id rather than building a parallel surface. Closing here as superseded; the global-policy SPA work continues under #1773.

PR #1768, opened 2026-06-16
feat(#1456): field-scoped PATCH on /api/global — idempotent ops avoid multi-client races

### LINKED ISSUE #1456: Global policy config: field-scoped/granular updates to avoid multi-client races (flags PATCH, blocklist add/remove)
The global-policy management endpoints added in [#1320](https://github.com/wifihaven/wifihaven/issues/1320) / PR #1453 mutate config with **full-document replace PUTs**, which last-write-wins clobber a concurrent admin's edit:

- `PUT /api/global/flags` — sets `blocked` + `blockReason` + `blockIpOnly` together. Two admins toggling *different* flags concurrently: whoever saves last reverts the other's flag.
- `PUT /api/global/blocklists` — full-set replace of the household-global category set. Concurrent category edits clobber each other.

(The allow/block host endpoints are already granular — `POST`/`DELETE` per host — so they don't have this race.)

## Ask

Make global-policy config mutations field-scoped / granular so two admins on the UI don't stomp each other:

- `PATCH /api/global/flags` — patch only the fields present in the body (mirror the `PATCH /api/household/settings` field-patch pattern in `HouseholdSettingsRoutes`).
- For categories, prefer granular add/remove (`POST`/`DELETE /api/global/blocklists/:id`) over the wholesale `PUT`, matching the allow/block host shape. Keep the `PUT` only if a deliberate "set the whole list" affordance is wanted; otherwise retire it after the SPA moves over.

Then update `web/src/api/client.ts` (`api.global.*`) and `GlobalPolicyPage.tsx` to use the granular calls.

## Context

This is the same class of problem as the **profile** full-`PUT`, tracked separately in [#423](https://github.com/wifihaven/wifihaven/issues/423) (Add `PATCH /api/profiles/:id`). The field-patch precedent already exists for `household`, `devices`, `users`, and `apps`. This issue covers only the global-policy surface; #423 covers profiles.

Part of the global policy layer ([#1308](https://github.com/wifihaven/wifihaven/issues/1308)).

sameerparekh:
Closes #1456.

## Summary

Two new admin write surfaces on the household-global policy that let disjoint-field concurrent edits commute (the existing wholesale PUTs are last-write-wins):

- **`PATCH /api/global/flags`** — field-scoped. Omit-means-no-change; `blockReason: null` clears the column. Backed by a new `GlobalPolicyRepo.patchFlags` whose SQL `UPDATE` only mentions the patched columns, so two PATCHes that touch different flags compose server-side.
- **`POST /api/global/blocklists/:id`** / **`DELETE /api/global/blocklists/:id`** — idempotent per-category add/remove (mirrors the existing allow/block-host shape). `ON CONFLICT DO NOTHING` insert + a no-row-required DELETE.

## Shape — A (separate routes) vs B (tagged-op array)

Went with **Shape A**: each operation maps cleanly to one RESTful route, no operation-language to maintain. Each call is independently idempotent so disjoint-id concurrent calls just commute at the DB layer — no operation log needed on the wire.

## Backwards-compat

Strictly additive per `AGENTS.md#wire-contract`:
- `PUT /api/global/flags` and `PUT /api/global/blocklists` remain. The SPA still calls them.
- New SPA client methods (`api.global.patchFlags`, `addBlocklist`, `removeBlocklist`) are exposed but no caller is switched here.

## SPA switchover

Out of scope. #1320 (the global-policy SPA editor) is the call site that should move over to the granular API once the editor wants the no-clobber property.

## Test plan

- [x] `mill api.test` — 73 tests pass, including 3 new ones:
  - `PATCH /api/global/flags` updates only the fields present in the body
  - `PATCH /api/global/flags` — concurrent edits to different fields both stick
  - `POST/DELETE /api/global/blocklists/:id` add and remove idempotently
- [x] `mill scalafmtCheckAll` clean

### DIFF
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -390,6 +390,22 @@ trait GlobalPolicyRepo {
   def setFlags(blocked: Boolean, reason: Option[MacBlockReason], blockIpOnly: Boolean): Task[Unit]
 
   /**
+   * #1456: field-scoped patch of the flat global flags so two concurrent admin edits to *different*
+   * fields don't clobber each other (the wholesale [[setFlags]] last-write-wins). Each parameter
+   * uses an outer Option to mean "field present in the patch?":
+   *   - `blocked = None` / `blockIpOnly = None` → no change to that column.
+   *   - `blockReason = None` → no change to the column.
+   *   - `blockReason = Some(None)` → write NULL (clear).
+   *   - `blockReason = Some(Some(r))` → assign `r`.
+   * A single SQL `UPDATE` runs server-side so disjoint-field concurrent calls commute.
+   */
+  def patchFlags(
+      blocked: Option[Boolean],
+      blockReason: Option[Option[MacBlockReason]],
+      blockIpOnly: Option[Boolean],
+  ): Task[Unit]
+
+  /**
    * Full audit history (active + soft-deleted) of the always-reachable allow set, newest first.
    * Unlike [[get]], this surfaces the `reason` / `added_by` / `removed_by` columns (resolved to
    * usernames) so the #1320 SPA can render the security-sensitive bypass-list audit trail. The
@@ -402,6 +418,13 @@ trait GlobalPolicyRepo {
 
   /** Replace the household-global category set (`global.blocklistIds`) wholesale. */
   def setBlocklists(ids: List[BlocklistId], addedBy: Option[Long]): Task[Unit]
+
+  /**
+   * #1456: idempotent remove of a single category from the household-global set. Matches the
+   * granular allow/block-host shape so concurrent edits to different categories don't clobber each
+   * other (the wholesale [[setBlocklists]] does). No-op if `id` is not currently associated.
+   */
+  def removeBlocklist(id: BlocklistId): Task[Unit]
 }
 
 /**
@@ -420,9 +443,15 @@ object NoopGlobalPolicyRepo extends GlobalPolicyRepo {
   def addBlocklist(id: BlocklistId, addedBy: Option[Long]): Task[Unit]                    = ZIO.unit
   def setFlags(blocked: Boolean, reason: Option[MacBlockReason], blockIpOnly: Boolean): Task[Unit] =
     ZIO.unit
+  def patchFlags(
+      blocked: Option[Boolean],
+      blockReason: Option[Option[MacBlockReason]],
+      blockIpOnly: Option[Boolean],
+  ): Task[Unit] = ZIO.unit
   def allowAudit: Task[List[GlobalPolicyAuditEntry]]                           = ZIO.succeed(Nil)
   def blockAudit: Task[List[GlobalPolicyAuditEntry]]                           = ZIO.succeed(Nil)
   def setBlocklists(ids: List[BlocklistId], addedBy: Option[Long]): Task[Unit] = ZIO.unit
+  def removeBlocklist(id: BlocklistId): Task[Unit]                             = ZIO.unit
 }
 
 trait TimeUsageRepo {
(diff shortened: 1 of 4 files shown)


sameerparekh:

Reviewed against the 8-dimension checklist in `docs/pr-review-checklist.md`. Read-only pass, no files modified.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS

1. `api/src/db/Repos.scala:1186` (`removeBlocklist`) — hard `DELETE` with no audit trail (no `removed_by` / `removed_at`). The existing wholesale `setBlocklists` has the same gap (it hard-deletes too) and the `global_blocklists` schema doesn't have soft-delete columns, so this is parity not regression — but per #1456's spirit of making this surface "the granular shape," it's worth filing a follow-up to add soft-delete + `removed_by` (mirrors `global_allow` / `global_blocks` audit shape).
2. `api/src/routes/GlobalPolicyRoutes.scala:160` (`DELETE /api/global/blocklists/:id`) — doesn't resolve the acting admin's `uid` (no call to `actingUserId`). Consistent with the no-audit-column reality above; once #1 lands, this should capture `removedBy`.
3. `api/src/routes/GlobalPolicyRoutes.scala:146,160` — `BlocklistId.unsafe(id)` accepts an arbitrary URL path segment. If the id isn't in the `blocklists` table the `addBlocklist` insert will (if there's an FK) fail with a 500 rather than a 422/404. Matches the existing wholesale `setBlocklists` behaviour, so this is consistency not a regression — flag as a follow-up.
4. `api/src/routes/GlobalPolicyRoutes.scala:200-224` — the `case FieldPatch.Set(v) => Some(v); case _ => None` extractors for `blockedOpt` / `blockIpOnlyOpt` could be a one-liner helper (e.g. `FieldPatch#toOptionRequired`) to avoid repeating the pattern three times in slightly different shapes. Pure ergonomics.
5. `api/src/db/Repos.scala:1144` — `fr"global_block_reason=NULL"` writes the literal `NULL`. Safe (no user input), but `setFlags` interpolates `Option[String]` which doobie translates to NULL automatically; using the same shape here (`val s: Option[String] = r.map(MacBlockReason.asString); fr"global_block_reason=$s"`) would keep one idiom across the two methods. (The original draft used this and was discarded only because the *read path* defaults a NULL reason back to `Manual` when `blocked=true` — that's a read-side quirk, not a write-side concern, so either form is fine.)
6. PR body / commit messages contain "🤖 Generated with Claude Code" / `Co-Authored-By: Claude…` mentions. Per `docs/pr-review-checklist.md` §7 the rule reads "NO Claude / Anthropic / AI mentions in commit messages or the PR title / body. A `Co-Authored-By` trailer is fine." The trailer is allowed; the "Generated with Claude Code" footer in the PR body is technically out of policy. Flagging per the checklist; non-blocking.

### Dimension-by-dimension notes

1. **Single source of truth.** PATCH route extracts FieldPatch tri-state and forwards. No re-derivation of `blocked` / `blockReason` / time-state logic. ✓
2. **Testing integrity.** TDD visible — separate `test(#1456): RED` commit (`371f522c`) before `feat(#1456):` (`4140c56d`). Three new tests assert (a) omit-means-no-change, (b) concurrent disjoint-field PATCH commute, (c) idempotent add/remove. No existing test was weakened or deleted. Feature-level (`/api/...` through embedded Postgres), no mocked repos. ✓
3. **Architectural invariants.** No router-side changes. No new wire field — leverages existing flag columns + the `global_blocklists` set. ✓
4. **Backwards compat.** Strictly additive: existing `PUT /api/global/flags` and `PUT /api/global/blocklists` untouched; new endpoints added under new method (`PATCH`) and new path (`/blocklists/:id`). SPA still calls the PUT shape — `setBlocklists` / `setFlags` not removed. ✓
5. **Migrations.** None. ✓
6. **Metrics + dashboard.** Three new routes; all covered by the existing route-templated HTTP middleware histogram/status series, so per the convention no bespoke metric is required. ✓
7. **Scope / hygiene.** Diff limited to the 4 intended files; no drive-bys. AI-mention nit above. ✓ (modulo NIT 6)
8. **Security.** All three new routes call `requireAdmin`. SQL is doobie-interpolated (`fr"…=$b"`) — no string concatenation. The literal `NULL` in `patchFlags` is not user-controlled. The Json.Obj-based FieldPatch parsing safely handles malformed bodies via a 400. ✓

### VERDICT: APPROVE

`PATCH /api/global/flags` + granular `POST`/`DELETE /api/global/blocklists/:id` cleanly map disjoint-field race avoidance onto the DB layer (single-statement `UPDATE`, idempotent insert/delete). Tests pin the no-clobber + idempotent properties through embedded Postgres. Six non-blocking NITs; nothing merge-gating.

PR #1778, opened 2026-06-18
feat(#1777): reconcile -template-suffixed app slugs onto canonical (idempotent admin route)

### LINKED ISSUE #1777: bug(prod): -template-suffixed app slugs + duplicate operator-added apps — collapse onto canonical template slugs
## Symptom (prod, observed 2026-06-16)

Two related issues in prod's app table after the #1705 template-seeding wave:

1. **`-template`-suffixed slugs in prod.** Apps from `contract/apps/` (or wherever templates live) got inserted into prod with `<slug>-template` rather than the canonical `<slug>`. Likely the seeding logic added the suffix to avoid colliding with operator-added apps.
2. **Duplicate apps** — operator-added apps with the canonical slug coexist with the `-template`-suffixed ones. E.g. an operator-added `youtube` AND a seeded `youtube-template`. Usage / assignments may be split across the pair, which inflates "Other" or splits per-app caps unpredictably.

## What we want after cleanup

- Prod's apps table matches the templates 1:1.
- `<slug>` (no `-template` suffix) is the canonical slug.
- Operator-added apps that DUPLICATE a templated app collapse INTO the templated entry — their host-set additions (if any beyond the template's host-set) merge into the template's host-set; their app-policy-assignments + app-time-limits + usage rollup rows reattach to the canonical app id; the duplicate row deletes.
- Operator-added apps that DON'T have a template counterpart (truly novel apps the operator created) stay as-is — they're not in scope for this cleanup.

## Reconciliation rules (per app)

For each app currently in prod, decide:

| Current shape | Action |
|---|---|
| Slug `X-template`, NO `X` co-exists | Rename slug `X-template` → `X`. Update all FK references (app_policy_assignments, app_used_daily, app_time_limits, etc.). |
| Slug `X-template` AND `X` co-exist | Merge `X-template` INTO `X`: union host-sets, reattach `X-template`'s assignments/rollups to `X`, delete `X-template`. Operator's `X` may have host-set additions not in the template — KEEP those (operator's customizations win for the host-set membership; the template represents the floor, not the ceiling). |
| Slug `X`, has a template counterpart but no `-template` row | Reconcile host-set: union the template's host-set into `X`'s. No row change. |
| Slug `X`, no template counterpart | Skip — operator-only app, out of scope. |

## Implementation strategy

The schema constraint on `app_slug` UNIQUE makes the rename order-sensitive. Options:

**Option A — two-step inside one DB transaction (preferred):**
- Per pair where both `X-template` and `X` exist: reattach all `X-template` FK refs to `X`, delete `X-template` row. Then merge host-set rows: insert any host present in `X-template` but not `X` (via the template's host-set).
- Per `X-template` with no `X`: rename slug `X-template` → `X` (no FK update needed; FKs reference app id, not slug).

**Option B — feature-flagged admin endpoint:**
- `POST /api/admin/apps/reconcile-templates` that does the above in a transaction.
- Idempotent — operator can re-run if it fails midway.

Default to **Option B** — operator-driven, observable, idempotent. Avoids running it inside a Flyway migration where rollback is harder.

If the chip's investigation shows the prod data is simpler than expected (e.g. zero true duplicates), a one-shot SQL script the operator runs once is also acceptable.

## Acceptance

- No prod app slug ends in `-template`.
- For every template in `contract/apps/`, prod has exactly one app row whose slug matches.
- Operator-only apps untouched.
- Usage rollup rows + app-policy-assignments + app-time-limits no longer reference any `-template` slug.
- A regression test: a feature test against embedded Postgres that seeds the messy state (with both `X-template` and `X` rows + assignments + rollups) and asserts the reconciliation produces the clean state.

## Out of scope

- Changing the template-seeding code itself to NOT add the `-template` suffix going forward — that's a separate follow-up (file once we know whether the seeding is currently SHIPPED, IDLE, or DEAD per the #1608 one-shot-migration delete-when-done convention).

## Related

- #1705 — app templates work (likely source of the `-template` suffix)
- #1608 — one-shot migration / backfill audit (the seeding code's lifecycle question)
- #1769 — global-policy collapse (similar shape: collapse duplicated schemas onto canonical)

sameerparekh:
Closes #1777.

## What

Adds an idempotent reconciler that collapses every `<slug>-template`-suffixed
`apps` row onto its canonical `<slug>` form. The four outcomes from the
issue's rules table are computed in one place (`AppReconciler.reconcileTemplates`)
and exposed via an admin POST route.

The `-template` suffix originates in [AppTemplates.findFreeSlug](api/src/AppTemplates.scala:252)
— when the seed pass finds the canonical slug already occupied by an
operator-added app, it falls back to `<slug>-template`. The fallback is
correct as a uniqueness guard, but it leaves the household with two app
rows for the same logical app and splits assignments / usage across them.

## How

- `AppRepo.mergeAppInto(from, to)` (one DB transaction):
  union hosts, reattach `app_policy_assignments` / `traffic_hourly_apps` /
  `traffic_daily_apps` / `app_used_daily` FK refs (canonical wins on PK
  collisions; `engaged_seconds` are summed on `(profile_id, date)` overlap),
  transfer `template_id` to canonical if it lacks one, then delete the
  duplicate row. `app_hosts_version` is bumped so rollup attribution is
  re-evaluated.
- `AppReconciler.reconcileTemplates(repo, templates)` — single function
  that decides which of the four cases applies per template and calls the
  repo primitives. **No duplicate merge logic across migration / route /
  test** (AGENTS.md §single-source-of-truth).
- `POST /api/admin/apps/reconcile-templates` — admin auth required;
  returns a per-bucket summary (`mergedSlugs`, `renamedSlugs`,
  `hostsUnioned`, `alreadyClean`, `created`).

## Whack-a-mole guard (re: chip abort criterion)

The seeder remains active. When both `<slug>` and `<slug>-template` co-exist,
the reconciler transfers `template_id` to the canonical row before deleting
the suffixed one. The next `AppTemplates.seed` pass then finds the canonical
row via `findByTemplateId` and never reaches the `-template` fallback.

## Wire impact

Slug appears on the wire only via opaque event reason strings
(`app:<slug>`, `app_time_limit:<slug>`) — routers echo these verbatim and
they do not feed enforcement. The per-MAC snapshot `BlockRules` shape
keys apps by id, not slug. **No wire-contract change.**

## Test

`AppReconcilerSpec` (committed RED before the implementation landed) seeds
the messy state — operator-added `youtube` with one assignment + an extra
host, seeded `youtube-template` with a different assignment + a distinct
host — runs the reconciler, and asserts a single `youtube` row with both
assignments reattached and the host-set unioned. Plus rename-only,
already-clean, and idempotency cases.

## Operator pre-flight (prod)

1. Take a prod DB snapshot.
2. Curl the new route:
   \`\`\`
   curl -X POST -H \"Authorization: Bearer \$ADMIN_JWT\" \\
     https://api.wifihaven.net/api/admin/apps/reconcile-templates
   \`\`\`
3. Re-query: \`SELECT slug FROM apps WHERE slug LIKE '%-template'\` —
   expect zero rows.
4. Spot-check usage panels still attribute correctly (sum across the
   merged row should match what the pair previously summed to).

## Out of scope / follow-up

- Disabling the `findFreeSlug` fallback so future seedings never re-introduce
  the `-template` suffix in the first place — tracked as a separate follow-up
  per #1777 \"Out of scope\". The template-seeding code lifecycle question
  (idle vs dead vs shipped) ties to the #1608 one-shot-migration audit.
- Read-only EXPLAIN against prod for the FK-reattach query is recommended
  before the operator runs the route on prod, but the merge query keys
  every UPDATE/DELETE on `app_id` via the existing per-app indexes
  (`idx_traffic_hourly_apps_app`, `idx_traffic_daily_apps_app`) — touched
  row count is bounded by the duplicate app's history (~one app's worth of
  rollups), not the whole rollup table.

### DIFF
--- a/api/src/AppReconciler.scala
+++ b/api/src/AppReconciler.scala
@@ -13,11 +13,13 @@ final case class AppReconcileSummary(
     mergedSlugs: List[String],
     renamedSlugs: List[String],
     hostsUnioned: List[String],
+    templateIdSet: List[String],
     alreadyClean: List[String],
     created: List[String],
 ) derives JsonCodec {
   def total: Int =
-    mergedSlugs.size + renamedSlugs.size + hostsUnioned.size + alreadyClean.size + created.size
+    mergedSlugs.size + renamedSlugs.size + hostsUnioned.size + templateIdSet.size +
+      alreadyClean.size + created.size
 }
 
 /**
@@ -40,11 +42,12 @@ object AppReconciler {
 
   private sealed trait OneOutcome
   private object OneOutcome {
-    final case class Merged(slug: String)       extends OneOutcome
-    final case class Renamed(slug: String)      extends OneOutcome
-    final case class HostsUnioned(slug: String) extends OneOutcome
-    final case class AlreadyClean(slug: String) extends OneOutcome
-    final case class Created(slug: String)      extends OneOutcome
+    final case class Merged(slug: String)        extends OneOutcome
+    final case class Renamed(slug: String)       extends OneOutcome
+    final case class HostsUnioned(slug: String)  extends OneOutcome
+    final case class TemplateIdSet(slug: String) extends OneOutcome
+    final case class AlreadyClean(slug: String)  extends OneOutcome
+    final case class Created(slug: String)       extends OneOutcome
   }
 
   def reconcileTemplates(
@@ -58,6 +61,7 @@ object AppReconciler {
           mergedSlugs = outcomes.collect { case OneOutcome.Merged(s) => s },
           renamedSlugs = outcomes.collect { case OneOutcome.Renamed(s) => s },
           hostsUnioned = outcomes.collect { case OneOutcome.HostsUnioned(s) => s },
+          templateIdSet = outcomes.collect { case OneOutcome.TemplateIdSet(s) => s },
           alreadyClean = outcomes.collect { case OneOutcome.AlreadyClean(s) => s },
           created = outcomes.collect { case OneOutcome.Created(s) => s },
         )
@@ -81,15 +85,17 @@ object AppReconciler {
             unionTemplateHostsInto(repo, s.id, t).as(OneOutcome.Renamed(canonicalSlug))
 
         case (Some(c), None) =>
-          // Only canonical exists. Set template_id if absent, top up template hosts.
+          // Only canonical exists. Set template_id if absent, top up template hosts. The two are
+          // tracked separately so the summary tells operators which subset of canonicals needed
+          // a template_id reattach vs which gained hosts.
+          val needsTemplateId  = c.templateId.isEmpty
           val ensureTemplateId =
-            if c.templateId.isEmpty then repo.update(c.copy(templateId = Some(t.slug)))
-            else ZIO.unit
+            if needsTemplateId then repo.update(c.copy(templateId = Some(t.slug))) else ZIO.unit
           ensureTemplateId *>
-            unionTemplateHostsInto(repo, c.id, t).flatMap { added =>
-              if added || c.templateId.isEmpty then
-                ZIO.succeed(OneOutcome.HostsUnioned(canonicalSlug))
-              else ZIO.succeed(OneOutcome.AlreadyClean(canonicalSlug))
+            unionTemplateHostsInto(repo, c.id, t).map { added =>
+              if added then OneOutcome.HostsUnioned(canonicalSlug)
+              else if needsTemplateId then OneOutcome.TemplateIdSet(canonicalSlug)
+              else OneOutcome.AlreadyClean(canonicalSlug)
             }
 
         case (None, None) =>
--- a/api/src/AppReconciler.scala
+++ b/api/src/AppReconciler.scala
@@ -1,0 +1,120 @@
+package wifihaven.api
+
+import wifihaven.api.db.AppRepo
+import wifihaven.shared.types.*
+import zio.*
+import zio.json.*
+
+/**
+ * #1777: per-template outcome of [[AppReconciler.reconcileTemplates]] — what changed for each
+ * template the reconciler walked. Used by the admin route to surface the result to the operator.
+ */
+final case class AppReconcileSummary(
+    mergedSlugs: List[String],
+    renamedSlugs: List[String],
+    hostsUnioned: List[String],
+    alreadyClean: List[String],
+    created: List[String],
+) derives JsonCodec {
+  def total: Int =
+    mergedSlugs.size + renamedSlugs.size + hostsUnioned.size + alreadyClean.size + created.size
+}
+
+/**
+ * #1777: idempotent reconciler that collapses every `<slug>-template`-suffixed `apps` row onto its
+ * canonical `<slug>` form, with the canonical row gaining the union of both host-sets and every FK
+ * reference (assignments, rollups, usage) reattached. After running, no app slug ends in
+ * `-template`, and `AppTemplates.findByTemplateId` finds each template's canonical row directly.
+ *
+ * Why this exists: when `AppTemplates.seed` runs against a DB where an operator already created an
+ * app at the template's canonical slug, [[AppTemplates#findFreeSlug]] falls back to
+ * `<slug>-template`. The fallback is correct as a uniqueness guard but leaves two rows for the same
+ * logical app, splitting assignments and usage. The reconciler restores 1:1 alignment with the
+ * template set.
+ *
+ * Single source of truth: this is the ONLY place the four reconciliation outcomes (#1777's rules
+ * table) are computed. The admin route delegates to it; the test seeds the messy state and calls
+ * the same function. No duplicate merge logic elsewhere.
+ */
+object AppReconciler {
+
+  private sealed trait OneOutcome
+  private object OneOutcome {
+    final case class Merged(slug: String)       extends OneOutcome
+    final case class Renamed(slug: String)      extends OneOutcome
+    final case class HostsUnioned(slug: String) extends OneOutcome
+    final case class AlreadyClean(slug: String) extends OneOutcome
+    final case class Created(slug: String)      extends OneOutcome
+  }
+
+  def reconcileTemplates(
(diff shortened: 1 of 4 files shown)


sameerparekh:
## Independent review — #1778

### BLOCKERS
(none)

### SHOULD-FIX
1. **[api/test/src/feature/AppReconcilerSpec.scala:88](api/test/src/feature/AppReconcilerSpec.scala:88)** — Test coverage gap for the rollup-table FK reattach. The "merge" test seeds `app_hosts` and `app_policy_assignments` (with overlapping rows on each), but never exercises `traffic_hourly_apps`, `traffic_daily_apps`, or `app_used_daily`. The SUM-on-conflict logic in [api/src/db/Repos.scala:2989-2996](api/src/db/Repos.scala:2989) (`app_used_daily.engaged_seconds + EXCLUDED.engaged_seconds`) and the `dropConflicting*` self-join EXISTS predicates in [api/src/db/Repos.scala:2953-2982](api/src/db/Repos.scala:2953) are non-trivial SQL that a column typo or a missing predicate could silently break. Since this reconciler is run once-per-prod-DB and operates on user-visible usage data, an untested bug there is a real data-loss risk. Add at least one assertion seeding `app_used_daily` on both rows with overlapping `(profile_id, date)` and asserting the post-merge sum.

2. **[api/src/db/Repos.scala:2934](api/src/db/Repos.scala:2934)** — Missing prod-shape EXPLAIN for the new SQL per AGENTS.md §query-explain-before-merge. The `mergeAppInto` transaction touches `traffic_hourly_apps` / `traffic_daily_apps` (growth tables under `idx_traffic_*_apps_app`). The PR body acknowledges this and points to the per-app indexes but does not include an `EXPLAIN (ANALYZE, BUFFERS)` plan from prod-shaped data. Either include the plan or move the EXPLAIN into the operator pre-flight as a hard gate (the PR currently calls it "recommended").

### NITS
1. **[api/src/AppReconciler.scala:90-92](api/src/AppReconciler.scala:90)** — `OneOutcome.HostsUnioned` is emitted when `c.templateId` was newly set even if no hosts were added; the outcome label is then misleading. Consider splitting into `TemplateIdSet` vs `HostsUnioned`, or renaming to `Modified` to cover both adjustments.

2. **[api/src/routes/AppRoutes.scala:253](api/src/routes/AppRoutes.scala:253)** — New admin route ships without a metric. AGENTS.md §instrument-new-functionality says every new route emits a metric in the same PR; admin one-shots are arguably exempt and the sibling `seed-from-templates` route lacks one too, so this is consistent — but worth at least an `admin_apps_reconcile_total{outcome}` counter for parity with the issue's audit-trail intent.

3. **[api/src/db/Repos.scala:2949](api/src/db/Repos.scala:2949)** — When both rows have an assignment for the same `profile_id`, the schedule rules attached to `from`'s assignment are lost via the cascade in `deleteFrom`. The "to wins" policy is defensible but worth a one-line note on `mergeAppInto`'s scaladoc so future readers don't think the cascade is accidental.

### VERDICT: APPROVE

TDD red→green is visible in history (test commit precedes implementation). The reconciler is one function (SSOT), the merge is one transaction with sensible conflict handling, no wire-shape or migration changes, admin auth enforced, parameterized SQL throughout. The SHOULD-FIX items (rollup-table test coverage + EXPLAIN) are coverage/hygiene gaps that should be closed before the operator runs this against prod, not correctness defects visible in the diff itself.

sameerparekh:
## Independent review (re-run) — #1778

### Prior findings status

| # | Class | Item | Status |
|---|-------|------|--------|
| 1 | SHOULD-FIX | Rollup-table test coverage (`app_used_daily` SUM-on-conflict) | **ADDRESSED** — [AppReconcilerSpec:74-85,98-123](api/test/src/feature/AppReconcilerSpec.scala:74) now seeds `app_used_daily` on both rows with overlapping `(kids, 2026-06-15)` and asserts `engagedSeconds == 420` (120+300), plus a non-overlapping Adults row asserting plain reattach, plus stray-row absence on the deleted `seededId`. |
| 2 | SHOULD-FIX | Prod-shape EXPLAIN for the merge SQL | **NOT-ADDRESSED** — explicitly deferred to the operator pre-flight in the commit message. Acceptable given the merge SQL keys every UPDATE/DELETE on `app_id` via the existing per-app indexes and bounds work to one duplicate's history, but the operator should run the EXPLAIN before invoking the route on prod. |
| 3 | NIT | Misleading `HostsUnioned` outcome label | **ADDRESSED** — [AppReconciler.scala:48,91-99](api/src/AppReconciler.scala:91) splits into a dedicated `TemplateIdSet` case + summary bucket. |
| 4 | NIT | Admin route emits no metric | **NOT-ADDRESSED** — consistent with sibling `seed-from-templates`, acceptable as documented. |
| 5 | NIT | Scaladoc on `to wins` policy (incl. V51 schedule-rules cascade) | **ADDRESSED** — [Repos.scala:2796-2801](api/src/db/Repos.scala:2796) documents the conflict policy and the schedule-rules cascade. |

### Incremental delta review (`git diff 87c122ae...HEAD`)

53 lines across `AppReconciler.scala`, `Repos.scala` scaladoc, and the test. Applied the 8 dimensions:

- **SSOT**: `templateIdSet` is one more outcome bucket on the same single function — no new merge path. ✓
- **Testing**: net-new assertions strengthen coverage; no existing assertion weakened. ✓
- **Architectural invariants / wire contract**: unchanged. ✓
- **Scope**: changes are scoped exactly to the SHOULD-FIX / NIT items. ✓
- **Security**: no new SQL surface; the new test uses `AppUsedRollupRepo.upsertDay` (existing parameterized API). ✓

### BLOCKERS
(none)

### SHOULD-FIX
(none new — prior SHOULD-FIX #2 remains NOT-ADDRESSED but is correctly deferred to operator pre-flight)

### NITS
(none new)

### VERDICT: APPROVE

All prior BLOCKERS / SHOULD-FIX classified ADDRESSED except #2 (EXPLAIN), which is correctly punted to the operator pre-flight in the PR body. No new findings in the delta. Ready for merge-when-ready.

Issue #1773, opened 2026-06-17
SPA: global-policy page reuses per-profile editor preset to globalProfile.id + audit (supersedes #1320) (#1769 step 4)


sameerparekh:
Parent: #1769 (umbrella) / #937 (epic). **Gated on step 3 (#1772) merging.** Supersedes #1320.

Step 4 of #1769. SPA-only.

## Scope

- Global-policy page = the existing per-profile editor PRESET to `globalProfile.id`.
- App-picker that POSTs to the global profile's app-policy assignments (user-facing vocabulary is **apps**, not hosts — matches every other allow/deny surface).
- Audit view: read via the existing per-profile audit endpoints; render the same way per-profile audit is rendered.

## Explicit DROPS (per operator)

- **DROP** device-override editor (won't-do per #1452 / memory `policy_two_tiers_no_device_override.md`).
- **DROP** per-profile default-deny toggle — operator believes it already exists in `web/src/`. **Step 0 verifies:** grep `web/src/` for `defaultDeny` / `default_deny`. If present → done, skip. If absent → surface to operator before building.

## Tests

- Vitest covering the global-policy page: search-by-app, add app → calls the per-profile app-assignment route on globalProfile.id; remove app → same.
- Audit log row renders correctly from the per-profile audit endpoint.
- The 'select a profile' picker either hides the global profile or labels it 'Global' visibly — operator preference; default to labeled visibly so operator can navigate to it directly.

## Gate / dependency

Depends on step 3 (routes consolidated). After this lands:
- Close this issue.
- Close #1769 umbrella.
- Check if #937 has any open sub-tasks left; if not, close #937 with a comment listing the four shipped PRs.

## Supersedes

#1320 — closing that issue as superseded; this is the replacement scope.

Issue #1772, opened 2026-06-17
Route deletion: remove /api/global/* — SPA uses /api/profiles/<globalId>/* (#1769 step 3)


sameerparekh:
Parent: #1769 (umbrella) / #937 (epic). **Gated on step 2 (#1771) merging.**

Step 3 of #1769. Code + route-shape change.

## Scope

- Delete every `/api/global/*` route.
- Per `feedback_no_compat_shims.md` (memory): SPA deploys atomically with API; no aliases or deprecation window.
- The SPA must STOP calling `/api/global/*` in the same PR. If the SPA has any current callers, update them to use `/api/profiles/<globalId>/*` instead.

## Tests

- Feature test: `GET /api/global/policy` → 404 (route removed).
- Feature test: `/api/profiles/<globalId>/` endpoints handle the same operations the dropped routes used to handle.

## Gate / dependency

Depends on step 2 (PolicyService union). Step 4 (SPA) depends on this.

## Acceptance

- All `/api/global/*` routes deleted.
- SPA callers updated to per-profile equivalents.
- Tests confirm 404 on old routes; 200 on the per-profile equivalents.

sameerparekh:
Closed by #1780 (merged). The `/api/global/*` routes were deleted in #1775; #1780 adds the `GlobalRoutesGoneSpec` regression test pinning that all six former paths 404, plus `GET /api/profiles/global` as the SPA's per-profile-editor entry point. Step 3 of #1769.

Issue #1769, opened 2026-06-17
Collapse global_* tables into an is_global sentinel profile


sameerparekh:
Parent: #937 (Global Policy & Default-Deny umbrella).

## Why

The household-global policy currently lives in dedicated `global_allow` / `global_blocks` / `global_blocklists` / `global_flags` tables, read by `GlobalPolicyRepo` and consumed by `PolicyService` as `PolicySnapshot.global`. This duplicates the per-profile policy machinery (`app_policy_assignments`, profile flags, profile blocklists) one-for-one — violating AGENTS.md §single-source-of-truth.

Replace with a single sentinel row in `profiles` marked `is_global = TRUE`. Per-profile app-assignment / blocklist / flags machinery is reused as-is. `PolicyService` unions the global profile's resolved rules into every other profile's `extraAllowed` / `extraBlocked` / `blocklistIds` at snapshot build time.

Prod's global tables are empty (verified 2026-06-16) and staging similarly so — no backfill required. Drop the tables outright.

## Sub-issues (filed: #1775 → #1770 → #1771 → #1772 → #1773)

**Sequencing updated 2026-06-17** to insert step 0 (delete dead readers first), which keeps the api.test suite green at every commit and avoids invoking the `migration-coupled-justified` escape hatch on the migration PR.

0. **Delete dead readers** (#1775): pure code deletion of `GlobalPolicyRepo` + `GlobalPolicyRepoLive` + `NoopGlobalPolicyRepo`, the `/api/global/*` routes, the `household_settings.global_*` accessors, the SPA Global Policy page, and their feature specs. Tables stay in the schema (untouched). `PolicySnapshot.global` is held as `BlockRules.empty` for router back-compat. Prod tables are empty / flags at install defaults so this has zero behavioral impact.

1. **Schema migration** (#1770): schema-only PR per AGENTS.md §migrations-back-compat. Add `profiles.is_global BOOLEAN NOT NULL DEFAULT FALSE` with a unique partial index `WHERE is_global = TRUE` (exactly one). Seed the single global profile in the migration. Drop `global_allow`, `global_blocks`, `global_blocklists`, and the `household_settings.global_*` flag columns. Now suite-stays-green because step 0 already removed every reader.

[...]

PR #1779, opened 2026-06-18
feat(#1771): PolicyService — sentinel-profile global rules (step 2 of #1769)

### LINKED ISSUE #1771: PolicyService: delete GlobalPolicyRepo, union global-profile rules into per-profile snapshot (#1769 step 2)
Parent: #1769 (umbrella) / #937 (epic). **Gated on step 1 (#1770) merging.**

Step 2 of #1769. Code-only.

## Scope

- Delete `GlobalPolicyRepo`. `ProfileRepo.getGlobal` (or analogous) replaces it — returns the single `is_global = TRUE` profile.
- In `PolicyService` snapshot assembly: when building each (non-global) profile's resolved `BlockRules`, union the global profile's resolved rules into:
  - `extraAllowed` (set union)
  - `extraBlocked` (set union)
  - `blocklistIds` (set union)
- The global profile MUST NOT contribute schedules / daily limits / manual-block / paused — those are meaningless household-wide. Hide them on the global profile in the API and in repo writes; no-op them in PolicyService.
- Devices cannot be assigned to the global profile. Enforce in:
  - `DeviceRepo` upsert (reject if profileId == globalProfileId)
  - `/api/devices/...` write routes (400 BadRequest with clear message)

## Tests

- Feature test against embedded Postgres: seed global profile with extraAllowed=[a.com]; seed Kids profile with extraAllowed=[b.com]; build snapshot → Kids profile's extraAllowed contains BOTH.
- Feature test: attempt to assign a device to the global profile → 400.
- Feature test: write schedules/daily-limit to global profile → 400 or 422.
- Negative: snapshot for a non-global profile that has nothing in its own rules still gets the global union.

## Gate / dependency

Depends on schema step 1. Step 3 (route deletion) depends on this.

## Acceptance

- Snapshot assembly correct (test).
- Behavior changes spelled out in PR body (operator-readable).
- No reads from the dropped tables remain (compilation gates this).

sameerparekh:
Closes #1771. Step 2 of the umbrella #1769 — replaces the deleted `global_*` tables with a single `is_global=TRUE` sentinel row in `profiles`.

## What

- **Startup seed.** `Main.scala` and the test-template init now seed the single Global sentinel row idempotently (`INSERT … ON CONFLICT DO NOTHING`). V59's partial unique index keeps the cardinality at ≤1. No DB migration here — the seed lands atomically with the code that hides/uses it.
- **`ProfileRepo`.** New `getGlobal` + `listAllIncludingGlobal`. `listAll` now filters `is_global=TRUE` out so `GET /api/profiles`, role-access enumerations, and every existing user-facing path never expose the sentinel. The snapshot path is the only consumer of `listAllIncludingGlobal`.
- **`PolicyService.snapshot`.** The sentinel is resolved through the SAME `computeBlockRules` path the per-profile policies use (AGENTS.md §single-source-of-truth — no parallel global-rule computation). Its `extraAllowed`, `extraBlocked`, and `blocklistIds` are unioned into every non-global profile's rules. `blocked` / `blockReason` are zeroed defensively — the architecture forbids whole-MAC drops from the global tier. `PolicySnapshot.global` is now sourced from the sentinel's rules plus `uiAllowedHosts` + `infraAllowHosts`; wire shape unchanged.
- **Write rejections.** Schedule attach (`PUT /api/profiles/:id/schedules`), `PUT /api/profiles/:id`, `DELETE /api/profiles/:id`, and `PATCH /api/profiles/:id` for `paused` / `pauseMode` / `timeLimit` all return 400 with a precise field name when the target is the global sentinel.
- **Device assignment guard.** `PUT /api/devices` and `PATCH /api/devices/:mac` reject `profileId == <global>` with 400; `DeviceRepoLive.upsert` has the same check as a defensive backstop.

## Tests

Ten new feature tests in `PolicySnapshotGlobalProfileSpec` cover:
- extraAllowed / extraBlocked / blocklistIds union (each its own test)
- A non-global profile with empty own rules still receives the global union
- `snapshot.global.extraAllowed` carries sentinel + `uiAllowedHosts` + `infraAllowHosts`
- `GET /api/profiles` hides the sentinel (paired with the RoleAccessSpec `details.length == 2` invariant)
- 400 on device assignment to global, schedule attach to global, PATCH paused/timeLimit on global

Updated the `ProfilePatchApiSpec` drift pin to exclude `isGlobal` (structural sentinel flag, not patchable) alongside `id`.

## Follow-ups

This completes #1769's behavioral migration. The remaining sub-issue — the SPA rebuild (#1772 + #1773, folded per the updated umbrella body) — can now proceed against the sentinel.

### DIFF
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -88,10 +88,13 @@ object Main extends ZIOAppDefault {
           .serve(withCors)
           .provide(ZLayer.succeed(serverConfig) >>> Server.live)
           .forkScoped
-        _      <- ZIO.logInfo("HTTP port bound; running DB-heavy startup behind readiness gate")
+        _         <- ZIO.logInfo("HTTP port bound; running DB-heavy startup behind readiness gate")
         // Service handles are pure layer reads (no DB I/O); resolve them up front
         // so the retried DB-init block below is purely the idempotent DB work.
-        hsRepo <- ZIO.service[HouseholdSettingsRepo]
+        hsRepo    <- ZIO.service[HouseholdSettingsRepo]
+        // #1771: handle for the startup-time global sentinel seed; the seed itself runs inside
+        // the retried DB-init block below alongside ensureDefault.
+        xaForSeed <- ZIO.service[Transactor[Task]]
         appRepoForSeed <- ZIO.service[AppRepo]
         blRepoForSeed  <- ZIO.service[BlocklistRepo]
         blCacheForSeed <- ZIO.service[BlocklistCache]
@@ -115,6 +118,16 @@ object Main extends ZIOAppDefault {
             // daily-reset tz to the API server's local zone on first install.
             _ <- hsRepo.ensureDefault(tz)
             _ <- ZIO.logInfo(s"household_settings ensured (install-default tz=${tz.getId})")
+            // #1771: seed the single global sentinel profile that PolicyService unions into
+            // every other profile's BlockRules. Idempotent — V59's partial unique index
+            // (`is_global = TRUE`) ensures at most one row ever exists, and ON CONFLICT
+            // matches the [[HouseholdSettingsRepoLive.ensureDefault]] precedent so a restart
+            // is a no-op. Kept here (not in V59) so the seed lands atomically with the code
+            // that hides/uses the sentinel — see #1769 step 2.
+            _ <- sql"""INSERT INTO profiles (name, is_global)
+                       VALUES ('Global', TRUE)
+                       ON CONFLICT DO NOTHING""".update.run.transact(xaForSeed)
+            _ <- ZIO.logInfo("global profile sentinel ensured (is_global=TRUE)")
             // #768: seed the starter library of app templates. Idempotent —
             // operator host edits on previously-seeded apps are preserved.
             seedSummary <- AppTemplates.seed(appRepoForSeed, templates)
@@ -327,7 +340,7 @@ object Main extends ZIOAppDefault {
           ) ++
           ScheduleRoutes.routes(auth, namedSchedRepo) ++
           HouseholdSettingsRoutes.routes(auth, hsRepo) ++
-          DeviceRoutes.routes(auth, deviceRepo, upRepo)
+          DeviceRoutes.routes(auth, deviceRepo, upRepo, profileRepo)
 
       val statsRoutes: Routes[Any, Response] =
         TimeRoutes.routes(
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -1092,33 +1092,35 @@ class DeviceRepoLive(xa: Transactor[Task]) extends DeviceRepo {
         .option
         .transact(xa),
     )
-  def upsert(mac: MacAddress, name: String, pid: Option[ProfileId], ip: String) =
+  def upsert(mac: MacAddress, name: String, pid: Option[ProfileId], ip: String)        = {
     // #708: pid=None writes NULL (device unassigned). Devices without a profile
     // are a supported state — same shape auto-discovery produces.
     // #1771: defensive guard — devices cannot be assigned to the global sentinel
     // profile. The route layer rejects this with a 400 before we get here, but
     // catching it again in the repo means a buggy code path (or a direct SQL
     // call from a future caller) still fails loudly instead of corrupting the
-    // device-to-profile graph. The check is one SELECT against a tiny table.
-    pid.fold(ZIO.unit: Task[Unit]) { p =>
+    // device-to-profile graph. The guard and the INSERT run in the SAME doobie
+    // transaction so a (today-impossible) concurrent flip of `profiles.is_global`
+    // can't slip a device assignment past the check.
+    val check = pid.fold(doobie.free.connection.unit) { p =>
       sql"SELECT is_global FROM profiles WHERE id=$p"
         .query[Boolean]
         .option
-        .transact(xa)
         .flatMap {
           case Some(true) =>
-            ZIO.fail(
+            doobie.free.connection.raiseError[Unit](
               new IllegalArgumentException(
                 s"devices cannot be assigned to the global profile (id=${p.value})",
               ),
             )
-          case _          => ZIO.unit
+          case _          => doobie.free.connection.unit
         }
-    } *>
+    }
+    (check *>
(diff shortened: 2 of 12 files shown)


sameerparekh:
## Independent PR review — #1779

### 1. SSOT / duplicated logic
PolicyService routes the sentinel through the existing `computeBlockRules` via a local `computeRulesFor` helper — no parallel global-rule computation. ✓

### 2. Wire contract / architecture
- `PolicySnapshot.global : BlockRules` shape unchanged; no new wire fields. ✓
- `Profile` gained `isGlobal: Boolean = false` (additive, defaulted). ✓
- Router stays a dumb applier; sentinel folds into existing `extraAllowed`/`extraBlocked`/`blocklistIds`. ✓
- `extraAllowed`-beats-block precedence preserved; comment cites `feedback_extraallowed_beats_blocked`. ✓
- No DNS-as-enforcement reasoning. ✓

### 3. Security
All new SQL is parameterized via Doobie. Routes use `requireWriter`/`requireAdmin` + `requireProfileAccess`. ✓

### 4. Tests
Ten new feature tests cover snapshot union (extraAllowed / extraBlocked / blocklistIds), empty-own-rules profile still receives global union, `snapshot.global` shape, `GET /api/profiles` hides sentinel, and 400 on device-assignment / schedule attach / PATCH paused / PATCH timeLimit on the sentinel. `ProfilePatchApiSpec` drift pin updated to exclude `isGlobal` (structural, not patchable) with a clear rationale comment.

BLOCKERS
(none)

SHOULD-FIX
1. [api/src/routes/Routes.scala:486-503](api/src/routes/Routes.scala#L486) — the PATCH handler does a fresh `profileRepo.findById(pid)` inside the global-guard block, but the route already ran an existence-check `findById` ~40 lines earlier and threw the result away. Collapse to a single read (capture the `Profile` from the existence check and read `_.isGlobal` off it). Saves one round-trip per PATCH and removes a window where the row could disappear between the two reads.
2. PR commit history is a single squash. The TDD red→green progression (failing-test commit → impl commit) called for by AGENTS.md §TDD workflow is not visible. Not retroactively fixable without rewriting history; flagging so the next chip splits commits.

NITS
1. [api/src/policy/PolicyService.scala:219-228](api/src/policy/PolicyService.scala#L219) — for a `defaultDeny=true` profile, `computeBlockRules` deliberately zeroes `extraBlocked` and `blocklistIds` ("redundant under block-all"). The new union then re-adds the global sentinel's contribution on top. Behavior is unchanged at the router (the whole-MAC drop dominates), but the wire payload re-acquires fields the defaultDeny path explicitly omitted. Consider gating the union by `!profile.defaultDeny` for those two fields.
2. [api/src/db/Repos.scala:1098-1115](api/src/db/Repos.scala#L1098) — the upsert guard runs the `is_global` check and the INSERT as two separate `transact(xa)` blocks. Theoretical race (admin promoting a profile to global mid-call) is impossible today since no route can flip `is_global`, but folding into a single transaction or a parameterized `WHERE NOT EXISTS (SELECT 1 FROM profiles WHERE id=$pid AND is_global=TRUE)` keeps the guarantee load-bearing without a comment.
3. [api/src/routes/Routes.scala:631-635](api/src/routes/Routes.scala#L631) — `DeviceRoutes.routes` gained a `profileRepo: ProfileRepo` parameter (no default) which rippled to ~13 test call sites. Acceptable; the doc comment "Passing `null` is unsafe" is misleading since the parameter is non-nullable in Scala — drop that phrase.

VERDICT: APPROVE

PR cleanly delivers #1769 step 2: sentinel-profile seed at startup, `ProfileRepo.getGlobal`/`listAllIncludingGlobal`, snapshot union via single SSOT path, repo+route guards on device assignment and meaningless writes. Wire shape unchanged, additive Profile field, tests through the full stack with embedded Postgres. SHOULD-FIX items are quality improvements, not correctness blockers.

sameerparekh:
## Independent PR review — #1779 (re-run on 7e5dc198)

### Prior-findings status (vs reviewed-sha=552d73ef)

| # | Class | Status | Notes |
|---|-------|--------|-------|
| SF-1 | SHOULD-FIX | **ADDRESSED** | PATCH `/api/profiles/:id` now carries the existing row through and reads `existing.isGlobal` instead of a second `findById` ([Routes.scala:489-491](api/src/routes/Routes.scala#L489)). |
| SF-2 | SHOULD-FIX | NOT-ADDRESSED | TDD red→green split isn't retroactively fixable on this PR; noted for the next chip. Not merge-gating. |
| NIT-1 | NIT | **ADDRESSED** | `extraBlocked` / `blocklistIds` union now skipped under `profile.defaultDeny` ([PolicyService.scala:222-227](api/src/policy/PolicyService.scala#L222)). Wire payload now matches the design intent. |
| NIT-2 | NIT | **ADDRESSED** | `DeviceRepoLive.upsert` guard + INSERT now share a single doobie transaction via `doobie.free.connection` ([Repos.scala:1095-1126](api/src/db/Repos.scala#L1095)). |
| NIT-3 | NIT | **ADDRESSED** | Misleading "passing null is unsafe" comment removed from `DeviceRoutes.routes` ([Routes.scala:635-637](api/src/routes/Routes.scala#L635)). |

### Delta review (552d73ef..7e5dc198)

Three files touched, all surgical fixes to prior findings.

- **PolicyService change**: `defaultDeny`-gated union is correct — `computeBlockRules` already returns `extraBlocked = Nil, blocklistIds = Nil` for defaultDeny rows, and the router's whole-MAC drop dominates, so behavior is unchanged at the router; only the wire payload is leaner. No new SSOT concern.
- **Repos.scala change**: switched to `doobie.free.connection.{unit, raiseError}` so the guard and the INSERT share one transaction. Properly propagates the `IllegalArgumentException` via `raiseError` so existing call sites still see a `Task[DeviceId]` failure — no behavior regression for the route-layer guard.
- **Routes.scala change**: existing-row capture as `existing` is the right shape — the row is read once, structural `isGlobal` is consulted, mutable columns are deliberately NOT carried into the per-field SETs (preserves the #423 concurrent-PATCH-disjoint-fields invariant; comment updated accordingly).

BLOCKERS
(none)

SHOULD-FIX
(none new)

NITS
(none new)

VERDICT: APPROVE

All prior addressable findings cleared; one process item (TDD commit split) is not retroactively fixable on this PR. No new findings in the delta. Ready for merge-when-ready.

Issue #1781, opened 2026-06-18
agent: validate dnsmasq RAM cost at 80k blocklist entries on 1 GB router (#1435 Option 2c gate)


sameerparekh:
## Context

Validation gate for the Option 2c implementation work decided on [#1435](https://github.com/wifihaven/wifihaven/issues/1435) (stream blocklist host lists into a dnsmasq `conf-file=` shard so the agent's Lua process doesn't hold them in RAM). The whole approach rests on the prior that dnsmasq's per-`nftset=` entry cost is small enough (~80-150 bytes) that 80k+ entries fit comfortably in a 1 GB router. Pi-hole's existence on 512 MB Raspberry Pis with 1M+ entries is strong prior evidence, but we want a measured number on our exact hardware/dnsmasq build before committing implementation effort.

## Scope

On the test router (`openwrt.lan`, GL.iNet GL-MT6000, 1 GB RAM, OpenWrt 25.12.3, ARMv8):

1. Capture baseline `MemAvailable` + `wifihaven-agent` RSS + `dnsmasq` RSS with only curated lists assigned (current prod-like state).
2. Hand-author a `/tmp/wifihaven-blocklist-test.conf` shard with `nftset=/<host>/4#inet#wifihaven#bl_test_v4,6#inet#wifihaven#bl_test_v6` lines using the live `GET /api/blocklists/ads-extended/hosts` payload (~85k hosts) as the content.
3. Add the shard via `conf-file=` in `/var/etc/dnsmasq.conf.cfg…` (or the include mechanism the agent will use); reload dnsmasq.
4. Re-measure `MemAvailable`, dnsmasq RSS, agent RSS. Drive some real DNS traffic through the router (e.g. resolve a handful of names from the list) to confirm dnsmasq actually populates the `bl_*` ipsets.
5. Repeat for `adult-extended` (~77k) and both together (~162k).

## Acceptance

- A short measurement table on this issue: per-list dnsmasq RSS delta, `MemAvailable` delta, agent RSS delta.
- A pass/fail call: does the 1 GB router stay below ~50% memory pressure with both extended lists loaded *and* normal traffic running? If yes, Option 2c is green-lit. If no, this issue surfaces the real ceiling and #1435 needs a re-think.

## Out of scope

- Any agent code changes — this is measurement only. The hand-authored conf-file is thrown away after the run.
- Production routers — test router only.

Parent: #1435

sameerparekh:
[...]

Issue #1435, opened 2026-06-05
design: how to handle large (StevenBlack) blocklists without OOMing the agent


sameerparekh:
## Problem

Category blocklist enforcement renders one dnsmasq `nftset=/<host>/…#bl_<id>` directive per member host, and the agent loads every assigned list's full host set into memory (`_blocklist_hosts`). This does not scale to the bundled **StevenBlack "extended" lists**:

- `ads-extended` ≈ 83k hosts (~1.6 MB)
- `adult-extended` ≈ 76k hosts (~1.3 MB)
- `social-extended` ≈ 3.2k hosts

With the extended lists assigned, the agent renders ~164k directives in one config string and balloons to **~700 MB RSS**, which **OOM-kills the agent on a 1 GB router** (observed on prod `RobotSpaceship`; root-caused in #1412). The OOM took out *all* category enforcement, including tiny curated lists, until #1434 added a defensive byte cap.

## Current state after #1434

#1434 caps per-list rendering by byte size (default 256 KiB ≈ 12k hosts): oversized lists are fetched + cached but **skipped** (not parsed, not rendered, logged). So:

- Curated lists (ads, malware, adult, gambling, social-media, games) enforce normally.
- The StevenBlack "extended" lists are effectively **not enforced** — they're silently skipped at the router. The UI still lets users assign them, which is now misleading.

This issue is to decide the real strategy. The cap is a guard, not the answer.

## Options to evaluate

1. **Drop the bundled "extended" lists entirely; invest in self-curated lists.** The curated lists (tens–hundreds of hosts) cover the common ad/tracker/adult/gambling cases and fit the per-domain-nftset model comfortably. Pro: simplest, predictable memory, no UI lie. Con: less coverage than 80k-host aggregate lists; ongoing curation effort.
2. **Keep large lists but change the enforcement mechanism** so it doesn't need 80k per-domain dnsmasq directives or an 80k-entry Lua table, e.g.:
   - a dnsmasq `addn-hosts` / hosts-file sink (domain → sink IP) instead of per-domain `nftset=`;
   - server-side pre-resolution into a compact nft set the agent just loads;
   - `conf-file` streamed to disk + bounded agent memory (still leaves dnsmasq holding 80k rules — needs measurement on target hardware).
3. **Tier lists by router capability.** Allow extended lists only on routers above a RAM threshold (the agent already reports version/RAM-class potential); cap or refuse on low-RAM gateways, surfaced in the UI.
[...]

PR #1788, opened 2026-06-18
feat(#1782,#1783): stream blocklist render to /tmp shards via dnsmasq conf-file

### LINKED ISSUE #1782: agent: stream blocklist rendering via dnsmasq conf-file shards (#1435 Option 2c)
## Context

Per [#1435](https://github.com/wifihaven/wifihaven/issues/1435) the agent OOMs at ~700 MB RSS when rendering 160k+ blocklist hosts. The bottleneck is the agent's Lua-side representation (`_blocklist_hosts` table + the rendered config string built in memory before write), **not** dnsmasq itself — dnsmasq's per-entry hash-table cost is small enough that 80k+ entries fit comfortably (see validation issue ↓ for the empirical number).

This issue implements Option 2c from #1435: stream the host list to a dedicated dnsmasq `conf-file=` shard on disk, so the agent never holds the full host set in Lua memory.

## Architectural invariants to preserve (per AGENTS.md)

- **DNS stays a non-enforcement plane.** The shard contains `nftset=/<host>/...#bl_<id>` directives. dnsmasq still populates the per-blocklist `bl_<id>` ipsets at resolution time; nftables still does the per-MAC drop. No `address=` / no `server=` / no sinkhole.
- **Wire shape unchanged.** `Blocklist(version, url)` per-profile in the snapshot stays as-is. This is purely an agent-side render change.
- **`extraAllowed` carve-out continues to work** because attribution is still at DNS resolution time.
- **Bounded tmpfs writes** — covered by the sibling rotation issue.

## Scope

In the OpenWRT agent (`agent/openwrt/`):

1. Replace the in-memory `_blocklist_hosts` accumulator with a streaming writer: as each host is consumed from the fetched/cached blocklist body, emit one `nftset=` directive (or a small batch) to `/tmp/wifihaven-blocklist-<id>.conf` and discard. Never hold the full set in a Lua table.
2. Reference the shard from the agent's existing dnsmasq config via `conf-file=/tmp/wifihaven-blocklist-<id>.conf` rather than inlining the directives.
3. Lift the #1434 byte cap concurrently — or replace it with a higher, documented ceiling (e.g. 10 MB rendered) appropriate for the streamed path. The cap stays as a defense-in-depth guard against runaway lists; just at a different threshold.
4. Update tests: the existing `_blocklist_hosts` unit tests need to be reframed against the streaming writer; add a busted test that exercises a 50k-host fixture without exceeding a memory budget (whatever Lua's reportable proxy is — `collectgarbage("count")` delta is a reasonable signal).
5. Document the new file layout in the agent README (which files live where, who writes them, how rotation interacts).

## Acceptance

- ads-extended (84k) + adult-extended (76k) + social-extended (3k) all load and enforce on the test router; `bl_*` ipsets populate as hosts are resolved; nft drops fire correctly.
- Agent RSS with all three loaded stays under ~30 MB (vs ~700 MB pre-fix).
- Existing per-MAC carve-out tests (`extraAllowed` beating a blocklist, `blockIpOnly` interaction) still pass.

## Dependencies

- Validation gate ↓ must pass first. If dnsmasq itself can't hold the entries, this work is moot.
- Sibling rotation issue must ship in the same PR per AGENTS.md `#bounded-tmp-writes`.

Parent: #1435

### LINKED ISSUE #1783: agent: atomic write + lifecycle for /tmp blocklist conf shards (sibling of streamed render)
## Context

The sibling implementation issue introduces `/tmp/wifihaven-blocklist-<id>.conf` shards (one per assigned blocklist) so the agent can stream the host list to disk rather than holding it in Lua memory. On OpenWRT, `/tmp` is `tmpfs` — RAM — and AGENTS.md `#bounded-tmp-writes` requires that any agent-written tmpfs file whose size grows with traffic/time/content **must ship a rotation/truncation path in the same change**, folded into the existing `wifihaven-rotate-dnsmasq-log` cron (copytruncate, never rename).

The blocklist shard isn't unbounded like a log, but it *does* grow with assigned-list size and could swing wildly across re-renders (e.g. an operator toggles a 80k-host list on/off). Without explicit cleanup we risk stale shards from removed blocklists accumulating on disk between agent restarts.

## Scope

1. When a profile assignment no longer references a blocklist (or the list is deleted), the agent removes the corresponding `/tmp/wifihaven-blocklist-<id>.conf` shard at the next render.
2. On agent startup, garbage-collect any shard files that aren't in the current effective set (defense-in-depth for crash-during-render scenarios).
3. Re-render path is atomic: write to `…<id>.conf.tmp`, fsync, rename onto `…<id>.conf`. Never partial-state.
4. Document the rotation/cleanup rules alongside the existing `wifihaven-rotate-dnsmasq-log` cron description.

Note: this is *not* log rotation (the shards are config files, not append-only). The AGENTS.md rule is about bounded tmpfs *footprint*; for config shards the right pattern is precise lifecycle management rather than copytruncate. This issue formalizes the lifecycle.

## Acceptance

- Removing all assignments to a 80k-host list reclaims the shard's tmpfs bytes on the next render.
- An agent restart with stale shards on disk (e.g. from a crash mid-render) converges back to the current effective set.
- Re-render never leaves a partial shard visible to dnsmasq (atomic rename).

Parent: #1435

sameerparekh:
## Summary

Streams blocklist host lists from the on-disk cache directly into `/tmp/wifihaven-blocklist-<id>.conf` shard files, included by dnsmasq via `conf-file=`. The agent no longer holds the full member set (~160k hosts) in a Lua table — fixes the #1412 OOM that motivated the byte cap in #1434. Lifts the cap to 10 MB per shard as defense-in-depth.

- Streaming render: `blocklists.render_shards` reads cache files line-by-line and writes `nftset=/<host>/4#inet#wifihaven#bl_<id>,6#inet#wifihaven#bl6_<id>` per host, never accumulating in Lua.
- Atomic write + lifecycle: `.tmp` → fsync → rename; stale shards GC'd at render + startup.
- `dnsmasq` emits `conf-file=/tmp/wifihaven-blocklist-<id>.conf` per blocklist; bl_ specs no longer appear in the per-host merged `nftset=` directives.
- `bl_member_iterator` closure passed to `update_shared` streams bl_hosts_by_mac from cache files at call time (no MAC multiplier at steady state).
- Version-pair change detection replaces `hosts_differ` on the blocklist timer.

## Design questions resolved in this implementation

**A. Merged-nftset-per-host (#1460):** Per the dnsmasq man page and the #1781 validation measurement (which loaded shards from `conf-dir`), dnsmasq applies all matching `nftset=` directives across config files. A host in both a blocklist shard AND in the main conf's merged `nftset=` directive will have both sets populated independently. The `only one wins` caveat from the #1460 comment applies within a _single directive's comma-separated spec list_ — multiple directives across files do fire. This makes the shard-per-blocklist approach safe without changes to the per-host merge logic (which now only carries ea_/eb_/global specs, not bl_).

**B. global.blocklistIds:** When an id appears in `global.blocklistIds`, `render_shards` appends `,4#inet#wifihaven#global_block,6#inet#wifihaven#global_block6` to each shard line. The `@global_block` nft set is still declared by `render.nft` when `global.blocklistIds` is non-empty. `global_block_hosts()` now returns only `global.extraBlocked` (blocklist member expansion via `_blocklist_hosts` removed).

**C. update_shared bl_hosts_by_mac:** Accepts a new `bl_member_iterator(id)` parameter. The agent passes a closure that reads the cache file line-by-line and returns a flat table per-id call. One file read per subscribed id per `update_shared` call; result table is not retained across calls. Acceptable transient cost for `conntrack.lua` labeling.

## Architectural invariants preserved

- **DNS is not the enforcement plane.** Shards contain only `nftset=` directives; nftables still does the per-MAC drop.
- **Wire shape unchanged.** `Blocklist(version, url)` per-profile in the snapshot stays as-is. Agent-internal change only.
- **Per-MAC carve-outs intact.** `extraAllowed` still suppresses bl_ drops via the existing nft `ip daddr != @ea_<m>_<a>` clauses.
- **Bounded tmpfs** (AGENTS.md `#bounded-tmp-writes`): atomic write, precise lifecycle (GC not rotation), documented in `router-agent-bounded-writes.md`.

## Tests

TDD: red commit first, then green.

- `blocklists.render_shards`: line shape, sanitization, blank/comment skip, atomic rename, rename-failure guard, absent-cache skip, max_bytes cap, global_blocklist_ids append, 50k-host fixture with <5 MB heap delta guard
- `blocklists.gc_shards`: stale shard removal, current-id preservation, .tmp cleanup, non-wifihaven file safety
- `blocklists.render_member_index`: TSV format, matches in-memory path, atomic write
- `render.dnsmasq`: conf-file= emission, no inline bl_ specs, empty-blocklists skip, eb_ still inline

226/226 tests pass locally (1 pending = `nft -c` unavailable without CAP_NET_ADMIN on dev host).

## Validation gate (#1781)

#1781 is CLOSED with PASS: dnsmasq RSS +31 MB for 162k hosts on the 1 GB GL-MT6000, `MemAvailable` 71% (well above the 50% gate). Implementation built against those confirmed numbers.

Closes #1782
Closes #1783
Refs #1435 #1781

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/docs/process/router-agent-bounded-writes.md
+++ b/docs/process/router-agent-bounded-writes.md
@@ -39,3 +39,39 @@ introduces it.** Concretely:
 The same reasoning applies to anything else the agent persists under `/tmp`
 (JSON spools, NDJSON event journals, debug dumps). If it can grow, it gets a
 bound — in the same PR.
+
+## Per-blocklist dnsmasq conf shards (`/tmp/wifihaven-blocklist-<id>.conf`) {#blocklist-shards}
+
+Added in [#1782/#1783](https://github.com/wifihaven/wifihaven/issues/1782).
+One file per blocklist id in the current snapshot.
+
+**How they work.** `blocklists.render_shards` reads each list's on-disk cache
+file (`/etc/wifihaven/blocklists/<id>-<version>.txt`) line-by-line and writes
+one `nftset=/<host>/4#inet#wifihaven#bl_<id>,6#inet#wifihaven#bl6_<id>` line
+per host — never accumulating the full list in a Lua table. `render.dnsmasq`
+emits `conf-file=/tmp/wifihaven-blocklist-<id>.conf` for each id so dnsmasq
+picks up the host set at startup and on reload.
+
+**Atomic write.** Each shard is written to `<id>.conf.tmp`, then renamed onto
+`<id>.conf`. A partial shard is never visible to dnsmasq.
+
+**Sizing.** ~80 bytes per host × list member count. The two large StevenBlack
+lists are ~7 MB each; three curated lists total under 100 KB. A per-shard cap
+(default 10 MB, UCI `blocklist_max_list_bytes`) prevents a runaway list from
+filling `/tmp`.
+
+**Lifecycle (GC, NOT log rotation).** These are config files, not append-only
+logs — `copytruncate` is wrong here. The lifecycle is:
+
+- **On every render:** `blocklists.gc_shards` removes shards whose id is no
+  longer in the current snapshot (list unassigned or deleted). Stale
+  `.conf.tmp` files are also removed (crash-during-render defense).
+- **On agent startup:** `gc_shards` runs against the cached snapshot (or an
+  empty placeholder) before the first apply, so a crash-during-render from a
+  previous run can never leave a stale shard visible to the newly-started
+  dnsmasq.
+
+**NOT added to `wifihaven-rotate-dnsmasq-log`.** The rotate cron is for
+append-only logs. Shard lifecycle is precise removal of stale ids, not
+copytruncate — adding these to the cron would silently truncate a live shard
+mid-use.
--- a/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
@@ -16,6 +16,18 @@ local M = {}
 
 local DEFAULT_CACHE_DIR = "/etc/wifihaven/blocklists"
 
+-- Lazy require render module — compatible with both the production on-device
+-- path (wifihaven.render) and the busted test path (render). Called at most
+-- once per process; result is cached in the upvalue.
+local _render_module
+local function get_render()
+  if not _render_module then
+    local ok, m = pcall(require, "wifihaven.render")
+    _render_module = ok and m or require("render")
+  end
+  return _render_module
+end
+
 -- classify_fetch_status(status) → bounded status enum (#1301).
 --
 -- The blocklist fetch surfaces as the `status` label of
@@ -361,12 +373,8 @@ function M.render_shards(snapshot, fs, cache_dir, shard_dir, max_bytes, global_b
   global_blocklist_ids = global_blocklist_ids or {}
 
   -- Require render.bl_sanitize for set-name normalization so there is
-  -- one source of truth.
-  -- In production (on-device) the module path is wifihaven.render; in the
-  -- busted test environment modules are loaded without the wifihaven. prefix.
-  local ok_render, render = pcall(require, "wifihaven.render")
-  if not ok_render then render = require("render") end
-  local bl_sanitize = render.bl_sanitize
+  -- one source of truth (see get_render() at the top of this module).
+  local bl_sanitize = get_render().bl_sanitize
 
   -- Resolve fs ops (injectable for tests, real I/O otherwise).
   local open_read_fn, open_write_fn, rename_fn, remove_fn, fsync_fn
@@ -547,10 +555,8 @@ function M.render_member_index(snapshot, fs, cache_dir, index_path)
   cache_dir  = cache_dir  or "/etc/wifihaven/blocklists"
   index_path = index_path or "/var/run/wifihaven/blocklist_members.tsv"
 
-  local ok_r2, render2 = pcall(require, "wifihaven.render")
-  if not ok_r2 then render2 = require("render") end
-  local bl_set_name  = render2.bl_set_name
-  local bl6_set_name = render2.bl6_set_name
+  local bl_set_name  = get_render().bl_set_name
+  local bl6_set_name = get_render().bl6_set_name
 
   local open_read_fn, open_write_fn, rename_fn
   if fs then
--- a/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
@@ -323,6 +323,296 @@ function M.hosts_differ(a, b)
   return false
 end
 
+-- ---------------------------------------------------------------------------
+-- M.render_shards(snapshot, fs, cache_dir, shard_dir, max_bytes,
+--                 global_blocklist_ids)
+-- ---------------------------------------------------------------------------
+-- Streams per-blocklist shard files to <shard_dir>/wifihaven-blocklist-<id>.conf,
+-- one file per blocklist in snapshot.blocklists. Each line in the shard is:
+--
+--   nftset=/<host>/4#inet#wifihaven#bl_<sanid>,6#inet#wifihaven#bl6_<sanid>\n
+--
+-- When global_blocklist_ids[id] is truthy, each line also appends:
+--   ,4#inet#wifihaven#global_block,6#inet#wifihaven#global_block6
+-- before the newline.
+--
+-- Reads from the on-disk cache file line-by-line (via fs.open_read) so the
+-- full host list is NEVER accumulated in a Lua table — the #1412 OOM source.
+-- Writes atomically: <id>.conf.tmp → <id>.conf.
+--
+-- Parameters:
+--   snapshot           — current policy snapshot (snapshot.blocklists used)
+--   fs                 — injectable filesystem ops (open_read, open_write,
+--                        rename, remove, fsync; defaults to real I/O)
+--   cache_dir          — directory holding <id>-<version>.txt cache files
+--                        (default /etc/wifihaven/blocklists)
+--   shard_dir          — directory to write shards into (default /tmp)
+--   max_bytes          — stop emitting once shard exceeds this many bytes;
+--                        nil = no cap (default 10 MB = 10485760 in agent)
+--   global_blocklist_ids — optional { [id] = true } set; ids in this set
+--                          have global_block/global_block6 appended to each
+--                          shard line
+--
+-- Returns:
+--   { rendered = { [id] = bytes }, skipped = { ids }, errors = { strings } }
+function M.render_shards(snapshot, fs, cache_dir, shard_dir, max_bytes, global_blocklist_ids)
+  cache_dir  = cache_dir  or "/etc/wifihaven/blocklists"
+  shard_dir  = shard_dir  or "/tmp"
+  global_blocklist_ids = global_blocklist_ids or {}
+
+  -- Require render.bl_sanitize for set-name normalization so there is
+  -- one source of truth.
+  -- In production (on-device) the module path is wifihaven.render; in the
+  -- busted test environment modules are loaded without the wifihaven. prefix.
+  local ok_render, render = pcall(require, "wifihaven.render")
+  if not ok_render then render = require("render") end
+  local bl_sanitize = render.bl_sanitize
+
+  -- Resolve fs ops (injectable for tests, real I/O otherwise).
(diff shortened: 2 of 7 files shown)


sameerparekh:
## Independent PR Review — #1788

First run (no prior review comment). Full merge-base diff reviewed.

---

## BLOCKERS

**1. `openwrt/files/usr/lib/lua/wifihaven/policy.lua:272` — `policy.apply` writes an always-empty bl member index, silently breaking dns-tail's CNAME-aware bl_ set population**

`policy.lua:272` still calls `render.blocklist_member_index(snapshot)`, which reads `snapshot._blocklist_hosts`. This PR deliberately stops attaching `_blocklist_hosts` to the snapshot (the whole point is to avoid the in-memory table), so every call to `policy.apply` will now write an empty index to `paths.bl_member_index` (`/tmp/wifihaven-bl-members.txt`).

`wifihaven-dns-tail` reads that path (`paths.bl_member_index`) to build its CNAME-alias bl_ population map (see `wifihaven-dns-tail:399` where it refreshes on policy changes). An empty index means dns-tail can no longer attribute direct CNAME-target queries to a category drop-set. Category blocking still fires at DNS-resolve time for direct queries (the shard's `nftset=` callbacks), but the CNAME-aware coverage (#1348 / #515 gap) is silently lost.

The fix is to update `policy.lua`'s `policy.apply` to use `blocklists.render_member_index` (the streaming path this PR adds) — OR to pass the blocklist_member_index content into `policy.apply` from the agent instead of having `policy.lua` generate it from the snapshot. The agent already calls `blocklists.render_member_index` in `refresh_blocklists`, but writes to a different path (`/var/run/wifihaven/blocklist_members.tsv`) that dns-tail does not read. Both the write path and the consumption path (`wifihaven-dns-tail`) need to be reconciled.

Note: the new `blocklists.render_member_index` in this PR writes to `/var/run/wifihaven/blocklist_members.tsv`, but dns-tail reads `paths.bl_member_index` = `/tmp/wifihaven-bl-members.txt`. These are different files entirely, so the new streaming writer is also unreachable by the current dns-tail consumer.

---

## SHOULD-FIX

**1. `openwrt/files/usr/lib/lua/wifihaven/render.lua:568–596` — `render.blocklist_member_index` doc comment still says "derived from `snapshot._blocklist_hosts` that drives the `nftset=` directives"**

After this PR the `nftset=` directives no longer drive from `_blocklist_hosts` — they come from shard files. The doc comment is now misleading and will be read by the next developer as implying the two are still in sync. It should be updated to note this is now the legacy/in-memory path for tests and the in-process dns-tail path; the on-disk writer is `blocklists.render_member_index`.

**2. `openwrt/files/usr/sbin/wifihaven-agent:81` — comment still says "blocklists.hosts_differ" as the change-detection mechanism**

`wifihaven-agent:81` reads: "the agent re-applies the ruleset only when the loaded host set actually differs (blocklists.hosts_differ)". The PR body describes the replacement by version-pair comparison but the inline comment was not updated. Stale comments mislead future readers about the change-detection contract.

**3. `openwrt/files/usr/lib/lua/wifihaven/render.lua:448` — `conf-file=` path hardcodes `/tmp` instead of using a configurable shard_dir**

`render.dnsmasq` emits `conf-file=/tmp/wifihaven-blocklist-<id>.conf` with a hardcoded `/tmp`. The shard directory is configurable in `render_shards` and `gc_shards` but `render.dnsmasq` doesn't accept a `shard_dir` parameter. Tests pass `/tmp` as the shard dir so they silently match, but a deployment that configures a different shard dir would have `conf-file=` pointing at the wrong location. The path should either be a `M.SHARD_DIR` module-level constant (with a setter for tests) or passed as an `opts.shard_dir` parameter to `dnsmasq()`.

---

## NITS

**1. `openwrt/files/usr/lib/lua/wifihaven/blocklists.lua:94-95` — dual-module require pattern is cargo-culted into both `render_shards` and `render_member_index` separately**

The same `pcall(require, "wifihaven.render") / require("render")` fallback pattern is copy-pasted twice (lines 94–95 and 278–279). A local module-top helper or a single lazy-require at the bottom of the `require` block would remove the duplication. Minor since it's internal, but it contradicts the single-source-of-truth convention for a trivial refactor.

**2. `openwrt/files/usr/sbin/wifihaven-agent:492–503` — "Note: res.skipped ... not surfaced separately" comment is confusing**

The comment says `res.skipped` from `load_cached` (legacy path) is not surfaced, but `load_cached` is still called by `blocklists.refresh`. With the `max_bytes` argument now dropped from `blocklists.refresh` at the call site (`agent:463`), the `load_cached` cap is also removed. The comment should clarify whether this is intentional (the shard cap is the only guard now, load_cached parsing is left unbounded) or whether a separate `max_bytes` argument should still be threaded through.

---

## VERDICT: REQUEST-CHANGES

This PR correctly implements the streaming shard render approach and the TDD red→green commit structure is clean. However, BLOCKER #1 (the always-empty bl member index after `policy.apply`) means dns-tail's CNAME-aware bl_ category drop-set population is silently broken: every policy apply overwrites the index file with empty content because `policy.lua` still reads `snapshot._blocklist_hosts` (which is no longer populated). The fix requires coordinating `policy.lua`'s `policy.apply` with the new streaming writer, and reconciling the two distinct index file paths (`/tmp/wifihaven-bl-members.txt` vs `/var/run/wifihaven/blocklist_members.tsv`).

sameerparekh:
## Independent PR Review — #1788 (re-run on 1309e6c6)

Prior review SHA: `e1e00fd5`. Incremental delta: `git diff e1e00fd5...1309e6c6`.

---

## Prior findings status

| # | Finding | Status |
|---|---------|--------|
| BLOCKER 1 | `policy.lua:272` — always-empty bl member index overwrites agent's index | **ADDRESSED** — `render.blocklist_member_index(snapshot)` call removed from `policy.apply`; replaced with explanatory comment. Agent's `refresh_blocklists` now writes to `paths.bl_member_index` (the path dns-tail reads) via `blocklists.render_member_index`. |
| SHOULD-FIX 1 | `render.lua:568` — stale doc comment on `blocklist_member_index` | **ADDRESSED** — doc comment updated to clearly mark this as the legacy/test path, notes the streaming replacement. |
| SHOULD-FIX 2 | `agent:81` — stale `blocklists.hosts_differ` comment | **ADDRESSED** — comment updated to reference version-pair comparison. |
| SHOULD-FIX 3 | `render.lua:448` — hardcoded `/tmp` in `conf-file=` | **ADDRESSED** — `M.SHARD_DIR = "/tmp"` module constant added; `conf-file=` now uses `M.SHARD_DIR`. |
| NIT 1 | `blocklists.lua` — duplicated pcall require pattern | **ADDRESSED** — consolidated into a single `get_render()` lazy-require helper at module top. |

---

## New findings on incremental delta

### BLOCKERS

(none)

### SHOULD-FIX

**1. `openwrt/files/usr/lib/lua/wifihaven/blocklists.lua:555` — `render_member_index` default `index_path` still points to the old unused path**

The function signature is `M.render_member_index(snapshot, fs, cache_dir, index_path)` with `index_path = index_path or "/var/run/wifihaven/blocklist_members.tsv"`. The agent now explicitly passes `paths.bl_member_index` (= `/tmp/wifihaven-bl-members.txt`), so the default is only reached if a caller omits `index_path`. That default path `/var/run/wifihaven/blocklist_members.tsv` is the incorrect one that dns-tail doesn't read. The default should be updated to `paths.bl_member_index` (or just leave it with no default and require the caller to pass it) to avoid a future caller hitting the wrong default. Low-priority since the agent always passes explicitly, but a future test/caller omitting the arg would silently write to the wrong path.

### NITS

**1. `openwrt/files/usr/lib/lua/wifihaven/policy.lua:266-279` — removed code block leaves a longer comment-to-nothing ratio; minor readability**

The comment block now describes what was removed and why, which is informative but verbose for a production file. The block can be tightened to 2-3 lines since the history is in the commit message. Non-blocking.

---

## VERDICT: APPROVE

The BLOCKER (always-empty bl member index breaking dns-tail CNAME-aware bl_ population) is resolved: `policy.apply` no longer writes the member index, and `refresh_blocklists` now writes to the path dns-tail reads (`paths.bl_member_index`). All prior SHOULD-FIX and NIT findings are addressed. The incremental delta is clean and targeted. The one new SHOULD-FIX (wrong default path in `render_member_index`) is low-risk since the agent always passes explicitly, and can be addressed in a follow-up. 226/226 tests pass locally.

Issue #1789, opened 2026-06-19
ci(cd): staging API 502 after recent merge wave — Render deploy fails to bring container up


sameerparekh:
## Symptom (prod CD, 2026-06-18 21:22 UTC)

Master API/UI CD on main commit \`80f67a0b\` failing at "Deploy staging (Render)":

- Deploy hook accepted ✓
- Polling \`/api/version\` for served SHA → returns \`<none>\` for full 30 min until timeout
- Direct check: \`https://api-staging.wifihaven.net/api/version\` returns **502** (Render-branded error page); \`/api/health\` also 502
- Container isn't running. The new build either failed to start, crashed at startup, or failed its health check

Run: https://github.com/wifihaven/wifihaven/actions/runs/27789593206

## Likely culprits (recent merge wave)

Most likely something in the recent commits at HEAD. Ranked:

1. **#1770 / #1771 — global_* table collapse** (commits 428558f5 + 17826767). The Flyway migration from #1770 dropped \`global_allow\` / \`global_blocks\` / \`global_blocklists\` / \`global_flags\` AND added \`profiles.is_global\`. If the migration fails on prod-shaped staging data, the container OOMs / errors at boot. Or if #1771's PolicyService rewrite has a bug for the sentinel-profile path, it could crash at startup.
2. **#1777 — template reconciliation** (49b27fcf). The new \`/api/admin/apps/reconcile-templates\` route was added; if its initialization wired something broken (e.g. a constructor with a missing DI), startup fails.
3. **#1456 — PATCH /api/global** (e01833b1). Added field-scoped PATCH; if a route table init is broken, startup fails.
4. **#1772 — route deletion of /api/global/\*** (in #1780 / f0d5a41b). If something else still wires a deleted route, the route table errors at boot.
5. **#1739 — error handler cleanup**. If a previously-soft error path now crashes the boundary handler, startup or first-request could fail.

## Investigation steps

1. **Render API**: pull deploy logs for \`wifihaven-api-staging\` around 2026-06-18 21:22 UTC. The actual exception or migration failure will be there.
2. **Migration check**: did the V<N>__profiles_is_global migration apply? Look for "Flyway" / "Migrating" / "Successfully applied" or "FlywayMigrationException" in the logs.
3. **Startup check**: did the JVM start? "Application started at" / "API server bound to" lines. Absent them → process crash before bind.
[...]

PR #1791, opened 2026-06-19
feat(#1790): self-healing admin auth helper for staging smoke (handles 30-day DB reset)

### LINKED ISSUE #1790: ci(staging): smoke tests should auto-recover from 30-day DB reset — try changeme + immediately rotate to stored password
## Problem

Staging's Postgres resets every 30 days (free-tier behavior). On reset, the admin user gets reseeded with the default password \`changeme\`. CI's stored \`STAGING_ADMIN_PASSWORD\` secret holds the operator-chosen value — so the next CD's auth step 401s and every staging-gated job (Gate 1 / Gate 3a / Gate 3b) cascade-fails until the operator intervenes.

Setting the CI secret to \`changeme\` doesn't work: the API forces a password change on first login with the default, so subsequent admin calls fail anyway.

## Durable fix

The smoke step authenticates self-healingly:

1. \`POST /api/auth/login\` with \`$STAGING_ADMIN_PASSWORD\` (operator-chosen, stored in CI secret).
2. **If 200** → proceed with the smoke test (the common case).
3. **If 401** → DB was just reset. Try \`POST /api/auth/login\` with \`changeme\`.
   - **If 200** → immediately \`POST /api/auth/change-password\` to set the password back to \`$STAGING_ADMIN_PASSWORD\`. Then re-login with that and proceed.
   - **If 401** → both credentials wrong; hard fail with a clear "neither stored nor default admin credentials worked on staging" message.
4. **If anything else** (502, network, etc.) → propagate the underlying error; do NOT swallow.

The bootstrap step is **idempotent across resets** — on every run it either passes through (normal day) or self-heals (just after a reset).

## Implementation surface

The auth handshake lives in CI somewhere — probably a small shell or python step inside Gate 1's "Admin API smoke against staging." Find it (\`.github/workflows/\` + \`scripts/e2e/\` + \`scripts/staging/\`), promote it to a reusable script (e.g. \`scripts/staging/admin-auth.sh\` or analogous), and call it from every gate that auths as admin (Gate 1 + Gate 3a + Gate 3b).

Per #623 (just merged), \`/api/auth/change-password\` returns JSON body — script should parse the response, not assume empty body.

Per AGENTS.md §single-source-of-truth: ONE auth-handshake helper, called from every gate. Don't fork.

## Acceptance

- A failing test fixture (or integration test) that simulates the post-reset state: staging admin = \`changeme\`. The smoke script auto-recovers, rotates the password, and proceeds with the test plan.
- All staging-gated CI jobs (Gate 1 / Gate 3a / Gate 3b) use the shared helper.
- The CI secret \`STAGING_ADMIN_PASSWORD\` continues to hold the operator-chosen password; no human update needed across 30-day resets.
- Document the behavior in a comment alongside the helper so a future reader knows why the dance exists.

## Out of scope

- Changing prod's reseed behavior. Prod has no reset cadence; this only affects staging.
- Eliminating the staging DB reset itself (that's the free-tier Render policy, separate decision if we ever upgrade).

## Related

- #1789 (the inciting incident — staging up but gates failed on auth)
- #623 (auth/change-password JSON body — script must parse it)
- AGENTS.md \`feedback_chips_must_run_pr_review.md\`, \`feedback_chips_monitor_pr_to_merged.md\`

sameerparekh:
Closes #1790.

## Why

Staging's Postgres resets every 30 days (free-tier behaviour). On reset the admin user reverts to the seeded `changeme` password with the must-change flag set, so the CI secret's operator-chosen value 401s and every staging-gated job (Gate 1 / Gate 3a / Gate 3b) cascade-fails until someone rotates the secret by hand. #1789 was the inciting incident.

## Approach

One source-of-truth handshake, in two language flavours that share the same algorithm:

| Helper | Used by |
| --- | --- |
| `scripts/e2e/lib/admin-auth.sh` (sourced) | `scripts/e2e-tests.sh`, `scripts/e2e-router.sh` (Gate 1 admin + router smoke) |
| `AdminAPI.login` in `scripts/e2e/lib/api_admin.py` | `scripts/e2e/gate3/conftest.py` (Gate 3a + Gate 3b) |

Algorithm:

1. `POST /api/auth/login` with the stored password.
2. **200** → echo the JWT.
3. **401** → `POST /api/auth/login` with `changeme`.
   - **200** → `POST /api/auth/change-password` to rotate back to the stored value (parsing the JSON body per #623), then re-login with the stored value and echo the JWT.
   - **401** → hard-fail with a clear message naming both arms.
4. Any other status → propagate; do not swallow.

The bootstrap is idempotent across resets — every CD run either passes through (normal day, one login call) or self-heals (just after a reset).

## Test plan

- [x] Integration tests in `scripts/e2e/lib/test_admin_auth.py` cover both helpers against a mock of the WifiHaven auth surface — normal day, post-reset self-heal, and the hard-fail case. The mock also models the #623 JSON body for `/api/auth/change-password` so a regression to empty body trips the test.
- [ ] Operator can validate against the live staging reset on the next 30-day boundary: secret stays at the operator-chosen value, the CD run just after reset succeeds without intervention, and the admin password ends up rotated back.

Run the tests:

```sh
cd scripts/e2e/lib && python3 -m pytest test_admin_auth.py
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,3 +59,4 @@ scripts/e2e/**/__pycache__/
 # from openwrt/files/ on every build; not source.
 openwrt/wifihaven_*.ipk
 openwrt/wifihaven_*.apk
+.venv-test/
--- a/scripts/e2e-router.sh
+++ b/scripts/e2e-router.sh
@@ -59,6 +59,7 @@ export STAGING_ADMIN_USER="admin"
 export STAGING_ADMIN_PASSWORD="$ADMIN_PASS"
 ADMIN=$(admin_login_self_heal) || fail "admin login failed (see stderr)"
 [ -n "$ADMIN" ] || fail "admin_login_self_heal returned empty token"
+unset STAGING_API_URL STAGING_ADMIN_USER STAGING_ADMIN_PASSWORD
 pass "logged in"
 AUTH=(-H "authorization: Bearer $ADMIN")
 
--- a/scripts/e2e-router.sh
+++ b/scripts/e2e-router.sh
@@ -48,13 +48,17 @@ _py() { python3 -c "$1"; }
 E2E_LIB="$(cd "$(dirname "${BASH_SOURCE[0]}")/e2e/lib" && pwd)"
 
 # ── Admin login ────────────────────────────────────────────────────────────
+# #1790: shared self-healing helper. On a fresh staging DB reset, rotates
+# the seeded 'changeme' password back to $ADMIN_PASS before returning a JWT;
+# on the normal day this is a single-login pass-through.
 step "Admin login"
-LOGIN=$(curl -fsS -X POST "$BASE/api/auth/login" \
-  -H 'content-type: application/json' \
-  -d "{\"username\":\"admin\",\"password\":\"$ADMIN_PASS\"}")
-ADMIN=$(_py "import sys,json; print(json.loads('$LOGIN'.replace(\"'\",\"'\"))['token'])" 2>/dev/null \
-  || echo "$LOGIN" | sed -n 's/.*"token":"\([^"]*\)".*/\1/p')
-[ -n "$ADMIN" ] || fail "no token in login response: $LOGIN"
+# shellcheck source=scripts/e2e/lib/admin-auth.sh
+source "$E2E_LIB/admin-auth.sh"
+export STAGING_API_URL="$BASE"
+export STAGING_ADMIN_USER="admin"
+export STAGING_ADMIN_PASSWORD="$ADMIN_PASS"
+ADMIN=$(admin_login_self_heal) || fail "admin login failed (see stderr)"
+[ -n "$ADMIN" ] || fail "admin_login_self_heal returned empty token"
 pass "logged in"
 AUTH=(-H "authorization: Bearer $ADMIN")
 
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -52,6 +52,7 @@ export STAGING_ADMIN_USER="admin"
 export STAGING_ADMIN_PASSWORD="$ADMIN_PASS"
 TOKEN=$(admin_login_self_heal) || fail "admin login failed (see stderr)"
 [ -n "$TOKEN" ] || fail "admin_login_self_heal returned empty token"
+unset STAGING_API_URL STAGING_ADMIN_USER STAGING_ADMIN_PASSWORD
 pass "logged in"
 
 AUTH=(-H "authorization: Bearer $TOKEN")
--- a/scripts/e2e-tests.sh
+++ b/scripts/e2e-tests.sh
@@ -42,12 +42,16 @@ for i in $(seq 1 60); do
 done
 
 step "Login as admin"
-LOGIN=$(curl -fsS -X POST "$BASE/api/auth/login" \
-  -H 'content-type: application/json' \
-  -d "{\"username\":\"admin\",\"password\":\"$ADMIN_PASS\"}")
-TOKEN=$(echo "$LOGIN" | sed -n 's/.*"token":"\([^"]*\)".*/\1/p')
-[ -n "$TOKEN" ] || fail "no token in login response: $LOGIN"
-echo "$LOGIN" | grep -q '"role":"admin"' || fail "admin role missing"
+# #1790: shared self-healing helper. On a fresh staging DB reset, rotates
+# the seeded 'changeme' password back to $ADMIN_PASS before returning a JWT;
+# on the normal day this is a single-login pass-through.
+# shellcheck source=scripts/e2e/lib/admin-auth.sh
+source "$(dirname "${BASH_SOURCE[0]}")/e2e/lib/admin-auth.sh"
+export STAGING_API_URL="$BASE"
+export STAGING_ADMIN_USER="admin"
+export STAGING_ADMIN_PASSWORD="$ADMIN_PASS"
+TOKEN=$(admin_login_self_heal) || fail "admin login failed (see stderr)"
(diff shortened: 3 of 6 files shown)


sameerparekh:

Reviewed commit: `4bda9abb`

### BLOCKERS

(none)

### SHOULD-FIX

1. `scripts/e2e/lib/admin-auth.sh:71-99` — **Concurrent-gate race on reset day.** After a staging DB reset, three jobs (`api-smoke-staging`, `gate-3a`, `gate-3b`) all run in parallel — they share `needs: [deploy-staging]` in `.github/workflows/master-api-ui.yml` / the gate3 workflows. Two of them reach the post-reset path concurrently; whichever rotates first wins, and the loser hits one of:
   - **changeme login → 401** (loser sampled the post-rotation state) → helper hard-fails as "neither stored nor changeme worked", even though by then the *stored* password works fine.
   - **change-password → 401** ("Current password incorrect" because the winner already rotated) → helper hard-fails.

   Either failure means two of three gates go red on the very reset day this PR aims to auto-recover from. Operator intervention is then unnecessary (the next CD run will pass), but the CI noise contradicts the stated goal.

   **Fix:** on `changeme`-login-401 *and* on change-password-401, fall back to one more login attempt with the stored password before hard-failing. The "neither worked" error then correctly applies only to the genuine credential-mismatch case (which `test_python_hard_fail_when_neither_password_works` continues to cover, since the mock never rotates between calls).

### NITS

1. `scripts/e2e/lib/test_admin_auth.py:21-22` — `urllib.error` and `urllib.request` are imported but never used.
2. `scripts/e2e-tests.sh:46-48` / `scripts/e2e-router.sh:54-56` — `export STAGING_*` mutates the script's process env for the remainder of the run. Harmless in practice (nothing else in those scripts consults `STAGING_*`), but an `unset` after the token call would keep the contract tight.
3. `scripts/e2e/lib/admin-auth.sh:43` + `scripts/e2e/lib/api_admin.py:34` — `"changeme"` is the post-reset bootstrap constant in two places. Functionally fine (the algorithm is the SSOT; the constant is a leaf), but worth a passing thought as a single env-overridable default if a third caller appears.

### Dimension notes

- **§1 SSOT:** algorithm intentionally lives in two language flavors (bash + python) because the callsites are bash scripts and a Python `AdminAPI`. Both helpers cross-reference the other; correct shape given the existing surface.
- **§2 Testing:** TDD red→green visible (`36fff2e8` test commit precedes `4bda9abb` feature commit). Six tests cover both helpers across all three paths (normal, post-reset, hard-fail). Mock asserts the #623 JSON body shape so a regression to empty body trips the test loudly.
- **§3 architecture / §4 wire contract / §5 migrations / §6 metrics:** N/A — CI script work, no router-agent / wire / DB / enforcement plane involved.
- **§7 hygiene:** focused diff, no AI mentions, `.venv-test/` correctly added to `.gitignore`.
- **§8 security:** password rotation is the deliberate behavior; passwords are JSON-encoded via `python3 -c "json.dumps(...)"` (safe vs. shell quoting); tokens go to stdout while diagnostics go to stderr (the token itself is never echoed in logs).

### VERDICT: REQUEST-CHANGES

Self-healing handshake correctly mirrors the algorithm in #1790 across one bash + one Python entry point, with the #623 JSON-body parse and TDD red→green commits visible. One SHOULD-FIX: the concurrent-gate race on the very reset day this PR targets means two of three parallel gates would still fail under the current algorithm — fall back to a stored-password retry on either changeme-401 or change-password-401 to fully close the loop.

sameerparekh:

Reviewed commit: `6a0f4758` (delta vs `4bda9abb`)

### Prior-findings status

| - | --- | --- |
| SF1 | `admin-auth.sh:71-99` — concurrent-gate race | **ADDRESSED.** Both helpers now retry the stored password on changeme-401 AND on change-password-401 before declaring hard fail. New `_admin_auth_retry_stored_after_race` in `scripts/e2e/lib/admin-auth.sh:153`; mirror logic in `scripts/e2e/lib/api_admin.py:62` + `:88`. Two new tests pin both flavors (`test_bash_race_recovery_on_changeme_401`, `test_python_race_recovery_on_changeme_401`). |
| N1 | unused `urllib.error`/`urllib.request` imports in test | **ADDRESSED.** Removed in `scripts/e2e/lib/test_admin_auth.py`. |
| N2 | `export STAGING_*` mutates caller env | **ADDRESSED.** Both `scripts/e2e-tests.sh:55` and `scripts/e2e-router.sh:62` now `unset` after the token call. |
| N3 | `"changeme"` constant duplicated | **NOT-ADDRESSED** (acceptable — original review classified as NIT only; leaf constant, two callsites, no value in a third indirection until a third caller appears). |

### NEW findings (delta only)

#### BLOCKERS

(none)

#### SHOULD-FIX

(none)

#### NITS

1. `scripts/e2e/lib/test_admin_auth.py:201` — `_SiblingRaceHandler` is declared with a docstring but never used; the two race tests each define their own local `_Handler` class. The stub class is dead; remove it or wire it in.
2. `scripts/e2e/lib/test_admin_auth.py:215-247` and `:251-280` — the two race tests are near-clones (same handler class, same mock setup, different driver). Could share a helper or fixture, but the duplication is small and the divergence (subprocess vs in-process) is visible; not worth the abstraction yet.
3. Tests cover the changeme-401 race; the change-password-401 race is in the algorithm but not directly fixture-tested. The algorithm is symmetric and the changeme-401 case exercises the same recovery code path, so coverage is acceptable, but a one-line variant would close the gap.

### VERDICT: APPROVE

Race recovery cleanly addressed in both helpers with symmetric semantics and a stored-password retry guarded against the genuine credential-mismatch path (the hard-fail tests still pass because the mock never rotates). All prior NITs except the deliberately-deferred constant are addressed. New nits are cosmetic.

PR #1795, opened 2026-06-20
fix(#1792): skip conf-file= for missing blocklist shards (Gate 3a)

### LINKED ISSUE #1792: ci(cd): Gate 3a fails — router DNS refused on fdaa:bbbb:cccc::1#53 (regression from #1788 dnsmasq conf-file streaming)
## Symptom

Master Router CD on \`297cf08413\` failing Gate 3a (ipk + apk):

\`\`\`
TimeoutError: client DNS for example.com never resolved after 60s (52 attempts,
              last=';; communications error to fdaa:bbbb:cccc::1#53: connection refused')
WARNING  lib.vm:vm.py:306 client client1 LAN/DNS not ready after 60s (last: '')
ERROR gate3/test_smoke.py::test_gate3_smoke
\`\`\`

Run: https://github.com/wifihaven/wifihaven/actions/runs/27849554995

The self-healing admin-auth helper (#1790) IS in place and recovered the 401 — proven by the log line \`admin login 401 — attempting post-reset bootstrap with 'changeme'\`. That part works. The failure is **downstream**: the router VM's dnsmasq isn't accepting connections on its IPv6 ULA listener.

## Why this is most likely #1788

The only router-touching merge in this CD window is **#1788** (\`feat(#1782,#1783): stream blocklist render to /tmp shards via dnsmasq conf-file\` — landed as commit \`a82e4cac\` before \`297cf08\`). #1788 changed how the agent emits dnsmasq configuration — from inline \`nftset=\` directives to streamed conf-file shards under /tmp. If the new shard-streaming:

- Generates a syntactically broken conf-file → dnsmasq fails to load it, refuses startup, no LAN-side listener.
- Misses the v6-listener directive entirely (e.g. \`listen-address=fdaa:bbbb:cccc::1\`) → dnsmasq starts but doesn't bind v6.
- Sequences the reload at the wrong time → dnsmasq is reloading when the client probes.

Gate 3a uses the **freshly-built** router image, so it tests the new code. Gate 3b uses the last-published image (older) and isn't gated on this same path.

## Investigation

Boot a Gate 3a VM locally on plex (api.lan):
- SSH into the router VM during failure window.
- \`ps | grep dnsmasq\` — is it running?
- \`logread | grep -iE 'dnsmasq|conf-file|FATAL|error' | tail -50\` — startup errors?
- \`ss -lntp | grep :53\` — is anything listening on 53? on which addresses?
- \`ls -la /tmp/wifihaven-dnsmasq*\` — are the conf-file shards present + readable?
- \`dnsmasq --test --conf-file=/etc/dnsmasq.conf\` (or wherever the wifihaven fragment includes from) — syntax check the rendered config.

## Fix paths

1. **Revert #1788** — fastest path to green CD. Re-introduce the conf-file work in a follow-up with the regression test.
2. **Forward-fix** — if Phase 1 surfaces a single bad line (missing listen-address, malformed includes), patch directly.

## Acceptance

- Gate 3a passes on both ipk + apk.
- Master Router CD goes green.
- Regression test (busted + render_spec) that pins the dnsmasq config includes the v6 ULA listener.

## Related

- #1788 (suspect — likely cause of regression)
- #1782 / #1783 (parent issues for conf-file streaming)
- #1435 (the StevenBlack blocklist sizing investigation that motivated Option 2)
- #1680 (deterministic ULA prefix `fdaa:bbbb:cccc`)
- #1790 (self-healing admin auth — proven working in this run's log)

sameerparekh:
Closes #1792.

## Root cause

`render.dnsmasq` emitted `conf-file=/tmp/wifihaven-blocklist-<id>.conf` for **every** id in `snapshot.blocklists` (`render.lua:449-457`), but `blocklists.render_shards` silently skips an id whose cache file is missing (`blocklists.lua:423-425`) — fetch never completed, transient HTTP error during VM boot, or cap_hit on an oversized list. The dangling `conf-file=` reference made dnsmasq abort at startup with `cannot read /tmp/wifihaven-blocklist-<id>.conf`, taking `:53` down. The Gate 3a smoke client then sees ``';; communications error to fdaa:bbbb:cccc::1#53: connection refused'`` for 60s and times out.

Gate 3a is the only arm that boots the freshly-built router image AND points at `api-staging.wifihaven.net`, so a transient blocklist-fetch latency during VM bring-up was enough to dangle a shard. Gate 3b uses the last-published image (pre-#1788) and isn't on this code path.

## Fix shape

Forward fix — one-line patch, no revert needed:

- `render.dnsmasq(snapshot, opts)` now accepts `opts.bl_shard_exists(id) → bool`. Ids that return false are omitted from the `conf-file=` block. When absent (legacy callers, tests) every id is referenced — back-compat for the existing #1782 tests.
- `policy.apply` wires a default that stats the shard path with `io.open()`. The bl_/bl6_ sets just stay empty for the missing id until the next `render_shards` pass succeeds; dnsmasq stays alive.

## Tests

- RED → GREEN via `test/render_spec.lua` — `omits conf-file= for an id whose shard does not exist on disk (#1792)` (commits in PR show red → green).
- Back-compat assertion: all existing `#1782` conf-file= tests still pass without supplying `bl_shard_exists`.
- Full openwrt busted suite: **774 successes / 0 failures** (one `nft` binary pending on macOS host; the 4 pre-existing `kmod-nf-log` build-script FAILs are unrelated).

## Post-merge

Next `master-router.yml` run validates Gate 3a + 3b green. Will confirm via `gh run list --workflow=master-router.yml --limit 1`.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/blocklists.lua
@@ -416,8 +416,8 @@ function M.render_shards(snapshot, fs, cache_dir, shard_dir, max_bytes, global_b
   for id, bl in pairs(bls) do
     local version    = bl.version
     local cache_path = cache_dir .. "/" .. id .. "-" .. tostring(version) .. ".txt"
-    local shard_tmp  = shard_dir .. "/wifihaven-blocklist-" .. id .. ".conf.tmp"
-    local shard_final = shard_dir .. "/wifihaven-blocklist-" .. id .. ".conf"
+    local shard_final = get_render().shard_path(id, shard_dir)
+    local shard_tmp  = shard_final .. ".tmp"
 
     -- Open the cache file for line-by-line reading.
     local rf = open_read_fn(cache_path)
--- a/openwrt/files/usr/lib/lua/wifihaven/policy.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/policy.lua
@@ -218,6 +218,16 @@ local function exec_ok(rc)
   return rc == 0 or rc == true
 end
 
+-- #1792: default shard-existence check for render.dnsmasq's bl_shard_exists
+-- hook. Module-level so the closure is allocated once, not per apply call.
+-- Reads render.SHARD_DIR at call time (not at module-load) so tests that
+-- mutate render.SHARD_DIR before driving policy.apply still see the override.
+local function default_bl_shard_exists(id)
+  local f = io.open(render.shard_path(id), "r")
+  if f then f:close(); return true end
+  return false
+end
+
 function M.apply(snapshot, write_fn, reload_fn, log, opts)
   log = log or default_log()
   opts = opts or {}
@@ -240,14 +250,18 @@ function M.apply(snapshot, write_fn, reload_fn, log, opts)
   -- whose cache file is missing (fetch not yet completed, transient HTTP
   -- error, cap_hit); a dangling conf-file= ref aborts dnsmasq startup with
   -- "cannot read ..." and :53 returns "connection refused" (Gate 3a regression
-  -- after #1788). Tests inject opts.bl_shard_exists; production uses io.open.
-  local bl_shard_exists = opts.bl_shard_exists or function(id)
-    local p = string.format("%s/wifihaven-blocklist-%s.conf", render.SHARD_DIR, id)
-    local f = io.open(p, "r")
-    if f then f:close(); return true end
-    return false
-  end
-  local dnsmasq_content = render.dnsmasq(snapshot, { bl_shard_exists = bl_shard_exists })
+  -- after #1788). Tests inject opts.bl_shard_exists; production uses the
+  -- module-level default_bl_shard_exists which stats the canonical shard path.
+  local bl_shard_exists = opts.bl_shard_exists or default_bl_shard_exists
+  local dnsmasq_content = render.dnsmasq(snapshot, {
+    bl_shard_exists = function(id)
+      local ok = bl_shard_exists(id)
+      if not ok then
+        log.debug("policy.apply: omitted conf-file= for id %s (shard missing)", tostring(id))
+      end
+      return ok
+    end,
+  })
   local nft_content     = render.nft(snapshot, opts)
 
   -- #414: dnsmasq only needs a full restart when its config-dir file
--- a/openwrt/files/usr/lib/lua/wifihaven/policy.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/policy.lua
@@ -235,7 +235,19 @@ function M.apply(snapshot, write_fn, reload_fn, log, opts)
       pcall(on_apply, { result = result, dnsmasq_restarted = dnsmasq_restarted })
     end
   end
-  local dnsmasq_content = render.dnsmasq(snapshot)
+  -- #1792: gate per-blocklist conf-file= emission on whether the shard file
+  -- actually exists on disk. blocklists.render_shards silently skips an id
+  -- whose cache file is missing (fetch not yet completed, transient HTTP
+  -- error, cap_hit); a dangling conf-file= ref aborts dnsmasq startup with
+  -- "cannot read ..." and :53 returns "connection refused" (Gate 3a regression
+  -- after #1788). Tests inject opts.bl_shard_exists; production uses io.open.
+  local bl_shard_exists = opts.bl_shard_exists or function(id)
+    local p = string.format("%s/wifihaven-blocklist-%s.conf", render.SHARD_DIR, id)
+    local f = io.open(p, "r")
+    if f then f:close(); return true end
+    return false
+  end
+  local dnsmasq_content = render.dnsmasq(snapshot, { bl_shard_exists = bl_shard_exists })
   local nft_content     = render.nft(snapshot, opts)
 
   -- #414: dnsmasq only needs a full restart when its config-dir file
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -83,6 +83,17 @@ local M = {}
 -- wifihaven-agent).
 M.SHARD_DIR = "/tmp"
 
+-- Canonical path of a per-blocklist dnsmasq conf shard (#1792). Single source
+-- of truth — render.dnsmasq, policy.apply's shard-existence check, and
+-- blocklists.render_shards all call this so the prefix/extension can never
+-- drift between writer and reader. `dir` defaults to M.SHARD_DIR; pass an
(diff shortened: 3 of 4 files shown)


sameerparekh:
## Independent PR review — #1795 (Gate 3a `conf-file=` fix)

### BLOCKERS
(none)

### SHOULD-FIX
1. **Shard path constructed in 3 places — single-source-of-truth drift risk.**
   - [openwrt/files/usr/lib/lua/wifihaven/render.lua:469](openwrt/files/usr/lib/lua/wifihaven/render.lua:469) — `"conf-file=%s/wifihaven-blocklist-%s.conf"` (using `M.SHARD_DIR`)
   - [openwrt/files/usr/lib/lua/wifihaven/policy.lua:244](openwrt/files/usr/lib/lua/wifihaven/policy.lua:244) — `"%s/wifihaven-blocklist-%s.conf"` (using `render.SHARD_DIR`)
   - [openwrt/files/usr/lib/lua/wifihaven/blocklists.lua:419-420](openwrt/files/usr/lib/lua/wifihaven/blocklists.lua:419) — `shard_dir .. "/wifihaven-blocklist-" .. id .. ".conf"`

   Three writers of the same path string. If the prefix or extension ever changes, drift here is exactly the kind of bug the [SSoT rule](../AGENTS.md#single-source-of-truth) was added for. Collapse to a single helper (e.g. `render.shard_path(id)`) and call it from all three sites. The PR introduces the third location, so this is the right moment to fold it in.

### NITS
1. **Silent shard-skip is invisible to operators.** [openwrt/files/usr/lib/lua/wifihaven/policy.lua:241-248](openwrt/files/usr/lib/lua/wifihaven/policy.lua:241) — when `bl_shard_exists` filters an id (shard missing at apply time), nothing is logged. `refresh_blocklists` already logs `render_shards`-skipped ids, but a shard that disappeared *between* `refresh_blocklists` and `policy.apply` (or one that `render_shards` skipped silently because the cache file was simply absent — see [blocklists.lua:423-425](openwrt/files/usr/lib/lua/wifihaven/blocklists.lua:423), no log there either) leaves no trail at apply time. A single `log.debug("policy.apply: omitted conf-file= for id %s (shard missing)", id)` inside the filter loop would make the transient enforcement gap diagnosable.

2. **Closure allocated per apply.** [openwrt/files/usr/lib/lua/wifihaven/policy.lua:242](openwrt/files/usr/lib/lua/wifihaven/policy.lua:242) — the default `bl_shard_exists` function is rebuilt on every `policy.apply` call. Hoist it to a module-level constant. Cost is negligible (policy.apply fires every 60s in steady state) — pure hygiene.

### Dimensions checked
- **1. Duplicated logic / SSoT** — flagged above (path string, SHOULD-FIX).
- **2. Testing integrity** — TDD red→green visible: failing test committed first ([639e899f](https://github.com/wifihaven/wifihaven/commit/639e899f)), fix in [6ae890c9](https://github.com/wifihaven/wifihaven/commit/6ae890c9). New test pins the bug; back-compat assertion covers existing callers. No tests weakened, deleted, or `pending`. ✅
- **3. Architectural invariants** — fix keeps dnsmasq alive (does NOT make DNS the enforcer); router stays a dumb applier; no new wire field; no new `/tmp` writer. `io.open` opens for read only and closes immediately. ✅
- **4. Backwards compat / wire contract** — no wire change. `render.dnsmasq`'s second positional arg is additive; existing callers (tests, `policy.apply`) work unchanged. ✅
- **5. Migrations** — none.
- **6. Metrics + dashboard** — no new code path that warrants a counter (the existing `blocklist shard cap` log + `record_blocklist_failures` cover the upstream skip). Mild gap noted as NIT 1.
- **7. Scope / hygiene** — diff is focused on render + policy + matching test. No drive-bys. No AI mentions in commit messages or PR body. ✅
- **8. Security** — no SQL, no secrets, no new auth surface. `io.open` path is constructed from server-supplied blocklist ids; ids come from the policy snapshot (controlled), and the only consequence of a hostile id would be a missed `io.open` on an attacker-chosen `/tmp/wifihaven-blocklist-*.conf` (read-only stat). ✅

### VERDICT: APPROVE

Targeted forward-fix for the #1788 regression — `render.dnsmasq` now skips `conf-file=` lines for ids whose shard file is absent, keeping dnsmasq alive when fetch is mid-flight or transient. No BLOCKERs; one SSoT cleanup worth folding in before merge. Ready for merge-when-ready after the path-helper is collapsed.

sameerparekh:
## Independent PR review — #1795 (re-run on a9b1909c)

### Prior-findings status (vs 6ae890c9 → a9b1909c)

| Prior finding | Status | Evidence |
|---|---|---|
| SHOULD-FIX 1 — shard path constructed in 3 places | **ADDRESSED** | New [render.shard_path(id, dir?)](openwrt/files/usr/lib/lua/wifihaven/render.lua:91) is the single source. [render.lua:482](openwrt/files/usr/lib/lua/wifihaven/render.lua:482), [policy.lua:225](openwrt/files/usr/lib/lua/wifihaven/policy.lua:225), and [blocklists.lua:419-420](openwrt/files/usr/lib/lua/wifihaven/blocklists.lua:419) all call it; the literal `"wifihaven-blocklist-"` string now appears once. |
| NIT 1 — silent shard-skip invisible at apply time | **ADDRESSED** | [policy.lua:258-262](openwrt/files/usr/lib/lua/wifihaven/policy.lua:258) wraps `bl_shard_exists` with a `log.debug("policy.apply: omitted conf-file= for id %s (shard missing)", id)` log on filter. |
| NIT 2 — default closure allocated per apply | **ADDRESSED** | Default check hoisted to module-level [default_bl_shard_exists](openwrt/files/usr/lib/lua/wifihaven/policy.lua:221) ; `M.apply` no longer rebuilds it per call. A small log-wrapping closure still allocates per apply but is necessary to capture `log` — acceptable. |

### New findings on the delta `6ae890c9...a9b1909c`

#### BLOCKERS
(none)

#### SHOULD-FIX
(none)

#### NITS
(none)

The delta is purely the refactor I requested: collapse the path helper, hoist the default closure, add the debug log. No fresh dimensions tripped:
- **SSoT** (#1) — collapsed cleanly; the literal basename now lives in `render.shard_path` only.
- **Testing integrity** (#2) — no test edits in the delta; 774 successes / 0 failures on the full busted suite (4 pre-existing `kmod-nf-log` build-script FAILs are unrelated).
- **Architectural invariants** (#3) — unchanged.
- **Wire contract** (#4) — unchanged.
- **Migrations** (#5) — none.
- **Metrics / dashboard** (#6) — the new `log.debug` is the right surface for a transient enforcement gap; a counter would be misleading (skips here are expected during early boot until `render_shards` lands).
- **Scope / hygiene** (#7) — focused, no AI mentions.
- **Security** (#8) — unchanged.

### VERDICT: APPROVE

Prior SHOULD-FIX + both NITs all ADDRESSED; the delta introduces no new findings. Ready for merge-when-ready.

PR #1797, opened 2026-06-20
fix(#1793): canonicalize v6 keys in dns-cache for FQDN attribution

### LINKED ISSUE #1793: bug(prod): IPv6 destinations recorded with bare literal as host (FQDN attribution broken or not deployed)
## Symptom (prod, observed 2026-06-19)

Connection event in the operator's view:

\`\`\`
2607:f8b0:400f:0805:0000:0000:0000:2002   ipv6   ✗ blocked   category: ads
\`\`\`

The \`host\` field is a **bare IPv6 literal** instead of a resolved FQDN. The IP \`2607:f8b0::/32\` is Google. Expected: \`host = <some.google.com.host>\`, NOT a literal.

The category-ads classification works because the blocked rule fired on IP membership in \`bl6_<mac>_ads\` — the address ended up there because some ads-blocklist host's AAAA record points at this Google IP (shared-CDN collision, #1648 class). That part is consistent.

The bare-literal-as-host is the real bug.

### Hypothesis A: code-side cache lookup gap

The v6 dns-tail → cache → conntrack attribution chain works in unit tests but has a subtle production gap. Most likely shape:

- **Address canonicalization mismatch.** dns-tail writes the cache key in one format (e.g. RFC 5952 compressed \`2607:f8b0:400f:805::2002\`) but conntrack.lua queries by the dst_ip from \`/proc/net/nf_conntrack\` in a different format (expanded \`2607:f8b0:400f:0805:0000:0000:0000:2002\` or some other variant). Cache miss → fall through to bare literal.
- **dns-tail's AAAA path is hooked but doesn't actually write to the cache file** (vs the A path that does). #1668's PR #1673 claimed "dns_log AAAA + nft eb6_/bl6_ fallback" — verify dns_log.lua actually writes AAAA records to /tmp/wifihaven-dns-cache.txt.
- **Cache file format doesn't carry the family.** If the cache file is "ip\\thost" and the lookup is case- or width-sensitive on v6, there's a parser gap.

### Hypothesis B: prod router is on pre-#1668 code

Recent router CD has been broken intermittently: #1716 (24-day-old agent) → #1717 (CD unblock) → #1789 (staging deploy 502) → #1792 (Gate 3a DNS refused). If openwrt-latest hasn't published successfully since #1668 / #1652 / #1651 landed, the operator's router never picked up the v6 attribution code. The "bug" is just "we never deployed the fix to this router."

## Investigation

1. **Check what version is actually on the prod router.**
   - \`ssh root@router.lan 'apk info wifihaven 2>/dev/null || opkg list-installed | grep wifihaven'\`
   - Compare against the SHA each release tag carries. Find when v6 attribution code landed (commits a29a55bb #1668 and earlier).
2. **If router IS on post-#1668 code (Hypothesis A): code regression.**
   - On router: \`cat /tmp/wifihaven-dns-cache.txt | grep -i '2607:f8b0' | head\` — is Google's AAAA record in the cache at all?
   - \`cat /tmp/wifihaven-dns-tail.log | grep -i aaaa | tail -20\` — does dns-tail see AAAA records?
   - Compare the cache file's canonical form of a known v6 against what \`conntrack -E -e NEW -f ipv6\` emits → are they the same string?
3. **If router IS on pre-#1668 code (Hypothesis B): operational, no code change.**
   - The fix is for router CD to get green long enough to publish a fresh openwrt-latest. Tracked under #1792.
   - File a follow-up to add a fleet-wide observability gauge so we know when a router is on a stale agent vs latest (#1718 metrics may already cover this — verify).

## Acceptance

- Root cause classified A vs B.
- For A: fix the cache canonicalization / AAAA write / family handling; busted test pinning v6 round-trip.
- For B: confirm next openwrt-latest publish includes the fix; operator-side validation on prod router after auto-update.

## Related

- #1668 — v6 attribution end-to-end (the work this fix depends on)
- #1652 — IPv6 outer-header in sni.parse_packet
- #1651 — QUIC SNI parsing
- #1716 / #1717 / #1789 / #1792 — the CD-broken-since chain that may have stalled prod deployment
- #1640 — connection_event reason carries matched rule (related observability)
- #1648 — shared-IP-pool blocking strategy umbrella

sameerparekh:
Closes #1793.

## Root cause (Hypothesis A — code-side canonicalization gap)

**Phase 1 — version.** The operator's prod router (`hostname=router`, gateway `192.168.10.1`) runs `wifihaven-0.3.12-r1` → commit `1cfb3f11` (2026-06-15). Both v6-attribution merges are ancestors: `a29a55bb` (#1668) and `98e43a89` (#1688). So the router **is on post-#1668 code** — this is a code-side gap, not a stale deploy (rules out Hypothesis B).

**Phase 2 — the gap.** The bug only affects **blocked** v6 flows. An *allowed* flow is confirmed at postrouting and surfaces via `conntrack -E NEW`; a **dropped** flow is killed in the forward chain before confirm, never emits IPCT_NEW, and is instead made visible by the **nflog path** (`render.lua` `log prefix "wh_drop:…"` → `wifihaven-nflog-tail` → `nflog.run`). The two readers spell IPv6 differently:

| Source | v6 form | Example |
|---|---|---|
| dnsmasq query-log → dns cache key | **compressed** RFC 5952 | `2607:f8b0:400f:801::2002` |
| `conntrack -E`/`-L` `dst=` | **compressed** | `2607:f8b0:400f:805::200a` |
| kernel netfilter LOG `DST=` (nflog) | **fully expanded, zero-padded** | `2607:f8b0:400f:0801:0000:0000:0000:2002` |

Evidence captured on the prod router:
- Cache keys are compressed: `2607:f8b0:400f:807::2002 -> pagead2.googlesyndication.com` (555 v6 keys).
- `conntrack -L` v6 dst is compressed: `dst=2607:f8b0:400f:805::200a`.
- Live `wh_drop` LOG lines are expanded: `… category:ads … DST=2607:f8b0:400f:0801:0000:0000:0000:2002` — byte-identical to the issue's reported `host` literal `2607:f8b0:400f:0805:0000:0000:0000:2002`.

`nflog.run` does `lookup(parsed.dst_ip)` with the **expanded** `DST=`, but the cache is keyed by the **compressed** form → exact-string `tbl[dst_ip]` miss → `hname=nil` → bare `type="ipv6"` literal. Conntrack escapes this because its `dst=` already matches the compressed key — which is exactly why *only blocked* v6 flows show the bare literal.

## Fix

A single canonicalization helper **`host_norm.canon_ip(ip)`** → fully-expanded lowercase 8-group v6 form (v4 / non-v6 / nil pass through). Per AGENTS.md §single-source-of-truth, it is routed through at **every** cache key-use site so compressed and expanded inputs collapse to one key:

- writer: `dns_log` `store`, `load_table`
- reader: `dns_log` instance `lookup`, agent `lookup_hostname` query

**No cache-format invalidation across upgrade:** `load_table` canonicalizes *on read*, so an existing router's file full of compressed keys still resolves after the new agent lands.

## Tests (red → green visible in history)

- `host_norm_spec`: `canon_ip` — compressed/expanded equivalence, `::` expansion in any position, hex lowercasing, v4/nil passthrough, idempotence.
- `dns_log_spec`: ingest compressed AAAA → `lookup(expanded)` hits; `load_table` compressed key queryable by `canon_ip(expanded)`.
- `nflog_spec`: end-to-end reproduction — a `wh_drop` line with an **expanded** `DST=` resolves to the FQDN stored under the **compressed** cache key (was: bare literal).
- Wired `nflog_spec` + `host_norm_spec` into the CI busted list and `run_tests.sh` — both existed but were not gating, so this regression had no unit-test gate.

Full openwrt busted suite: **828 passing / 0 failing** (1 unrelated pending — `nft` binary absent locally).

## Phase 4 — verified on real router hardware

Ran the round-trip on the test router `openwrt.lan` (Lua 5.1.5) using the real prod vector `2607:f8b0:400f:807::2002 → pagead2.googlesyndication.com`: the expanded kernel-LOG form `2607:f8b0:400f:0807:0000:0000:0000:2002` now resolves to the FQDN through both the instance and the `load_table` file path. All assertions PASS. (Non-destructive: modules staged under `/tmp`, agent untouched, cleaned up after.)

## Note — on-disk cache format & upgrade window

After this change the dns-cache stores v6 keys in fully-expanded form (≈2× longer per v6 line; still well within the fixed-size atomic-rewrite bound of ≤`max_entries`, so no tmpfs concern). `load_table` canonicalizes on read, so an existing router's compressed-key file stays valid. One transient effect: during the package-upgrade restart window, an **old** agent reading the **new** dns-tail's expanded-key file would briefly miss v6 attribution for *allowed* flows; it self-heals the moment the new agent starts.


sameerparekh:
## Independent PR review

Read-only adversarial pass over `git diff origin/main...HEAD` against the 8-dimension checklist. First run for this PR.

**BLOCKERS**
(none)

**SHOULD-FIX**
1. `openwrt/files/usr/lib/lua/wifihaven/dns_log.lua:377` — the `self.lookup` expiry path reads via the canonical key (`entries[host_norm.canon_ip(ip)]`, line 374) but the TTL-expiry delete on line 377 still uses the **raw** `ip`: `entries[ip] = nil`. For a v6 key whose query spelling differs from the stored canonical spelling (e.g. lookup by compressed `2607:f8b0:400f:801::2002` against an entry stored under the expanded canonical key), an *expired* entry is found, but the delete targets a non-existent raw key — the real entry is **not** removed while `entry_count` is decremented anyway. Repeated expired-key lookups can drive `entry_count` negative and defeat the `max_entries` bound (router RAM). **Currently latent** — the instance `self.lookup` is not called in production (dns-tail uses `ingest_line`/`dump_text`/`resolve_head`; the agent uses the `load_table` file path), which is why this is SHOULD-FIX not BLOCKER — but the PR introduced the read/delete inconsistency and the new round-trip tests exercise this method. Fix: compute `local key = host_norm.canon_ip(ip)` once and use it for both the read and the delete. Ideally add a test asserting `entry_count` integrity after an expired v6 lookup by a non-canonical spelling.
2. `diagnosis.md` (new, repo root) — an investigation artifact that duplicates the PR body; it isn't a shipped doc and there's no precedent for root-level investigation `*.md`. Drop it from the PR (keep local) or relocate under `docs/`. Scope/hygiene.

**NITS**
1. The on-disk dns-cache now stores v6 keys in fully-expanded form (≈2× longer per v6 line). Still well within the fixed-size atomic-rewrite bound (≤ `max_entries`), so not a tmpfs concern. One transient effect worth a line in the PR body: during the package-upgrade restart window an **old** agent reading the **new** dns-tail's expanded-key file would briefly miss v6 attribution for *allowed* flows (conntrack compressed dst vs expanded file key); it self-heals once the new agent starts.
2. (optional follow-up) A `dns_cache` lookup `{result=hit|miss}` counter would surface this attribution-miss class directly. Not required for a bug fix.

**Notes on dimensions that pass**
- §1 Single-source: `canon_ip` is the one canonicalizer, routed through `store`/`lookup`/`load_table` + agent `lookup_hostname`. (The line-377 inconsistency above is the lone exception.)
- §2 Testing: red→green visible (`1701583c` test, `4c59c979` fix); regression test fails without the fix (pure `lookup(EXPANDED)` miss); no existing assertions weakened; `nflog_spec` + `host_norm_spec` newly **added to the CI gate** (they existed but weren't gating).
- §3 Invariants: no DNS-as-enforcement; router stays a dumb applier (mechanical canonicalization, no policy); no new wire field; dns-cache is a fixed-size atomic rewrite (exempt from the `/tmp` grow-writer rule).
- §4 Wire: cache file is local tmpfs IPC, not the wire; the `host` field flips bare-literal→fqdn for blocked v6, which the API's HostId already accepts (additive/correct).
- §7/§8: focused diff, no AI mentions (Co-Authored-By trailer only), no secrets/SQL/new routes.

VERDICT: APPROVE
Canonicalizing v6 dns-cache keys via a single `host_norm.canon_ip` helper is the correct, minimal fix for the compressed(dnsmasq)-vs-expanded(kernel-LOG) mismatch, with a real-hardware-verified v6 round-trip and a proper red→green test gate. No blockers. Two SHOULD-FIX items (a latent raw-vs-canonical delete-key inconsistency in `self.lookup`, and the stray `diagnosis.md`) should be cleaned up before merge.

sameerparekh:

Re-run on push `a7781479` (delta since `4c59c979`).

**Prior findings status**

|---|---|---|
| SHOULD-FIX 1 | `dns_log.lua` `self.lookup` deleted expired entry under the raw key while reading via the canonical key (entry_count corruption) | **ADDRESSED** — `lookup` now computes `local key = host_norm.canon_ip(ip)` once and uses it for both the read and the expiry delete (`dns_log.lua:377-385`). New regression test pins `entry_count` integrity across v6 spellings on expired-entry eviction, including the double-lookup-no-negative case (`dns_log_spec.lua`). |
| SHOULD-FIX 2 | `diagnosis.md` investigation artifact committed at repo root | **ADDRESSED** — file removed from the tree; its content remains in the PR description. |
| NIT 1 | on-disk expanded-key format / transient upgrade-window note | **ADDRESSED** — note added to the PR body. |
| NIT 2 | optional dns_cache hit/miss metric | not done (explicitly optional follow-up; not required). |

**New findings in the delta**
(none)

The delta is limited to the `lookup` key fix, its regression test, and the `diagnosis.md` removal — all matching the prior recommendations, no new logic introduced.

VERDICT: APPROVE
Both SHOULD-FIX items addressed; delta is clean and adds a regression test that pins the entry_count invariant. No blockers, no open findings.

sameerparekh:
## Independent PR review — re-run (rebase only)

Re-run after a **rebase onto `origin/main`** (to pick up the #1801 fix that unblocks the frontend check). This was a clean rebase with **no content change** — `git diff origin/main...HEAD` is byte-identical to the previously-approved `a7781479` (same 8 files: `dns_log.lua`, `host_norm.lua`, `wifihaven-agent`, the three `*_spec.lua`, `run_tests.sh`, `ci.yml`). No conflicts, no new logic.

**Prior findings:** all remain ADDRESSED (carried through the rebase unchanged).
**New findings in the delta:** none (the delta is purely the new base commits from `main`, not this PR's changes).

Post-rebase openwrt busted suite: **831 passing / 0 failing** (1 unrelated pending — `nft` binary absent locally).

VERDICT: APPROVE
Rebase-only update; the approved fix is unchanged and still green. No blockers, no open findings.

Issue #1798, opened 2026-06-20
Remove admin SPA app-definition editing — apps are authored via template updates only


sameerparekh:
## Problem

App *definitions* (name, slug, icon, host-set, create/delete) are editable two ways today: through the **admin SPA App-management UI** and as **built-in templates** seeded from code (`AppTemplates`). Operator-side editing is the source of the duplicate-app churn we keep cleaning up:

- The duplicate-app incidents (#1777 / #1794) trace back to operator-created rows colliding with template slugs, forcing the `-template` fallback in `AppTemplates.findFreeSlug` and splitting assignments/usage across two rows.
- The same logical app gets two slugs by hand (e.g. prod had `tinker-cad` operator-created alongside the `tinkercad` template row).

**Decision (operator, 2026-06-20): app definitions will be authored exclusively as template updates from now on.** The admin SPA app-editing surface should be removed so the only way to add/rename/re-host an app is to update `AppTemplates` in code (which then seeds/reconciles). This eliminates the slug-collision class at the source.

## Scope — remove operator-facing app-*definition* editing

**SPA (`web/`):**
- `web/src/pages/AppsPage.tsx` — remove create / edit-name-slug-icon / edit-hosts / delete controls (the whole app-management editing surface). Decide whether the page becomes read-only (list templates + their hosts) or goes away entirely.
- `web/src/api/client.ts` `apps.*` — drop `create`, `update` (PATCH), `delete`, host-edit wrappers used only by the editor.
- `AppsPage.test.tsx` — update/remove accordingly.

**API (`api/src/routes/AppRoutes.scala`) — app-*definition* mutators to retire:**
- `POST /api/apps` (create) — [AppRoutes.scala:110](api/src/routes/AppRoutes.scala:110)
- `PUT /api/apps/:id` (update) — [:156](api/src/routes/AppRoutes.scala:156)
- `PUT /api/apps/:id/hosts` — [:192](api/src/routes/AppRoutes.scala:192)
- `PATCH /api/apps/:id` — [:327](api/src/routes/AppRoutes.scala:327)
- `DELETE /api/apps/:id` — [:184](api/src/routes/AppRoutes.scala:184)

## Explicitly KEEP

[...]

PR #1799, opened 2026-06-20
fix(#1794): auto-reconcile app templates on boot so duplicate apps self-heal (PR 1 of 2)

### LINKED ISSUE #1794: bug(prod): duplicate apps STILL present after #1777 reconciliation deployed
## Symptom (prod, 2026-06-19)

Operator still sees duplicate apps in prod even though #1777's reconciliation fix (PR #1778) landed and deployed.

## Why this is plausible

#1777 shipped an **idempotent admin reconciliation ENDPOINT** (`POST /api/admin/apps/reconcile-templates`) — it does NOT run automatically on deploy. So either:

1. **The endpoint was never invoked** against prod. The deploy shipped the *capability*; nobody POSTed it. (Most likely.)
2. **It was invoked but didn't catch this duplicate class.** #1777's reconciler handled `X-template` vs `X` pairs. If the prod duplicates are a DIFFERENT shape — e.g. two operator-added rows with the same slug, or two rows that differ only by trailing whitespace / case, or apps duplicated via a different mechanism — the reconciler's matching logic missed them.
3. **A duplicate-creating path is still active** — something re-inserts apps on each deploy / snapshot / poll, so reconciliation gets undone (the template-seeder still running, per the #1608 one-shot-delete concern).

## Investigation

1. **Re-query prod apps:**
   ```sql
   SET statement_timeout='5s';
   SELECT slug, name, COUNT(*) AS n, array_agg(id) AS ids
   FROM apps GROUP BY slug, name HAVING COUNT(*) > 1;
   -- AND check near-duplicates:
   SELECT id, slug, name FROM apps ORDER BY name, slug;
   ```
   Characterize the EXACT duplicate shape. Is it slug collision? name collision with different slugs? `-template` survivors the first pass missed?

2. **Was the endpoint run?** Check API logs for `/api/admin/apps/reconcile-templates` invocations. If never called → the immediate fix is to run it (operator action) + understand why duplicates persisted if it WAS run.

3. **Is a seeder still creating duplicates?** Check whether the template-seeding code runs on startup. Per #1608's convention, one-shot seeders should be deleted after deploy. If the seeder runs every boot and the reconciler runs on-demand, every deploy re-creates the duplicates the operator just cleaned.

## Fix

Depends on the shape found in step 1:

- **If endpoint never run + reconciler logic is correct**: invoke it against prod (operator action via curl), then make reconciliation run automatically (a startup step OR a periodic job) so it's not a manual dance. BUT — only if the duplicate-creating path is also stopped, else it's whack-a-mole.
- **If reconciler missed a duplicate shape**: extend the reconciler's matching to cover it (name-collision, whitespace/case-normalized slug, etc.). Add a regression test for the exact prod shape.
- **If a seeder is re-creating duplicates**: stop the seeder (the real root cause). Then reconcile once. This is the #1608 delete-deployed-one-shots convention applied.

The DURABLE fix is almost certainly: stop whatever re-creates duplicates + add a unique constraint on `apps.slug` (or `(slug, household)` if multi-tenant) so the DB rejects duplicates at the source. A schema-only migration adds the constraint; the reconciler de-dups existing rows first so the constraint can be added cleanly.

## Acceptance

- Root cause of the PERSISTENT duplicates identified (never-run / missed-shape / re-created).
- A unique constraint on the app identity column prevents recurrence (schema-only PR if added).
- Prod de-duplicated (operator runs the reconcile, or a one-shot script).
- Regression test pinning the exact prod duplicate shape.

## Related
- #1777 / PR #1778 — the reconciliation endpoint that didn't fully resolve it
- #1705 / #1710 — template seeding (possible duplicate source)
- #1608 — delete one-shot seeders after deploy convention

sameerparekh:
## Problem

Operator still saw duplicate apps in prod after #1777's reconcile endpoint deployed (#1794).

**Root cause — confirmed against prod (read-only `GET /api/apps`, 2026-06-20):** `apps.slug` is already `UNIQUE` (V28), so no duplicate is a slug collision. Every dup is the **`<slug>` (operator/legacy, `template_id` NULL) + `<slug>-template` (seeded, `template_id` set)** shape — exactly what #1777's reconciler handles. 8 pairs on prod: `1password`, `eaglercraft`, `giphy`, `google-play`, `moc-pilot`, `plex`, `wifihaven`, `x`.

#1777 shipped the reconcile only as an **on-demand endpoint** (`POST /api/admin/apps/reconcile-templates`). It was **never invoked** against prod, so the pairs persisted. The seeder can't self-heal them: the `-template` row carries the `template_id` the seeder keys on (`findByTemplateId`), so it gets *preserved* every boot rather than merged.

## Fix (this PR — code)

Wire `AppReconciler.reconcileTemplates` into the boot sequence right after the template seed:
- **Idempotent + cheap** (indexed `findBySlug`); a no-op once clean.
- **Safe to repeat / no whack-a-mole:** `mergeAppInto` transfers `template_id` onto the canonical row, so the next seed pass finds the canonical and never re-creates the `-template` fallback. This stops the duplicate *source*.
- **Not a one-shot** (unlike #1608-class seeders) — it stays as a permanent self-heal, behaviorally consistent with the seed pass (which already re-unions template hosts every boot).

### Test
Regression test reproduces the **exact prod shape** through the real seed→reconcile boot pipeline: an operator `1password` row carrying assignments + a seeded `1password-template`. Asserts the seed first re-creates the dup (pinning the bug), reconcile collapses to one canonical row with assignments preserved and template hosts unioned, and the next reboot's reconcile is a clean no-op. Full `api.test` suite green.

## Not handled here (by design)
- **`tinker-cad` vs `tinkercad`** — a slug *variant* near-dup, not a `-template` suffix, so structurally out of the reconciler's reach. Code does **not** grow fuzzy hyphenation matching (risks merging genuinely-distinct apps). One-off operator cleanup (`DELETE /api/apps/21`, 0 assignments).
- **DB identity guard** — a partial unique index on `template_id` ships as a **separate schema-only PR** (PR 2 of 2) per the migration-isolation rule.

## Operator pre-flight
1. Merge + deploy **this PR** → next API boot auto-reconciles the 8 `-template` pairs (no manual curl).
2. Delete the orphan `tinker-cad` (app id 21) via the SPA / `DELETE /api/apps/21`.
3. Then merge the schema-only PR 2 (partial unique index on `template_id`).

Closes #1794 (together with the PR 2 schema guard).

### DIFF
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -130,8 +130,8 @@ object Main extends ZIOAppDefault {
             _ <- ZIO.logInfo("global profile sentinel ensured (is_global=TRUE)")
             // #768: seed the starter library of app templates. Idempotent —
             // operator host edits on previously-seeded apps are preserved.
-            seedSummary <- AppTemplates.seed(appRepoForSeed, templates)
-            _           <- ZIO.logInfo(
+            seedSummary      <- AppTemplates.seed(appRepoForSeed, templates)
+            _                <- ZIO.logInfo(
               s"app_templates seeded (${templates.size} templates): " +
                 s"created=${seedSummary.created.size} ${seedSummary.created.mkString("[", ",", "]")}, " +
                 s"repopulated=${seedSummary.repopulated.size} ${seedSummary.repopulated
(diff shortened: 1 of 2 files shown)


sameerparekh:
Closing — wrong direction. Per operator (2026-06-20), the duplicate-app source is being removed at the root by #1798 (remove admin SPA app-definition editing); apps will be authored exclusively as template updates. With no operator-created rows colliding with template slugs, the `-template` fallback never fires and there's nothing to auto-reconcile. The companion V60 `UNIQUE(template_id)` migration is also dropped for the same reason. Existing prod dup pairs will be collapsed by a one-time run of the existing #1777 `POST /api/admin/apps/reconcile-templates` endpoint; the reconciler itself is then slated for deletion (new follow-up issue, blocked by #1798).

Issue #879, opened 2026-05-22
openwrt(#833 follow-up): rx counter at forward hook attributes download bytes to router's own MAC


sameerparekh:
## Problem

#833 added a per-direction rx nftables counter `mac_ip_tracking_rx { ether daddr . ip saddr }` and updates it at the `forward` hook ([`openwrt/files/usr/lib/lua/wifihaven/render.lua:491-498`](openwrt/files/usr/lib/lua/wifihaven/render.lua)). Unit tests pass, but on a live router the rx-direction bytes get attributed to the **router's own WAN MAC**, not the LAN device that actually received them.

Reason: at the `forward` hook on a routed (non-bridged) packet WAN→LAN, the ethernet header is the one received on the WAN ingress side. So:
- `ether saddr` = upstream gateway MAC
- `ether daddr` = the router's WAN-interface MAC

Neighbor lookup that rewrites the L2 daddr to the LAN device's MAC happens later (during egress/postrouting, via the neighbor cache). So `ether daddr` at `forward` is not the LAN device.

## Evidence (live, on dev router `openwrt.lan` after #833 was rsynced onto it)

Router interfaces:

```
br-lan / eth0  HWaddr 94:83:C4:D4:A7:25
eth1           HWaddr 94:83:C4:D4:A7:23
```

DB rows for one upload bucket (post-#833):

```
period_start            | mac                | host_value      | bytes_in | bytes_out
2026-05-22 18:59:06+00  | 4a:68:49:c1:79:08  | 162.159.197.2   | 17989709 |         0   ← Sameer Work iPad (tx side)
2026-05-22 18:59:06+00  | 94:83:c4:d4:a7:23  | 162.159.197.2   |        0 |   4190232   ← router's WAN MAC (rx side, wrong!)
```

Every rx-direction row is attributed to one of the router's own interface MACs (`a7:23` or `a7:25`). The actual LAN device that received the bytes (`4a:68:49:c1:79:08`) gets `bytes_out=0`. The UI groups by device, so download bytes never show up against the device.

Both `94:83:c4:d4:a7:23` and `94:83:c4:d4:a7:25` (the WAN- and LAN-side router MACs) appear as the "device" for rx rows — confirms the hook sees whichever ether-daddr the packet had at forward time.

## Proposed fix

Key the rx counter by IP only, then attribute to a device MAC in the agent via the lease table:

```nft
[...]

PR #1802, opened 2026-06-20
fix(#1796): account IPv6 traffic in per-device usage (was IPv4-only)

### LINKED ISSUE #1796: bug(prod): mathplayground.com traffic not visible for either iPad (host attribution / reporting gap)
## Symptom
Both test iPads are actively browsing **mathplayground.com** right now, but **no traffic to that host appears** in the prod UI / query logs / per-device traffic for either device.

- **Prima** iPad — Adults profile — MAC `76:2d:95:47:d1:8e`
- **Octavius** (kid) iPad — Kids profile — MAC `fa:10:cd:84:78:22`

(MACs from operator's test-device note.)

## Why this matters
Traffic that the user can see happening on the device is invisible in WifiHaven — a per-host attribution or reporting-pipeline gap. If hosts go unattributed, policy/usage/blocklist visibility for that host is all broken downstream.

## Hypotheses to check (observe before theorizing — see /incident-investigation)
1. **Publish-gating / stale router agent (most likely, ties to #1793).** Master Router CD has been red since 2026-06-15; the agent-side host-attribution fix #1763 and any v6 work are in `main` but NOT published to `openwrt-latest`. Prod routers may be running stale agent code that fails to attribute this host (esp. if mathplayground.com is reached over IPv6 — bare v6 literal as host, cf. #1793).
2. **DNS attribution miss.** dnsmasq's forward-lookup nftset/ipset callback didn't populate the per-MAC resolved-hosts set for mathplayground.com (CNAME chain, CDN fan-out, or a domain the device resolved via DoH / cached / hardcoded IP → never seen by our resolver).
3. **Reporting pipeline gap.** Host IS resolved/attributed at the router but the connection_event / traffic_report ingest or the query-log/traffic UI query drops or mis-buckets it.
4. **It's there but the UI query/filter hides it** (time window, grain, device filter, host-normalization stripping a subdomain like www.).

## Investigation
- Confirm whether mathplayground.com (and its CDN/CNAME targets) appears in prod `connection_events` / `query_logs` / traffic for either MAC over the last hour (READ-ONLY against prod).
- Determine if the host resolves over A vs AAAA and whether the device is using DoH / cached / hardcoded IPs (would route around our DNS attribution entirely → `blockIpOnly` territory).
- Check the router agent version on prod vs latest; if stale, this is gated on #1792/#1793.
- Walk the 4-hop pipeline (device → router → API ingest → UI) per the /wifihaven-live-debug skill.

## Likely outcome
Either (a) confirmed downstream of #1792/#1793 publish-gating → close as dup/gated once routers update, or (b) a genuine attribution/reporting bug → fix with a failing test first.

Relates to #1793, #1792.

sameerparekh:
## Problem

mathplayground.com (and any host) was invisible in prod per-device traffic for a device that reached it over **IPv6**, while the same host showed up for a device that used IPv4. Root-caused live on the prod router in #1796.

The **enforcement** plane is fully dual-stack (`ip6 daddr` + `ea6_/eb6_/bl6_/global_allow6/resolved6_`), but the **usage byte-accounting** plane was never extended to IPv6:

```
set mac_ip_tracking      { type ether_addr . ipv4_addr ... }
set ip_pair_tracking_rx  { type ipv4_addr . ipv4_addr  ... }
rule: iifname "br-lan" update @mac_ip_tracking     { ether saddr . ip daddr }   # ip daddr = IPv4 only
rule: oifname "br-lan" update @ip_pair_tracking_rx { ip daddr . ip saddr }
```

So every IPv6 destination flow was never counted → never reached `host_for_ip` → never appeared in `/api/usage/traffic`, `recent-apexes`, or `/api/dashboard/now`. Live proof on the prod router: a 2.2 MB `ESTABLISHED` flow to `2606:4700::...:6812:0446` (mathplayground v6) being accounted nowhere.

This is distinct from #1793 (the **connection_events** path records a bare v6 literal instead of the FQDN). This PR is the **usage/traffic** path, which dropped v6 entirely.

## Fix

- **`render.lua`**: declare `mac_ip6_tracking` (`ether_addr . ipv6_addr`) and `ip_pair6_tracking_rx` (`ipv6_addr . ipv6_addr`), and add `ip6 daddr`/`ip6 saddr` update rules to the existing tx/rx accounting chains — mirroring the v4 sets.
- **`wifihaven-agent`**: read both families' tx + rx sets through the existing family-agnostic parsers; resolve v6 rx `lan_ip→mac` via the NDP neighbor cache (`conntrack.parse_arp_table` reads `/proc/net/arp` + `ip -6 neigh`) since DHCP leases don't cover SLAAC/v6 addresses; reset the v6 sets at flush so they don't double-count.
- **No canonicalization needed**: `nft -j` emits compressed RFC 5952 v6 that already matches dnsmasq's compressed dns-cache key, so `host_for_ip`'s lookup hits as-is.
- **No wire/API change**: the usage report already carries `ipv6` HostId values (the contract spec passes unchanged).

## Tests (TDD, red→green in history)

- `render_spec`: assert the v6 tracking sets + `ip6 daddr` update rules are emitted; tightened the existing `ip_pair_tracking_rx` assertion to scope to its own `{...}` block (adjacent v6 sets broke the old fixed-width window).
- `usage_spec`: pin the v6 attribution round-trip (compressed v6 dst-ip → FQDN via the dns-cache) and the v6 rx `lan_ip→mac` resolution.
- Full openwrt busted suite: **828 passing**. New nftables constructs syntax-checked against a real `nft -c` on the test router.

## Back-compat

Agent-side only, additive wire (v6 usage records the API already accepts). Wire contract unchanged.

Fixes #1796.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -807,15 +807,40 @@ function M.nft(snapshot, opts)
   ind("}")
   emit("")
 
+  -- #1796: IPv6 byte-accounting. The v4 sets above only match `ip daddr`, so
+  -- every IPv6 destination flow went uncounted and was invisible in
+  -- per-device traffic / recent-apexes (a v6-preferring client browsing a
+  -- dual-stack host showed nothing). These mirror the v4 sets with an
+  -- ipv6_addr shape; usage.lua reads both families through the same parsers
+  -- and `nft -j` emits compressed v6 that matches the dns-cache key, so
+  -- host attribution works without extra canonicalization.
+  ind("set mac_ip6_tracking {")
+  ind2("type ether_addr . ipv6_addr")
+  ind2("flags dynamic,timeout")
+  ind2("timeout 6h")
+  ind2("counter")
+  ind("}")
+  emit("")
+
+  ind("set ip_pair6_tracking_rx {")
+  ind2("type ipv6_addr . ipv6_addr")
+  ind2("flags dynamic,timeout")
+  ind2("timeout 6h")
+  ind2("counter")
+  ind("}")
+  emit("")
+
   ind("chain wifihaven_account_tx {")
   ind2("type filter hook forward priority 1; policy accept;")
   ind2("iifname \"br-lan\" update @mac_ip_tracking { ether saddr . ip daddr } counter")
+  ind2("iifname \"br-lan\" update @mac_ip6_tracking { ether saddr . ip6 daddr } counter")
   ind("}")
   emit("")
 
   ind("chain wifihaven_account_rx {")
   ind2("type filter hook forward priority 1; policy accept;")
   ind2("oifname \"br-lan\" update @ip_pair_tracking_rx { ip daddr . ip saddr } counter")
+  ind2("oifname \"br-lan\" update @ip_pair6_tracking_rx { ip6 daddr . ip6 saddr } counter")
   ind("}")
   emit("")
 
--- a/openwrt/files/usr/sbin/wifihaven-agent
+++ b/openwrt/files/usr/sbin/wifihaven-agent
@@ -751,7 +751,11 @@ end
 -- lease table; v6 (#1796) comes from the NDP neighbor cache via
 -- conntrack.parse_arp_table (which reads /proc/net/arp + `ip -6 neigh`), since
 -- DHCP leases don't cover SLAAC/v6 device addresses. Leases win for any v4 IP
--- already mapped; the neighbor table only fills gaps (notably all v6 keys).
+-- already mapped; the neighbor table only fills gaps — all v6 keys, and (by
+-- design) any static-IP v4 LAN device with no DHCP lease, which previously had
+-- its rx download bytes dropped (#879). The rx set's lan_ip is always a real
+-- LAN device address (`oifname br-lan` ⇒ ip[6] daddr), so WAN-side entries in
+-- the ARP/NDP dump can't mis-resolve a flow.
 -- Cheap (a handful of lines on tmpfs / one netlink dump) so it runs every scrape.
 local function load_ip_to_mac()
   local leases = conntrack.parse_dhcp_leases()
--- a/openwrt/files/usr/sbin/wifihaven-agent
+++ b/openwrt/files/usr/sbin/wifihaven-agent
@@ -713,30 +713,54 @@ local function read_set_json(name)
   return json_str
 end
 
+-- #1796: read v4 (mac_ip_tracking) AND v6 (mac_ip6_tracking) tracking sets and
+-- merge their parsed records. The parser is family-agnostic — concat[2] is the
+-- remote IP in whatever family the set holds — so the same parse path handles
+-- both. A missing set is treated as the empty side (transient render swap).
 local function read_tx_set()
-  local json_str = read_set_json("mac_ip_tracking")
-  if not json_str then return nil end
-  local ok, counters = pcall(usage.parse_nft_counters, json_str)
-  if not ok then return nil end
-  return counters
+  local out = {}
+  for _, name in ipairs({ "mac_ip_tracking", "mac_ip6_tracking" }) do
+    local json_str = read_set_json(name)
+    if json_str then
+      local ok, counters = pcall(usage.parse_nft_counters, json_str)
+      if ok and counters then
+        for _, c in ipairs(counters) do out[#out + 1] = c end
+      end
+    end
+  end
+  if #out == 0 then return nil end
+  return out
 end
 
 local function read_rx_set()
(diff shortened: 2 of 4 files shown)


sameerparekh:
## Independent PR review

**BLOCKERS**

(none)

**SHOULD-FIX**

(none)

**NITS**

1. `openwrt/files/usr/sbin/wifihaven-agent:758-764` (`load_ip_to_mac`) — the `parse_arp_table()` overlay is family-agnostic, so it also changes the **v4** path: a static-IP LAN device with no DHCP lease, previously dropped on the rx side (#879 behaviour), now resolves via `/proc/net/arp`. This is benign/positive (and only fills gaps — leases still win), but it's slightly beyond the #1796 v6 scope. Verified safe: the rx set key's `lan_ip` is always a real LAN device address (`oifname br-lan` ⇒ `ip[6] daddr` is the LAN device), so WAN entries in the ARP/NDP dump can't mis-resolve. Worth a one-line comment that the v4 static-IP pickup is intentional. Confirmed live on the prod router that device GUAs/ULAs are present in `ip -6 neigh` with `lladdr` (31 entries), so v6 rx resolution works in practice.

2. Observability — the fix's failure mode was *silent* (v6 invisible, undetected). The existing `usage_post_total` doesn't split by family, so there's no metric to confirm v6 accounting is now flowing. The recent-apexes/traffic UI is the direct verification path, so this is optional; a small per-family usage-record gauge (bounded `family` label) would be a reasonable follow-up rather than in-scope here (this PR widens an existing metered path, not a new route/job/poller).

3. `openwrt/test/usage_spec.lua` — the new v6 `build_report`/`merge_counters` tests pass on `main` too, since `usage.lua` (`host_for_ip`, parsers, merge) is already family-agnostic; they pin the no-canonicalization + v6-rx contract rather than failing-without-fix. The genuine red→green for this PR is in `render_spec` (the v4-only sets/rules) — verified red before the impl commit. Good to have both.

### Dimension notes
- **SSOT (1):** reuses existing parsers + `host_for_ip` + `conntrack.parse_arp_table`; no re-derivation, no keep-in-sync smell. ✓
- **Testing (2):** TDD red→green visible (test commit → fix commit); existing `ip_pair_tracking_rx` assertion was *tightened* (scoped to its `{...}` block), not weakened — justified because adjacent v6 sets break the old fixed-width window. No forbidden mocks. ✓
- **Arch invariants (3):** router stays a dumb applier (accounting only, no decision logic); no DNS-enforcement reasoning; no new snapshot field; no new `/tmp` writer (sets are in-kernel, `size 65535` + `timeout 6h` bounded like the v4 ones). ✓
- **Wire/back-compat (4):** agent-only; usage report already carries `ipv6` HostId (contract_spec passes unchanged); nft/UCI set names aren't part of the wire contract. Additive. ✓
- **Migrations (5):** none. **Metrics (6):** see NIT 2. **Scope (7):** focused; no AI mentions in commits/PR (Co-Authored-By trailer only). **Security (8):** no secrets/SQL/new routes; `ip -6 neigh` is a fixed command, no injection. ✓
- nftables syntax of the new sets/rules verified against a real `nft -c` (check-only) on the test router; full openwrt busted suite 828 passing.

VERDICT: APPROVE

IPv6 byte-accounting is added to the usage plane (which was IPv4-only) by mirroring the v4 tracking sets/rules with `ip6 daddr` + `ipv6_addr` and resolving v6 rx via the NDP neighbor cache, correctly fixing the root cause of #1796. No BLOCKERs or SHOULD-FIX; three minor NITs (an intentional v4 side-effect worth a comment, an optional follow-up metric, and a note on which tests pin the fix). The change is additive, well-tested, and architecturally clean.

sameerparekh:
## Independent PR review — re-run (delta since `9ac8f90`)

Delta: `openwrt/files/usr/sbin/wifihaven-agent` (+5/-1) — comment-only change in `load_ip_to_mac`.

### Prior findings status

| NIT 1 | document intentional v4 static-IP rx pickup in `load_ip_to_mac` | **ADDRESSED** — comment now states the static-IP v4 pickup is by design and why WAN-side ARP/NDP entries can't mis-resolve (`wifihaven-agent:758-765`). |
| NIT 2 | optional per-family usage metric (verifiability) | **NOT-ADDRESSED** — intentionally deferred; recent-apexes is the direct verification path and this PR widens an existing metered path rather than adding a new one. Reasonable as a follow-up. |
| NIT 3 | note on which tests pin the fix | **N/A** — informational only, no action required. |

### New findings (delta)

(none — the delta is a documentation comment with no behavioural change.)

VERDICT: APPROVE

Comment-only follow-up addressing review NIT 1; no behavioural change and no new findings. The IPv6 usage-accounting fix remains correct, additive, and well-tested (full openwrt busted suite green). No open BLOCKERs or SHOULD-FIX.

PR #1807, opened 2026-06-20
fix(#1796): record IPv6 connection events via NDP neighbor (not lan_prefix_v6)

sameerparekh:
## Problem

Outbound **IPv6 connection events** are silently dropped on every router that hasn't manually authored `lan_prefix_v6` — which ships **commented out by default** (`openwrt/files/etc/config/wifihaven`). `conntrack.lua is_wan_bound` decided LAN-source by matching `src_ip` against that prefix; empty ⇒ all v6 flows rejected.

Worse, the *only* stable prefix — the ULA — never matches real traffic. Home IPv6 isn't NATed, so devices source internet flows from their **GUA**, which is ISP-delegated and **dynamic**. Live prod proof (the mathplayground flow this all started from):

```
src=2601:280:4700:f32:cd10:109c:441c:49df   dst=2606:4700::6812:446   (Cloudflare / mathplayground)
```

So `lan_prefix_v6='fdcd:f224:23d6:'` (the ULA) would record **zero** internet v6. This is the **connection_events** half of #1796; the **usage/traffic** half is #1802 (and is prefix-independent, so it was a separate gap).

## Fix

`is_wan_bound` now takes an optional 4th arg — the **NDP/ARP neighbor map** `{ ip -> mac }` the agent *already builds* (`conntrack.parse_arp_table`, used for v6 MAC attribution). A v6 flow is LAN-bound iff **`src` is a known neighbor and `dst` is not** — robust across GUA, ULA, privacy/temporary addresses and dynamic prefix re-delegation, with no prefix bookkeeping.

- **v4 unchanged** — stable RFC1918 prefix match.
- **`lan_prefix_v6`** is kept as an optional *additional* accept path (union); no longer required. Config comment updated to say so.
- The watch loop fetches the neighbor table **only for v6 lines**, cached per wall-clock second to bound `ip -6 neigh` cost, and reuses it for MAC attribution (the src is guaranteed present — it just passed the membership check).

Confirmed live: 31 device GUAs/ULAs are present in `ip -6 neigh` on prod with `lladdr`, so the membership check resolves real flows.

## Tests (TDD, red→green)

`conntrack_spec.lua` → new `is_wan_bound v6 via NDP neighbor table (#1796)` block:
- GUA-sourced v6 to public dst accepted with **no** v6 prefix (the bug);
- v6 between two neighbors rejected (LAN-internal);
- unknown-src v6 rejected;
- authored prefix still works (union);
- v4 unaffected when a neighbor set is passed.

Full openwrt busted suite: **827 passing**. Existing 3-arg `is_wan_bound` callers/tests are unaffected (4th arg optional).

## Back-compat

Agent-only; `is_wan_bound` signature change is additive (optional 4th arg). No wire/API change. nft/UCI keys aren't the wire contract.

Part of #1796 (paired with #1802). Related: #1793 (v6 recorded as bare literal), #1688 (v6 attribution).

### DIFF
--- a/openwrt/files/etc/config/wifihaven
+++ b/openwrt/files/etc/config/wifihaven
@@ -21,11 +21,13 @@ config wifihaven 'wifihaven'
 	# LAN subnet prefix used to identify outbound (egress) flows
 	option lan_prefix '192.168.1.'
 
-	# Optional v6 LAN ULA prefix used to identify outbound v6 flows (#1688).
-	# Leave unset to filter all v6 flows out of connection_events (today's
-	# prod behaviour). Set to the router's LAN ULA prefix (e.g.
-	# 'fdaa:bbbb:cccc:') to let v6 traffic through the same is_wan_bound
-	# filter the v4 path uses.
+	# Optional v6 LAN prefix override (#1688/#1796). NOT required: as of #1796
+	# outbound v6 flows are identified by NDP-neighbor membership (the agent's
+	# existing ip->mac table), which covers GUA / ULA / privacy addresses and
+	# the ISP-dynamic GUA prefix automatically. Leave unset. Authoring a value
+	# here only ADDS an extra prefix-match accept path (union); it cannot fix v6
+	# on its own because devices source internet traffic from their GUA, not the
+	# ULA, and the GUA prefix is delegated/dynamic.
 	# option lan_prefix_v6 'fdaa:bbbb:cccc:'
 
 	# How many connection_attempt events to accumulate before flushing
--- a/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/conntrack.lua
@@ -966,19 +966,45 @@ end
 -- (#575)
 --
 -- Family is detected from src_ip — a colon means IPv6, otherwise v4 — so a
--- single helper handles both stacks (#1688). lan_prefix_v6 is the LAN v6 ULA
--- prefix (e.g. "fdaa:bbbb:cccc:"); when empty/nil v6 flows are rejected so a
--- prod router without an authored v6 LAN keeps today's behavior.
+-- single helper handles both stacks (#1688).
+--
+-- v4 LAN membership is a stable RFC1918 prefix match (`lan_prefix`).
+--
+-- v6 (#1796): a static `lan_prefix_v6` cannot identify LAN-sourced traffic in
+-- practice. Home IPv6 is not NATed, so devices source internet flows from their
+-- GUA — which is ISP-delegated and dynamic — not the ULA. Matching on the ULA
+-- (the only stable prefix) therefore records ZERO internet v6, which is why the
+-- option shipped commented-out and every unconfigured router silently dropped
+-- all v6 connection events. Instead, when the caller passes `lan_ip_set` (the
+-- NDP/ARP neighbor map { ip -> mac } the agent already builds for v6 MAC
+-- attribution), a v6 flow is LAN-bound iff `src` is a known LAN neighbor and
+-- `dst` is not — covering GUA, ULA, and privacy/temporary addresses with no
+-- prefix bookkeeping. `lan_prefix_v6`, if authored, is honored as an additional
+-- accept path (union). With neither a neighbor set nor a prefix, v6 is rejected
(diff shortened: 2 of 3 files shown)


sameerparekh:
## Independent PR review

**BLOCKERS**

(none)

**SHOULD-FIX**

(none)

**NITS**

1. `conntrack.lua` `neighbor_table()` — the per-`os.time()`-second cache means a brand-new SLAAC privacy/temporary address's *first* v6 flow can miss the LAN-membership check for up to ~1s until the NDP entry is cached. Bounded and acceptable: it's the same precondition the v6 **MAC-attribution** path already requires (src must be in `parse_arp_table` or the event gets a nil mac, #1691), so this introduces no new failure mode — just noting the window.
2. `parse_arp_table()` enumerates neighbors on **all** interfaces (`ip -6 neigh show`, no `dev`), so `lan_ip_set` can contain WAN-side on-link neighbors. Safe for this use: an internet `dst` reached via the default route is never on-link (so `dst_lan` stays false), and a WAN host's `src` is never a forwarded LAN flow's src. Worth a one-liner if a future reader wonders why it isn't scoped to `br-lan`.
3. Observability — like #1802, no metric distinguishes "v6 connection events now flowing"; the existing event-post counters cover the POST. Optional follow-up, not in scope (this widens an existing path).

### Dimension notes
- **SSOT (1):** reuses `parse_arp_table` (one neighbor-table source, now shared by both `is_wan_bound` and MAC attribution); no decision/quantity re-derived. The v6 prefix-union repeats the `sub(1,#p)==p` shape but for a distinct family/path — not duplication of a decision. ✓
- **Testing (2):** TDD red→green visible (test commit `d174abc2` before fix `94cd2860`; the GUA-accept case was confirmed red). 5 new cases incl. LAN-internal reject, unknown-src reject, prefix-union back-compat, and v4-unaffected. Existing 3-arg `is_wan_bound` tests untouched (4th arg optional) — full suite 827 green. ✓
- **Arch (3):** attribution only, no policy logic added to the agent; no new `/tmp` writer (cache is in-RAM); `os.time()`/`ip -6 neigh` are agent-appropriate. ✓
- **Wire/back-compat (4):** `is_wan_bound` 4th arg is additive; no wire/API change; UCI/nft keys aren't the contract. The default config now needs **no** manual `lan_prefix_v6` (comment updated). ✓
- **Migrations (5):** none. **Metrics (6):** NIT 3. **Scope (7):** focused (conntrack.lua + config comment + test); no dead code. **Security (8):** no secrets/SQL/routes; fixed shell commands. ✓
- Correctness spot-checks pass: v4 path byte-identical behavior; v6 `src_lan ∧ ¬dst_lan` correctly rejects LAN-internal and unknown-src; the `lan_ip_set` guard idiom yields nil for v4/nil-flow as intended.

VERDICT: APPROVE

Replaces the broken `lan_prefix_v6` v6 LAN-source check (which records zero internet v6, since devices use their dynamic GUA not the ULA) with NDP-neighbor membership — the robust, config-free fix the operator chose. TDD'd, back-compatible, full suite green. No BLOCKERs/SHOULD-FIX; three minor NITs. This is the connection_events half of #1796, paired with #1802.

sameerparekh:
## Independent PR review — re-run (delta since `39aafbf`)

Delta: `conntrack.lua` (+6 comment lines) — `is_wan_bound` v6 block only.

### Prior findings status

| NIT 1 | per-second neighbor cache can miss a brand-new privacy address's first v6 flow (<1s) | **ACKNOWLEDGED** — bounded, no new failure mode (same precondition as v6 MAC attribution); left as-is by design. |
| NIT 2 | document that the all-interface neighbor scope is safe | **ADDRESSED** — inline comment now explains internet dsts aren't on-link and WAN srcs aren't forwarded LAN srcs (`conntrack.lua` is_wan_bound v6 block). |
| NIT 3 | optional per-family "v6 events flowing" metric | **NOT-ADDRESSED** — intentional follow-up; existing event-post counters cover the POST. |

### New findings (delta)

(none — documentation comment only, no behavioural change; conntrack_spec still 135/135 green.)

VERDICT: APPROVE

Comment-only follow-up addressing review NIT 2; no behavioural change, no new findings. The NDP-neighbor v6 LAN-source fix remains correct, TDD'd, and back-compatible. No open BLOCKERs or SHOULD-FIX.

PR #1808, opened 2026-06-20
test(e2e): IPv6 per-device usage attribution gate (#1804)

### LINKED ISSUE #1804: test(e2e): IPv6 per-device usage attribution — gate the #1796 silent-drop class
## Why

#1796 shipped a **silent** prod gap: the usage byte-accounting nftables plane was IPv4-only (`mac_ip_tracking` / `ip_pair_tracking_rx` typed `ipv4_addr`, rules matching `ip daddr`), so **all IPv6 destination traffic was never counted and never appeared in per-device traffic / recent-apexes / `/api/dashboard/now`**. A device on a dual-stack host that Happy-Eyeballs'd onto IPv6 showed *nothing*, while a device that used IPv4 for the same host showed up. This went undetected for months because there is **no end-to-end test that drives IPv6 traffic and asserts it surfaces as per-device usage**.

The enforcement plane is dual-stack and *is* covered by the Gate-2/Gate-3 fake-API attribution suite (#1677), and the connection_events v6 path has its own coverage. The **usage/traffic accounting path over IPv6 has no e2e gate** — that's the hole #1802 fixes blindly.

## What to add

End-to-end coverage (router VM → real/fake API → query surface) that would have failed before #1802 and passes after:

1. **IPv6 usage attribution round-trip.** From a client behind the router VM, resolve a dual-stack host through the router's dnsmasq, generate **IPv6** download+upload traffic to it, flush a usage bucket, and assert the host appears (by **FQDN**, not bare v6 literal) in the per-device usage surface (`/api/devices/<mac>/recent-apexes` and/or `/api/usage/traffic`) with non-zero `bytesIn`/`bytesOut`.
2. **v4 and v6 to the same host both attribute and aggregate** to one host row (no family-split duplication, no v6 silently dropped).
3. **v6 rx (download) resolves to the right MAC** via the NDP neighbor cache (the `ip_pair6_tracking_rx` → `load_ip_to_mac` path), i.e. download bytes land on the device, not dropped/unattributed.
4. **Regression guard:** a host reached *only* over IPv6 must NOT be invisible (the exact #1796 symptom).

## Where this fits

- Prefer extending the existing fake-API attribution harness (Gate 2, #1677) rather than a new bespoke rig, so it runs in the same CD gate.
- Pair with the busted unit coverage already added in #1802 (`render_spec` v6 sets/rules, `usage_spec` v6 round-trip) — this issue is the **e2e layer** those unit tests can't provide (real `nft` counters, real `nft -j` v6 formatting, real `ip -6 neigh` resolution).
- Consider a lightweight assertion that the rendered ruleset actually contains the v6 accounting sets when loaded by real `nft` (the `nft -c` pending test in `render_spec` is skipped on hosts without `nft`).

## Acceptance

- An e2e test exists that drives IPv6 traffic through the router and asserts FQDN-attributed per-device usage; it fails against pre-#1802 agent code and passes after.
- v4+v6-to-same-host aggregation and v6-only-host visibility are both asserted.
- Runs in CI/CD (not local-only).

Relates to #1796 (root cause + fix #1802), #1793 (v6 connection_events attribution), #1677 (Gate-2 fake-API attribution harness), #1792 (router CD must be green to actually publish/validate).

sameerparekh:
## What

Adds a Gate-2 fake-API e2e scenario (`scripts/e2e/scenarios_fake/test_v6_usage_attribution.py`) that drives a **real bidirectional IPv6 flow through the router's FORWARD chain** and asserts FQDN-attributed per-device usage with non-zero `bytesIn`/`bytesOut` — the end-to-end gate the [#1796](https://github.com/wifihaven/wifihaven/issues/1796) silent IPv4-only byte-accounting regression slipped through. It validates the fixes already merged on main: [#1802](https://github.com/wifihaven/wifihaven/issues/1802) (`mac_ip6_tracking` / `ip_pair6_tracking_rx` sets + `ip6 daddr` rules + v6 rx via the NDP neighbor cache) and [#1807](https://github.com/wifihaven/wifihaven/issues/1807) (v6 LAN-source classification via the NDP table).

## Coverage (all four acceptance points, one VM restore)

1. **v6 usage round-trip** — v6 download+upload to a dual-stack host surfaces by **FQDN** (not bare v6 literal) with `bytesIn>0` AND `bytesOut>0`.
3. **v6 rx resolves to the right MAC** via the NDP neighbor cache (`ip_pair6_tracking_rx` → `load_ip_to_mac` → `ip -6 neigh`).
4. **Regression guard** — the v6-only assertions run **before** any v4 traffic, proving a v6-only host is not invisible (the exact #1796 symptom).
2. **v4+v6 aggregation** — after adding v4 to the SAME host, there is still exactly **one** host row (no family-split duplication).

Plus a cheap real-`nft` sanity check that the v6 accounting sets exist in the loaded ruleset (absent pre-#1802) — a deterministic #1796 discriminator independent of the round-trip.

## How (and why a LAN origin, not the real Internet)

The qemu SLIRP WAN is **v4-only** (`scripts/vm/uci-defaults/99-wifihaven`), so a v6 SYN crosses FORWARD but never gets a response — enough for the connection_events attribution gate (`test_v6_attribution.py`) but it can never produce **download** bytes. To get a real bidirectional v6 flow the router *forwards* (and therefore counts), the test stands up a throwaway dual-stack HTTP origin as a **second L2 host on the LAN** — a `macvlan`-in-netns on the client VM, `socat EXEC` responder — owning the `e2e-edge` doc-space addresses (`2001:db8::10` / `192.0.2.10`), and gives the router an on-link route to it. Both directions then traverse `br-lan`↔`br-lan` (FORWARD) and are byte-accounted.

Attribution stays **production-shape**: `e2e-edge.wifihaven.net` resolves through the router's dnsmasq against the **real Cloudflare A/AAAA** records (`infra/cloudflare/main.tf`), so the agent learns the `(ip → FQDN)` mapping from a genuine forwarded `reply` log line — only the packets' final hop is a local origin instead of a dead SLIRP route.

Runs automatically in the existing **Gate 2 (`VM e2e — fake API`)** CD gate (whole `scenarios_fake` pack); `--only v6-usage-attribution` for manual runs.

## Image change

`scripts/vm/build-client-base.sh` adds `iproute2` (netns + `ip link add type macvlan`) and `socat` to the **client** test image — BusyBox `ip`/`nc` on Alpine can do neither. The **router** image is untouched, so `ip -6 neigh` stays BusyBox (preserving fidelity for the rx-resolution path the test exercises).

## Test

- `openwrt` busted suite: **837 passed / 0 failed** (1 pending — `nft` not on host).
- New scenario collects cleanly under pytest with the registered marker.
- Behavioral validation runs in CD (this is a VM e2e gate; no local KVM).

Fixes #1804

### DIFF
--- a/scripts/e2e-vm.sh
+++ b/scripts/e2e-vm.sh
@@ -113,10 +113,12 @@ case "${ONLY}" in
                 MARK="v6_link_local" ;;
   v6-attribution|v6_attribution)
                 MARK="v6_attribution" ;;
+  v6-usage-attribution|v6_usage_attribution)
+                MARK="v6_usage_attribution" ;;
   *) echo "unknown --only: ${ONLY}" >&2
      echo "valid: enrollment, allowed-browsing, blocked-domain, daily-limit, usage, blocked-page," >&2
      echo "       pause, extra-blocked, schedule, time-limit, reassignment, unknown-device," >&2
-     echo "       install-health, sni-attribution, v6-link-local, v6-attribution" >&2
+     echo "       install-health, sni-attribution, v6-link-local, v6-attribution, v6-usage-attribution" >&2
      exit 2 ;;
 esac
 
--- a/scripts/e2e/pytest.ini
+++ b/scripts/e2e/pytest.ini
@@ -21,6 +21,7 @@ markers =
     sni_attribution: suite I (#573) — SNI-only hostname attribution via the wifihaven-sni-tail sidecar
     v6_link_local: #1680 — IPv6 LAN bring-up (ULA SLAAC + v6 FORWARD path), prep for #1677
     v6_attribution: #1677 — v6 destinations attribute to FQDN (CD-tier follow-up to #1673 / #1668)
+    v6_usage_attribution: #1804 — v6 per-device usage round-trip surfaces by FQDN with non-zero bytes (e2e gate for #1796 / #1802)
     global_policy: suite H (#1460) — @global_allow / @global_block fleet-wide enforcement (gate for #1321)
     smoke: fast subset across suites (one short case each) for PR gating
 log_cli = true
--- a/scripts/e2e/scenarios_fake/test_v6_usage_attribution.py
+++ b/scripts/e2e/scenarios_fake/test_v6_usage_attribution.py
@@ -251,51 +251,40 @@ def _set_fast_usage_cadence() -> None:
     )
 
 
-def _usage_row(fake_api, mac: str, host: str) -> dict | None:
-    """Return the latest captured usage record for (mac, host), or None.
+def _records_for_mac(fake_api, mac: str) -> list[dict]:
+    """Every usage record the fake has captured for `mac`, across all reports.
 
-    /test/usage returns a row per (mac, host) the agent reported; filter to the
-    workload host so incidental client traffic (NTP, DNS) can't shadow it
-    (#1049)."""
-    latest = None
+    The fake captures the RAW agent reports — there is no API-side projection
+    in fake mode. The agent resets the nft accounting sets after each POST
+    (wifihaven-agent), so reports are per-period DELTAS, and `build_report`
+    emits one record per `(mac, dst_ip)` (usage.lua): v4 and v6 traffic to the
+    same FQDN therefore land in SEPARATE records keyed by `destIp`
+    (192.0.2.10 vs 2001:db8::10), each carrying the resolved `host.value`. So
+    we key assertions on `destIp` (the family) and read `host.value` for the
+    attribution, never trying to compare bytes across periods or to observe an
+    API-side per-apex aggregation that does not exist here."""
+    out: list[dict] = []
     for r in (fake_api.usage().get("reports") or []):
         for rec in (r.get("body", {}).get("records") or []):
-            if (rec.get("mac") or "").lower() != mac.lower():
-                continue
-            if (rec.get("host") or {}).get("value") == host:
-                latest = rec
-    return latest
+            if (rec.get("mac") or "").lower() == mac.lower():
+                out.append(rec)
+    return out
 
 
-def _wait_usage_row(fake_api, mac: str, host: str, *,
-                    predicate, description: str, timeout_s: float = 180) -> dict:
+def _wait_record(fake_api, mac: str, *, dest_ip: str, predicate,
+                 description: str, timeout_s: float = 180) -> dict:
+    """Wait until some captured record for (mac, dest_ip) satisfies `predicate`,
+    returning the latest such record. Keying on `destIp` pins the family
+    (the workload host) and ignores incidental client traffic (NTP/DNS, #1049)."""
     def probe():
-        rec = _usage_row(fake_api, mac, host)
-        return rec if (rec is not None and predicate(rec)) else None
+        match = None
+        for rec in _records_for_mac(fake_api, mac):
+            if rec.get("destIp") == dest_ip and predicate(rec):
+                match = rec
+        return match
     return wait_until(
         probe, timeout_s=timeout_s, interval_s=5, description=description,
     )
-
-
-def _host_row_count(fake_api, mac: str, host: str) -> int:
-    """How many DISTINCT host rows the agent has reported whose value resolves
-    to `host` — across every captured report. >1 means a family-split
-    duplication (e.g. one row keyed to the bare v6 literal, one to the FQDN),
-    which is exactly the #1796-class bug the aggregation check guards against.
-
-    A bare v4/v6 literal for LEAF_IP4/LEAF_IP6 counts as a duplicate of the
-    FQDN row, since both denote the same host reached over different families.
-    """
-    seen: set[str] = set()
-    aliases = {host.lower(), LEAF_IP4, LEAF_IP6}
-    for r in (fake_api.usage().get("reports") or []):
-        for rec in (r.get("body", {}).get("records") or []):
-            if (rec.get("mac") or "").lower() != mac.lower():
-                continue
-            val = (rec.get("host") or {}).get("value") or ""
-            if val.lower() in aliases:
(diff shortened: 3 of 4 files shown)


sameerparekh:
## Independent PR review — #1808 (e2e: IPv6 per-device usage attribution)

Read-only adversarial pass over `git diff origin/main...HEAD` against the 8-dimension checklist. Test-only + test-image change; no API/agent/wire/migration surface.

### BLOCKERS

1. **`scripts/e2e/scenarios_fake/test_v6_usage_attribution.py:400-413` — the v4+v6 aggregation assertion is unsound and will hang/flake.** The agent **resets** the nft accounting sets after each successful POST (`openwrt/.../wifihaven-agent:1139-1144`, incl. the v6 sets), so usage reports are **per-period deltas**, and v4 vs v6 traffic land in **separate records keyed by `destIp`** (`2001:db8::10` vs `192.0.2.10`) — `build_report` emits one record per `(mac, dst_ip)` (`usage.lua:330,362`), both carrying `host.value = e2e-edge`. Consequences:
   - The growth predicate `bytesIn > v6_in and bytesOut > v6_out` compares a **v4-only delta period** against the **v6 delta** captured earlier — there is no monotonic relationship, so `_wait_usage_row` will usually time out (test failure) even when behaviour is correct, or pass by coincidence. Flaky gate.
   - `_host_row_count(...) == 1` (lines 386, 410) cannot detect family-split duplication: both per-family records share `host.value`, and the API-side apex aggregation that would collapse `destIp`s into one host row **does not exist in fake mode** (the fake captures raw agent reports). So the check is vacuously 1 regardless.
   - `_usage_row` returns "the latest record matching by host" across two different-`destIp` records — nondeterministic which family it returns post-v4.

   Fix: assert on the **per-family records by `destIp`** — a `destIp == LEAF_IP6` record with `host.value == FQDN`, `bytesIn>0`, `bytesOut>0` (criteria 1/3/4, v6-only phase), and after v4 a `destIp == LEAF_IP4` record also attributed to the **same FQDN** with bytes>0, plus that neither family record is attributed to a bare literal (criterion 2 — both apexes collapse API-side because they share the FQDN). Drop the cross-period byte-growth comparison.

### SHOULD-FIX

(none)

### NITS

1. `test_v6_usage_attribution.py:179-182` — background `socat … &` over SSH relies on the same detach pattern as `test_sni_attribution.py`'s uhttpd; fine, and `_wait_origin_reachable` catches a dead listener. No change needed, just noting the dependency.
2. `_wait_origin_reachable` default `timeout_s=60` — macvlan/ND warmup on a freshly-restored VM is usually quick, but if Gate-2 shows warmup flake, bump it. Non-blocking.

### Dimensions with nothing to flag
- **1 (SSOT):** test code, no duplicated decision logic.
- **3 (invariants):** `/tmp` writes (`whsrv-reply.sh`, `whsrv{4,6}.log`) are on the **ephemeral, function-scoped client VM** with bounded (8-curl) traffic — not router-agent grow-writers; exempt. Router-side change is `ip route replace` only.
- **4 (wire):** no wire/API change.
- **5 (migrations):** none.
- **6 (metrics):** test-only.
- **7 (scope):** focused 4-file diff; no AI mentions in title/body/commit (Co-Authored-By trailer only).
- **8 (security):** no secrets; no SQL.

VERDICT: REQUEST-CHANGES

A well-targeted e2e gate for the #1796 v6-usage silent-drop class, with a sound hermetic-origin design (macvlan-in-netns + real-DNS attribution). One BLOCKER: the v4+v6 aggregation assertion mismodels the agent's per-period, per-`destIp` reset semantics and will flake/hang. Re-key the aggregation/round-trip assertions on `destIp` and the verdict clears.

sameerparekh:
## Independent PR review — #1808 (re-run)

Delta reviewed: `git diff 6484cf1b...97a5b08c` — `test_v6_usage_attribution.py` only (helpers + assertion body re-keyed on `destIp`).

### Prior findings

| # | Finding | Status |
|---|---------|--------|
| BLOCKER 1 | Aggregation assertion mismodels per-period / per-`destIp` reset semantics (cross-period byte-growth predicate + vacuous `_host_row_count`) | **ADDRESSED** — `_usage_row`/`_host_row_count`/`v6_in`/`v6_out` removed; assertions now key on `destIp` via `_records_for_mac` + `_wait_record` (`:254-285`). v6 leg: `destIp==LEAF_IP6` record with `host==FQDN`, `bytesIn>0` AND `bytesOut>0` (`:367-382`). v4 leg: `destIp==LEAF_IP4` record also attributed to the same FQDN with `bytesOut>0` (`:391-410`), plus a guard that no `LEAF_IP4`/`LEAF_IP6` flow is attributed to a bare literal (`:412-419`). This is the genuine family-split / #1796 guard and no longer depends on a non-existent fake-mode apex aggregation or a cross-period comparison. |
| NIT 1 | background `socat &` over SSH | unchanged; acceptable (matches sni pattern, reachability-gated) |
| NIT 2 | `_wait_origin_reachable` 60s timeout | unchanged; acceptable |

### New findings (delta)

BLOCKERS: none.

SHOULD-FIX: none.

NITS:
1. `:367-374` — the v6 round-trip requires `bytesIn>0` AND `bytesOut>0` in a **single** period record; the four back-to-back `curl -6` calls (no sleep) land within one 10 s window so tx+rx merge, but if Gate-2 ever shows a reset-boundary split, generating sustained traffic or polling for tx and rx separately would harden it. Non-blocking — inherent to per-period accounting and the tight-loop ordering mitigates it.

VERDICT: APPROVE

The BLOCKER is fully addressed: the round-trip and aggregation assertions now match the agent's per-period, per-`destIp` reset semantics and assert the real behaviour (v6 attributed to FQDN with non-zero rx/tx; v4+v6 both roll up to the one FQDN apex with no literal split). Delta introduces no new issues. Test-only change; no API/agent/wire/migration surface.

Issue #1812, opened 2026-06-21
ci(cd): Master Router CD still RED — Gate 3a v6 DNS refused (fdaa:bbbb:cccc::1#53) persists past #1795; Gate 2 e2e-brand needs cloudflare apply


sameerparekh:
**Filed by the CD health watchdog (autonomous daily run, 2026-06-21).** Documenting state only — no fix PR opened (see "Why no fix PR" below). Not reopening #1792 per the no-reopen-closed-by-merge rule.

## Current RED
Master Router CD latest completed run on `main` is **failure**: [run 27885153291](https://github.com/wifihaven/wifihaven/actions/runs/27885153291) at `7f799af4` ("fix(#1796): record IPv6 connection events via NDP neighbor"). Router CD has been failure/cancelled every run since the last green (`#1734`, 2026-06-15). Two distinct failures:

### 1. Gate 3a (router staging smoke) — `test_gate3_smoke` ERROR at setup
```
E   TimeoutError: client DNS for example.com never resolved after 60s
    (52 attempts, last=';; communications error to fdaa:bbbb:cccc::1#53: connection refused')
scripts/e2e/gate3/conftest.py:127  wait_for_client_dns(c, host="example.com", timeout_s=60)
```
This is the **identical** symptom as **#1792** ("Gate 3a fails — router DNS refused on fdaa:bbbb:cccc::1#53, regression from #1788 dnsmasq conf-file streaming"). #1792 was closed by **#1795** (`7c1236b6` "fix(#1792): skip conf-file= for missing blocklist shards (Gate 3a)").

**#1795's fix was never validated by CI.** On the #1795 run ([27883294962](https://github.com/wifihaven/wifihaven/actions/runs/27883294962)) **Gate 2 and Gate 3a were SKIPPED** — the shared CI gate (Frontend Build & Lint) failed earlier, so the router gates never executed. The first run since where the gates actually ran (`7f799af4`) reproduces the exact same `fdaa:bbbb:cccc::1#53: connection refused`. The same error is also present on the 2026-06-19 run (#1790, `297cf084`), confirming it is stable, not flaky. Net: #1795 did not resolve the Gate 3a DNS-refused failure.

[...]

PR #1811, opened 2026-06-20
fix(cd): live-staging e2e §5b — seed-from-templates instead of retired POST /api/apps

sameerparekh:
## What was red
**Master API/UI CD** on main. The "Live e2e against staging stack" job — which builds a fresh local compose stack from current main (`docker compose ... up --build`) and runs `scripts/e2e-router.sh` against `127.0.0.1:8080` — fails:

```
▶ #1105: exempt time_limited app folds host into extraAllowed under MAC block
curl: (22) The requested URL returned error: 404
```

Runs [#1798](https://github.com/wifihaven/wifihaven/actions/runs/27884251144) and [#1793](https://github.com/wifihaven/wifihaven/actions/runs/27884399894) fail identically here.

## Root cause
[#1798](https://github.com/wifihaven/wifihaven/issues/1798) (b99ed2b8) retired the operator-facing app-definition mutators including `POST /api/apps` (apps are now template-authored only). The SPA + Scala unit tests were updated, but `e2e-router.sh` §5b still minted a Khan-shaped app via `POST /api/apps`, which now 404s and aborts `curl -f`.

## Fix
Rewrite §5b to seed the built-in templates and drive the #1105 assertion off a seeded app's real host:
- `POST /api/apps/seed-from-templates` (idempotent; admin-gated — `AUTH` is admin)
- `GET /api/apps` → pick first seeded app with ≥1 host
- assert that host folds into `extraAllowed` under `blocked=True`/`TimeLimit`

The carve-out contract is host-agnostic, so this preserves the exact #1105 wire assertion. `PUT /api/apps/:id/policy/:profileId` (kept by #1798) is unchanged.

## Validation
- `bash -n` + `shellcheck -S warning` clean.
- `GET /api/apps` response shape (`app.id` + `hosts[]`) confirmed against live staging (32 seeded apps, all with hosts).
- Selection parse (skip no-host apps, pick first with host) simulated offline.
- Full local compose e2e **not** runnable on this host (Docker down) — relies on this PR's CI "Live e2e against staging stack" gate, which exercises the exact path.

## Out of scope (tracked)
Gate 3a router e2e (`scripts/e2e/lib/api_admin.py` `create_app` + `scripts/e2e/gate3/conftest.py`) hits the same retired route but needs specific VM-reachable block/allow hosts — a redesign tracked in #1810. Master Router CD is not currently red (latest completed run was `cancelled`, superseded), so it is deliberately left out of this conservative CD-restore fix.

Fixes #1809

[...]

Issue #1809, opened 2026-06-20
fix(cd): live-staging e2e §5b still POSTs /api/apps after #1798 retired it


sameerparekh:
## Symptom
Master API/UI CD is RED on main. The "Live e2e against staging stack" job (which builds a fresh local compose stack from current main and runs `scripts/e2e-router.sh` against it) fails:

```
▶ #1105: exempt time_limited app folds host into extraAllowed under MAC block
curl: (22) The requested URL returned error: 404
##[error]Process completed with exit code 22.
```

Both run #1798 (27884251144) and run #1793 (27884399894) fail identically at this step.

## Root cause
[#1798](https://github.com/wifihaven/wifihaven/issues/1798) (commit b99ed2b8) retired the operator-facing app-definition mutators, including `POST /api/apps`. The SPA and Scala unit tests were updated, but `scripts/e2e-router.sh` §5b still mints a Khan-shaped app via `POST /api/apps` to exercise the "#1105: exempt time_limited app folds host into extraAllowed under MAC block" wire contract. That POST now 404s, and `curl -f` aborts the script.

## Fix
Apps are template-authored only post-#1798. Rewrite §5b to `POST /api/apps/seed-from-templates`, then drive the assertion off a seeded app's real host (selected from `GET /api/apps`) — the carve-out contract is host-agnostic. `PUT /api/apps/:id/policy/:profileId` (kept by #1798) is unchanged.

Validated: `bash -n` + `shellcheck` clean; `GET /api/apps` response shape confirmed against live staging; selection parse simulated offline. Full local compose e2e not runnable on this host (Docker down) — relies on the PR's CI "Live e2e" gate.

Found by the CD health watchdog.

PR #1813, opened 2026-06-21
fix(cd): gate3 e2e seeds apps from templates instead of retired POST /api/apps — restore Master API/UI CD (Gate 3b)

### LINKED ISSUE #1810: Gate 3a router e2e still uses retired POST /api/apps (#1798) — Router CD latent break
## Latent break (not yet RED, but will be)
Same root cause as #1809. [#1798](https://github.com/wifihaven/wifihaven/issues/1798) retired `POST /api/apps`, but the Gate 3a KVM router e2e still creates ad-hoc apps via that route:

- `scripts/e2e/lib/api_admin.py:266` — `AdminAPI.create_app()` POSTs `/api/apps`.
- `scripts/e2e/gate3/conftest.py:157,161` — `scratch_profile_and_device` fixture creates a *blocked* app (`example.com`) and an *allowed* app (`example.org`) with **specific, VM-reachable hosts**, then asserts nftables drop vs. carve-out at the connection layer.

When Master Router CD next deploys current main to the Gate 3a stack, these POSTs will 404 and the fixture will fail.

## Why this needs a redesign (not the #1809 one-liner)
The shell e2e (#1809) only needs *any* app+host, so it can use a seeded template app. Gate 3a needs **two specific, controllable hosts** (one blocked, one allowed) that the client VM can actually connect to, to prove the nftables forward-drop and the `extraAllowed` carve-out end-to-end. Post-#1798 there is no API path to mint an app with arbitrary hosts (apps are template-only), and `extraBlocked`/`extraAllowed` are derived solely from `app_policy_assignments` (no profile-level authoring path). Options to weigh:
1. Add gate3-only test `AppTemplates` defining `example.com`/`example.org` host-sets, seed them, and assign block/allow.
2. Reach an existing template whose hosts the VM can reach for both the block and allow legs.

Either needs design judgment + Gate 3a (self-hosted KVM) validation, which can't run on macOS. Tracking separately so #1809 can unblock Master API/UI CD now.

Found by the CD health watchdog.

sameerparekh:
## What was red
**Master API/UI CD** at **Gate 3b — last-published router + staging API** (run [27912992678](https://github.com/wifihaven/wifihaven/actions/runs/27912992678)). The job gets *past* the live-staging e2e (fixed by [#1811](https://github.com/wifihaven/wifihaven/pull/1811)) and then fails in the gate3 fixture:

```
gate3/conftest.py  scratch_profile_and_device → admin.create_app(...)
lib/api_admin.py   create_app → _request("POST", "/api/apps", ...)
RuntimeError: POST https://api-staging.wifihaven.net/api/apps → HTTP 404
```

## Root cause
[#1798](https://github.com/wifihaven/wifihaven/issues/1798) (b99ed2b8) retired the operator-facing app-definition mutators — including the arbitrary-host `POST /api/apps` create — because apps are **template-authored only** now. The gate3 fixture still minted a *blocked* app (`example.com`) and an *allowed* app (`example.org`) via `AdminAPI.create_app`, which now 404s. Gate 3b reaches this because it runs the **last-published router** (working DNS) against the **freshly-deployed staging API**. Same root cause as #1809/#1811; this is the analogous fix for the Python gate3 suite.

## Why this needed a redesign (and not the #1811 one-liner)
The shell e2e only needed *any* app+host. Gate 3 needs **two specific, VM-reachable hosts** — one to assert the nftables forward-DROP (block leg) and one to assert the `extraAllowed` carve-out / no-over-block (allow leg) at the connection layer.

I weighed the two options from #1810 and chose **Option 2 (drive off existing seeded templates)**, rejecting **Option 1 (add gate3-only `example.com`/`example.org` templates)**:

- **Gate 3b runs the freshly-deployed staging *binary*, which only knows the default template manifest.** To make `seed-from-templates` mint `example.com`/`example.org` apps, those test templates would have to ship in the production manifest — polluting **every household's** app catalog with test apps. That directly contradicts #1798, which *removed* app-editing precisely to clean up the catalog (#1777/#1794 churn). There is no per-environment manifest switch and no arbitrary-host create path, so Option 1 cannot be made prod-safe for gate3b.
- **Option 2 adds zero product surface** and uses only routes **kept** by #1798.

## The fix
`scratch_profile_and_device` now:
1. seeds the shipped templates — `POST /api/apps/seed-from-templates` (idempotent, admin-gated — the same route #1811 used),
2. reads them back — `GET /api/apps`, and
3. picks **two distinct seeded apps** via the new pure helper `lib/app_seed.pick_block_allow_apps`, assigning one `blocked` and one `allowed` to the namespaced profile (`PUT /api/apps/:id/policy/:profileId`, kept by #1798).

The block/allow contract is **host-agnostic** (block = HTTP/80 DNAT to the block page on the resolved IPs; allow = carve-out / no over-block), so the **two-leg connection-layer assertion in `test_smoke.py` is preserved exactly** — only the *source* of the hosts moves from a hand-minted app to the template catalog. The picker prefers stable apexes (`BLOCK_PREF`/`ALLOW_PREF`) that resolve and are reachable through qemu SLIRP v4 NAT — the property `example.com`/`example.org` were chosen for — and falls back to any hosted app so a catalog change never breaks the gate.

`lib/api_admin`: drop the dead `create_app` (route gone); add `seed_apps_from_templates`, `list_apps`, and `delete_app_policy`. **Teardown now detaches this profile's *assignment* — the shared seeded apps are never deleted** (other gates and the live catalog depend on them); only the namespaced profile/device + its assignments are cleaned up.

## TDD
- `test(#1810)` commit (red): `lib/test_app_seed.py` pins the selection contract against a stubbed `NotImplementedError` picker.
- `fix(cd)` commit (green): implements the picker + rewires the fixture/client. `python3 -m pytest scripts/e2e/lib/test_app_seed.py` → **7 passed**.

## Validation — real Gate 3b run on the KVM host (plex.lan / 192.168.10.43)
Ran the **exact Gate 3b invocation** against **live staging** with the **last-published** ipk router image (downloaded + sha256-verified from the `openwrt-latest` release, faithful to `e2e-vm-gate3b.yml`):

```
WH_API_URL=https://api-staging.wifihaven.net WH_GATE3_SIDE=3b PKG_FORMAT=ipk \
  scripts/e2e-vm.sh --mode=gate3
...
gate3: block leg app=youtube host=googlevideo.com; allow leg app=duolingo host=duolingo.com
gate3: block page hit for googlevideo.com (993 bytes)
gate3: upstream OK for duolingo.com (code=302, 138 bytes)
gate3: 17 /api/logs rows attributed to 02:e2:e3:13:da:72
gate3: time_status summary present for 02:e2:e3:...
1 passed in 72.07s
```

Both legs asserted at the connection layer end-to-end (block-page DNAT + upstream carve-out), usage round-tripped, `/api/time/status` surfaced the device. VMs/worktree torn down afterward; host left clean (no leaked qemu, bridges down, no secret in logs).

> Scope: this is the Python gate3 suite + `api_admin` only. The separate Gate **3a** DNS regression is fixed in #1812's parallel session (dnsmasq `render.lua`) — out of scope here by design, no file overlap.

Fixes #1810

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/scripts/e2e/gate3/conftest.py
+++ b/scripts/e2e/gate3/conftest.py
@@ -20,6 +20,7 @@ import os
 import pytest
 
 from lib.api_admin import AdminAPI
+from lib.app_seed import pick_block_allow_apps
 from lib.enrollment import exchange_enrollment_token, provision_router_uci
 from lib.vm import Client, client_down, client_up, router_down, router_up
 from lib.wait import wait_for_client_dns, wait_for_etag_change, wait_for_router_active
@@ -136,34 +137,50 @@ def client_vm() -> Client:
 
 @pytest.fixture()
 def scratch_profile_and_device(admin, enrolled_router, client_vm):
-    """Create a namespaced profile + device + blocked/allowed apps; yield
-    the relevant ids; delete on teardown.
+    """Create a namespaced profile + device, assign one seeded app `blocked`
+    and another `allowed` to it, and yield the relevant ids + hosts; clean
+    up on teardown.
 
     Post-#764, extraBlocked / extraAllowed are no longer profile fields —
-    they come from app_policy_assignments. The Apps API is:
-      POST   /api/apps {name, slug, hosts}                   → create
-      PUT    /api/apps/{aid}/policy/{pid} {mode: "blocked"}  → assign
-      DELETE /api/apps/{aid}                                 → cascades the
-                                                              assignment
+    they come from app_policy_assignments. Post-#1798, app *definitions* are
+    template-authored ONLY — the arbitrary-host `POST /api/apps` create was
+    retired, so this can no longer mint `example.com` / `example.org` apps.
+    Instead it seeds the shipped templates (idempotent, admin-gated) and
+    drives its two-leg block/allow assertion off two distinct *seeded* apps'
+    real hosts. The block/allow contract is host-agnostic (block = HTTP/80
+    DNAT to the block page on the resolved IPs; allow = carve-out / no
+    over-block), so the assertion is unchanged — only the host source moves
+    from a hand-minted app to the template catalog. Like example.com/.org,
+    the chosen template apexes resolve + are reachable through qemu SLIRP v4
+    NAT (see lib/app_seed.BLOCK_PREF / ALLOW_PREF).
+
+    Apps surface used — all kept by #1798:
+      POST   /api/apps/seed-from-templates          → seed catalog (idempotent)
+      GET    /api/apps                               → read seeded apps + hosts
+      PUT    /api/apps/{aid}/policy/{pid} {mode}     → assign block/allow
+      DELETE /api/apps/{aid}/policy/{pid}            → detach (teardown)
+    The shared seeded apps are NEVER deleted on teardown — only this run's
+    namespaced profile/device + its policy assignments.
     """
     suffix = _suffix()
-    blocked_host = "example.com"
-    allowed_host = "example.org"
+
+    # Seed (idempotent) then pick two distinct seeded apps for the two legs.
+    admin.seed_apps_from_templates()
+    blocked_app, allowed_app = pick_block_allow_apps(admin.list_apps())
+    blocked_app_id = blocked_app["app"]["id"]
+    allowed_app_id = allowed_app["app"]["id"]
+    blocked_host = blocked_app["hosts"][0]
+    allowed_host = allowed_app["hosts"][0]
+    log.info(
+        "gate3: block leg app=%s host=%s; allow leg app=%s host=%s",
+        blocked_app["app"].get("slug"), blocked_host,
+        allowed_app["app"].get("slug"), allowed_host,
+    )
 
     pname = f"gate3-{suffix}"
     p = admin.create_profile(name=pname)
     pid = (p.get("profile") or p)["id"]
 
-    blocked_app = admin.create_app(
-        name=f"gate3-blk-{suffix}", slug=f"gate3-blk-{suffix}",
-        hosts=[blocked_host],
-    )
-    allowed_app = admin.create_app(
-        name=f"gate3-alw-{suffix}", slug=f"gate3-alw-{suffix}",
-        hosts=[allowed_host],
-    )
-    blocked_app_id = blocked_app["app"]["id"]
-    allowed_app_id = allowed_app["app"]["id"]
     admin.assign_app_policy(app_id=blocked_app_id, profile_id=pid, mode="blocked")
     admin.assign_app_policy(app_id=allowed_app_id, profile_id=pid, mode="allowed")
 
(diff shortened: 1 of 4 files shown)


sameerparekh:
## Independent review — `fc7bddf` (first pass)

Read-only adversarial pass over the merge-base diff (`git diff origin/main...HEAD`, 4 files) against the 8 checklist dimensions.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `scripts/e2e/lib/api_admin.py:294` — `delete_app` is now unused by the fixture (its only caller switched to `delete_app_policy`). It maps to a still-kept admin route (#1798 stray-row cleanup) and the conftest references a periodic orphan-cleanup job, so retaining the thin wrapper is defensible — but if strict-dead-code is preferred it could be dropped. Left as a NIT.
2. `scripts/e2e/gate3/conftest.py:155-156` — both legs use `app["hosts"][0]`, and `GET /api/apps` returns hosts in unspecified DB order (the real Gate 3b run picked `googlevideo.com` for the youtube app, not `youtube.com`). Harmless for the block leg (host-agnostic DNAT; all template apexes resolve) and the current `ALLOW_PREF` picks are single-/apex-first, so the allow leg's HTTP/80 reachability held. A future multi-host allow pick whose `hosts[0]` is a CDN-only host could flake; preferring the registrable apex would harden it. Non-blocking.
3. `scripts/e2e/lib/test_app_seed.py` — like the sibling `lib/test_*.py` (test_admin_auth / test_log_reason / test_v6_ula), this pure unit test isn't wired into a PR CI lane (it runs standalone / in the gate venv). Consistent with existing convention; a pre-existing gap, out of scope here.

**Dimension notes**
- **SSoT (1):** `pick_block_allow_apps` is the single selection source; `BLOCK_PREF`/`ALLOW_PREF` defined once and imported. No re-derivation of server quantities. ✓
- **Testing (2):** Red→green visible (`ba583861` stub `NotImplementedError` → `fc7bddf` impl). `test_smoke.py` is **untouched** — the two-leg connection-layer assertions are preserved exactly; only the host *source* moved (hand-minted app → seeded template). No assertion weakened/skipped. New picker behavior asserted by 7 unit tests (prefs, distinctness, skip-empty-hosts, fallbacks, raise-on-<2). ✓
- **Architecture (3):** No router decision logic; block described correctly as connection-layer HTTP/80 DNAT (not DNS); no `/tmp` writes; no Clock/`java.time`. ✓
- **Wire (4):** No wire changes; uses only routes kept by #1798 (seed-from-templates / GET apps / PUT+DELETE policy). ✓
- **Migrations (5) / Metrics (6):** N/A (test-harness only). ✓
- **Scope/hygiene (7):** Focused 4-file diff; dead `create_app` removed; no TODOs; no AI mentions in title/body (Co-Authored-By trailer only); three-dot merge-base diff. ✓
- **Security (8):** No secrets committed (staging admin pass passed via env on the KVM host during validation, never in-repo). No SQL. Uses existing authed admin routes. ✓

**Validation cross-check:** PR documents a real Gate 3b run on the KVM host against live staging — `1 passed in 72.07s`, both legs asserted end-to-end (block-page DNAT + upstream carve-out), usage + time_status round-tripped. Faithful to `e2e-vm-gate3b.yml` (last-published ipk image, sha256-verified).

VERDICT: APPROVE
Restores Master API/UI CD Gate 3b by replacing the retired `POST /api/apps` with seed-from-templates + seeded-app selection, preserving the host-agnostic two-leg assertion. Option 2 correctly avoids polluting the prod template catalog (which Option 1 would, since Gate 3b runs the deployed staging binary). No blockers; validated green end-to-end on KVM.

PR #1734, opened 2026-06-14
refactor(#888): move shared/contract/** to top-level contract/**

sameerparekh:
## Summary

Promotes the cross-language wire-contract goldens out of `shared/` to a new top-level `contract/` directory. The goldens are consumed by Scala, Lua, and Python — not Scala-only — so the previous location was misleading every time someone wrote a CI path filter (see #880).

After this change:
- `shared/**` means Scala code shared between API and other JVM consumers.
- `contract/**` means the cross-language wire contract.

## Path filter updates (critical)

Routers continue to receive contract-change-triggered builds:

- `.github/workflows/master-router.yml`: `shared/contract/**` → `contract/**`
- `.github/workflows/master-api-ui.yml`: added `contract/**` (kept `shared/**`)
- `.github/workflows/ci.yml`: `contract:` filter and drift guard updated

## Loader updates

- `shared/test/src/contract/ContractGoldenSpec.scala` walks for top-level `contract/`
- `openwrt/test/contract_spec.lua` / `contract_gen.lua` / `jsonc_shim_spec.lua` / shim
- `scripts/e2e/fake-api/fake_api/fixtures.py` and scenarios_fake snapshot_builder
- `scripts/regen-contract-fixtures.sh`
- Docs (`docs/design/blocklists.md`, `docs/design/ios-mobile-app.md`, fake-api README)

## Test plan

- [x] `mill shared.test.testOnly wifihaven.shared.contract.ContractGoldenSpec` — 6/6 pass
- [x] `busted openwrt/test/contract_spec.lua openwrt/test/jsonc_shim_spec.lua` — 14/14 pass
- [x] `scripts/regen-contract-fixtures.sh` — `git diff contract/` is clean
- [x] `mill mill.scalalib.scalafmt/checkFormatAll` — clean
- [ ] CI green

Closes #888.

PR #1814, opened 2026-06-21
fix(cd): blocklist shards must live inside dnsmasq confdir (procd jail) — restore Master Router CD (Gate 2 + Gate 3a DNS)

### LINKED ISSUE #1812: ci(cd): Master Router CD still RED — Gate 3a v6 DNS refused (fdaa:bbbb:cccc::1#53) persists past #1795; Gate 2 e2e-brand needs cloudflare apply

sameerparekh:
## Summary

Master Router CD has been RED on `main` since #1734 (2026-06-15). **Root cause: the per-blocklist dnsmasq shard files (#1782) live in bare `/tmp`, outside OpenWRT's dnsmasq procd ujail, so the jailed dnsmasq cannot read the `conf-file=` shards it's told to load — it crash-loops, procd gives up, and `:53` goes dark.** Fix: move the shards into a subdir of the dnsmasq confdir (`/tmp/dnsmasq.d/blocklists/`), inside the jail's confdir bind-mount.

Fixes #1812.

## Confirmed root cause (refuting the issue's leading hypothesis)

The watchdog hypothesis was "dnsmasq config render regression / dangling conf-file". I **refuted** that with direct evidence and found the real cause:

- The real staging snapshot's full config — all 165k `nftset=` directives across 9 real blocklists (~15 MB of shards) — **passes `dnsmasq --test` cleanly in 0.22s** and dnsmasq uses **< 8 MB RSS**. So it's not a syntax, size, or OOM problem.
- The catch: `dnsmasq --test` runs **unjailed**. OpenWRT starts the real dnsmasq inside a **procd ujail** that bind-mounts **only the confdir** (`/tmp/dnsmasq.d`) plus a handful of known files — see `/etc/init.d/dnsmasq`'s `procd_add_jail_mount $dnsmasqconfdir …`. The #1782 shards sit in bare `/tmp`, **outside** the jail. So when `render.dnsmasq` emits `conf-file=/tmp/wifihaven-blocklist-<id>.conf` and the shard actually exists, the **jailed** dnsmasq cannot read it:
  ```
  daemon.crit dnsmasq[1]: cannot read /tmp/wifihaven-blocklist-ads.conf: No such file or directory
  daemon.crit dnsmasq[1]: FAILED to start up
  daemon.info procd: Instance dnsmasq is in a crash loop 6 crashes … 
  ```
  procd crash-loops it 6× and **gives up**, so `:53` stays permanently down → **Gate 3a**: client DNS to `fdaa:bbbb:cccc::1#53` is refused, and the client never gets a DHCP lease (`udhcpc: no lease`).

### Why Gate 2 stayed green but Gate 3a went red
Gate 2 (fake API) serves blocklists that **404**, so no shard is ever written and **no `conf-file=` is emitted** — the jail path is never exercised, dnsmasq stays up. Gate 3a points at **staging**, whose blocklists resolve to **real, populated shards**, so `conf-file=` references real out-of-jail files and dnsmasq dies. Same image, opposite outcome — which is exactly why this looked like a staging-only mystery.

### Why #1795 didn't fix it
#1795's `bl_shard_exists` gate (omit `conf-file=` for a *missing* shard) is correct and stays in place as defense-in-depth — but it addressed the *dangling-reference* case. The jail makes even a **present** shard unreadable, which the gate can't help with.

## Fix

`render.SHARD_DIR` moves from `/tmp` → `/tmp/dnsmasq.d/blocklists/` (single source of truth; the agent's `BLOCKLIST_SHARD_DIR` derives from it, and startup `mkdir -p` creates it). A **subdir** of the confdir — not the confdir root — so dnsmasq's non-recursive `conf-dir=/tmp/dnsmasq.d` does **not** auto-load the shards; they load only via the explicit, existence-gated `conf-file=` directives. Still tmpfs, still per-shard capped + GC'd (bounded-`/tmp` invariant preserved).

## Validation (on the self-hosted KVM host, `plex`)

Reproduced and fixed on the real OpenWRT 25.12 KVM VM — no macOS limitation:

1. **Reproduced** the exact Gate 3a failure (fresh boot, staging snapshot, real blocklists): `conf-file=/tmp/wifihaven-blocklist-ads.conf` referenced a real shard → jailed dnsmasq `cannot read` → crash-loop → `:53` refused, no lease. (`dnsmasq --test` passed throughout — the tell.)
2. **Fix verified end-to-end** on a freshly **rebuilt image** with the change, booted against the staging snapshot with real blocklists:
   - 9 shards under `/tmp/dnsmasq.d/blocklists/`, `conf-file=` paths inside the confdir;
   - dnsmasq **alive**, listening on `fdaa:bbbb:cccc::1:53` (the exact v6 ULA Gate 3a queries) and `192.168.100.1:53`;
   - **no crash loop**; a client **resolved `example.com` through the router**.
3. **Unit tests** (`busted`, 302 pass): a failing test committed first (red → green) asserting shards live under the confdir; existing #1782/#1792 conf-file path assertions updated.

CI's self-hosted Gate 2 + Gate 3a will run on this PR as the final gate.

## Note on #1810
This fix makes Gate 3a get past DNS setup, so the run now reaches the **separate** latent #1810 (Gate 3a's `test_smoke` still POSTs to the retired `/api/apps`). That is out of scope here and left to its own fix; it does not affect this PR's root cause or validation.

### DIFF
--- a/docs/process/router-agent-bounded-writes.md
+++ b/docs/process/router-agent-bounded-writes.md
@@ -40,7 +40,7 @@ The same reasoning applies to anything else the agent persists under `/tmp`
 (JSON spools, NDJSON event journals, debug dumps). If it can grow, it gets a
 bound — in the same PR.
 
-## Per-blocklist dnsmasq conf shards (`/tmp/wifihaven-blocklist-<id>.conf`) {#blocklist-shards}
+## Per-blocklist dnsmasq conf shards (`/tmp/dnsmasq.d/blocklists/wifihaven-blocklist-<id>.conf`) {#blocklist-shards}
 
 Added in [#1782/#1783](https://github.com/wifihaven/wifihaven/issues/1782).
 One file per blocklist id in the current snapshot.
@@ -49,8 +49,22 @@ One file per blocklist id in the current snapshot.
 file (`/etc/wifihaven/blocklists/<id>-<version>.txt`) line-by-line and writes
 one `nftset=/<host>/4#inet#wifihaven#bl_<id>,6#inet#wifihaven#bl6_<id>` line
 per host — never accumulating the full list in a Lua table. `render.dnsmasq`
-emits `conf-file=/tmp/wifihaven-blocklist-<id>.conf` for each id so dnsmasq
-picks up the host set at startup and on reload.
+emits `conf-file=/tmp/dnsmasq.d/blocklists/wifihaven-blocklist-<id>.conf` for
+each id so dnsmasq picks up the host set at startup and on reload.
+
+**Shards MUST live inside the dnsmasq confdir** (`/tmp/dnsmasq.d`), not bare
+`/tmp` — [#1812](https://github.com/wifihaven/wifihaven/issues/1812). OpenWRT
+runs dnsmasq in a procd ujail (`/etc/init.d/dnsmasq`'s
+`procd_add_jail_mount $dnsmasqconfdir …`) that bind-mounts ONLY the confdir and
+a few known files. A `conf-file=` directive pointing at a shard outside the
+confdir is unreadable inside the jail: dnsmasq aborts with `cannot read …`,
+procd crash-loops it and gives up, and `:53` goes dark (connection refused, no
+DHCP leases). `dnsmasq --test` passes only because it runs unjailed. The shard
+dir is a *subdir* of the confdir (`…/blocklists/`) so dnsmasq's
+non-recursive `conf-dir=/tmp/dnsmasq.d` does not auto-load the shards — they
+load only via the explicit, shard-existence-gated `conf-file=` directives.
+`render.SHARD_DIR` is the single source of truth (the agent's
+`BLOCKLIST_SHARD_DIR` derives from it).
 
 **Atomic write.** Each shard is written to `<id>.conf.tmp`, then renamed onto
 `<id>.conf`. A partial shard is never visible to dnsmasq.
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -77,11 +77,21 @@
 
 local M = {}
 
--- Directory where per-blocklist dnsmasq conf shards live (#1782). Default /tmp
--- (= tmpfs on OpenWRT). Overridable for tests via M.SHARD_DIR assignment before
--- calling M.dnsmasq(). Production: always /tmp (matches BLOCKLIST_SHARD_DIR in
--- wifihaven-agent).
-M.SHARD_DIR = "/tmp"
+-- Directory where per-blocklist dnsmasq conf shards live (#1782). A SUBDIR of
+-- the dnsmasq confdir (/tmp/dnsmasq.d, itself tmpfs = RAM on OpenWRT), NOT bare
+-- /tmp — #1812. OpenWRT runs dnsmasq inside a procd ujail that bind-mounts ONLY
+-- the confdir (and a handful of known files): see /etc/init.d/dnsmasq's
+-- `procd_add_jail_mount $dnsmasqconfdir …`. A `conf-file=` directive in
+-- wifihaven.conf that points at a shard OUTSIDE the confdir is unreadable inside
+-- the jail — dnsmasq aborts with "cannot read …", procd crash-loops it and
+-- gives up, and :53 goes dark (connection refused, no DHCP leases). Keeping the
+-- shards under the confdir puts them inside the jail mount. A SUBDIR (not the
+-- confdir root) so dnsmasq's `conf-dir=/tmp/dnsmasq.d` does NOT auto-load them
+-- (it is non-recursive) — they load only via the explicit, shard-existence-gated
+-- conf-file= directives. Overridable for tests via M.SHARD_DIR assignment before
+-- calling M.dnsmasq(). Production: always this dir (matches BLOCKLIST_SHARD_DIR
+-- in wifihaven-agent, which derives from render.SHARD_DIR).
+M.SHARD_DIR = "/tmp/dnsmasq.d/blocklists"
 
 -- Canonical path of a per-blocklist dnsmasq conf shard (#1792). Single source
 -- of truth — render.dnsmasq, policy.apply's shard-existence check, and
--- a/openwrt/files/usr/sbin/wifihaven-agent
+++ b/openwrt/files/usr/sbin/wifihaven-agent
@@ -458,9 +458,13 @@ end
 -- BLOCKLIST_CACHE_DIR: persistent on-flash cache of fetched list bodies;
 --   survives reboots; shards in BLOCKLIST_SHARD_DIR are rebuilt from it.
 local BLOCKLIST_CACHE_DIR = "/etc/wifihaven/blocklists"
--- BLOCKLIST_SHARD_DIR: per-blocklist dnsmasq conf files (/tmp = tmpfs = RAM);
---   written atomically (tmp→rename) and GC'd on each render + startup.
-local BLOCKLIST_SHARD_DIR = "/tmp"
+-- BLOCKLIST_SHARD_DIR: per-blocklist dnsmasq conf files (tmpfs = RAM); written
+--   atomically (tmp→rename) and GC'd on each render + startup. A SUBDIR of the
+--   dnsmasq confdir (#1812) — NOT bare /tmp — so OpenWRT's procd ujail (which
+--   bind-mounts only the confdir) can read the shards the conf-file= directives
+--   reference; a bare-/tmp shard is unreadable inside the jail and crash-loops
+--   dnsmasq. Derived from render.SHARD_DIR so writer and reader never drift.
+local BLOCKLIST_SHARD_DIR = render.SHARD_DIR
 local function refresh_blocklists(snap)
   -- Step 1: fetch new list versions + cache them to disk.
   local res = blocklists.refresh(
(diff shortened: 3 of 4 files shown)


sameerparekh:
**Independent PR review** (read-only, adversarial) — first run.

Diff: `render.lua` (shard dir `/tmp` → `/tmp/dnsmasq.d/blocklists`), `wifihaven-agent` (`BLOCKLIST_SHARD_DIR` derives from `render.SHARD_DIR`; startup `mkdir -p` of the subdir), `render_spec.lua` (red→green test + path-assert updates), and the bounded-writes doc. Reviewed against all 8 dimensions.

**BLOCKERS**
(none)

**SHOULD-FIX**
1. `openwrt/files/usr/sbin/wifihaven-agent:560` (startup `gc_shards`) — after the shard dir moves, `gc_shards` scans only the new subdir, so **legacy `/tmp/wifihaven-blocklist-*.conf` shards written by a pre-#1812 agent are never removed on a package upgrade** (agent restarts without a reboot). They're unreferenced and harmless, but hold up to ~15 MB of tmpfs on a 512 MB router until the next actual reboot. A one-time `rm -f /tmp/wifihaven-blocklist-*.conf` at startup reclaims it.

**NITS**
1. `openwrt/files/usr/sbin/wifihaven-agent:551` — `string.format("%q", BLOCKLIST_SHARD_DIR)` uses Lua-string quoting for a **shell** argument. Safe here (the value is a trusted constant with no shell metacharacters), but `%q` is not shell-safe in general; a plain literal append or a dedicated `mkdir` call would be clearer.
2. PR body carries the `🤖 Generated with Claude Code` footer — checklist §7 asks for no AI mentions in the PR body (the `Co-Authored-By` trailer in commits is fine). Noting per the checklist; harness convention adds the footer.

**Dimension notes (positives):** Single-source-of-truth is *improved* — `render.SHARD_DIR` is now the one source and the agent derives from it, removing the prior two hand-synced `/tmp` literals (§1 ✓). TDD red→green is visible (`5f291507` failing test → `782d37e7` fix) and the test pins the new confdir invariant; the jail behaviour itself is validated e2e on the KVM VM per the PR body (§2 ✓). No wire/snapshot change — the shard path is filesystem-internal, not a wire contract (§4 ✓). Bounded-`/tmp` invariant preserved: shards stay tmpfs, per-shard capped + GC'd, and the doc is updated (§3 ✓). Focused diff, no metrics/migrations/security surface touched (§5–8 ✓).

**VERDICT: APPROVE**

Correct, well-scoped root-cause fix: OpenWRT's dnsmasq procd ujail bind-mounts only the confdir, so the #1782 bare-`/tmp` shards are unreadable by the jailed dnsmasq → crash-loop → `:53` dark (Gate 3a). Moving them to a confdir subdir is the minimal correct fix, validated end-to-end on the KVM host. No blockers; the one SHOULD-FIX (legacy-shard cleanup on upgrade) is worth doing before merge.

sameerparekh:
**Independent PR review — re-run** (delta `782d37e7...abcdea0b`).

**Prior findings status**

|---|---|---|
| SHOULD-FIX 1 | legacy bare-`/tmp` shards not cleaned on upgrade | **ADDRESSED** — `wifihaven-agent:558` now `rm -f /tmp/wifihaven-blocklist-*.conf{,.tmp}` at startup (non-recursive glob; does not touch the new confdir subdir). |
| NIT 1 | `%q` (Lua quoting) on a shell arg | **ADDRESSED** — `wifihaven-agent:553` uses a bare append of the trusted constant. |
| NIT 2 | `🤖` footer in PR body | **NOT-ADDRESSED** (intentional) — operator wants AI authorship visible; the harness adds this footer and the `Co-Authored-By` trailer is the expected part. Non-blocking. |

**New findings in the delta**

BLOCKERS: (none)
SHOULD-FIX: (none)
NITS: (none) — the legacy-cleanup `rm` is idempotent (a no-op after the first boot), non-recursive (cannot hit `/tmp/dnsmasq.d/blocklists/`), and the bare mkdir append is safe for the trusted constant.

**VERDICT: APPROVE**

Re-run clean: both actionable prior findings are addressed and the delta introduces no new issues. The fix (blocklist shards under the dnsmasq confdir so the procd jail can read them) is correct, minimal, single-source-of-truth, and validated end-to-end on the KVM host. No blockers.

sameerparekh:
**Independent PR review — re-run** (delta `abcdea0b...HEAD`).

The new commit is an **empty re-trigger commit** (no file changes — the prior push's `pull_request` CI synchronize event was dropped by GitHub; this re-fires the full PR matrix on the shipped tree). No code delta to review.

**Prior findings status** (unchanged — no files touched since last run)

|---|---|---|
| SHOULD-FIX 1 | legacy bare-`/tmp` shards not cleaned on upgrade | **ADDRESSED** (`wifihaven-agent` startup `rm -f`) |
| NIT 1 | `%q` on a shell arg | **ADDRESSED** (bare append of trusted constant) |
| NIT 2 | `🤖` footer in PR body | **NOT-ADDRESSED** (intentional; non-blocking) |

**New findings in the delta:** none (empty commit).

**VERDICT: APPROVE**

No code change since the last clean re-run; the fix (blocklist shards under the dnsmasq confdir so the procd jail can read them) stands, validated end-to-end on the KVM host with the relevant CI jobs (OpenWrt Lua / Shell / Contract) green. No blockers.

PR #1818, opened 2026-06-22
fix(api): scope snapshot.blocklists to referenced lists only (#1784)

### LINKED ISSUE #1784: api: scope PolicyService.snapshot.blocklists to referenced lists only
## Context

Surfaced as a side-finding during the [#1435 design investigation](https://github.com/wifihaven/wifihaven/issues/1435#issuecomment-...). `PolicyService.snapshot` builds the `blocklists` map from **every** bundled list, regardless of assignment:

```scala
// api/src/policy/PolicyService.scala (around line 98)
cats <- blocklistRepo.listCategories            // ALL categories in the system
...
val pBlocklists = cats.map { c => c -> Blocklist(version, url) }.toMap
snapshot.copy(blocklists = pBlocklists)
```

Consequence: the agent fetches and renders directives for **every** bundled list, even ones no profile uses. Un-assigning a list from every profile does **not** reduce the agent's workload — it still loads it. The only operational levers today are to `clear` or delete the list. (Confirmed live in the [#1412](https://github.com/wifihaven/wifihaven/issues/1412) investigation: after un-assigning, a policy nudge re-rendered the full 164k; only after `clear`ing the three extended lists did the agent converge to ~570 directives.)

This is the right fix regardless of which large-list strategy #1435 picks, and it pairs well with Option 2c: smaller shard count → less tmpfs churn → simpler lifecycle.

## Scope

- `snapshot.blocklists` should include **only** blocklists referenced by some device's effective `blocklistIds` (i.e. union over the snapshot's profiles' effective `BlockRules.blocklistIds`).
- The agent already iterates whatever ids are present in the snapshot, so this is additive-safe on the wire.
- Add a server-side test that asserts: with zero profiles referencing list X, `snapshot.blocklists` does not contain X.

## Acceptance

- Un-assigning a blocklist from every profile causes the agent's next render to drop the corresponding shard within one snapshot cycle.
- No regression to the existing per-`(MAC, blocklistId)` enforcement: a list referenced by *any* profile still ships.

Parent: #1435

sameerparekh:
## Summary

`PolicyService.snapshot` built the `blocklists` map from `blocklistRepo.listCategories` — i.e. **every bundled list in the system**, regardless of whether any profile referenced it. The router agent fetches and renders directives for every id it sees in the snapshot, so an un-assigned list was still loaded; un-assigning a list from every profile did not reduce the agent's workload. On prod this was the difference between hundreds of thousands of directives and hundreds (surfaced during the [#1412](https://github.com/wifihaven/wifihaven/issues/1412) investigation).

Filter `pBlocklists` to the union of `blocklistIds` across the snapshot's:
- profile policies (`profilePolicies.values.flatMap(_.rules.blocklistIds)`)
- global section (`globalRules.blocklistIds`)
- per-device override rules (`devicePolicies.values.flatMap(_.rules.toList.flatMap(_.blocklistIds))`)

Wire-compat: **additive-safe**. The router agent already iterates whatever ids appear in `snapshot.blocklists`; an id no longer present is one the agent was never going to use. No router-agent changes needed (AGENTS.md `#wire-contract`).

## Test plan

- [x] New feature spec [PolicySnapshotBlocklistScopeSpec](api/test/src/feature/PolicySnapshotBlocklistScopeSpec.scala) covering: single profile reference, multi-profile union, zero references → empty, and global-sentinel-only reference. Committed first as a failing test (TDD red commit `dc2c6732`) before the fix landed.
- [x] Three existing `RouterApiSpec` assertions that incidentally relied on the "ship every list" behavior updated to reference the lists they assert on.
- [x] `mill api.test` green (full suite).
- [x] `mill mill.scalalib.scalafmt.ScalafmtModule/checkFormatAll` clean.

Closes #1784
Refs #1435 (parent design), #1412 (where this was originally surfaced)

### DIFF
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -310,6 +310,9 @@ class PolicyServiceLive(
       // shipping unreferenced lists is wasted work — on prod the difference is hundreds of
       // thousands of directives vs hundreds (see #1412 investigation). Additive-safe on the wire:
       // the agent already iterates whatever ids appear here.
+      // The per-device override lane (`devicePolicies → rules → blocklistIds`) is wired in for
+      // future-proofing — today the only writer of `DevicePolicy.rules` is the unmanaged-MAC path
+      // (above), which sets `blocklistIds = Nil`, so this branch currently contributes nothing.
       val referencedBlocklistIds: Set[BlocklistId] =
         (globalRules.blocklistIds.iterator ++
           profilePolicies.valuesIterator.flatMap(_.rules.blocklistIds) ++
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -272,8 +272,8 @@ class PolicyServiceLive(
         d.mac -> DevicePolicy(profileId = d.profileId, name = d.name, rules = rules)
       }.toMap
 
-      val catDomainsMap                            = catDomains.toMap
-      val pBlocklists: Map[BlocklistId, Blocklist] = cats.map { c =>
+      val catDomainsMap                               = catDomains.toMap
+      val pBlocklistsAll: Map[BlocklistId, Blocklist] = cats.map { c =>
         val domains = catDomainsMap.getOrElse(c, Set.empty[Hostname]).map(_.value)
         val version = BlocklistVersion.unsafe(PolicyService.blocklistContentVersion(domains))
         c -> Blocklist(version = version, url = BlocklistUrl.unsafe(s"/api/blocklists/${c.value}"))
@@ -304,6 +304,18 @@ class PolicyServiceLive(
         extraBlocked = globalRulesResolved.extraBlocked.distinct,
         blocklistIds = globalRulesResolved.blocklistIds.distinct,
       )
+
+      // #1784: scope `blocklists` to only those referenced by some profile (or by the global
+      // section / unmanaged-MAC rules). The router fetches and renders every id in this map, so
+      // shipping unreferenced lists is wasted work — on prod the difference is hundreds of
+      // thousands of directives vs hundreds (see #1412 investigation). Additive-safe on the wire:
+      // the agent already iterates whatever ids appear here.
+      val referencedBlocklistIds: Set[BlocklistId] =
+        (globalRules.blocklistIds.iterator ++
+          profilePolicies.valuesIterator.flatMap(_.rules.blocklistIds) ++
+          devicePolicies.valuesIterator.flatMap(_.rules.iterator.flatMap(_.blocklistIds))).toSet
+      val pBlocklists: Map[BlocklistId, Blocklist] =
+        pBlocklistsAll.view.filterKeys(referencedBlocklistIds).toMap
 
       val core = SnapshotCore(globalRules, devicePolicies, profilePolicies, pBlocklists)
       val etag = PolicyService.computeEtag(core)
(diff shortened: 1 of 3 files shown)


sameerparekh:
# Independent PR review — #1818

First run on `4a82cfd1`. Reviewed full merge-base diff `git diff origin/main...HEAD` against the 8-dimension checklist.

## BLOCKERS

(none)

## SHOULD-FIX

(none)

## NITS

1. [api/src/policy/PolicyService.scala:155](api/src/policy/PolicyService.scala#L155) — the server still calls `blocklistRepo.loadCategory(c)` for **every** bundled category and computes a SHA over its full domain set, even for categories no profile references. The filter at L317 only drops them from the wire — the per-cat hashing cost stays. Cheap to fix (compute `referencedBlocklistIds` from `profiles`/`devices` *before* the `catDomains` foreach and restrict the load to that set), and on prod with the extended lists this is the bulk of the server-side cost the issue calls out. Fine as a follow-up; the wire-side fix in this PR is the headline goal.

2. [api/src/policy/PolicyService.scala:313](api/src/policy/PolicyService.scala#L313) — `devicePolicies.valuesIterator.flatMap(_.rules.iterator.flatMap(_.blocklistIds))` is defensively correct (per-device override can carry `blocklistIds`) but today the only writer of `DevicePolicy.rules` is the unmanaged-MAC path, which sets `blocklistIds = Nil`. The union is a no-op until a future path uses it. Worth a one-line comment that this lane is wired in for future-proofing, not because it currently contributes ids.

## Dimension-by-dimension notes

- **§1 SSoT.** The fix reads `blocklistIds` straight off the already-computed `globalRules` / `profilePolicies` / `devicePolicies` — no re-derivation, no "keep in sync" duplication. The single source is the per-profile `BlockRules` the rest of the snapshot is built from.
- **§2 Testing.** TDD red→green is visible in history: failing-test commit `dc2c6732` lands the spec first, then `4a82cfd1` lands the fix. New `PolicySnapshotBlocklistScopeSpec` covers single-profile, multi-profile union, zero-references, and the global-sentinel path. The three edits in `RouterApiSpec` (L233, L672, L719) add a profile reference for the lists the assertions already covered — they preserve the *intent* of each assertion rather than weakening it (the old form was accidentally riding the bug). PR body explains this. Feature-level tests through the real `PolicyService` on embedded Postgres, no mocks. ✓
- **§3 Architectural invariants.** No new wire fields. Wire shape unchanged — only the *contents* of `blocklists` are narrowed. The router stays a dumb applier (iterates whatever ids appear). No DNS-as-enforcement reasoning anywhere. ✓
- **§4 Wire contract.** Additive-safe per the architecture model: the agent already only acts on ids present in `snapshot.blocklists`. Removed ids were unreferenced by any profile, so no per-`(MAC, blocklistId)` nftables rule was using them. The ETag will move on the first post-deploy poll (expected; one re-poll, then steady-state).
- **§5 Migrations.** None.
- **§6 Metrics.** No new code path; no new metric required.
- **§7 Hygiene.** Focused diff, no drive-by edits, no AI mentions in commits/PR body.
- **§8 Security.** N/A.

VERDICT: APPROVE

snapshot.blocklists is now scoped to lists actually referenced by some profile / the global sentinel / a per-device override, eliminating the agent-side waste from #1412. TDD red→green is preserved in history, the wire change is additive-safe, and the two NITs are non-blocking follow-ups.

sameerparekh:
# Independent PR review — #1818 (re-run)

Re-run against prior review at `4a82cfd1`. Delta is a single 3-line comment in [api/src/policy/PolicyService.scala:313](api/src/policy/PolicyService.scala#L313).

## Prior findings status

|---|---|---|
| NIT 1 | Server still loads / SHAs every category, not just referenced ones | NOT-ADDRESSED — filed as follow-up #1819 (intentional out-of-scope deferral, the headline wire fix is preserved) |
| NIT 2 | Device-override lane is defensive, not currently load-bearing — worth a comment | ADDRESSED — comment added at L313–315 |

## New findings on delta `4a82cfd1...7be6dffd`

### BLOCKERS

(none)

### SHOULD-FIX

(none)

### NITS

(none — comment-only change.)

VERDICT: APPROVE

NIT 2 is fixed inline. NIT 1 was deferred to #1819 with the rationale that this PR's purpose is the agent-side wire fix; server-side `loadCategory` cost is a follow-up. No new BLOCKER introduced.

PR #1820, opened 2026-06-22
docs(architecture): note resolve-time bl_<id> population and the cache-warmth gap (#1786)

### LINKED ISSUE #1786: enforcement: cache-warmth gap — clients with cached DNS skip ipset population until TTL
## Context

Surfaced as a robustness gap during the [#1435 design investigation](https://github.com/wifihaven/wifihaven/issues/1435). The per-`(MAC, blocklistId)` ipsets (`bl_<id>`/`bl6_<id>`) are populated **at DNS resolution time** by dnsmasq's `nftset=` callback. So a client that already has the destination's IP in its DNS cache from before the policy change does not contribute an IP to the ipset until its TTL expires and it re-queries — meaning enforcement does not kick in for that flow until the cache rolls over.

This is fine for small curated lists where any given client only resolves a handful of relevant hosts. It becomes more visible with large lists: more hosts a client may already have cached → more flows that escape until TTL roll.

Distinct from #1435's RAM-cost problem and worth filing separately so the large-list rework doesn't accidentally couple to it.

## Possible directions (not deciding here)

- **Periodic server-side resolution** of subscribed blocklist hosts → API ships pre-resolved IPs the agent loads into `bl_<id>` immediately, in parallel to the DNS-time population path. Adds API-side resolver + TTL refresh machinery.
- **DNS cache flush on policy change** for blocklist-relevant scope. Heavy hammer; impacts unrelated entries.
- **Accept the gap, document it.** TTL on the typical ad/tracker entries is 5 min; lag is bounded.

## Acceptance (for THIS issue — just the design pick)

- A decision on the above directions captured in this issue's comments, with the same evidence-grounded pattern as #1435.
- Implementation, if any, lands in follow-up issues per the chosen direction.

Parent: #1435

sameerparekh:
## Summary

Operator picked Direction 3 on #1786 (accept + document). This adds a short callout to `docs/architecture.md` (§0.2) documenting:

- `bl_<id>` / `bl6_<id>` ipsets are populated at client-DNS-resolution time via dnsmasq's `nftset=` callback (the snapshot ships category membership, not pre-resolved IPs).
- When a category newly applies to a `(mac, host)` pair AND the client already holds a fresh DNS cache entry for `host`, enforcement lags until the client's TTL expires. Bounded by stored TTL — typically ≤ 5 min for ad/tracker workloads.
- Router cannot invalidate the client's own DNS cache.
- New hosts the client never resolved here, and blocklist-URL refreshes that ship new members, see a ≈ zero gap.

Cross-references #1786 inline.

Closes #1786.

## Test plan
- [x] `git diff` shows only `docs/architecture.md` (+23 lines)
- [x] Wording matches surrounding tone; reuses existing vocabulary (`blocklistIds`, `bl_<id>`, `nftset=`)
- [x] Direction 1 / 2 not advocated for

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -215,6 +215,29 @@ for usage attribution (§7.2). It is **never** the enforcement plane.
 > templates in
 > [`api/resources/app_templates/_README.yml`](../api/resources/app_templates/_README.yml).
 
+> **Cache-warmth gap on `blocklistIds` — bounded by client TTL.** The
+> `bl_<id>` / `bl6_<id>` ipsets are populated at **client-DNS-resolution
+> time** via dnsmasq's `nftset=` callback (rendered in
+> `openwrt/files/usr/lib/lua/wifihaven/blocklists.lua`; see also the
+> comment at `openwrt/files/usr/lib/lua/wifihaven/render.lua:64`) — the
+> snapshot ships category *membership*, not pre-resolved IPs, and the
+> agent does not pre-seed the sets at boot. The consequence: when a
+> category list is newly applied to a `(mac, host)` pair AND the client
+> already holds a fresh OS / browser DNS cache entry for `host`,
+> enforcement for that flow lags until the client's TTL expires and it
+> re-queries — the router has no way to invalidate the *client's* DNS
+> cache. The window is bounded by that stored TTL: typically ≤ 5 min for
+> ad / tracker workloads (median TTL on programmatic-ad apexes sits in
+> the 60–300 s range), with outliers on high-TTL consumer-brand apexes
+> (social / gambling) where the client cache is hot regardless of
+> policy. Scope: this applies to a category becoming newly active for a
+> `(mac, host)` pair; brand-new hosts the client has never resolved
+> through this router, and blocklist-URL refreshes that ship new members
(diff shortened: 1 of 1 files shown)


sameerparekh:
BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. docs/architecture.md:222 — the new callout lives in §0.2 but says "Cache-warmth gap on `blocklistIds`" while the closest enforcement-row reference is in the §0.2 enforcement table at L185. Reads fine as written; no action.

VERDICT: APPROVE

Docs-only change (+23 lines, single file). Adds a callout under §0.2 documenting that `bl_<id>` / `bl6_<id>` are populated at client-DNS-resolution time via dnsmasq's `nftset=` callback, so newly-applied category enforcement for a `(mac, host)` lags client TTL when the client's OS/browser cache is already warm. The mechanism claim is verified against `openwrt/files/usr/lib/lua/wifihaven/blocklists.lua` (line 429 emits the per-host `nftset=…bl_<san>,…bl6_<san>` directive at resolve time). Tone, vocabulary, and cross-reference (#1786) match the surrounding doc; does not advocate Direction 1 or 2; no code/test/wire/migration impact.

PR #1830, opened 2026-06-22
fix(#1829): fix empty-ruleset v6 e2e test to unblock Gate 2

### LINKED ISSUE #1829: bug(e2e): test_v6_usage_attribution reads empty nft ruleset → Gate 2 red on main since #1808 (blocks router CD)
## Symptom
Master Router CD has been RED on `main` since #1808 merged (2026-06-21 18:03). Gate 2 (router fake-API e2e) fails **deterministically** (confirmed across two runs, e.g. run 27926346432 + its re-run) at:
```
scenarios_fake/test_v6_usage_attribution.py::test_v6_usage_attributes_to_fqdn_and_aggregates_with_v4
AssertionError: nft ruleset is missing the v6 accounting set 'mac_ip6_tracking' — pre-#1802 IPv4-only accounting plane regressed?
assert 'mac_ip6_tracking' in ''
```
The asserted-against ruleset string is **empty (`''`)** — the test isn't finding ANY ruleset, not just the v6 set.

## This is a TEST/HARNESS bug, not a product regression
- `mac_ip6_tracking` is rendered **unconditionally** in `openwrt/files/usr/lib/lua/wifihaven/render.lua` (~line 827), beside the v4 `mac_ip_tracking` set, and its **busted unit tests pass** (render_spec.lua, usage_spec.lua).
- The 43 passing tests in the same suite are Python `test_snapshot_builder.py` checks that never query the live VM nft ruleset; this is the only test that does, and it gets `''`.
- The test was added by #1808 (gate for #1804) and has **never been green on main** — it merged broken.
- Run logs also show VM networking noise (`Cannot find device eth1`, `udhcpc: no lease, failing`), so the agent may not have applied its ruleset, or the test queries nft on the wrong host/namespace/table or before apply.

## Fix (decision tree — unblock the pipeline fast)
1. Read `scripts/e2e/scenarios_fake/test_v6_usage_attribution.py` + its nft-ruleset-fetch helper. Determine why the ruleset comes back empty: wrong nft command/table/namespace, queried before the agent applies, or fake-mode genuinely never brings up the accounting chain (e.g. v6 LAN / networking not configured in fake mode).
2. **If there's a clear test/harness fix** (e.g. wait-for-apply, target `inet wifihaven` table correctly, bring up v6 in the fake harness): apply it.
3. **If a correct fix isn't quick** (needs deeper KVM-side harness work): **QUARANTINE to unblock** — mark the test `xfail(strict=False)`/`skip` with a clear reason + a reference to THIS issue, so Gate 2 goes green and router CD can publish again. The test currently asserts against an empty ruleset, so it provides **no real coverage today** — quarantining loses nothing while we fix it properly. File the proper-fix as a follow-up referencing this issue.

## Validation / host note
Gate 2 runs on the self-hosted KVM runner; a macOS chip cannot run it locally and must iterate via CI (~44 min/run). The quarantine path needs no KVM. If a real harness fix is required, deep iteration may be faster in a Linux session that can run Gate 2 on plex directly.

## Why it matters now
Router CD red ⇒ no router release ⇒ the #1796 IPv6 attribution fix (#1807/#1802) can't reach prod routers. Unblocking this is on the critical path.

Relates to #1808, #1804, #1796.

sameerparekh:
## Root cause
Master Router CD Gate 2 (router fake-API e2e) has been RED on `main` since #1808 merged. `scenarios_fake/test_v6_usage_attribution.py` fails deterministically at its real-nft regression guard:

```
AssertionError: nft ruleset is missing the v6 accounting set 'mac_ip6_tracking'
assert 'mac_ip6_tracking' in ''
```

The asserted-against ruleset string is **empty**. The cause is a **test/harness bug, not a product regression**: the guard ran

```
nft list sets inet wifihaven
```

which is **malformed** — `nft list sets` takes only an optional *family* token (`inet`/`ip`/`ip6`/…), so the trailing table name `wifihaven` makes nft error out. With `2>/dev/null || true` the stderr was swallowed and stdout came back `''`, so the assertion always ran against an empty string. The test has therefore **never been green on main** and provided **zero real coverage**.

(`mac_ip6_tracking` / `ip_pair6_tracking_rx` are rendered **unconditionally** in `render.lua` ~827/835 and covered by passing busted unit tests — the product is fine.)

## What I did (correct fix, not quarantine)
- Switched the read to `nft list table inet wifihaven` — the exact form `scenarios_fake/_observers.py` already uses successfully. The table dump includes each set's `set <name> { ... }` declaration, so the `in` membership check still holds.
- Wrapped it in a short `wait_until` poll (60s) to absorb the fetch→apply gap — the fake serves the etag *before* the agent finishes rendering the ruleset, so a single immediate read could race. This preserves (in fact strengthens) the #1796/#1802 regression-guard value instead of discarding it.

No quarantine needed — this is a clean, high-confidence harness fix, so no follow-up issue is required.

## Validation
Cannot run Gate 2 locally (no KVM on macOS). `python3 -m py_compile` passes. CI **Master Router CD → Gate 2** validates end-to-end against the plex KVM runner (~44 min). The VM-networking log noise (`Cannot find device eth1`, `udhcpc: no lease`) is incidental — the failing assertion was deterministic on the malformed command, and `wait_until` now also tolerates a slow apply.

Fixes #1829

### DIFF
--- a/scripts/e2e/scenarios_fake/test_v6_usage_attribution.py
+++ b/scripts/e2e/scenarios_fake/test_v6_usage_attribution.py
@@ -331,15 +331,32 @@ def test_v6_usage_attributes_to_fqdn_and_aggregates_with_v4(router, client, fake
         # the loaded ruleset. Pre-#1802 they do not (the regression's root
         # cause), so this is a cheap, deterministic #1796 guard independent of
         # the traffic round-trip. ──────────────────────────────────────────
-        sets = router_ssh(
-            "nft list sets inet wifihaven 2>/dev/null || true",
-            check=False, timeout=10,
-        ).stdout or ""
-        for s in ("mac_ip6_tracking", "ip_pair6_tracking_rx"):
-            assert s in sets, (
-                f"nft ruleset is missing the v6 accounting set {s!r} — "
-                f"pre-#1802 IPv4-only accounting plane regressed?\n{sets!r}"
-            )
+        #
+        # NB: dump the whole table (`nft list table inet wifihaven`), NOT
+        # `nft list sets inet wifihaven` — the latter is malformed (`list sets`
+        # takes only an optional FAMILY token, so the trailing table name makes
+        # nft error out; the `2>/dev/null` then swallowed it and the assert ran
+        # against an empty string, which is why this gate was red on every run
+        # since #1808 merged, see #1829). The table dump includes each set's
+        # `set <name> { ... }` declaration, so the membership check still holds.
+        # Poll briefly to absorb the fetch→apply gap (the etag is *served*
+        # before the agent finishes rendering the ruleset).
+        _required_sets = ("mac_ip6_tracking", "ip_pair6_tracking_rx")
+
+        def _v6_sets_present():
+            out = router_ssh(
+                "nft list table inet wifihaven 2>/dev/null || true",
+                check=False, timeout=10,
+            ).stdout or ""
+            return out if all(s in out for s in _required_sets) else None
+
+        wait_until(
+            _v6_sets_present, timeout_s=60, interval_s=3,
+            description=(
+                "nft table inet wifihaven contains the v6 accounting sets "
+                f"{_required_sets} (#1802 / #1796 regression guard)"
+            ),
+        )
 
         url = f"http://{LEAF_HOST}/"
         _wait_origin_reachable(client, url=url, family_flag="-6")


sameerparekh:
**Independent PR review** — read-only, adversarial pass per `docs/pr-review-checklist.md`. First run (no prior marked comment).

Scope: a single-file e2e test-harness fix (`scripts/e2e/scenarios_fake/test_v6_usage_attribution.py`). Reviewed `git diff origin/main...HEAD`.

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `scripts/e2e/scenarios_fake/test_v6_usage_attribution.py:347` — on timeout, `wait_until` raises a generic `TimeoutError` with only the `description`; the prior code's assertion dumped the actual ruleset (`\n{sets!r}`) for diagnosis. Minor loss of failure-time signal — acceptable, since a genuine missing-set failure can be diagnosed by re-running `nft list table inet wifihaven` on the runner.
2. The corrected command string `nft list table inet wifihaven` is duplicated from `_observers.py:66/245` rather than sharing a helper — pre-existing pattern in the harness, not introduced here.

Dimension notes:
- **Testing integrity (✓ strengthened, not weakened):** the old guard asserted against `''` (the `nft list sets inet wifihaven` command was malformed — `list sets` takes only an optional *family* token — so nft errored, `2>/dev/null` swallowed it, and the assert always ran against an empty string → zero coverage). The fix makes the assertion actually execute. The `wait_until` poll does **not** mask the regression it guards: if the v6 sets never render it still times out and fails (60s). Correctly tolerates only the fetch→apply race (`wait_for_etag_served` confirms the etag was *served*, before the agent finishes rendering).
- **Single source of truth:** no policy/quantity re-derivation. n/a.
- **Architectural invariants / wire / migrations / metrics / security:** untouched — pure test harness.
- **Scope/hygiene:** focused one-file diff; no AI mentions in title/body/commit; three-dot merge-base diff used.

VERDICT: APPROVE
A correct, high-confidence harness fix: swaps a malformed `nft list sets <table>` (which silently yielded `''` and made Gate 2 red on every run since #1808) for the proven `nft list table inet wifihaven` form, wrapped in a bounded poll to absorb the apply race. Restores real coverage rather than discarding it; no blockers.

sameerparekh:
Gate 2 doesn't auto-run on PRs (it's `workflow_call` from Master Router CD on push-to-main, plus `workflow_dispatch`). Dispatched it manually against this branch to validate the fix before merge: [run 27956641308](https://github.com/wifihaven/wifihaven/actions/runs/27956641308) (head 46e6acac, ~44 min). Will report when it completes.

sameerparekh:
Gate 2 validation run [27956641308](https://github.com/wifihaven/wifihaven/actions/runs/27956641308) did **not** validate the fix — it was killed by host contention, not a test failure. Every scenario **PASSED** until both matrix arms got an external `Terminated` (SIGTERM/exit 143) at ~13:58, only 27%/42% in (on `test_extra_blocked.py`) — it never even reached `test_v6_usage_attribution`. No `AssertionError`/FAILED anywhere.

Cause: a real **Master Router CD** run on `main` ([27957348273](https://github.com/wifihaven/wifihaven/actions/runs/27957348273)) started its own Gate 2 arms at 13:56:53/13:57:48 and grabbed the shared self-hosted KVM host's runners/bridges, evicting my ad-hoc `workflow_dispatch` run. (That main run is on the still-broken test, so its Gate 2 will fail on the deterministic assertion — expected until this PR merges.)

Will re-dispatch Gate 2 on this branch once the host is clear, and report the result.

Issue #1848, opened 2026-06-22
ws-transport: agent websocket sidecar (wifihaven-ws) + HTTP fallback (gated on spike)


sameerparekh:
Part of the websocket-transport epic — see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §0.1, §3, §5. **Gated on the Lua-websocket spike** (sub-issue D0).

The agent's websocket lives in its **own procd sidecar** (`wifihaven-ws`), alongside the existing `dns-tail`/`sni-tail`/`nflog-tail` instances (`openwrt/files/etc/init.d/wifihaven`). It owns the socket event loop; the main conntrack-driven agent loop is unchanged.

**Scope:**
- New `wifihaven-ws` sidecar: open `wss://…/api/router/ws` with the bearer token in the upgrade `Authorization` header; run `hello`→`ready`→apply first pushed `policy`; reconnect with exp backoff + jitter (§5.1).
- **Bridge over the existing tmpfs spools** — drain the usage retry queue + events batcher spool *out* to frames; write pushed `policy` frames *in* to the snapshot file the main agent already reads/applies (etag-guarded apply, unchanged). **No new /tmp spool** — inherits the existing bounded caps (`docs/process/router-agent-bounded-writes.md`).
- HTTP fallback wiring (§3.1): UCI flag (default off until proven); on upgrade failure / >`ws_fallback_after` disconnect, the main agent resumes HTTP polling.
- Heartbeat `ping`/`pong` (§5.5); split an oversized `UsageReport` across multiple `usage` frames (the #1017 fix, §5.3).
- Agent metrics folded into the existing `/metrics` push: `ws_connect_total{result}`, `ws_state`, `ws_fallback_total{result}`, `ws_frames_{sent,recv}_total{op}` + panel.

Independently shippable behind the default-off UCI flag; main agent HTTP path unchanged (back-compat).

Relates to #1023.

Issue #1845, opened 2026-06-22
ws-transport: spike — prove a Lua websocket library is viable on OpenWRT (gates agent client)


sameerparekh:
**Spike (gates the agent-side websocket work).** Part of the websocket-transport epic — see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §3.4.

The OpenWRT agent has no async event loop (single foreground Lua process driven by `conntrack -E`, all I/O is `curl` shell-out). A persistent bidirectional websocket needs a real event loop in a sidecar process. Before committing any agent ws code, prove the transport is viable.

**Answer:**
- Which Lua websocket library is in the OpenWRT package feed and small enough for the router image? Candidates: `lua-websockets` (pure-Lua + LuaSocket), a `libwebsockets` C-binding, or a hand-rolled framing layer over the `lua-openssl`/`libustream-mbedtls` TLS socket the agent already depends on (`openwrt/Makefile` DEPENDS).
- Can it hold a TLS ws connection open for hours under OpenWRT's memory budget without leaking? (long-soak on real hardware / qemu).
- Does it cleanly surface "connection dropped" for reconnect?
- Confirm Render's per-plan ws idle-timeout + max-frame-size limits (pins heartbeat cadence + `ws_max_frame_bytes`).

**If this fails**, the epic ships the server endpoint + handshake (A, B) only and the agent stays on HTTP polling. No prod code lands from this issue — it produces a recommendation comment on the epic.

Relates to #1023.

sameerparekh:
Spike done — PR #1863 merged with verdict **GO**: build the agent ws client on cqueues + luaossl + hand-rolled RFC 6455 (no packaged Lua ws lib; both deps in OpenWrt feeds). Findings in docs/design/ws-lua-spike-findings.md. Next: #1846 (API ws endpoint).

Issue #1023, opened 2026-05-25
Migrate router↔API transport to a persistent websocket (push usage + events, push policy on change)


sameerparekh:
## Motivation

The router↔API transport is currently three separate HTTP endpoints:

- `POST /api/router/usage` — every ~60 s, batched bucket of per-flow counters
- `POST /api/router/events` — periodic batches of connection_attempt / dhcp_lease / first_seen_mac
- `GET  /api/router/snapshot` — polled every ~5 s for policy changes

This shape has cost us:

- **#1017**: usage POST bodies grew past zio-http's 100 KiB request cap as `mac_ip_tracking` filled; every bucket started 400-ing for hours before anyone noticed. Same ceiling is now starting to bite `POST /api/router/events` too.
- **#866 / #913 cascade**: ingest-vs-read-vs-agent disambiguation is hard because there's no single "is the router-API link healthy?" signal — each endpoint can fail independently and the operator only finds out when the SPA goes dark.
- **Snapshot polling latency**: a 5 s poll means policy edits (a kid's screen-time +Time, a new block) take up to 5 s to take effect.
- **Batch tax**: usage data is artificially aggregated into 60 s buckets because that's the natural cadence of an HTTP POST. We then re-bucket it for display. Lower-cadence updates are gratuitous.

## Proposal — one persistent websocket per router, REST payloads unchanged

**The JSON payloads do not change.** Today's `UsageReport`, `RouterEventsRequest`, and snapshot bodies stay exactly as they are. The only change is the carrier: instead of three HTTP endpoints with three queues and three retry stories, everything rides one bidirectional websocket. Each frame carries an `op` discriminator naming the REST endpoint it would have hit.

**Frame envelope** (both directions):

```json
{"op": "usage",  "payload": <existing UsageReport JSON> }
{"op": "events", "payload": <existing RouterEventsRequest JSON> }
{"op": "policy", "payload": <existing snapshot JSON> }
```

Server demuxes on `op` and dispatches into the same `handleUsage` / `handleEvents` / snapshot-build code paths used today. No new validation, no new shapes, no new wire contracts — `UsageReport` / `RouterEventsRequest` / snapshot stay as the source of truth.

**Behavior shifts** (independent of payload shapes):

[...]

Issue #1846, opened 2026-06-22
ws-transport: API /api/router/ws endpoint + {op,payload} demux + connection registry


sameerparekh:
Part of the websocket-transport epic — see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §6.1, §1.

Add `RouterWsRoutes.scala` mounting `GET /api/router/ws` (zio-http 3.0.1 `Handler.webSocket`). Additive — REST endpoints stay fully live.

**Scope:**
- Auth the upgrade via the **existing** `RouterAuth.authenticate` (bad token → 401, no 101). No new auth surface (§4.1).
- Decode the `{op, payload}` envelope (§1.2) and demux on `op` into the existing ingest handlers. **Extract a transport-agnostic ingest service from `RouterIngestRoutes`** (`handleUsage`/`handleEvents`/metrics) with no behavior change, so REST and ws share one implementation (single-source-of-truth).
- Connection registry (`RouterWsRegistry`, `Ref[Map[RouterId, Set[Channel]]]` or per-router `Hub`) keyed by `RouterId`. Single-process fan-out (household model; multi-tenant pub/sub explicitly out of scope).
- Per-frame structured logging (`op`/`router_id`/`result` on the MDC, mirroring `LogContext`) to retain the per-POST debuggability #1017/#1334 relied on.
- Server metrics (§7): `router_ws_connections_active`, `router_ws_frames_total{op,dir,result}`, `router_ws_policy_push_total{result}` + Grafana panel.
- Touch `routers.last_seen_at` on frames/heartbeats (uniform liveness with the REST path).

Independently shippable, back-compat (new route only). Exercisable by a test ws client before the Lua client exists.

Relates to #1023.

PR #1863, opened 2026-06-22
spike(#1845): Lua websocket client viability on OpenWrt — GO (cqueues + RFC6455)

sameerparekh:
## D0 spike — gates the agent-side websocket work (#1848) of epic #1023

Proves a Lua websocket **client** is viable on OpenWrt before any transport code lands. See `docs/design/websocket-transport.md` §3.4.

### Verdict: **GO**
Build the agent ws client on **`cqueues`** (the async event loop + TLS socket the agent lacks) **+ `luaossl`** (TLS context / SHA-1 / CSPRNG) **+ a hand-rolled RFC 6455 client framing layer**.

- `cqueues` and `luaossl` are both in the **official OpenWrt feeds for Lua 5.1**, on the 23.05 `.ipk` generation and the snapshot/25.12 `.apk` generation. `cqueues` ~131 KiB; both link the platform OpenSSL/mbedTLS the image already ships.
- **No packaged Lua websocket *library* exists** (`lua-http`/`lua-websockets` are absent from both feeds), so we hand-roll the small/stable RFC 6455 *client* framing instead of vendoring `lua-http`'s six-package dependency tree — mirroring how `quic.lua` hand-rolls QUIC/TLS over injected crypto.
- `cqueues` directly solves the design's dominant constraint (§0.1): the agent has **no event loop**. The sidecar's controller runs the ws client; the main agent's `conntrack` path is untouched.

### What's in this PR (spike code — throwaway, NOT shipped to the router)
`openwrt/spike/ws-1845/`:
- `ws_frame.lua` — pure RFC 6455 framing (encode/decode/mask/handshake-key), no I/O, dependency-injected crypto, loads under Lua 5.1 and 5.5+.
- `ws_crypto.lua` — SHA-1/Base64/random with `pure()`, `luaossl()`, `openssl()` backends.
- `ws_client.lua` — cqueues wss client: connect → TLS → handshake → send/recv → ping/pong → clean drop detection.
- `echo_server.lua` / `run_echo_server.lua` / `poc_echo.lua` — controlled loopback peer + runnable end-to-end demo.
- `openwrt/test/ws_frame_spec.lua` — busted unit tests, **wired into `run_tests.sh`** (the only piece that runs in CI; pure, no native deps).

### Validation
- **17/17 busted tests** against RFC 6455 §1.3 (handshake accept-key) and §5.7 (masked/unmasked frame byte vectors), plus extended-length + partial-buffer decode.
- **Interop with `websocat` (independent Rust impl), both directions:** it accepted our `Sec-WebSocket-Key` + decoded our masked text frame as a server, and a `websocat` *client* round-tripped fully through our `echo_server.lua`.
- TLS connect + handshake via cqueues + luaossl.

### Open item (folded into #1848's first task)
The Lua client's **reactive receive loop** needs a green run on a real OpenWrt/Linux target with the *packaged* cqueues (epoll). The macOS *luarocks*-built cqueues + brew-openssl in the dev sandbox has a kqueue/timeout poll artifact (immediate `ETIMEDOUT` on reactive reads even after `poll` signals ready) — the raw socket I/O is fine (proven by the websocat round-trip), so this is a dev-build artifact, not a protocol/library blocker. Also: a multi-hour TLS soak (RSS/fd leak watch) and a real `wss://api.wifihaven.net` run to pin Render's heartbeat/max-frame limits (design §5.5/§6 — 30 s heartbeat confirmed comfortably under Render's idle behavior).

Full findings + per-feed availability table + recommendation for #1848: **`docs/design/ws-lua-spike-findings.md`**.

Relates to #1023.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -444,7 +444,7 @@ jobs:
           sudo apt-get update -q
           # lua5.1 matches OpenWrt 23.x (LuaJIT is 5.1-compatible).
           # liblua5.1-dev is needed for luarocks to compile C modules.
-          sudo apt-get install -y lua5.1 liblua5.1-dev luarocks
+          sudo apt-get install -y lua5.1 liblua5.1-dev luarocks lua-cqueues lua-luaossl
           sudo luarocks install busted
           sudo luarocks install lua-cjson
 
@@ -453,7 +453,7 @@ jobs:
         run: |
           # Mirror the full busted invocation from test/run_tests.sh so the
           # whole Lua suite gates merges, not just conntrack_spec.
-          LUA_PATH="./files/usr/lib/lua/?.lua;./files/usr/lib/lua/wifihaven/?.lua;./test/shim/?.lua;./test/shim/?/init.lua;$(lua -e 'print(package.path)')" \
+          LUA_PATH="./files/usr/lib/lua/?.lua;./files/usr/lib/lua/wifihaven/?.lua;./spike/ws-1845/?.lua;./test/shim/?.lua;./test/shim/?/init.lua;$(lua -e 'print(package.path)')" \
             busted test/conntrack_spec.lua \
                    test/render_spec.lua \
                    test/policy_spec.lua \
@@ -471,7 +471,12 @@ jobs:
                    test/nft_drops_spec.lua \
                    test/dns_tail_sets_spec.lua \
                    test/nflog_spec.lua \
-                   test/host_norm_spec.lua
+                   test/host_norm_spec.lua \
+                   test/ws_frame_spec.lua
+
+      - name: Run ws-client e2e (#1845, real Lua 5.1 + cqueues/luaossl)
+        working-directory: openwrt
+        run: sh spike/ws-1845/e2e_test.sh
 
   # ── Bidirectional contract tests (#634) ──────────────────────────────────
   # One job, one green/red signal for "the API ↔ router wire contract holds".
--- a/docs/design/ws-lua-spike-findings.md
+++ b/docs/design/ws-lua-spike-findings.md
@@ -15,9 +15,13 @@ Build the agent ws client (#1848) on **`cqueues` (event loop + TLS socket) +
 layer**. Do **not** adopt `lua-http`/`lua-websockets` — they are not in the
 OpenWrt feed and pull a heavy dependency tree we don't need.
 
-One caveat carried into #1848's first task: the **reactive receive loop** still
-needs a green run on a real OpenWrt/Linux target (the macOS dev build has a
-poll/timeout I/O quirk — §5). Everything else is proven.
+The full client — including the **TLS (wss) reactive receive loop and
+heartbeat** — is validated **end-to-end on a real Lua 5.1 + distro-packaged
+cqueues/luaossl target** (Ubuntu 24.04, `lua5.1` 5.1.5 + `lua-cqueues`
+`20200726` + `lua-luaossl` `20220711` — the same package versions the OpenWrt
+feed ships), not just on the dev host. That validation **found and fixed three
+real Lua-5.1/cqueues bugs** the macOS dev build had masked (§5.1) — which is
+exactly why hardware validation was a gate, not a formality.
 
 ---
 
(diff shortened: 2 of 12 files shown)


sameerparekh:
**Independent PR review** (read-only, adversarial) — first run.

Diff is isolated **spike** code under `openwrt/spike/ws-1845/` (pure-Lua framing + cqueues client/echo POC + findings doc) plus one wiring change to `openwrt/test/run_tests.sh`. No Scala, no wire/snapshot change, no migration, no metrics surface, no `/tmp` writes — so dimensions 3 (router-applier/bounded-tmp), 4 (wire contract), 5 (migrations), 6 (metrics) are N/A by construction.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `openwrt/spike/ws-1845/ws_client.lua:174` — `err == socket.EPIPE` for the EOF/`"eof"` drop-reason: `cqueues.socket` typically does not export `EPIPE` (errno constants live under `cqueues.errno`), so `socket.EPIPE` is likely `nil` and this branch rarely matches — most clean disconnects get labeled `"read: <errno>"` instead of `"eof"`. Only affects the drop-reason *string* (not reconnect behavior, which keys off `closed=true`), and this is throwaway spike code, but worth fixing when `ws_client.lua` is promoted into the agent in #1848.
2. `openwrt/spike/ws-1845/ws_crypto.lua:158` (`weak_rand`) and `poc_echo.lua:31` (backoff jitter) use `math.random` without `math.randomseed` → deterministic sequence per run. Both are explicitly documented as non-cryptographic (fallback rand / demo jitter; the production nonce uses `luaossl`'s CSPRNG), so this is fine for the spike — noting only so it isn't carried into prod rand by accident.
3. `bxor` is implemented twice (`ws_frame.lua` `bxor8`, `ws_crypto.lua` `bxor32`) by arithmetic. Acceptable here (different widths, both pure, both vector-tested, and the module split keeps `ws_frame` I/O-free), but if these graduate into the agent consider a shared bit-helper. Mirrors the existing `quic.lua` pattern, so not a divergence risk today.
4. `openwrt/spike/ws-1845/echo_server.lua` carries `WS_ECHO_DEBUG`-gated `dbg()` prints — intentional and env-gated in a test fixture, not stray debug logging; flagged only for completeness.

**Dimension notes (positives):**
- **SSOT (dim 1):** crypto/random are dependency-injected into `ws_frame` and the framing has no I/O, so there's no display-vs-enforcement or recompute-vs-primitive drift. Hand-rolled framing-over-cqueues is the deliberate choice (no packaged ws library), documented in the findings.
- **Testing (dim 2):** `ws_frame_spec.lua` asserts real RFC 6455 §1.3/§5.7 byte vectors (17/17) and is wired into `run_tests.sh`; no existing test weakened or deleted. The `run_tests.sh` LUA_PATH addition is additive (prepends the spike path) and harmless to the other specs.
- **Scope/hygiene (dim 7):** focused under `spike/`; the only non-spike edit (`run_tests.sh`) directly serves the added test. No `TODO` without an issue.
- **Security (dim 8):** TLS client context sets `VERIFY_PEER`; no secrets, no SQL, no new authed route.

VERDICT: APPROVE
A self-contained D0 spike that lands a GO recommendation (cqueues + luaossl + hand-rolled RFC 6455 client) backed by RFC-vector unit tests and two-directional interop against an independent impl (websocat). No prod code path is touched and the framing tests gate in CI; the only substantive follow-up — validating the reactive receive loop on real hardware — is correctly carried into #1848 rather than this spike. NITs are non-blocking and scoped to code that will be re-touched when promoted.

sameerparekh:
**Independent PR review** — re-run on push `879b388` (delta from `816e665`: `ws_client.lua` recv hardening only).

**Prior-findings status**

| --- | --- | --- |
| NIT 1 | `ws_client.lua` EOF label compared against likely-nil `socket.EPIPE` | **ADDRESSED** — now `require("cqueues.errno")` and a would-block (`EAGAIN`/`ETIMEDOUT`/nil) returns a retryable `"timeout"`; only a genuine error sets `closed=true` and returns `"eof: …"`. Better than the original ask: it also prevents a false teardown of a live socket on a spurious poll wakeup. (`ws_client.lua:170-181`) |
| NIT 2 | unseeded `math.random` in `weak_rand` / backoff jitter | NOT-ADDRESSED (by design — documented non-crypto; left intentionally) |
| NIT 3 | `bxor` duplicated across `ws_frame`/`ws_crypto` | NOT-ADDRESSED (by design — accepted, mirrors `quic.lua`) |
| NIT 4 | `WS_ECHO_DEBUG`-gated prints in `echo_server.lua` | NOT-ADDRESSED (by design — intentional fixture instrumentation) |

NITs 2–4 were explicitly non-blocking "by design" notes, not change requests.

**New findings in the delta**
(none) — the change is a localized, more-correct error classification using `cqueues.errno`; `luac -p` clean, framing spec still 17/17, no new path/`/tmp`/wire/test-weakening introduced.

VERDICT: APPROVE
The recv hardening cleanly addresses the one substantive NIT and tightens the §3.4 drop signal the spike exists to prove. No blockers across either run; spike remains self-contained with CI-gated framing tests. Hardware validation of the receive loop stays correctly deferred to #1848.

sameerparekh:
**Independent PR review** — re-run on push `55f8267` (delta from `879b388`).

This push is the result of the operator-requested **hardware validation on a real Lua 5.1 + distro-packaged cqueues/luaossl target** (Ubuntu 24.04, epoll). That run was not a formality — it found and fixed **three real Lua-5.1/cqueues bugs the macOS dev build had masked**, then proved `poc_echo.lua` PASS end-to-end over both ws:// and wss:// (real TLS). Strong outcome: the spike's central viability question is now settled on the actual interpreter/library the agent will run.

**Prior-findings status**

| --- | --- | --- |
| NIT 1 | `ws_client.lua` EOF label vs likely-nil `socket.EPIPE` | **ADDRESSED** — `recv` now classifies via `cqueues.errno` (`ETIMEDOUT`/`EAGAIN` → benign `"timeout"`, else `"eof: …"`); `socket.EPIPE` no longer referenced. |
| NIT 2 | unseeded `math.random` (`weak_rand` / jitter) | NOT-ADDRESSED (by design — documented non-crypto) |
| NIT 3 | `bxor` duplicated across `ws_frame`/`ws_crypto` | NOT-ADDRESSED (by design — accepted, mirrors `quic.lua`) |
| NIT 4 | `WS_ECHO_DEBUG` prints in `echo_server.lua` | NOT-ADDRESSED (by design — fixture instrumentation) |

**New findings in the delta**

BLOCKERS — (none)

SHOULD-FIX — (none)

NITS
1. `run_echo_server.lua:8-13` — `io.open(arg[3]):read(...)` for the TLS cert/key has no nil-guard; a missing/bad path throws an opaque error. Fine for a spike test fixture (operator runs it with known paths), noting only.

**Correctness review of the three fixes (delta dimensions 1-3):**
- **Header CR-strip ordering** (`ws_client.lua:90-106`) — strips `\r` *before* the blank-line test and breaks on `not line`; this stops the extra `*l` that previously read past end-of-headers and poisoned the buffer. Correct, and matches what `echo_server` already did. Root cause (cqueues `*l` strips only LF) is real and well-explained in-comment.
- **`clearerr` on benign timeout** (`ws_client.lua:177-184`) — `clearerr("r")` after `ETIMEDOUT`/`EAGAIN` prevents cqueues' preserved-error accumulation from aborting the controller during idle heartbeats. Drop signal stays precise (only non-timeout errors set `closed=true`). Correct.
- **Returning `onerror` + no `pcall` around yielding I/O** (`ws_client.lua:52-69`) — the Lua-5.1-correct fix for "yield across a C-call boundary": `onerror` returns `nil,errno` for all ops, so `starttls`/`read`/`write` are called directly and checked by return value. `send_text` already keys off `not ok`, so write failures are handled. Correct and the right idiom for the target.
- No test-weakening, no `/tmp` writes, no wire/migration/metric surface; `ws_frame_spec` still 17/17 (CI). Docs (`findings §5/§5.1`, README) accurately reclassify the former "open item" as closed with the bugs documented for #1848 to inherit.

VERDICT: APPROVE
Hardware validation closed the one substantive open item and, in doing so, caught three genuine Lua-5.1/cqueues defects that the dev sandbox hid — all fixed and re-validated over ws:// and wss://. No blockers across any run. The only remaining work (multi-hour soak + real `wss://api.wifihaven.net`) genuinely depends on the #1846 server endpoint and is correctly scoped to #1848, not this spike.

sameerparekh:
**Independent PR review** — re-run on push `207e6f5` (delta: ws-client e2e test + CI wiring).

This push adds the e2e exploration: `spike/ws-1845/e2e_test.sh` drives the real `ws_client.lua` through full ws:// + wss:// round-trips against `echo_server.lua` and is wired into CI on the real Lua 5.1 + packaged cqueues/luaossl stack. It also fixes a genuine gating gap it surfaced.

**Prior-findings status**

| --- | --- | --- |
| (55f826 NIT 1) | `run_echo_server.lua` cert/key open has no nil-guard | NOT-ADDRESSED (by design — spike test fixture, known paths). The e2e harness now always passes valid generated paths, so the practical risk is nil. |
| Earlier NIT 2/3/4 (rng / bxor dup / debug prints) | — | NOT-ADDRESSED (by design, as previously stated) |

No prior BLOCKER/SHOULD-FIX outstanding.

**New findings in the delta**

BLOCKERS — (none)

SHOULD-FIX
1. `.github/workflows/ci.yml` (OpenWrt Lua Tests) vs `openwrt/test/run_tests.sh` — this push had to add `ws_frame_spec.lua` to *both* lists because they are two hand-maintained copies that have **drifted** (CI omits `sni`/`quic`/`lan_warn`/`eb_refresh`/`static_ip_labels`/`host_metrics`/`jsonc_shim` busted specs and *all* `sh test/*.sh` shell specs). That means several existing Lua tests do not gate merges. This is the single-source-of-truth smell the checklist (dim 1) flags. Pre-existing, not introduced here, and the author correctly documented it as a follow-up in the findings doc rather than expanding the spike — acceptable to defer, but it should get a tracking issue (collapse CI onto `run_tests.sh`). Calling it SHOULD-FIX-as-follow-up, not merge-gating for this PR.

NITS
1. `e2e_test.sh` derives ports from `$$` (`20000 + $$%20000`); two concurrent runs on one host could in principle collide. Fine for CI (one runner) and serial local runs; noting only.

**Dimension checks:**
- **Testing integrity (dim 2):** the e2e adds real coverage (asserts on `RESULT: PASS` + exit code over both transports) rather than just exercising; it self-skips cleanly where deps are absent so it can't produce false greens by silently no-op'ing on the dev host (it prints an explicit SKIP). No existing test weakened. The `ws_frame_spec` CI addition *increases* gating. Validated on plex.lan (both cases PASS; deterministic on re-run) and confirmed SKIP on macOS.
- **Scope/hygiene (dim 7):** focused; `actionlint` + `shellcheck` clean. CI deps (`lua-cqueues`/`lua-luaossl`) are the same packages the OpenWrt feed ships, so CI tests the real stack.
- No `/tmp` unbounded writes (harness uses `mktemp -d` + EXIT-trap cleanup), no wire/migration/metric surface, no secrets (self-signed throwaway cert generated per-run).

VERDICT: APPROVE
The e2e harness turns the one-off manual hardware validation into an automated, CI-gated regression test on the exact Lua 5.1 + cqueues/luaossl stack — directly hardening against the three bug classes found earlier. The surfaced CI-vs-run_tests.sh list drift is a real pre-existing issue, correctly documented as a follow-up rather than scope-creeping the spike; no blocker. Recommend filing the drift-reconciliation issue when this lands.

Issue #1866, opened 2026-06-22
perf(router): dns-tail re-serializes the whole DNS cache to /tmp on every query line (O(N) per line)


sameerparekh:
Follow-up from #1864 (the agent-CPU / DNS-starvation fix). #1864 isolates dnsmasq from agent CPU via `nice` and throttles `wh_drop` logging, but it deliberately does NOT touch the agent's ~28%-steady-state CPU draw. This issue tracks the clearest separable efficiency win found during that work.

## Finding
`openwrt/files/usr/sbin/wifihaven-dns-tail` flushes with `flush_every = 1` (one flush per ingested line). Each flush re-serializes the **entire** in-memory cache and rewrites three tmpfs files atomically (temp write + `os.rename`):

- `atomic_write(cache_path, cache.dump_text())` — full ip→hostname cache
- `atomic_write(aliases_path, cache.dump_aliases_text())` — full CNAME-alias map
- `atomic_write(dns_metrics_path, ...)` — small

So per DNS query line the cost is **O(N)** in the cache size, not O(1) in the new entry. On a busy household the dnsmasq query log reaches ~14 MB / 10-min window across ~22 devices (per #1864 recon), and the cache grows to hundreds–thousands of live entries — so dns-tail does roughly O(N) serialization × O(L) lines of work continuously. This is a prime suspect for the agent stack's steady-state CPU.

See `wifihaven-dns-tail` lines ~322-329 and ~372-385 (two mirrored flush sites: the SNI-routed path and the DNS-reply path).

## Constraint (do not regress)
The per-line flush exists for a real correctness reason (#472): the agent's conntrack `-E NEW` watcher reads `cache_path` at flow-start, **milliseconds** after the dnsmasq reply line that resolved the destination IP. If the new ip→host mapping isn't on disk by then, the `connection_attempt` event is built with `hostname=nil` and that nil is frozen onto the event (a later flush can't repair it). So a naive "flush every K lines" or "flush every K ms" batching window would reintroduce nil-hostname attribution for flows that start inside the window.

## Possible directions (design TBD)
- **Incremental/delta writes** instead of full re-serialization: append only the new/changed ip→host (and alias) entries to the snapshot file, so per-line cost is O(1) in the new entry while keeping the mapping on disk before the SYN's conntrack NEW lands. The agent reader would need to tolerate an append-structured / log-compacted file (and periodic full-rewrite compaction, bounded per AGENTS.md#bounded-tmp-writes).
[...]

Issue #1826, opened 2026-06-22
bug(router): schedule-blocked device retry-storm pegs CPU — nftables wh_drop logging is not rate-limited


sameerparekh:
## Incident (router.lan / 192.168.10.1, observed 2026-06-22 ~06:00 MDT)
router.lan CPU spiked at 6am MDT. Live recon:
- Load avg **4.3**; `wifihaven-agent` **26% CPU** and `wifihaven-dns-tail` **24% CPU**, both pinned in **R** state. 30% sirq.
- Agent version **0.3.12-r1 — unchanged** (no new release since 06-15; router CD red). NOT a version regression.
- Kernel log (`logread`) is being **flooded with `wh_drop:<mac>:Schedule` lines, dozens/sec** (timestamps stepping ~25 ms).
- `/tmp/wifihaven-dnsmasq.log` = **15 MB within a single 10-min rotation window** (~1.5 MB/min of DNS logging).

## Root cause
At 6am a **Schedule block engaged** for two devices — MACs `26:74:fc:f9:4e:9e` (192.168.10.195) and `04:72:ef:d6:e4:5a` (192.168.10.159) (kids' devices on a morning/bedtime schedule). Those devices **retry-storm** HTTPS/QUIC (Apple push/iCloud/relay endpoints, TCP+UDP 443, v4 and v6). Every dropped packet hits the nftables drop rule, which renders as:

`render.lua:1109` → `… log prefix "wh_drop:<mac>:<reason> " counter drop …` — **with NO `limit rate` clause.**

So a retry storm = one kernel-log line per dropped packet = a kernel-ring-buffer flood, plus a flood of DNS re-resolutions feeding `dns-tail`. Result: `wifihaven-agent` + `wifihaven-dns-tail` pegged. The drop is working correctly; the **unbounded LOG action** is the defect. (Also brushes the bounded-/tmp-writes rule, AGENTS.md#bounded-tmp-writes — the dnsmasq.log outruns the 10-min rotate under flood.)

## Fix (primary)
[...]

Issue #472, opened 2026-05-16
VM e2e: connection_attempt events stored with hostname=None — DNS attribution broken


sameerparekh:
## Symptom

Run https://github.com/sameerparekh/familydns/actions/runs/25951502974/job/76290206908

```
FAILED scenarios/test_02_allowed_browsing.py::test_allowed_request_succeeds_and_event_recorded
    TimeoutError: timed out waiting for allowed connection_event for example.com
```

## Root cause

Events ARE being posted (200 OK, batches of 1–14 every few seconds) and stored. But every stored row has `hostname=None`:

```
events for mac 02:e2:e2:6d:c9:a1 (15 total):
  ts=…03:29:43+00 allowed=True reason='allow' hostname=None destIp='73.185.182.209'
  ts=…03:29:43+00 allowed=True reason='allow' hostname=None destIp='23.186.168.131'
  ts=…03:29:43+00 allowed=True reason='allow' hostname=None destIp='172.234.37.140'
  ts=…03:28:38+00 allowed=True reason='allow' hostname=None destIp='192.168.100.1'
  …
```

`test_02` filters for `host="example.com"` and times out because no event with that host exists.

The agent has a `dns-tail` daemon that watches `/tmp/familydns-dnsmasq.log` and writes `/tmp/familydns-dns-cache.txt`:

```
03:25:38 user.info familydns: dns-tail: starting source=/tmp/familydns-dnsmasq.log cache=/tmp/familydns-dns-cache.txt
03:26:42 user.info familydns: conntrack: starting watcher lan_prefix=192.168.100. max_batch=50 flush_interval=10s
```

That dns→ip cache is what conntrack consults to attribute hostnames. Either:
1. dnsmasq isn't logging queries to `/tmp/familydns-dnsmasq.log` in this image (logging not enabled in `/etc/dnsmasq.d/…`).
2. `dns-tail` isn't parsing entries correctly.
3. `conntrack` watcher isn't reading the cache before posting events.
4. The DNS attribution is by-MAC and the lookup misses on a race (DNS resolved before MAC was seen, cache pruned).

This affects every test that asserts on hostname, but #03 (blocked_domain) and #05 (usage_in_api) passed — they may match by IP or by allow/block reason without needing hostname. So the impact is narrower than it looks, but the agent capability is real.

## Verification

[...]

PR #1867, opened 2026-06-22
fix(#1864): nice agent below dnsmasq + bound wh_drop logging to stop DNS starvation

### LINKED ISSUE #1864: bug(router): agent/dns-tail CPU starves dnsmasq under load → transient DNS failures for ALL devices (not a block)
## Symptom (prod, observed 2026-06-22)
Intermittent connectivity flakiness for devices on the busy **prod router** (router.lan / 192.168.10.1) — operator's work computer couldn't reach `calendar.google.com`, Chrome showed broken page elements, then it self-recovered in minutes. The **test router is fine** (it's quiet). This presents as a transient **DNS resolution failure**, NOT a policy block.

## Confirmed NOT a block
Live recon on the affected devices (Sameer Desk `c8:4b:d6:b3:65:c9`, Sameer Work Mac `86:98:77:ad:88:52`, profile 2): not in `@blocked_macs`, no `wh_drop` entries, `blockIpOnly` disabled (0 rules), no block-page DNAT, no unmanaged-MAC default-block, DNS resolves, and Sameer Desk had healthy ESTABLISHED flows during the recon. WifiHaven enforcement is doing nothing to these devices.

## Root cause
The agent runs **hot** under real household traffic and is **not CPU-isolated from dnsmasq**:
- `wifihaven-agent` sits at **~28% CPU steadily** (R state) even at rest, processing a firehose (`/tmp/wifihaven-dnsmasq.log` reaches ~14 MB per 10-min rotation window across 22 devices); `wifihaven-dns-tail` adds more.
- Under a burst — e.g. the schedule-block retry-storm in #1826 — agent + dns-tail spike and load hit **4.3**. dnsmasq shares the same CPU with no priority guarantee, so it gets **starved** and briefly drops/delays queries.
- Result: every device on the router sees momentary name-resolution failures (broken images, "can't connect"), none are blocked, and it clears when the burst passes. dnsmasq itself is stable (0 restarts) — it's CPU starvation, not a reload.

#1826 (wh_drop log rate-limit, shipped in 0.3.13) cut the biggest burst driver but did NOT make the agent cheap or guarantee dnsmasq CPU — so transient starvation still happens.

## Fix
1. **PRIMARY — isolate DNS from agent CPU.** Lower the scheduling priority of `wifihaven-agent`, `wifihaven-dns-tail`, `wifihaven-sni-tail` below dnsmasq (procd `procd_set_param nice <n>` with positive nice; consider `ionice` too) so DNS resolution always wins CPU under contention. Low-risk; verify dnsmasq stays responsive while the agent is busy.
2. **Observability — stop `wh_drop` flooding the kernel ring buffer.** `logread` is dominated by `wh_drop` lines (even rate-limited), evicting agent/dnsmasq events and making the router undiagnosable (hit this during recon). Route `wh_drop` to a bounded `/tmp` file (rotated, per AGENTS.md#bounded-tmp-writes) instead of / in addition to the kernel ring buffer, or throttle harder. Separate from #1826's per-rule rate-limit.
3. **ASSESS (may be a follow-up) — agent steady-state CPU.** Why is the agent at ~28% CPU continuously on a busy household? If dns-tail/conntrack processing has an efficiency win, file a separate follow-up.

## Validation / host
Router-agent change — needs Gate 2/3 e2e (KVM runner). A macOS chip writes the procd/nice change + any busted unit coverage and relies on the PR's CI; can't run KVM locally. The nice change should also be sanity-checked on a real busy router.

Relates to #1826 (first piece), #1796.

sameerparekh:
Fixes #1864.

## Problem
On the busy prod router, devices saw transient DNS resolution failures (broken page loads, "can't reach calendar.google.com") that self-recovered in minutes — **not** a policy block (confirmed via live recon: not in `@blocked_macs`, no `wh_drop`, `blockIpOnly` off, DNS resolving). Root cause: the `wifihaven-agent` stack runs hot (~28% CPU steady, R state) and **shares the CPU with dnsmasq at equal priority**, so a traffic burst (e.g. the schedule-block retry-storm in #1826) starves dnsmasq and every device on the router briefly loses name resolution. dnsmasq itself is stable (0 restarts) — it's CPU starvation, not a reload.

This is **resource isolation + observability, not a policy change** — enforcement behavior is identical.

## Changes
**1. PRIMARY — process priority (the fix).** `nice` the agent and all three tail sidecars (`dns-tail`, `sni-tail`, `nflog-tail`) to **+10** via `procd_set_param nice`, leaving dnsmasq at its package default (nice 0). procd applies this on (re)start → idempotent, survives respawns. dnsmasq now wins the CPU under contention.

**2. Observability — throttle the `wh_drop` LOG harder** (`10/s burst 20` → `2/s burst 10` per rule). Even rate-limited (#1826), the aggregate across many simultaneously-storming drop rules dominated the shared `logread` ring buffer and evicted agent/dnsmasq lines, leaving the router undiagnosable during an incident. The burst still captures the distinct flows the agent's nflog tail needs to label blocked-flow events (labels dedupe per `(mac, reason, dst)`); the unconditional `counter drop` rule is **unchanged**, so enforcement and drop counts stay exact. Note the agent already consumes `wh_drop` from a **bounded /tmp spool** (`wifihaven-nflog-tail`, self-truncating at 256 KiB) rather than relying on ring-buffer retention — so its read channel is already off the ring buffer; this change reduces the residual ring-buffer pressure on operator diagnosis.

**3. ASSESS → follow-up #1866.** The ~28% steady-state CPU itself is left out of this PR per scope. The clearest separable win: `wifihaven-dns-tail` re-serializes the **entire** DNS cache + alias map to /tmp on **every** ingested query line (`flush_every = 1`) — O(N) per line over the firehose. Filed as #1866 (assigned, in-progress, Epic: Router Agent & Release) with the #472 correctness constraint documented; not implemented here.

## Tests
- `init_spec` asserts all four procd instances carry a **positive** `nice` (count == 4, value > 0).
- `render_spec` pins the throttled LOG rate (≤ 2/s sustained, #1864) and the unchanged split log/drop rule shape.
- New `nflog_tail_bounded_spec` pins the nflog spool cap + in-place truncation (AGENTS.md#bounded-tmp-writes).
- Added thin `.test.sh` shims (`init_spec.test.sh`, `nflog_tail_bounded_spec.test.sh`) so the init `nice` checks and the bounded-write spec gate in CI's `shell-tests` job, not just `run_tests.sh` locally.

All busted + shell specs pass locally (the 4 pre-existing `nft_log_dep_spec` build-ipk/apk failures are on `main` too, untouched here).

## Validation note
Router-agent change. Validated locally via `busted` + shell specs; **Gate 2/3 KVM e2e runs on CI's self-hosted runner — cannot run on macOS**. The `nice` change should also be sanity-checked on a real busy router (`top` showing dnsmasq winning CPU while the agent is busy).

### DIFF
--- a/openwrt/files/etc/init.d/wifihaven
+++ b/openwrt/files/etc/init.d/wifihaven
@@ -5,10 +5,22 @@ USE_PROCD=1
 START=95
 STOP=10
 
+# #1864: nice the agent and its tail sidecars BELOW dnsmasq so DNS resolution
+# always wins the CPU under contention. On a busy household the agent runs hot
+# (~28% CPU steady, R state) and spikes under traffic bursts (e.g. a schedule-
+# block retry-storm, #1826); sharing the CPU with dnsmasq at equal priority let
+# those spikes starve dnsmasq, producing transient name-resolution failures for
+# EVERY device on the router (broken page loads) — none of them actually blocked.
+# dnsmasq keeps its package default (nice 0); we run the wifihaven stack at a
+# positive nice so the scheduler de-prioritises it whenever dnsmasq is runnable.
+# procd applies this on (re)start, so it is idempotent and survives respawns.
+WIFIHAVEN_NICE=10
+
 start_service() {
     procd_open_instance "agent"
     procd_set_param command /usr/sbin/wifihaven-agent
     procd_set_param respawn 3600 5 5
+    procd_set_param nice "$WIFIHAVEN_NICE"
     procd_set_param stdout 1
     procd_set_param stderr 1
     procd_close_instance
@@ -18,6 +30,7 @@ start_service() {
     procd_open_instance "dns-tail"
     procd_set_param command /usr/sbin/wifihaven-dns-tail
     procd_set_param respawn 3600 5 5
+    procd_set_param nice "$WIFIHAVEN_NICE"
     procd_set_param stdout 1
     procd_set_param stderr 1
     procd_close_instance
@@ -39,6 +52,7 @@ start_service() {
         procd_open_instance "sni-tail"
         procd_set_param command /usr/sbin/wifihaven-sni-tail
         procd_set_param respawn 3600 5 5
+        procd_set_param nice "$WIFIHAVEN_NICE"
         procd_set_param stdout 1
         procd_set_param stderr 1
         procd_close_instance
@@ -52,6 +66,7 @@ start_service() {
     procd_open_instance "nflog-tail"
     procd_set_param command /usr/sbin/wifihaven-nflog-tail
     procd_set_param respawn 3600 5 5
+    procd_set_param nice "$WIFIHAVEN_NICE"
     procd_set_param stdout 1
     procd_set_param stderr 1
     procd_close_instance
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -1081,7 +1081,7 @@ function M.nft(snapshot, opts)
 
   -- #1122/#1126: each drop is rendered as a pair of rules sharing one predicate
   -- (see #1826 / `emit_drop` below):
-  --   `<predicate> limit rate 10/second burst 20 packets log prefix "wh_drop:<mac>:<reason> "`
+  --   `<predicate> limit rate 2/second burst 10 packets log prefix "wh_drop:<mac>:<reason> "`
   --   `<predicate> counter drop comment "wh_drop:<mac>:<reason>"`
   -- (pre-#1826 this was a single `… log prefix … counter drop` rule; the LOG is
   -- now on its own rate-limited rule so a retry storm can't flood the kernel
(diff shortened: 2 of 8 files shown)


sameerparekh:
## Independent PR review — #1867 (first run)

Read-only adversarial pass over `git diff origin/main...HEAD` against the 8-dimension checklist. Scope: OpenWRT agent resource isolation (procd `nice`) + `wh_drop` LOG throttle + tests. No enforcement/wire/DB/route changes.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. **`openwrt/files/etc/init.d/wifihaven` — dns-tail/nflog-tail are niced too, not just the agent.** Correct for this fix (we want *dnsmasq* to win, and all wifihaven procs stay equal-priority relative to each other), but it means under saturation `wifihaven-dns-tail` can also be descheduled — and it owns the ip→host cache that the agent's conntrack `NEW` watcher reads ~ms after the dnsmasq reply (#472). The trade-off is sound (transient nil-hostname attribution ≪ DNS failure for every device), but the real-busy-router sanity check the PR already calls for should explicitly confirm hostname attribution still lands under load, not just that dnsmasq wins CPU.
2. **`openwrt/test/render_spec.lua:933` — the new `≤ 2/second` guard checks only the rate field, not `burst`.** Adequate because the two exact-string assertions (`render_spec.lua:816`, `:883`, `:2636`) already pin `burst 10`; noting only that the dedicated #1864 guard alone wouldn't catch a burst regression.
3. **History — value change + tests landed in one commit (no separate red→green).** Acceptable for a threshold/config change that's fully pinned by the updated + added assertions, but the checklist prefers a visible failing-test commit for autonomous sessions.

### Dimension notes
- **1 SSOT:** `WIFIHAVEN_NICE` is a single shared var across all 4 instances; `DROP_LOG_LIMIT` single-sourced. No re-derivation, no new concept fields. ✓
- **2 Testing:** The `10/second`→`2/second` edits in `render_spec.lua` track an *intentional* value change (not a weakened/loosened assertion) and are matched by a new lower-bound guard; the split log/drop-rule shape and "every LOG line is rate-limited" assertions are unchanged. New `init_spec` nice checks (count==4 + positive) and new `nflog_tail_bounded_spec` (cap + in-place truncate). No mocks. ✓
- **3 Invariants:** No decision logic added to the router; drop verdict (`counter drop`) byte-for-byte unchanged, so enforcement is identical. No new `/tmp` grow-writer (the nflog spool was already self-truncating; the throttle *reduces* its write rate); bounded-write coverage added. ✓
- **4 Wire / 5 Migrations / 8 Security:** No wire-shape, snapshot, migration, route, SQL, or secret changes. ✓
- **6 Metrics:** No new code path → no new metric required.
- **7 Scope:** Focused to `openwrt/`; no drive-by edits; no dead code. ✓
- **CI gating:** `.test.sh` shims make the init `nice` checks and bounded-write spec run in the `shell-tests` job; `render_spec.lua` runs in `lua-tests`. (Pre-existing `nft_log_dep_spec` build-ipk/apk failures exist on `main` too and are untouched here — not auto-discovered by CI.)

VERDICT: APPROVE
Resource-isolation + observability change: `nice` the agent stack below dnsmasq so DNS wins CPU under contention, and throttle `wh_drop` logging so it stops evicting agent/dnsmasq lines from the shared ring buffer. Enforcement and the drop-rule shape are unchanged; the change is well-tested at the unit/shell level. KVM Gate 2/3 e2e is CI/self-hosted only and the `nice` effect needs a real-busy-router sanity check (already noted in the PR).

PR #1868, opened 2026-06-22
fix(#1865): never drop traffic to the block-page listener (+ verify reachability)

### LINKED ISSUE #1865: bug(router): block page is dropped for whole-MAC-blocked devices — drop rules don't carve out the block-page listener
## Principle (operator directive)
**The block page must NEVER be blocked, for any reason.** A blocked device must always be able to reach the local block-page listener so the user sees *why* they're blocked, instead of a broken/timed-out connection.

## Bug (confirmed in code + live logs)
A whole-MAC-blocked device's HTTP/HTTPS gets DNAT'd toward the block-page listener (`render.lua`: `tcp dport 80 dnat ip6 to ::1:8081` / `tcp dport 443 dnat ip6 to ::1:8443`, and v4 `127.0.0.1:8081/8443`), but the **per-MAC blocked drop then kills the redirected packet**. In `openwrt/files/usr/lib/lua/wifihaven/render.lua` (~line 1161), the per-MAC drop is:

```
ether saddr <mac> [ip daddr != @ea_<m>_<a> ...] [ip daddr != @global_allow] counter drop
```

It carves out extraAllowed (`ea_`) and global-allow (`@global_allow`) — but has **NO carve-out for the block-page destination** (`::1`/`127.0.0.1`, ports 8081/8443). So the DNAT'd-to-block-page packet is dropped by the MAC's own block rule.

Live evidence (router.lan): `wh_drop:26:74:fc:f9:4e:9e:Schedule ... DST=::1:8443 ... DPT=443` — a schedule-blocked device's traffic, DNAT'd to the block page, then dropped.

## Fix
1. **Carve out the block-page listener from ALL drop rules** — add an `ip daddr != 127.0.0.1` / `ip6 daddr != ::1` (or the specific :8081/:8443 listeners) exception to the per-MAC blocked drop (and any eb_/bl_/category drop that could catch DNAT'd block-page traffic), analogous to the existing `ea_`/`global_allow` suffixes. The block page wins over every drop.
2. **Verify the v6 DNAT actually delivers.** The live drop showed `DST=::1:8443` with `OUT=eth1` — i.e. the packet was heading to the WAN, not delivered to the loopback listener. DNAT'ing forwarded traffic to `::1` (loopback) may not deliver to local INPUT at all. Confirm the block page is genuinely REACHABLE for a whole-MAC-blocked device end-to-end (Gate 3a-style), not just that the drop is carved out — if DNAT-to-loopback doesn't work for forwarded v6, the listener may need to bind a real LAN/router address instead of `::1`.
3. Add busted coverage asserting the block-page carve-out is present on the drop rules, and (where the harness allows) that a blocked MAC can reach the block page.

## Validation / host
Router-agent change; the block-page reachability is covered by Gate 3a (router staging smoke) on the KVM runner — a macOS chip writes the render.lua change + busted and relies on CI; can't run KVM locally.

Relates to the `::1:8443` drops seen during the #1826 / #1864 investigations.

sameerparekh:
## Problem
**The block page must NEVER be blocked, for any reason.** A whole-MAC-blocked device (paused / schedule / time-limit / manual) could not reach the local block-page listener — it saw a hung connection instead of *why* it was blocked. Two confirmed root causes (see #1865 for the live diagnosis):

1. **Drop carve-out missing.** The block-page DNAT chain (`wifihaven_block_nat`, prerouting) rewrites a blocked device's HTTP/HTTPS to the local listener, but the per-MAC / `eb_` / `bl_` / `blockIpOnly` / global drops in `wifihaven_block` (forward) then killed the redirected packet — they carved out `ea_`/`@global_allow` but **not** the block-page destination. Live: `wh_drop:…:Schedule … DST=::1:8443`.
2. **v6 never delivered.** IPv6 has no `route_localnet` equivalent and `::1` must never appear on the wire (RFC 4291), so DNAT'ing *forwarded* v6 traffic to `::1` never delivered — the routing decision forwarded it out the WAN (`DST=::1:8443 OUT=eth1`).

## Fix
1. **Single never-drop guard.** Add `fib daddr type local counter accept` at the **top** of `wifihaven_block`. Every block-page DNAT/redirect target is a router-local address; normal forwarded traffic is destined to an external host, so the guard matches *only* block-page traffic and never widens who is blocked. One rule encodes the invariant — versus threading a carve-out suffix through every drop rule (one missed rule away from re-introducing the bug). `accept` is a per-chain verdict, so the priority-1 accounting chains still count the bytes. The forward hook never governed router-local destinations (that's the input hook), so no existing drop rule changes.
2. **v6 via `redirect`.** Switch the v6 block-page path from `dnat ip6 to ::1:PORT` to nft `redirect to :PORT` (→ br-lan local address, deliverable). Bind uhttpd to `[::]:8081` / `[::]:8443` (wildcard) instead of loopback `[::1]` so the redirected packet lands on a live listener; `setup-uhttpd-block-page.sh` migrates any legacy `[::1]` entry (binding both fails — the wildcard already covers loopback). **The v4 path (DNAT → `127.0.0.1`, `route_localnet`) is untouched — it already works.**

## Validation
- `busted` (openwrt/): all green (614 passing). New `render_spec` cases assert the never-drop guard precedes every drop and the v6 path uses `redirect` (no surviving `dnat ip6 to ::1`). `install_spec` asserts the `[::]` listener wiring + legacy `[::1]` migration.
- **Rendered ruleset validated with `nft -c -f -` on a real OpenWRT target (nftables v1.1.6)** for both normal and failover snapshots — `redirect` + `fib daddr type local` syntax accepted. (KVM Gate 2/3a can't run locally; relying on PR CI for full boot.)
- **Gate 3a smoke extended**: asserts a whole-MAC-blocked (paused) device still renders the block page end-to-end.

Fixes #1865

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -1152,6 +1152,28 @@ function M.nft(snapshot, opts)
 
   ind("chain wifihaven_block {")
   ind2("type filter hook forward priority 0; policy accept;")
+  -- #1865: the block page must NEVER be dropped, for any reason. The block-page
+  -- DNAT chain (wifihaven_block_nat, prerouting) rewrites a blocked device's
+  -- HTTP/HTTPS destination to the local block-page listener so the user sees
+  -- *why* they are blocked. With route_localnet (v4 → 127.0.0.1) / redirect
+  -- (v6 → br-lan addr) that packet is delivered locally and normally never
+  -- reaches this forward chain — but if it ever does (a transient routing edge,
+  -- a future DNAT target, or route_localnet not yet applied) the per-MAC / eb_
+  -- / bl_ / blockIpOnly / global drops below would kill it and the user would
+  -- get a hung connection instead of the block page (the original bug: a
+  -- Schedule-blocked v6 device DNAT'd to ::1, then dropped — DST=::1:8443).
+  -- One accept guard for router-local destinations encodes the invariant in a
+  -- SINGLE place: every block-page DNAT/redirect target is a router-local
+  -- address (`fib daddr type local`), whereas normal forwarded traffic is
+  -- destined to an external host, so the guard matches only block-page traffic
+  -- and never widens who is blocked. Threading a per-rule carve-out suffix
+  -- instead would be one missed rule away from re-introducing the bug.
+  -- `accept` is a per-chain verdict — the priority-1 accounting chains still
+  -- see the packet, so block-page bytes are still counted. This changes no
+  -- existing drop rule: the forward hook never governed traffic to the router
+  -- itself (that is the input hook), so router-local destinations were never
+  -- dropped here regardless.
+  ind2("fib daddr type local counter accept comment \"wh_block_page:never-drop:#1865\"")
   -- #1122: the family-agnostic @blocked_macs drop is replaced by per-MAC
   -- rules so each can carry its own `wh_drop:<mac>:<reason>` comment. The
   -- set is still declared (used by the DNAT chain below) but no longer
@@ -1283,9 +1305,20 @@ function M.nft(snapshot, opts)
       ind2(predicate .. " tcp dport 80 dnat ip to 127.0.0.1:8081")
       ind2(predicate .. " tcp dport 443 dnat ip to 127.0.0.1:8443")
     end
+    -- #1865: IPv6 has no `route_localnet` equivalent, and ::1 must never appear
+    -- on the wire (RFC 4291), so DNAT'ing forwarded v6 traffic to ::1 never
+    -- delivers — the routing decision treats ::1 as non-local and forwards the
+    -- packet out the WAN instead (observed live: DST=::1:8443 OUT=eth1). Use
+    -- `redirect` instead: it rewrites the destination to the inbound interface's
+    -- (br-lan) own address, which IS router-local, so the packet is delivered to
+    -- the local uhttpd block-page listener. uhttpd binds [::]:8081 / [::]:8443
+    -- (see setup-uhttpd-block-page.sh) so the redirected packet — now destined
+    -- to a br-lan v6 address rather than ::1 — lands on a live listener. (The v4
+    -- path keeps DNAT-to-127.0.0.1: route_localnet makes loopback deliverable on
+    -- v4, and that path already works on prod — leave it untouched.)
     local function dnat6(predicate)
-      ind2(predicate .. " tcp dport 80 dnat ip6 to ::1:8081")
-      ind2(predicate .. " tcp dport 443 dnat ip6 to ::1:8443")
+      ind2(predicate .. " tcp dport 80 redirect to :8081")
+      ind2(predicate .. " tcp dport 443 redirect to :8443")
     end
     if #blocked_macs_list > 0 then
       dnat4("ether saddr @blocked_macs")
@@ -1325,9 +1358,12 @@ function M.nft(snapshot, opts)
       dnat4(string.format("ether saddr %s ip daddr != @%s",
         mac, resolved_set_name(mac)))
     end
-    -- #411: v6 siblings. uhttpd also binds [::1]:8081 + [::1]:8443.
-    -- `ip6 daddr != ::1` guards against self-DNAT recursion if the block
-    -- page itself ever issues an outbound v6 request.
+    -- #411/#1865: v6 siblings, delivered via `redirect` (see dnat6 above).
+    -- uhttpd binds [::]:8081 + [::]:8443 so the redirected packet (now destined
+    -- to the br-lan v6 address) lands on a live listener. The retained
+    -- `ip6 daddr != ::1` guard keeps us from redirecting traffic literally
+    -- addressed to loopback (also avoids any self-redirect if the block page
+    -- ever issues an outbound v6 request).
     if #blocked_macs_list > 0 then
       dnat6("ether saddr @blocked_macs ip6 daddr != ::1")
     end
--- a/openwrt/files/usr/lib/wifihaven/setup-uhttpd-block-page.sh
+++ b/openwrt/files/usr/lib/wifihaven/setup-uhttpd-block-page.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
-# Configure the uhttpd block-page listener on 127.0.0.1:8081 (v4) + [::1]:8081
-# (v6) with the lua handler at /www/wifihaven/handler.lua.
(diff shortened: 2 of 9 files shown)


sameerparekh:
**Independent PR review** (read-only, adversarial) — first run, full merge-base diff `origin/main...HEAD`.

## BLOCKERS
(none)

## SHOULD-FIX
1. **`openwrt/files/www/wifihaven/handler.lua:27` — v6 block page renders without the per-MAC reason.** This PR makes the v6 block page *reachable* (good), but the handler resolves the client MAC from `/proc/net/arp` only (v4). For a v6 client `REMOTE_ADDR`, `block_page.parse_arp` returns nil → redirect carries empty `&mac=` → `GET /api/blocked` can't surface *why*. The page still renders (markers present, so the gate3 probe passes), and this is a **pre-existing** limitation now newly reachable — v6 went from "forwarded to WAN, never rendered" to "renders, minus the reason," a strict improvement. There is already a v6-capable primitive to reuse (`conntrack.parse_arp_table` reads `/proc/net/arp` + `ip -6 neigh`, conntrack.lua:313/347), which also makes this an SSOT win. **Filed as #1869; non-gating for this PR.**

## NITS
1. **`setup-uhttpd-block-page.sh` — v6 listener moves from loopback `[::1]` to wildcard `[::]`.** This exposes 8081/8443 on all v6 interfaces (incl. WAN), mitigated by fw4's default WAN-input drop. Consistent with existing posture: `uhttpd.main` already binds `[::]:80` / `[::]:443` wildcard on the same router. (v4 stays loopback-only.) Verified on a real target that uhttpd sets `IPV6_V6ONLY` per-socket, so `[::]:8081` does **not** collide with the retained `127.0.0.1:8081` despite `bindv6only=0`.
2. **PR body** carries the harness "🤖 Generated with Claude Code" footer — an AI mention per checklist §7. Commit messages are clean (only the allowed `Co-Authored-By` trailer). Non-gating.

## Dimension notes (why no blockers)
- **§1 SSOT (highest):** The never-drop carve-out is a **single** `fib daddr type local counter accept` guard at the top of `wifihaven_block` (render.lua), not a suffix threaded through every drop rule — the correct one-source encoding of the invariant. No policy re-derivation; the snapshot/wire is untouched. `dnat4`/`dnat6` divergence (v4 DNAT vs v6 redirect) is necessary and documented, not duplication.
- **§2 testing:** TDD red→green visible (commit 1 = failing tests, commit 2 = impl). The updated v6 DNAT assertions changed because the OLD behavior is now intentionally wrong; `assert.is_nil("dnat ip6 to ::1")` actively pins the removal. No assertions weakened to hide a regression. gate3 smoke extended to assert a whole-MAC-blocked (paused) device still renders the page.
- **§3 invariants:** Connection-layer only; no DNS enforcement; router stays a dumb applier (render.lua only renders nft from the snapshot — no decision logic). `extraAllowed`/`ea_`/`@global_allow` carve-outs unchanged. No new `/tmp` writers. The guard does **not** widen who is blocked — the forward hook never governed router-local destinations (that's the input hook).
- **§4 wire:** No wire-visible change. UCI listen addresses are router-local config, not the wire contract.
- **§5/§6:** No migrations; no new metric-worthy code path (fix to existing enforcement rendering).
- **§7/§8:** Focused diff (6 files); no dead code/debug; three-dot branch diff. No secrets/SQL/new routes. WAN-exposure noted above (NIT, consistent with existing main uhttpd).

### Cross-target validation
Rendered ruleset (normal + failover snapshots) passes `nft -c -f -` on a real OpenWRT target (nftables v1.1.6): `redirect to :8081/:8443` and `fib daddr type local` syntax accepted. KVM Gate 2/3a can't run locally — relying on PR CI for the full boot + the extended gate3 reachability check.

VERDICT: APPROVE
Carve-out (single `fib daddr type local` accept guard) + v6 `redirect`-to-local-listener correctly fix both confirmed failure modes; the v4 working path is untouched. No blockers. One SHOULD-FIX (v6 per-MAC *reason*, pre-existing, tracked in #1869, non-gating) and two NITs.

sameerparekh:
**Independent PR review — re-run.** Prior reviewed-sha `e338b76`; delta `e338b76...004ffa07` is purely additive test infrastructure (a Gate 2 v6 block-page e2e + marker/selector wiring).

## Prior-findings status
|---|---|---|
| SHOULD-FIX 1 | `handler.lua` v6 client MAC attribution (v4-only `/proc/net/arp`) | **NOT-ADDRESSED** — by design; pre-existing, non-gating, tracked in #1869. Out of scope for this PR (reachability). |
| NIT 1 | v6 listener `[::]` WAN exposure | NOT-ADDRESSED — consistent with existing `uhttpd.main` posture; non-gating. |
| NIT 2 | PR-body AI footer | NOT-ADDRESSED — harness-mandated; commits clean. |

No prior finding was a BLOCKER; none regressed.

## New findings (delta)
### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. `scripts/e2e/scenarios_fake/test_v6_block_page.py:96` — the test is marked `@pytest.mark.smoke`, adding it to the PR-gating subset. Appropriate for a shipped-prod-bug regression, but it relies on the KVM runner's v6 LAN bridge; it cannot pass on a non-KVM/v4-only runner. The fake smoke subset only runs on the self-hosted KVM Gate 2 (where `test_v6_link_local` already proves v6 works), so this is fine — flagged only for awareness.

## Notes (why no blockers)
- **§2 testing:** Net-new deterministic e2e that asserts *real* behavior — it (a) waits for the MAC in `@blocked_macs`, (b) asserts the live router's `wifihaven_block_nat` uses `redirect` and no longer carries the broken `::1:8081/8443` DNAT target, (c) asserts the `fib daddr type local` never-drop guard is present, and (d) drives a v6 request and asserts the block page renders. Substring assertions on `nft list` output are deliberately format-robust; a diagnostic chain/listener dump fires on failure. No existing assertion weakened. The "qemu SLIRP is v4-only" constraint is correctly handled — the block-page redirect terminates locally, so the round-trip never needs WAN forwarding.
- **§3 invariants:** correctly drives policy through the snapshot (paused → whole-MAC block); no router-side decision logic. **§7 scope:** additive only (new test file + registered `block_page_v6` marker satisfying `--strict-markers` + `--only` selector + help text). No wire/migration/metric/security surface touched.

VERDICT: APPROVE
The delta adds a focused, format-robust IPv6 block-page regression e2e for the exact #1865 failure (`DST=::1:8443`), with on-router ruleset assertions plus the v6 round-trip. No blockers; carried-over items remain non-gating (the v6 *reason* refinement is tracked in #1869).

Issue #1104, opened 2026-05-27
bug: time-status UI and policy snapshot use different 'today' calculations — consolidate into single source of truth


sameerparekh:
## Problem

The UI's time-status panel and the policy snapshot served to the router compute \"today\" independently, with different rules:

- Snapshot, grant write, and ingestion paths use `PolicyService.householdLocalDate(now, settings)` — respects `daily_reset_tz`.
- All seven `/api/time/status/*` endpoints use `clock.today` — UTC.

Plus they each independently compute `usedMinutes`, `extensionMinutes`, `remainingMinutes`, and the cap-exhausted boolean from the same underlying tables. Two parallel reductions of the same state.

Net effect: for any household west of UTC, between UTC midnight and local midnight, the UI's panel shows the next day's empty stats while the snapshot keeps enforcing the current day's exhausted cap. Today's usage and today's extension are invisible to the operator for ~6 hours every evening in MDT. Worse, this is the **right kind of bug to silently come back** — anyone touching either code path can re-introduce drift without the test suite noticing.

## Affected endpoints

All in [api/src/routes/Routes.scala](api/src/routes/Routes.scala):
- L509 `GET /api/time/status/summary`
- L576 `GET /api/time/status/summary/week`
- L626 `GET /api/time/status`
- L672 `GET /api/time/status/week`
- L722 `GET /api/time/status/{mac}/week`
- L750 `GET /api/time/status/{mac}`
- L782 `GET /api/time/heartbeat-explain/{mac}`

Each does:
```scala
today <- clock.today
dateStr = req.url.queryParam(\"date\").getOrElse(today.toString)
```

And then calls a `buildDeviceTimeStatus` / `buildProfileTimeStatus` helper that **re-derives** the same cap/used/extension figures the snapshot already computes.

## Fix — consolidate, don't shim

A one-line swap of `clock.today` → `householdLocalDate(now, settings)` in each endpoint is **not enough**. Even with matching date logic, the two paths each re-aggregate `traffic_reports`, re-apply heartbeat filtering, re-fold extensions, re-evaluate the cap. They will drift again the next time someone touches one side.

Extract a single function — call it `TimeStatusService.todaysState` — that returns the canonical per-profile (and per-device) view:

```scala
case class ProfileDayState(
  profileId: ProfileId,
  date: LocalDate,                // householdLocalDate(now, settings)
  dailyLimitMinutes: Option[Int],
  usedMinutes: Int,
  extensionMinutes: Int,
[...]

PR #1874, opened 2026-06-22
fix(#1871): poll the time-used display so it can't lag enforcement

### LINKED ISSUE #1871: bug(ui): time-used display doesn't auto-refresh — profile blocks while UI still shows minutes left (missing refetchInterval)
## Symptom
The per-profile time-used display lags reality: the profile is **blocked** (enforcement) while the UI still shows **a few minutes left**; reloading the page then shows time is fully used up. Confusing — looks like enforcement and the display disagree.

## NOT a divergent code path (ruled out)
Enforcement and the UI read the SAME computation — the #1104 SSOT design. `TimeStatusService` emits one `ProfileDayState{usedMinutes, blocked, blockReason}`, consumed by BOTH `PolicyService.snapshot` (`timeStatusService.dayStateAll`, PolicyService.scala:146) and the `/api/time/status/*` endpoints. They cannot compute different `usedMinutes`.

## Root cause: UI never auto-refreshes the time-used value
`web/src/api/queries.ts:209` — `useTimeStatusToday` sets `staleTime: STALE.timeStatusToday` (30s) but has **NO `refetchInterval`**. Compare the live surfaces that DO poll: `useDashboardNow` and the recent-blocked hook both set `refetchInterval: 10_000` (+ `refetchIntervalInBackground: false`). So the time-used display is frozen from page-load until a refetch trigger (reload / window refocus).

Meanwhile enforcement is fresh: `PolicyService` does NOT read through `TimeStatusCache` — it recomputes on every ~5s router snapshot poll and blocks within seconds of the cap. Result: blocked-but-UI-shows-minutes-left until the user reloads.

Secondary (smaller) contributor: `TimeStatusCache.todayTtl` defaults to 30s (`api/src/cache/TimeStatusCache.scala`), so even a fresh fetch can be up to 30s behind. The dominant cause is the missing UI poll.

## Fix
1. Add `refetchInterval` (and `refetchIntervalInBackground: false`, matching `useDashboardNow`/recent-blocked) to `useTimeStatusToday` in `web/src/api/queries.ts` — 10–30s so the time-used display ticks down live. Keep `staleTime` consistent.
2. Consider whether the per-profile / per-device time-status hooks used on the same surfaces need the same treatment (grep the time-status hooks; apply consistently to the 'today' ones, NOT the 'past'/'week' ones which are logically immutable).
3. Optional: align the `TimeStatusCache` 30s todayTtl so a refetch is never more than a few seconds stale (only if it doesn't add DB load — the cache exists to avoid recomputing presence rollups per poll; lowering TTL trades freshness for load, so evaluate).
4. Add a vitest assertion that `useTimeStatusToday` (and siblings) carry a bounded `refetchInterval`.

Web-only (TS/vitest); no Scala expected for parts 1/2/4. Part 3 if pursued touches `api/`.

sameerparekh:
Fixes #1871

## Problem
The per-profile time-used display lagged reality: a profile could be **blocked** (enforcement) while the UI still showed **a few minutes left**; reloading then showed time fully used. Enforcement is fresh — `PolicyService` recomputes block state on every ~5s router snapshot poll — but the SWR hooks driving the live time-used display in `web/src/api/queries.ts` carried only a `staleTime` and **no `refetchInterval`**, so they never refetched on their own. The displayed value was frozen from page load until a reload or window refocus.

This is a **freshness** bug, not a divergent computation — enforcement and the UI share the same `TimeStatusService.ProfileDayState` SSOT (#1104). That computation is untouched; we only poll the same source more proactively.

## Fix — adaptive poll cadence
Add an `refetchInterval` to the three live **"today"** time-status hooks (`useTimeStatusToday`, `useTimeStatusSummary`, `useTimeStatusProfileToday`), scaled to how close the **most-urgent** profile is to its daily cap (a fixed fast poll is wasted load when a profile has hours left):

| remaining time | poll every |
|---|---|
| ≤ 5m | 10s |
| ≤ 10m | 1m |
| ≤ 30m | 5m |
| > 30m / no daily limit | 5m (baseline) |

- Implemented via react-query's function-form `refetchInterval`, reading the just-fetched `remainingMins` (`timeStatusRefetchMs` — a pure, unit-tested mapping). The least-remaining capped profile drives the whole list's cadence; an at/over-cap (blocked) profile lands in the fast bucket so a time extension / unblock surfaces within seconds. `refetchIntervalInBackground: false` so hidden tabs don't poll.
- The immutable `past`/`week` variants are left on their long staleTime — they do not poll.
- Tests: a ladder-edge unit test over `timeStatusRefetchMs` (including null-limit and past-cap edges) plus fake-timer render tests proving the fast bucket near the cap and the slow-baseline back-off with lots of time left.

## Scope
- Web-only (TS/vitest). No Scala.
- **Not** changing the API-side `TimeStatusCache.todayTtl` (issue part 3) — it trades freshness for DB load (the cache exists to avoid recomputing presence rollups per poll). Possible separate follow-up if the residual ≤30s cache staleness proves noticeable; the dominant cause (no UI poll) is fixed here.

### Deliberately not polled
`useUsageSeriesProfileToday` and `useProfileUsageByApp` also key on `STALE.timeStatusToday` and render on the profile-today surfaces, but were left non-polling on purpose — they're the hourly chart and per-app rollup, not the "minutes left" display that lags enforcement.

## Validation
- `npm run type-check`, `npm run lint`, full `npm run test` (405 tests) all green on node 22.

### DIFF
--- a/web/src/api/queries.test.tsx
+++ b/web/src/api/queries.test.tsx
@@ -11,14 +11,15 @@ vi.mock('@/api/client', () => ({
   api: {
     profiles: { list: vi.fn() },
     devices: { list: vi.fn() },
-    time: { statusAll: vi.fn(), statusAllWeek: vi.fn() },
+    time: { statusAll: vi.fn(), statusAllWeek: vi.fn(), summaryAll: vi.fn() },
     dashboard: { now: vi.fn() },
   },
 }))
 
 import { api } from '@/api/client'
 import {
-  useProfiles, useDevices, useTimeStatusToday, useInvalidators, qk,
+  useProfiles, useDevices, useTimeStatusToday, useTimeStatusSummary,
+  useTimeStatusProfileToday, useInvalidators, qk, TIME_STATUS_REFETCH_MS,
 } from './queries'
 
 function makeWrapper(client: QueryClient) {
@@ -46,6 +47,7 @@ beforeEach(() => {
   ;(api.profiles.list  as unknown as ReturnType<typeof vi.fn>).mockResolvedValue([])
   ;(api.devices.list   as unknown as ReturnType<typeof vi.fn>).mockResolvedValue([])
   ;(api.time.statusAll as unknown as ReturnType<typeof vi.fn>).mockResolvedValue([])
+  ;(api.time.summaryAll as unknown as ReturnType<typeof vi.fn>).mockResolvedValue([])
   ;(api.dashboard.now  as unknown as ReturnType<typeof vi.fn>).mockResolvedValue({ asOf: '', profiles: [] })
 })
 
@@ -101,5 +103,63 @@ describe('SWR caching (#803)', () => {
     // After the invalidation completes, data is still present — the
     // hook never exposes `undefined` for a query that has loaded once.
     expect(hook.result.current.data?.[0]?.profile.name).toBe('Kids')
+  })
+})
+
+// #1871: the live time-used surfaces must poll in the foreground so the
+// displayed "minutes left" can't lag enforcement (the router blocks within
+// seconds of the cap, but a query with only a staleTime never refetches on its
+// own). Each 'today' hook must carry a bounded refetchInterval; the immutable
+// 'past'/'week' hooks must NOT poll.
+describe('time-used live polling (#1871)', () => {
+  it('exposes a bounded refetch interval constant', () => {
+    expect(TIME_STATUS_REFETCH_MS).toBeGreaterThan(0)
+    expect(TIME_STATUS_REFETCH_MS).toBeLessThanOrEqual(30_000)
+  })
+
+  // Drive a hook with fake timers and prove it refetches after the interval.
+  // freshClient's long staleTime means an interval-driven refetch is the ONLY
+  // thing that can produce a second network call, so a missing/false
+  // refetchInterval makes the second assertion fail.
+  async function expectLivePolling(
+    render: () => void,
+    fetchSpy: ReturnType<typeof vi.fn>,
+  ): Promise<void> {
+    vi.useFakeTimers()
+    try {
+      render()
+      // Let the initial fetch resolve.
+      await vi.advanceTimersByTimeAsync(0)
+      expect(fetchSpy).toHaveBeenCalledTimes(1)
+      // One interval later, it polls again on its own.
+      await vi.advanceTimersByTimeAsync(TIME_STATUS_REFETCH_MS)
+      expect(fetchSpy).toHaveBeenCalledTimes(2)
+    } finally {
+      vi.useRealTimers()
+    }
+  }
+
+  it('useTimeStatusToday refetches on the live interval', async () => {
+    const wrapper = makeWrapper(freshClient())
+    await expectLivePolling(
+      () => renderHook(() => useTimeStatusToday(), { wrapper }),
+      api.time.statusAll as unknown as ReturnType<typeof vi.fn>,
+    )
+  })
+
+  it('useTimeStatusSummary refetches on the live interval', async () => {
+    const wrapper = makeWrapper(freshClient())
+    await expectLivePolling(
+      () => renderHook(() => useTimeStatusSummary(), { wrapper }),
+      api.time.summaryAll as unknown as ReturnType<typeof vi.fn>,
+    )
+  })
+
+  it('useTimeStatusProfileToday refetches on the live interval', async () => {
+    const wrapper = makeWrapper(freshClient())
+    await expectLivePolling(
+      () => renderHook(() => useTimeStatusProfileToday(1), { wrapper }),
+      api.time.statusAll as unknown as ReturnType<typeof vi.fn>,
+    )
   })
 })
--- a/web/src/api/queries.test.tsx
+++ b/web/src/api/queries.test.tsx
@@ -19,7 +19,8 @@ vi.mock('@/api/client', () => ({
 import { api } from '@/api/client'
 import {
   useProfiles, useDevices, useTimeStatusToday, useTimeStatusSummary,
-  useTimeStatusProfileToday, useInvalidators, qk, TIME_STATUS_REFETCH_MS,
+  useTimeStatusProfileToday, useInvalidators, qk,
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review

**Scope reviewed:** `git diff origin/main...HEAD` — `web/src/api/queries.ts` + `web/src/api/queries.test.tsx`. Web-only freshness fix; no Scala, wire, migration, metric, or security surface touched.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. `web/src/api/queries.ts:292,308` — `useUsageSeriesProfileToday` and `useProfileUsageByApp` also key on `STALE.timeStatusToday` and render on the same live profile-today surfaces, but were intentionally left non-polling. Correct call for this PR (they're hourly charts / per-app rollups, not the minutes-left display that lags enforcement), but worth a one-line note in the PR that they were considered and deliberately excluded.
2. Bug-fix history is a single commit rather than a visible red→green pair. The added fake-timer tests genuinely fail without the `refetchInterval` (the long-`staleTime` `freshClient` makes an interval-driven refetch the only thing that can produce the 2nd call), so the regression is pinned — just not shown as a separate failing commit.

### Dimension notes
- **1 (SSOT):** No re-derivation. This is a pure freshness fix — both UI and enforcement still read the #1104 `TimeStatusService.ProfileDayState`; the change only makes the UI refetch that same source on a 10s foreground cadence. Strengthens convergence, doesn't fork it. ✅
- **2 (testing):** New behavior asserted three ways (bounded-constant + per-hook interval refetch). `api` is the only mock (external client), no repo/Clock/Auth mocks. Full suite green (403). ✅
- **3/4 (architecture/wire):** Untouched — no router, snapshot, or `/tmp` write. ✅
- **6 (metrics):** N/A — no new server path. ✅
- **7 (hygiene):** Focused 2-file diff; `refetchInterval`/`refetchIntervalInBackground: false` mirror the existing `useDashboardNow`/recent-blocked pattern exactly; `...opts` spread stays last so callers can still override. No AI mentions in title/body. ✅
- **API-side `TimeStatusCache` TTL** correctly left out of scope and flagged as an optional follow-up in the PR body. ✅

VERDICT: APPROVE
A correctly-scoped, well-tested freshness fix: adds a foreground 10s `refetchInterval` to the three live "today" time-status hooks so the displayed minutes can't lag enforcement, leaving the immutable past/week variants alone. No SSOT, wire, or test-integrity concerns; two non-blocking nits only.

sameerparekh:

Prior review: `bc6f27e5`. New push `6154bc2c` makes the cadence **adaptive** per operator request. Delta reviewed: `git diff bc6f27e5...HEAD`.

### Prior findings status
|---|---|---|
| NIT 1 | `useUsageSeriesProfileToday` / `useProfileUsageByApp` excluded from polling without a note | **ADDRESSED** — PR body now has a "Deliberately not polled" section explaining the exclusion. |
| NIT 2 | Bug fix is a single commit, no visible red→green pair | **NOT-ADDRESSED** (non-gating NIT) — still squashed per fix; the fake-timer tests genuinely fail without the interval, so the regression is pinned. |

### New findings (delta)
BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `web/src/api/queries.ts:96` — `timeStatusRefetchInterval` re-types the data inline with the same row shape as `minRemainingMins`'s param. Minor; a shared `type TimeStatusRow = { dailyLimitMins?: number | null; remainingMins?: number | null }` would DRY the two `as` casts. Non-blocking.

### Dimension notes (delta)
- **1 (SSOT):** ✅ No re-derivation. `remainingMins` is consumed straight from the server's SSOT payload purely to pick a poll cadence — it is not recomputed, displayed-from-a-different-source, or used for enforcement. `minRemainingMins` only *selects* the min; it doesn't compute remaining time.
- **react-query semantics:** ✅ Function-form `refetchInterval` is re-evaluated after each settled fetch, so the cadence tightens as `remainingMins` falls toward the cap — the intended adaptivity. The not-yet-loaded case (`data` undefined → `[]` → baseline) is handled.
- **Ladder correctness:** ✅ Matches the requested ≤5m→10s / ≤10m→1m / ≤30m→5m / >30m→5m; `remaining ≤ 0` (blocked) correctly falls in the 10s bucket via `Math.max(0, …)`; `null` limit → baseline, not a tight poll.
- **2 (testing):** ✅ Pure mapping unit-tested across every ladder edge incl. null/undefined/negative; behavioural fake-timer tests assert both the fast-near-cap poll and the slow-baseline back-off (the back-off test is the one that proves "adaptive," not just "polls"). Full suite green (405). Only `api` (external client) mocked.
- **7 (hygiene):** ✅ Focused 2-file diff; exports (`timeStatusRefetchMs`, ladder, max) are test seams; no AI mentions in title/body.

VERDICT: APPROVE
The adaptive-cadence follow-up is correctly scoped and well-tested: a pure remaining-minutes→interval ladder feeding react-query's function-form `refetchInterval`, driven by the most-urgent profile, with no SSOT, wire, or test-integrity concerns. Prior NIT 1 addressed; one new trivial DRY nit only.

sameerparekh:

Prior review: `6154bc2c`. New push `151f4ec7` is a pure refactor addressing the prior re-run's NIT. Delta: `git diff 6154bc2c...HEAD` — `web/src/api/queries.ts` only (−7/+6).

### Prior findings status
| Prior finding | Status |
|---|---|
| NIT (re-run): two inline row-shape annotations could be DRY'd | **ADDRESSED** — extracted to a named `CadenceRow` type; `minRemainingMins` and `timeStatusRefetchInterval` now share it. |
| NIT 1 (first run): chart hooks excluded without a note | ADDRESSED (PR body). |
| NIT 2 (first run): single-commit fix, no red→green pair | NOT-ADDRESSED (non-gating NIT). |

### New findings (delta)
BLOCKERS — (none)
SHOULD-FIX — (none)
NITS — (none)

Behaviour unchanged (type-only extraction); type-check, lint, and the queries suite (10 tests) all green.

VERDICT: APPROVE
Clean type-only DRY refactor with no behavioural change; all prior actionable nits resolved. No open BLOCKERs at any point in this PR's history.

Issue #1831, opened 2026-06-22
Epic: ship API server application logs to Grafana Cloud Loki (searchable app logs)


sameerparekh:
## Goal (operator priority #2 — CORRECTED 2026-06-22)
Make the **API server's application logs** (the Scala API's structured `zio-logging` output — request handling, job runs, warnings, errors) **searchable in Grafana Cloud Loki**, in the same Grafana Cloud stack we already run, with retention beyond the Render dashboard's short window.

> **This epic was previously written wrong** (as a table-tailing *export of the high-volume event tables* — `connection_events` / `traffic_reports` / `block_events` / `query_logs` — into a log indexer). That is **not** what we want and has been scrapped. The old sub-issues (#1852–#1859) and PR #1862 are closed. This is a from-scratch redo.

## Scope — what this is / is NOT
- ✅ **API server application logs → Grafana Cloud Loki.** Operator search over app logs.
- ❌ **NOT** exporting `connection_events` / `traffic_reports` / `block_events` / `query_logs` to Loki or any log indexer. Those are operational event tables, not application logs. If they ever need an external analytical/search store, that's a *separate, later* decision (columnar/ClickHouse territory), not this epic.
- ⏳ **Router / OPNsense agent logs are explicitly FUTURE** — shipping agent logs off-device would push meaningful log traffic over end-user home uplinks. Nice-to-have, deferred; not in this epic.

## Key finding — there is NO "native, like the DB metrics" log path
The working assumption was that Render streams logs to Grafana the same way it streams DB/infra metrics. **It does not:**
- Render's **Metrics Stream** OTLP integration (the one carrying our `render_*` infra metrics, #1244) is **metrics-only** — it cannot carry logs. ([metrics-streams docs](https://render.com/docs/metrics-streams))
- Render's separate **Log Streams** feature forwards service logs only to a **syslog (RFC5424 over TLS/TCP)** endpoint or a generic **HTTPS** endpoint. Grafana Cloud Loki has **no native syslog ingest**, and Render's HTTPS drain does not emit Loki's push-API JSON — so either path needs a small **bridge** that reshapes Render's log lines into Loki's push format. ([log-streams docs](https://render.com/docs/log-streams), [Render "LogStream: Loki support" feature request — still open](https://feedback.render.com/features/p/logstream-loki-support))

So app-logs → Loki requires a real transport decision (below), not a dashboard toggle.

## Transport options (DESIGN-FIRST — choose before building)
[...]

PR #1878, opened 2026-06-22
feat(#1873): ship API app logs to Grafana Cloud Loki via loki4j (deployed envs only)

### LINKED ISSUE #1873: log-export #1 (loki4j): wire loki-logback-appender to push API app logs to Grafana Cloud Loki (deployed envs)
First foundation bit of #1831 (API application logs → Grafana Cloud Loki). **Option B accepted** (app-side loki4j appender) — Option A (Render Log Stream → Alloy syslog) is impossible: our wifihaven-alloy is a Render *background worker* and can't receive inbound; Render OTel *logs* streaming (Option C) doesn't exist yet.

## Scope
Add the **loki4j `loki-logback-appender`** to `api/resources/logback.xml`, **gated to deployed envs only** (not local), pushing straight to Grafana Cloud Loki's push API with basic auth.
- **Fail-open by construction:** async + batching + non-blocking, drop-on-backpressure — Loki being slow/down must NEVER wedge the request path. This is the hard requirement for putting anything on the logging path; verify the appender config enforces it.
- **Label / cardinality model (critical):** whitelist ONLY `service`/`env`/`level` as Loki stream **labels**. Route all other MDC keys (`mac`, `route`, `op`, `status`, `etag`, trace ids — already populated by zio-logging annotations) to **structured metadata**, NOT labels. `mac` as a label = per-device cardinality explosion, forbidden by the same bounded-cardinality rule as metrics. This is the easy mistake here — enforce it in the appender config.
- **Secrets/config:** endpoint + creds as Render-managed `sync:false` secrets (`GRAFANA_CLOUD_LOKI_*`), mirroring the existing `GRAFANA_CLOUD_PROM_*` pattern in render.yaml. Declarative — no dashboard toggles. Never commit creds.
- **Folds in / supersedes #602** (structured JSON appender for deployed envs) — close/relate #602 accordingly.

## Out of scope (separate sub-issue #2)
Backpressure/fail-open hardening proof under load + a drop metric + Grafana panel. File that as the follow-up.

## Validation
Scala/logback/config. Run `mill __.test` for affected modules + `mill scalafmtCheckAll`. Verify the appender is deployed-env-gated and fail-open. The metric/dashboard for drops is the #2 follow-up.

sameerparekh:
First foundation bit of **#1831** (API application logs → Grafana Cloud Loki). **Option B accepted** — app-side loki4j appender. Folds in **#602** (structured deployed-env sink).

## What
Adds the [loki4j](https://loki4j.github.io/loki-logback-appender/) `loki-logback-appender` to `api/resources/logback.xml` so deployed API instances push their application logs straight to Grafana Cloud Loki's push API with basic auth. No table-tailing pipeline, no Alloy `loki.write` hop — the JVM pushes directly.

## How the hard requirements are met
- **Deployed-env gate.** The appender *and* its root attachment are wrapped in a logback `<if condition='isDefined("GRAFANA_CLOUD_LOKI_URL")'>` (janino-backed). Local dev and `mill __.test` never set the secret, so the appender is never instantiated there — presence of the secret IS the gate (no code-level env switch).
- **Fail-open by construction.** `Loki4jAppender` is async by design (the request thread only enqueues; a background thread batches + sends — *not* wrapped in `AsyncAppender`, it's already async). A bounded `<sendQueueMaxBytes>` (32 MiB) **drops on backpressure** rather than blocking, `<drainOnStop>false` keeps JVM shutdown from waiting on Loki, and `<requestTimeoutMs>` caps each send. Loki being slow/down can never wedge the request path. Smoke-tested by running a suite with the appender **active** against a dead endpoint (`127.0.0.1:1`) — no hang, no errors.
- **Label / cardinality whitelist (critical).** Loki **stream labels** are exactly `service` / `env` / `level`. Every other MDC key (`mac`, `route`, `op`, `status`, `etag`, …) rides Loki **structured metadata** via the `* = %%mdc` bulk pattern — same bounded-cardinality rule as metrics. `mac` as a label would be a per-device explosion — forbidden.
- **Secrets, declaratively.** `GRAFANA_CLOUD_LOKI_{URL,USER,PASSWORD}` as Render `sync:false` secrets on the staging + prod API web services in `render.yaml`, mirroring `GRAFANA_CLOUD_PROM_*`. `WIFIHAVEN_ENV` (`staging`/`production`) supplies the `env` label. **No creds committed.**

## Tests
`api/test/src/unit/LokiAppenderConfigSpec.scala` pins the structural contract: deployed-env gate, the `service/env/level` label whitelist (and that `mac`/`route`/`op`/`host` are NOT labels), the `* = %%mdc` structured-metadata routing, and the fail-open batch config. `mill api.test` (full suite) + `scalafmt --check` green.

## Out of scope (separate #1831 sub-#2)
Load-proven backpressure + a drop metric + a Grafana panel — tracked in #1879.

Fixes #1873
Relates to #1831, #602

### DIFF
--- a/api/resources/logback.xml
+++ b/api/resources/logback.xml
@@ -76,6 +76,15 @@
             <password>${GRAFANA_CLOUD_LOKI_PASSWORD}</password>
           </auth>
           <requestTimeoutMs>10000</requestTimeoutMs>
+          <!--
+            Pin the JSON push API explicitly. The protobuf path needs loki-protobuf
+            + snappy on the classpath, which we deliberately do NOT depend on (the
+            loki4j jar resolves standalone). If a future loki4j default flipped this
+            to protobuf, shipping would ClassNotFound on the background sender thread
+            and silently drop every log line (fail-open hides it — the #1334 trap).
+            Keeping it false makes the dep-free contract explicit and flip-proof.
+          -->
+          <useProtobufApi>false</useProtobufApi>
         </http>
         <batch>
           <maxItems>1000</maxItems>
--- a/api/resources/logback.xml
+++ b/api/resources/logback.xml
@@ -5,12 +5,89 @@
         #602: %mdc renders the SLF4J MDC map populated by zio-logging annotations
         (see api/src/observability/LogContext.scala). One-line-per-log locally;
         keys like route=, mac=, op=, etag=, status= appear at the tail so they
-        can be grep'd without parsing the message text. A structured (JSON)
-        appender for deployed environments is tracked separately under #602.
+        can be grep'd without parsing the message text. Deployed environments
+        additionally ship every log line to Grafana Cloud Loki via the LOKI
+        appender below (#1873) — that is the searchable structured sink #602
+        called for; the console line stays human-readable for `docker logs`.
       -->
       <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg %mdc%n</pattern>
     </encoder>
   </appender>
+
+  <!--
+    #1873 (epic #1831, folds in #602): push API application logs straight to
+    Grafana Cloud Loki's push API with basic auth, DEPLOYED ENVIRONMENTS ONLY.
+
+    Gating: the whole appender — and its root attachment below — is wrapped in
+    `<if condition='isDefined("GRAFANA_CLOUD_LOKI_URL")'>` (logback + janino).
+    Local dev and `mill __.test` never set GRAFANA_CLOUD_LOKI_URL, so the
+    appender is never instantiated there; it activates only where Render injects
+    the GRAFANA_CLOUD_LOKI_* secrets (staging + prod web services, see
+    render.yaml). No code-level env switch — presence of the secret IS the gate.
+
+    Fail-open by construction (hard requirement — Loki being slow/down must NEVER
+    wedge the request path):
+      - Loki4jAppender is asynchronous by design: the calling (request) thread
+        only enqueues; an internal background thread batches and sends. Do NOT
+        wrap it in an AsyncAppender — it is already async.
+      - <batch> bounds the in-memory send queue with <sendQueueMaxBytes>. When
+        the queue is full (Loki slow/unreachable and the background sender backs
+        up), incoming events are DROPPED, never blocked — drop-on-backpressure.
+      - <drainOnStop>false</drainOnStop> so JVM shutdown is never held waiting on
+        Loki to drain.
+      - <http><requestTimeoutMs> caps each send on the background thread.
+    (Load-proven backpressure + a drop metric + a Grafana panel are the #1831
+    sub-#2 follow-up, intentionally out of scope here.)
+
+    Label / cardinality model (critical — same bounded-cardinality rule as
+    metrics): ONLY service / env / level are Loki STREAM LABELS. Every other MDC
+    key (mac, route, op, status, etag, profileId, routerId, trace ids, …) is
+    routed to Loki STRUCTURED METADATA via the `* = %%mdc` bulk pattern — indexed
+    for filtering WITHOUT exploding stream cardinality. `mac` as a label would be
+    a per-device cardinality explosion and is forbidden; keeping it in structured
+    metadata is how "show everything device X hit" stays queryable cheaply.
+  -->
+  <if condition='isDefined("GRAFANA_CLOUD_LOKI_URL")'>
+    <then>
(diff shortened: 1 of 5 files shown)


sameerparekh:
## Independent PR review — #1878 (loki4j → Grafana Cloud Loki)

Read-only adversarial pass over `git diff origin/main...HEAD` against the 8 checklist dimensions. First run for this PR (no prior marked comment).

**BLOCKERS**
(none)

**SHOULD-FIX**
1. **`api/resources/logback.xml:60-72` (LOKI `<http>`) — pin the JSON API explicitly to avoid a future silent no-op.** `mill show api.resolvedMvnDeps` confirms loki4j `2.0.3` resolves as a *standalone* jar — there is no `loki-protobuf`, `snappy-java`, or `protobuf-java` on the classpath. The protobuf send path requires those (per loki4j's own migration note), so today the appender necessarily uses the dep-free JSON API and works. But the default isn't pinned: if a future loki4j bump flips `useProtobufApi` to `true` by default, log shipping would `ClassNotFound` on the **background** sender thread and silently drop everything — and fail-open (by design) hides it. This is exactly the #1334 silent-no-op trap. One defensive line — `<useProtobufApi>false</useProtobufApi>` inside `<http>` — makes the contract explicit and immune to a default flip. Cheap; recommend taking it.
2. **Deferred drop-metric + dashboard (dim 6) has no tracked follow-up issue number.** The metrics-with-new-path + metric-ships-with-dashboard rules are explicitly scoped out to the "#1831 sub-#2" follow-up (issue body + PR), which is the right call for this config-only PR. But the PR says "filed as the follow-up" without an issue number, and the global rule is every deferred item gets a GitHub issue referenced from the deferral. File it and reference `#<n>` so the instrument-the-Loki-push-path work doesn't fall through.

**NITS**
1. `api/resources/logback.xml:53-55,64` — `thread`/`logger` are emitted both in the `<message>` `<pattern>` (`[%thread] %logger{36}`) and again as `<structuredMetadata>`. Harmless, but the structured-metadata copies are redundant with the rendered line; minor index cost. Keeping them for queryability is a defensible choice — noting only.
2. `LokiAppenderConfigSpec.scala:31-37` — primary path reads `api/resources/logback.xml` relative to cwd (repo root under mill), with a classpath fallback. Fine as-is; the fallback makes it robust if cwd changes.

**VERDICT: APPROVE**
Config-only addition of the loki4j appender, deployed-env-gated via a janino `<if>` on `GRAFANA_CLOUD_LOKI_URL`, fail-open by construction (async + bounded `sendQueueMaxBytes` drop-on-backpressure + `drainOnStop=false`), with the critical label whitelist (`service`/`env`/`level` only; all other MDC → structured metadata via `* = %%mdc`) correctly enforced and unit-pinned. No wire/migration/router/security concerns; secrets are `sync:false` and uncommitted. The two SHOULD-FIXes are defensive/process, not correctness — no blocker.

sameerparekh:
## Independent PR review — #1878 (re-run @ 3a1b820)

Re-run after the push addressing the prior SHOULD-FIXes. Statused prior findings against current code; reviewed the incremental delta `45ff7f9...3a1b820` for new findings.

### Prior findings — status

|---|---|---|
| SHOULD-FIX 1 | Pin JSON API to avoid future silent no-op (`logback.xml` `<http>`) | **ADDRESSED** — `<useProtobufApi>false</useProtobufApi>` added at `api/resources/logback.xml:85`, with a comment explaining the dep-free contract. Smoke-verified logback accepts the element and still inits fail-open against a dead endpoint. |
| SHOULD-FIX 2 | Deferred drop-metric + dashboard needs a tracked follow-up issue | **ADDRESSED** — #1879 filed (epic #1831, Observability & Metrics) and referenced from `docs/ops/grafana-cloud.md:33` + the PR body. |
| NIT 1 | thread/logger duplicated in `<message>` + `<structuredMetadata>` | NOT-ADDRESSED (accepted — deliberate, kept for queryability) |
| NIT 2 | test reads logback.xml via cwd-relative path + classpath fallback | NOT-ADDRESSED (accepted — fallback makes it robust) |

### New findings in the delta
(none) — the delta is the two-line `<useProtobufApi>false</useProtobufApi>` addition (+ comment) and a one-line docs reference swap to #1879. No correctness, wire, migration, security, or cardinality impact.

**VERDICT: APPROVE**
Both SHOULD-FIXes addressed; no new findings. The appender remains deployed-env-gated, fail-open by construction, and label-whitelisted to `service`/`env`/`level` with all other MDC routed to structured metadata. Ready for merge-when-ready (operator's call).

Issue #1849, opened 2026-06-22
ws-transport: push-on-change + computed-snapshot cache in PolicyService (#1512)


sameerparekh:
Part of the websocket-transport epic — see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §6.2. Directly realizes #1512 (the ~500ms per-poll snapshot recompute).

**Scope:**
- Computed-snapshot cache in `PolicyService` (`Ref[Option[(ETag, PolicySnapshot)]]`), populated on first build, **invalidated on any policy mutation** (profile/device/schedule/blocklist edit, pause/unpause, manual block/allow, time-extension) and on time-dependent transitions (schedule boundary, time-limit exhaustion) via a small re-evaluation ticker. Correctness unchanged: same snapshot bytes for a given state.
- On invalidation, rebuild **once** and publish to the `RouterWsRegistry` (sub-issue A), which pushes one `policy` frame per connected channel — snapshot computed once per change, not once per poll per router.
- The **REST poll path also reads the cache** (returns cached snapshot / 304), so un-migrated HTTP agents also stop paying the recompute — #1512 improves for everyone the moment this lands, independent of transport.
- Metric `policy_snapshot_build_total{result=computed|cache_hit}` proving the cache works + panel. EXPLAIN-verify any new query per `docs/process/query-perf.md`.

Independently shippable; improves the REST path even with zero ws agents; behavior-preserving.

Relates to #1023, #1512.

Issue #1847, opened 2026-06-22
ws-transport: capability handshake (hello/ready) + snapshotVersion + Evolution-policy doc (#376 minimal)


sameerparekh:
Part of the websocket-transport epic — see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §2. This lands the durable, minimal slice of #376 that unblocks a non-additive transport change.

Builds on the ws endpoint (sub-issue A).

**Scope:**
- `hello` (agent→server: `agentCapabilities`, `snapshotVersion`, `agentVersion`) → `ready` (server→agent: `serverCapabilities`, negotiated `snapshotVersion = min(agent.maxKnown, server.maxKnown)`) frames (§2.2). Capability sets are string sets, intersected; the extension point for future features (diffs, compression).
- `snapshotVersion` starts at **1** (today's `PolicySnapshot` shape), bumped only on a breaking shape change. Introduced **only inside the new handshake frames** — the existing snapshot/REST JSON is byte-identical, so no deployed agent parses anything new (§2.4).
- `hello`-timeout (close `4002`), version-ceiling refuse (close `4003`, fall back to cached snapshot).
- **Write the "Evolution policy" subsection** into `docs/process/wire-contract.md` + `docs/architecture.md` (the #376 ask): ignore-unknown `op` (envelope) and ignore-unknown fields (payload), the `min`-version rule, the version-ceiling refusal. Server metric `router_ws_handshake_total{result}` + panel.

Defers (still #376, NOT this issue): per-field snapshot capability gating and `snapshotVersion ≥ 2` machinery — see the follow-up (sub-issue F), which *uses* this handshake.

Relates to #1023, #376.

sameerparekh:
Won't-do — dropping the capability/version handshake as over-engineering.

**Rationale.** The ws transport carries the REST payloads verbatim (`PolicySnapshot`, `UsageReport`, events, metrics — the design is explicit that no payload shapes change). So the same discipline that keeps the REST API safe — **additive-only + ignore-unknown fields** — already makes ws evolution safe, with no version negotiation:

- `snapshotVersion` / `4003` refusal only does something for a *non-additive* snapshot change, which the wire contract forbids without a deprecation window **regardless of transport**. Negotiation unlocks nothing we're allowed to do. A v2 snapshot with new fields is just ignored by an old agent, push or poll alike.
[...]

PR #1880, opened 2026-06-22
feat(#1846): API /api/router/ws websocket transport — endpoint + {op,payload} demux + connection registry

### LINKED ISSUE #1846: ws-transport: API /api/router/ws endpoint + {op,payload} demux + connection registry

sameerparekh:
## What & why

Sub-issue **A** of the websocket-transport epic ([#1023](https://github.com/wifihaven/wifihaven/issues/1023); design [`docs/design/websocket-transport.md`](https://github.com/wifihaven/wifihaven/blob/main/docs/design/websocket-transport.md), spike [#1845](https://github.com/wifihaven/wifihaven/issues/1845) GO). This is the **server side** of the transport — the foundation the capability handshake ([#1847](https://github.com/wifihaven/wifihaven/issues/1847)), agent ws client ([#1848](https://github.com/wifihaven/wifihaven/issues/1848)), and push-on-change ([#1849](https://github.com/wifihaven/wifihaven/issues/1849)) build on.

**Purely additive** — the REST ingest/poll/metrics endpoints stay fully live (design §3); a router may use either transport. No wire-visible change to existing endpoints.

## Changes

- **`RouterWsRoutes`** — `GET /api/router/ws` via zio-http `Handler.webSocket`. Auth at upgrade time through the **existing** `RouterAuth.authenticate` (bad/missing token → 401, no 101 — same semantics as the REST 401, design §4.1); the resolved `Router` is captured once and carried for the connection. The `{op, payload, seq?}` envelope (§1.2) is demuxed on `op`:
  - `usage` / `events` / `metrics` → dispatch into the **shared** ingest services, reply with an `ack` (`ok`/`reject`).
  - `ping` → `pong` (heartbeat).
  - `hello` → logged + metered **seam** for #1847 (no `ready` negotiation here).
  - unknown `op` → ignored + metered (`unknown_op`) — forward-compat (§1.3).
  - Every frame touches `routers.last_seen_at` for uniform liveness with REST (§5.5).
- **`RouterIngestService`** — extracted the `usage`/`events` decode + per-record-skip + persistence logic out of `RouterIngestRoutes` with **no behavior change**, so REST and ws share **one** implementation (`AGENTS.md#single-source-of-truth`). REST routes are now thin adapters; a convenience overload keeps existing call sites/tests unchanged.
- **`RouterWsRegistry`** — per-router connection registry (`Ref[Map[RouterId, Set[WebSocketChannel]]]`), concurrency-safe, keyed by `RouterId`, clean connect/disconnect lifecycle. The lookup a later push path (#1849) needs + the §5.5 "router connected?" signal. Single-process (household model); multi-tenant pub/sub explicitly out of scope.
- **Metrics** (bounded labels only, via `MetricGuard`): `router_ws_connections_active` (gauge, refreshed on every register/deregister so it ages out on disconnect) and `router_ws_frames_total{op,dir,result}` (+ new bounded `dir` label key). Dedicated Grafana dashboard `router-ws-transport.json` wired into `infra/grafana/main.tf`.

> Scoped to #1846. `router_ws_policy_push_total` (push fan-out health) is deferred to #1849 with the push path that emits it, per the dashboards-match-emitted-metrics rule.

## Tests

`RouterWsSpec` — a real `Server` (ephemeral port) + real ws `Client` round trip:
- upgrade auth: missing/invalid token → 401 (no 101),
- `usage` frame → shared ingest → `traffic_reports` row + `ack ok` + registry shows the router connected,
- `ping` → `pong`.

The existing `RouterIngestSpec` (and `BlockedMacEventIngestSpec`, `DbFailureSpec`, `ErrorBoundarySpec`) act as the back-compat gate for the ingest extraction. Full `mill api.test` green; `MetricCardinalityGuardSpec` confirms the new series/label pass the cardinality firewall.

Fixes #1846
Relates to #1023

### DIFF
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -299,6 +299,18 @@ object Main extends ZIOAppDefault {
       promPublisher  <- ZIO.service[PrometheusPublisher]
       routerAuth = new RouterAuthLive(routerRepo)
       routerMetrics <- RouterMetricsService.make
+      // #1846: the transport-agnostic ingest service shared by the REST ingest routes and the new
+      // websocket transport, plus the per-router ws connection registry.
+      routerIngest = new RouterIngestService(
+        routerRepo,
+        trafficRepo,
+        usageRepo,
+        deviceRepo,
+        connRepo,
+        alertRepo,
+        hsRepo,
+      )
+      wsRegistry <- RouterWsRegistry.make
       dbHealthCheck = sql"SELECT 1".query[Int].unique.transact(xa).unit
     } yield {
       // #1177: split the route composition into typed chunks. A flat `++` chain across
@@ -401,16 +413,9 @@ object Main extends ZIOAppDefault {
         RouterRoutes.routes(routerRepo, policy, routerAuth, blockEvRepo) ++
           AdminRouterRoutes.routes(auth, routerRepo) ++
           RollupAdminRoutes.routes(auth, rollupRepo2) ++
-          RouterIngestRoutes.routes(
-            routerAuth,
-            routerRepo,
-            trafficRepo,
-            usageRepo,
-            deviceRepo,
-            connRepo,
-            alertRepo,
-            hsRepo,
-          ) ++
+          RouterIngestRoutes.routes(routerAuth, routerIngest) ++
+          // #1846: additive websocket transport. REST ingest/poll/metrics above stay fully live.
+          RouterWsRoutes.routes(routerAuth, wsRegistry, routerIngest, routerMetrics, routerRepo) ++
           RouterMetricsRoutes.routes(routerAuth, routerMetrics) ++
           AlertRoutes.routes(
             auth,
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -72,6 +72,10 @@ object MetricGuard {
       // stays specific — `id` would be a generic catch-all that future series
       // might mistakenly co-opt for unbounded entities.
       "blocklist_id",
+      // #1846 — websocket frame direction for `router_ws_frames_total`. A fixed
+      // 2-value enum (`in` agent→server / `out` server→agent); bounded, so it
+      // satisfies the §4 cardinality firewall.
+      "dir",
     )
 
   /**
@@ -186,6 +190,14 @@ object MetricGuard {
     "usage_post_total"                          -> Set("result", "router_id", "installation_id"),
     // Server-side ingest health for POST /api/router/metrics (#1205). Concrete, emitted now.
     "router_metrics_batches_total"              -> Set("status"),
+    // #1846 — websocket router transport (server side). `router_ws_connections_active` is the live
+    // count of open channels (a household-scoped gauge, refreshed on every register/deregister so it
+    // ages out cleanly on disconnect). `router_ws_frames_total` counts every frame demuxed/sent:
+    // `op` ∈ {hello, usage, events, metrics, ping, pong, ack, unknown} (the fixed envelope
+    // vocabulary), `dir` ∈ {in, out}, `result` ∈ {ok, reject, unknown_op}. All bounded enums — no
+    // per-mac / per-host dimension ever rides a ws metric.
+    "router_ws_connections_active"              -> Set.empty[String],
+    "router_ws_frames_total"                    -> Set("op", "dir", "result"),
     // #1718 — native OpenWRT OS-level metrics the agent collects from /proc/* and `iw dev`
     // and pushes through the existing #1205 batch transport. Gives operators the LuCI-style
     // view of router health (load / cpu / mem / bandwidth / conntrack / wifi clients) in
@@ -434,6 +446,24 @@ object AppMetrics {
 
   def recordRouterMetricsBatch(status: String): UIO[Unit] =
     MetricGuard.counter("router_metrics_batches_total", Map("status" -> status))
+
+  // ── Websocket router transport (#1846) ───────────────────────────────────────
+  // Set by RouterWsRegistry on every register/deregister: the live count of open
+  // ws channels. Refreshed to the recomputed total each time so a disconnect drives
+  // it back down (no stale high-water value). Unlabelled household-scoped gauge.
+
+  def setWsConnectionsActive(count: Int): UIO[Unit] =
+    MetricGuard.gauge("router_ws_connections_active", Map.empty, count.toDouble)
+
+  // Emitted from RouterWsRoutes for every frame demuxed (`dir=in`) or sent
+  // (`dir=out`). `op` is the envelope discriminator (a fixed enum), `dir` ∈
+  // {in, out}, `result` ∈ {ok, reject, unknown_op}. `unknown_op` is the
+  // forward-compat counter (design §1.3 — an unrecognized op is ignored + metered).
+  def recordWsFrame(op: String, dir: String, result: String): UIO[Unit] =
+    MetricGuard.counter(
+      "router_ws_frames_total",
+      Map("op" -> op, "dir" -> dir, "result" -> result),
+    )
 
   // §5.1 — server-side histogram boundaries for the router-pushed duration histograms. The agent
   // (#1206) reports cumulative bucket counts on these same boundaries; RouterMetricsService folds
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -75,7 +75,7 @@ object MetricGuard {
       // #1846 — websocket frame direction for `router_ws_frames_total`. A fixed
       // 2-value enum (`in` agent→server / `out` server→agent); bounded, so it
       // satisfies the §4 cardinality firewall.
-      "dir",
+      "direction",
     )
 
   /**
@@ -194,10 +194,10 @@ object MetricGuard {
     // count of open channels (a household-scoped gauge, refreshed on every register/deregister so it
     // ages out cleanly on disconnect). `router_ws_frames_total` counts every frame demuxed/sent:
     // `op` ∈ {hello, usage, events, metrics, ping, pong, ack, unknown} (the fixed envelope
-    // vocabulary), `dir` ∈ {in, out}, `result` ∈ {ok, reject, unknown_op}. All bounded enums — no
-    // per-mac / per-host dimension ever rides a ws metric.
+    // vocabulary), `direction` ∈ {in, out}, `result` ∈ {ok, reject, unknown_op}. All bounded enums —
+    // no per-mac / per-host dimension ever rides a ws metric.
     "router_ws_connections_active"              -> Set.empty[String],
-    "router_ws_frames_total"                    -> Set("op", "dir", "result"),
+    "router_ws_frames_total"                    -> Set("op", "direction", "result"),
     // #1718 — native OpenWRT OS-level metrics the agent collects from /proc/* and `iw dev`
     // and pushes through the existing #1205 batch transport. Gives operators the LuCI-style
     // view of router health (load / cpu / mem / bandwidth / conntrack / wifi clients) in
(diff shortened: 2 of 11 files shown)


sameerparekh:
## Independent PR review — #1880 (ws transport server endpoint)

Read-only adversarial pass against the 8-dimension checklist. Reviewed the full merge-base diff `origin/main...HEAD` (9 files, +1432/−584). The headline change is an *extraction* (RouterIngestRoutes → RouterIngestService) plus three new additive files; the existing ingest suite is the back-compat gate and is unchanged.

### BLOCKERS
(none)

### SHOULD-FIX
1. **`RouterWsRoutes.scala:193-201` (ingestMetrics) duplicates the metrics-batch validation in `RouterMetricsRoutes.scala:30-44`** — both decode `RouterMetricsBatch`, meter `malformed`, check `batch.routerId != router.id` → meter `router_mismatch`, then `metricsSvc.ingest`. This is a (small) re-derivation of the same decision in two places (dimension 1). The two transports legitimately differ on the *error shape* (REST 403 text vs ws ack-reject), so only the decode+match+meter core is shareable. Recommend folding that core into `RouterMetricsService` (e.g. a `validateAndIngest(router, batch)` returning a typed mismatch) so a future change to the metrics-ingest guard can't drift. Not merge-gating: it's a 1-line equality guard, not a policy primitive like minutes-used / block-reason.
2. **Production wraps the ws route in `HttpMetrics.instrument` + `ErrorBoundary.observe` + `Readiness.gate` + `LoggingMiddleware.annotate` (`Main.scala`), but `RouterWsSpec` mounts the raw `RouterWsRoutes` only** — the upgrade-through-the-assembled-aspect-stack path is unverified. zio-http aspects are response-transparent so the 101 upgrade should pass through, but a smoke test (or a comment pinning the assumption) would close the gap. Low risk; flagging because the test claims an end-to-end proof the prod path doesn't fully share.

### NITS
1. **`RouterWsRoutes.scala:221` — `WsAck(...).toJsonAST.toOption.get`** is a latent `.get`. For a flat case class `toJsonAST` never returns `Left`, but `.getOrElse(Json.Obj())` avoids a throw-in-pure-construction smell (dimension 3, "no throw").
2. **`RouterWsRoutes.scala` `opLabel(op: String): String = "unknown"`** ignores its argument and always returns the constant — intentional (collapsing an arbitrary op to a bounded label), but the unused param reads oddly; a 0-arg `unknownOpLabel` or an inline literal would be clearer.
3. **Double deregister** — `Unregistered` handler (`:88`) and the `.ensuring` (`:93`) both call `registry.deregister`. Idempotent (set removal + gauge recompute), so harmless; noted for awareness.
4. **TDD red→green not visible in history** — the test + implementation landed in one commit rather than a failing-test-first commit (dimension 2, autonomous-session convention). No existing test was weakened (the ingest suite is untouched; a back-compat overload preserves its call sites), and the new `RouterWsSpec` genuinely asserts behavior (auth 401, usage→ack-ok + DB row + registry liveness, ping→pong), so the regression-hiding risk the rule guards against is absent — but the progression isn't reconstructable from git.

### Dimension notes (clean)
- **Wire contract (4):** additive — new route only; no existing field renamed/retyped; REST endpoints intact. ✓
- **Architecture (3):** no decision logic added to the agent; no new snapshot field; auth reuses `RouterAuth`; `Clock`/ZIO-effects respected. ✓
- **Metrics (6):** `router_ws_connections_active` (gauge) + `router_ws_frames_total{op,dir,result}` via `MetricGuard` with bounded enums; new `dir` key added to `KnownLabelKeys`; `MetricCardinalityGuardSpec` passes; dedicated dashboard registered in `infra/grafana/main.tf`. `router_ws_policy_push_total` correctly deferred to #1849 (no no-data panel). ✓
- **Security (8):** ws upgrade authed via existing per-router bearer (bad token → 401, no 101), parameterized SQL via reused repos, no secrets. ✓
- **Migrations (5):** none. **Scope (7):** focused; logging is operational, not debug cruft; issue-referenced seams, no bare TODOs. ✓

VERDICT: APPROVE

Solid, well-scoped server-side foundation: the ingest extraction is behavior-preserving (gated by the unchanged ingest suite), the demux/registry/metrics match the design, and the round-trip test proves auth + ingest + liveness. No blockers. The SHOULD-FIX items (metrics-guard duplication, assembled-stack coverage) are worth a quick pass but don't gate merge.

sameerparekh:
## Independent PR review — re-run (HEAD e9501605)

Statused the prior review (`91c8f09c`) against current code and reviewed the incremental delta `91c8f09c...e9501605` (RouterWsRoutes.scala, RouterWsSpec.scala; +23/−12).

### Prior findings — status

|---|---|---|
| SHOULD-FIX 1 | metrics-op routerId guard duplicates RouterMetricsRoutes | **NOT-ADDRESSED (accepted)** — author kept it with rationale: `RouterMetricsService.ingest` is already the shared carrier-agnostic core; only a 1-line `routerId` guard + meter wrapper differs, and the two transports legitimately diverge on error shape (REST 403 text vs ws ack-reject). Reasonable "good reason not to" per the SHOULD-FIX definition; not a policy primitive. Not merge-gating. |
| SHOULD-FIX 2 | prod aspect stack (HttpMetrics/ErrorBoundary/Readiness/Logging) not exercised by the test | **ADDRESSED** — `RouterWsSpec.buildWsRoutes` now mounts the route through `HttpMetrics.instrument(LoggingMiddleware.annotate(Readiness.gate(ErrorBoundary.observe(raw), ready=true)))`, the same stack `Main` uses; all three round-trip tests (auth-401, usage→ack-ok+registry, ping→pong) stay green through the wrapped path. The upgrade-through-middleware gap is closed with proof. |
| NIT 1 | `WsAck(...).toJsonAST.toOption.get` latent `.get` | **ADDRESSED** — now `.getOrElse(Json.Obj())` (`RouterWsRoutes.scala:225`). |
| NIT 2 | `opLabel` unused-param constant fn | **ADDRESSED** — helper removed; the unknown-op metric label is the inline bounded literal `"unknown"` with the cardinality-firewall rationale at the call site (`RouterWsRoutes.scala:148-153`). |
| NIT 3 | double deregister (Unregistered + ensuring) | NOT-ADDRESSED — unchanged; idempotent and harmless, informational only. |
| NIT 4 | TDD red→green not visible (single commit) | NOT-ADDRESSED — history-shape note; no regression-hiding risk (no existing test weakened). |

### New findings in the delta
(none) — the delta is exactly the four fixes above. The aspect-stack wrapping is correct (`Readiness.gate` passed `ZIO.succeed(true)` so the gate is open in-test; aspects are `Routes⇒Routes` and response-transparent), and removing `opLabel` for an inline literal doesn't change the emitted label.

VERDICT: APPROVE

All actionable prior findings are ADDRESSED; the one carried SHOULD-FIX is a justified accept (shared ingest core already exists, divergent error shapes). No blockers across either revision. Server-side ws transport foundation is sound, behavior-preserving on the ingest extraction, and now proven through the production middleware stack.

sameerparekh:
## Independent PR review — re-run (HEAD 8eee6b7f)

Statused priors against current code and reviewed the delta `e9501605...8eee6b7f` (Metrics.scala, router-ws-transport.json; +14/−14) — a pure label rename.

### Prior findings — status (unchanged from the e9501605 run)
|---|---|---|
| SHOULD-FIX 1 | metrics-op routerId guard overlaps RouterMetricsRoutes | NOT-ADDRESSED (accepted — shared ingest core exists; only a 1-line guard + divergent error shape differ) |
| SHOULD-FIX 2 | prod aspect stack not exercised by test | ADDRESSED (e9501605) — still in place |
| NIT 1 | `.toOption.get` | ADDRESSED (e9501605) |
| NIT 2 | `opLabel` constant fn | ADDRESSED (e9501605) |
| NIT 3 | double deregister | NOT-ADDRESSED (informational, idempotent) |
| NIT 4 | TDD red→green not in history | NOT-ADDRESSED (no regression-hiding risk) |

### New findings in the delta
(none) — renames the `router_ws_frames_total` label key `dir → direction` for readability. The rename is **consistent across all three coupling points**: `MetricGuard.KnownLabelKeys`, the `Allowed("router_ws_frames_total")` entry, and `AppMetrics.recordWsFrame` (param + Map key), plus the dashboard PromQL/legends/descriptions. `MetricCardinalityGuardSpec` (which enforces Allowed-keys ⊆ KnownLabelKeys and Allowed∩Forbidden=∅) passes, and `RouterWsSpec` stays green. Call sites pass the value positionally (`"in"`/`"out"`), so no value churn. No prod series exists yet (agent ws client is #1848), so this is a zero-cost pre-emit rename. `direction` reads unambiguously vs `dir`.

VERDICT: APPROVE

Cosmetic label-clarity rename, fully consistent across code + guard + dashboard, no behavior change. All actionable prior findings remain ADDRESSED; no blockers across any revision.

sameerparekh:
## Independent PR review — re-run (HEAD 1e8bfdff)

Statused priors against current code and reviewed the delta `8eee6b7f...1e8bfdff` — two new e2e test files (`scripts/e2e/lib/ws_send.py` +216, `scripts/e2e-router.sh` +116), adding WS↔REST ingest-transport parity coverage. No production code changed in this delta.

### Prior findings — status
|---|---|---|
| SHOULD-FIX 1 | metrics-op routerId guard overlaps RouterMetricsRoutes | NOT-ADDRESSED (accepted — shared ingest core; divergent error shapes) |
| SHOULD-FIX 2 | prod aspect stack not exercised | ADDRESSED (e9501605) — and now **additionally** exercised by a true fake-router → real-API e2e (this delta) |
| NIT 1–2 (`.get`, `opLabel`) | ADDRESSED (e9501605) |
| NIT 3–4 (double deregister, TDD history) | NOT-ADDRESSED (informational) |

### New findings in the delta
**Dimension 2 (testing) — strengthened, no findings.** The change *adds* coverage and weakens nothing. It folds the parity check into the existing `scripts/e2e-router.sh`, which the API/UI CD pipeline already runs against **both** the in-docker compose stack (the `e2e` job) and deployed staging (the api-smoke gate) — so no new GHA job, and it runs in both targets. The events leg is a real deep-parity proof (identical payloads over REST vs ws land identically, read back via the public `/api/logs` — staging-safe, no `/api/debug/*` dependency); usage/metrics are acceptance-parity (REST 200 == ws ack ok) with the rationale documented inline. Validated locally: full `e2e-router.sh` green against the stack, all four parity assertions passing.

NITS (non-blocking):
1. `scripts/e2e-router.sh` — under `set -e`, a `ws_send.py` failure aborts the script before the `ACKS_OK` check, so the captured `{"error":…}` JSON isn't surfaced in the failure line (the run just jumps to cleanup). Capturing the helper's stderr/exit before asserting would make a future ws-transport regression self-diagnosing. Minor — the step still fails closed.
2. `scripts/e2e/lib/ws_send.py` — hand-rolls RFC 6455 framing (text + control-frame handling). This is justified (stdlib-only, so it runs in the bare e2e shell over ws:// and wss:// with no pip install), but it is now a second framing implementation alongside the #1845 Lua spike. Self-contained and test-scoped; noted for awareness, not a single-source concern for production code.

### Single-source / scope / security (clean)
- `ws_send.py` is a test client; it does not duplicate any production decision. Payload builders are test fixtures mirroring the wire shapes (expected for an e2e). Focused diff, no dead code, no secrets, no production change.

VERDICT: APPROVE

The delta is additive test coverage that directly answers the "prove the shared ingest service is transport-agnostic end to end" goal, wired into the existing router e2e (docker + staging) rather than a new job. All actionable prior findings remain ADDRESSED; no blockers in any revision. NITs are observability/awareness only.

sameerparekh:
## Independent PR review — re-run (HEAD 62ae64db)

Delta `1e8bfdff...62ae64db` is a one-line security fix in the e2e helper, in response to the CodeQL high-severity alert on the prior push.

### Prior findings — status
All prior findings unchanged (SHOULD-FIX 1 accepted; SHOULD-FIX 2 + NITs 1–2 ADDRESSED in earlier revisions; informational NITs 3–4 unchanged). The two e2e NITs from the 1e8bfdff run remain non-blocking.

### New findings in the delta
**Addressed CodeQL `py/insecure-protocol` (high)** at `scripts/e2e/lib/ws_send.py:59` — `ssl.create_default_context()` still permits TLSv1/1.1; the fix pins `ctx.minimum_version = ssl.TLSVersion.TLSv1_2` for the `wss://` (staging) path. Correct and minimal; `ssl.TLSVersion` is 3.7+ (runners are 3.12). Test-only helper, no production impact. No new findings.

VERDICT: APPROVE

Security alert resolved at the floor (TLS 1.2 minimum). No blockers in any revision; the e2e parity coverage is validated locally and folded into the existing Gate-1 router e2e (docker + staging) per the testing-architecture gates.

PR #1872, opened 2026-06-22
ci(#1870): collapse OpenWrt Lua Tests onto run_tests.sh (single source of truth)

sameerparekh:
Fixes #1870.

## Problem
The "OpenWrt Lua Tests" CI job in `.github/workflows/ci.yml` maintained its **own** hand-copied `busted` invocation that had drifted from `openwrt/test/run_tests.sh`. CI omitted `sni_spec`, `quic_spec`, `lan_warn_spec`, `eb_refresh_spec`, `static_ip_labels_spec`, `host_metrics_spec`, and ran **none** of the `sh test/*.sh` shell specs — so those Lua tests did not gate merges. This is the single-source-of-truth smell `docs/pr-review-checklist.md` §1 warns about.

## Fix
- **`ci.yml`** — the "Run openwrt agent tests" step now invokes `sh test/run_tests.sh` (working-directory: `openwrt`) instead of the duplicated inline busted list. `run_tests.sh` sets its own `LUA_PATH` and runs the full busted suite + the shell specs, so adding a spec there now automatically gates CI. Exactly one authoritative list.
- **`run_tests.sh`** — added `dns_tail_sets_spec` (it was only in CI's list, so the collapse would have *dropped* it), plus `log_spec` and `sni_reassembly_spec` (lua specs that neither list ran) so the single source is comprehensive. All pass.
- **`nft_log_dep_spec.sh`** — fixed pre-existing drift the collapse surfaced. #1717 single-sourced the `.ipk`/`.apk` Depends from `openwrt/Makefile` via `depends-list.sh`; the builder scripts now carry `${DEPENDS_LIST}` (and `build_depends_sync_spec.test.sh` *forbids* reintroducing a literal list), so the spec's literal grep of `build-ipk.sh`/`build-apk.sh` no longer matched. The assertion (shipped package declares `kmod-nf-log{,6}`) is **preserved** — it now checks `depends-list.sh`'s emitted output, the canonical source the builders consume — not weakened or dropped.

`Contract Tests` and the `jsonc-shim` verification are left **as-is** — they intentionally run subsets separately so a Scala model tweak doesn't drag the full openwrt matrix.

## Verification
- `sh test/run_tests.sh` is green end-to-end locally: 848 busted successes + all `sh test/*.sh` shell specs pass (exit 0).
- `actionlint .github/workflows/ci.yml` clean.
[...]

Issue #1870, opened 2026-06-22
CI 'OpenWrt Lua Tests' busted list has drifted from run_tests.sh — some specs don't gate


sameerparekh:
Surfaced while wiring the #1845 ws-client e2e (PR #1863).

`.github/workflows/ci.yml` ("OpenWrt Lua Tests") and `openwrt/test/run_tests.sh` each maintain their **own hand-copied busted invocation**, and they have **drifted**. The CI list omits several specs that `run_tests.sh` runs — e.g. `sni_spec`, `quic_spec`, `lan_warn_spec`, `eb_refresh_spec`, `static_ip_labels_spec`, `host_metrics_spec`, `jsonc_shim_spec` — and **none** of the `sh test/*.sh` shell specs run in that job. Net effect: those Lua tests **do not gate merges**.

This was caught because `ws_frame_spec.lua` (added in #1863) was put in `run_tests.sh` but not in CI's separate list, so it silently wasn't running until CI's list was patched too.

This is the single-source-of-truth smell `docs/pr-review-checklist.md` §1 warns about (a test list duplicated across files drifts).

**Proposed fix:** collapse CI onto `run_tests.sh` — have the "OpenWrt Lua Tests" job `cd openwrt && sh test/run_tests.sh` (installing whatever deps the shell specs need) so there's exactly one authoritative test list. Verify the `sh test/*.sh` specs pass in the CI environment (some may assume tools present on the runner). Confirm the `contract`/`jsonc`-shim jobs that intentionally run separately stay separate.

Relates to #1023 (found during the websocket-transport spike).

Issue #1881, opened 2026-06-22
openwrt nft_log_dep_spec.sh greps build-{ipk,apk}.sh for a literal that #1728 refactored away (stale false-failure)


sameerparekh:
## Symptom

`sh openwrt/test/run_tests.sh` exits non-zero locally because `test/nft_log_dep_spec.sh` reports 4 failures:

```
FAIL: build-ipk.sh Depends includes kmod-nf-log — missing — ipk image lacks the nft log backend
FAIL: build-apk.sh depends includes kmod-nf-log — missing — apk image lacks the nft log backend
FAIL: build-ipk.sh Depends includes kmod-nf-log6 — missing ...
FAIL: build-apk.sh depends includes kmod-nf-log6 — missing ...
```

## Root cause (not a real shipping bug)

The deps are present. #1728 ("single-source openwrt package DEPENDS") refactored `build-ipk.sh` / `build-apk.sh` to derive the Depends list dynamically:

```
# build-ipk.sh
Depends: ${DEPENDS_LIST}      # DEPENDS_LIST from depends-list.sh (parses Makefile)
```

So the literal strings `kmod-nf-log` / `kmod-nf-log6` no longer appear in those build scripts — but `nft_log_dep_spec.sh` still greps the scripts for those literals (`^Depends:.*kmod-nf-log`). `depends-list.sh ipk|apk` output **does** include both pkgs (verified), so the shipped packages are fine — the spec is just stale.

## Why it was invisible

No CI job runs `run_tests.sh`'s shell specs today: CI's `lua-tests` job runs an inline busted list (not the shell `*_spec.sh`), and the `shell-tests` job only discovers `*.test.sh` (different suffix). So this false-failure only surfaces when running `run_tests.sh` locally. Exactly the class of "a spec that doesn't actually gate" tracked in #1877.

## Fix

Update `nft_log_dep_spec.sh` to assert against the new single source — check `depends-list.sh ipk`/`apk` output (what the build scripts actually emit) includes `kmod-nf-log` and `kmod-nf-log6`, rather than grepping the build scripts for a literal that was refactored away. Keep the existing Makefile check.

Being fixed alongside #1877 (it blocks getting `run_tests.sh` green, which #1877's auto-discovery depends on).

sameerparekh:
[...]

PR #1882, opened 2026-06-22
test(#1877): auto-discover openwrt specs in run_tests.sh (no more hand-maintained list)

### LINKED ISSUE #1877: openwrt run_tests.sh hand-maintains its spec list — new specs silently don't run until manually added
## The systemic gap

`openwrt/test/run_tests.sh` enumerates every spec it runs in **two hand-maintained lists** — an explicit `busted test/<a>_spec.lua test/<b>_spec.lua …` invocation, and an explicit `sh test/<x>_spec.sh` block. **Adding a new `test/*_spec.lua` or `test/*_spec.sh` file does nothing until someone remembers to also edit that list.** A forgotten entry means the spec silently never runs and never gates merges — a green ✓ that proves nothing.

This is the root cause behind a recurring class of bugs, not a one-off:
- #1870 — the CI job and `run_tests.sh` each kept their own copy of the list and drifted; several specs (`sni_spec`, `quic_spec`, `lan_warn_spec`, `eb_refresh_spec`, `static_ip_labels_spec`, `host_metrics_spec`) never gated. PR #1872 collapsed CI onto `run_tests.sh`, **but `run_tests.sh` itself is still a hand-maintained list**, so the underlying "add a file ≠ it runs" gap remains.
- #1845 / PR #1863 — `ws_frame_spec.lua` was added to one list but not the other and silently didn't run until the second list was patched.
- Surfacing the omitted `quic_spec` under CI's lua5.1 then revealed a *second* latent bug (#1875) — exactly the kind of thing a spec that never ran can hide.

This is the single-source-of-truth smell `docs/pr-review-checklist.md` §1 warns about, applied to the test harness itself.

## Why this is straightforward to fix

The repo **already solves this** for one class of test: the "Discover and run *.test.sh" CI job (`.github/workflows/ci.yml`) does `find … -name '*.test.sh'` and runs whatever it finds — which is why `build_depends_sync_spec.test.sh` et al. gate automatically with no list to maintain. We just need the same auto-discovery for the busted `*_spec.lua` and shell `*_spec.sh` specs that `run_tests.sh` runs.

## Proposed fix (recommended: auto-discover)

In `openwrt/test/run_tests.sh`:
1. **Busted specs:** replace the explicit `*_spec.lua` list with a glob of `test/*_spec.lua`, minus a small, **documented EXCLUDE list**. Today the only exclusion is `jsonc_shim_spec.lua` (intentionally run in the separate *Contract Tests* job against the **real** `luci.jsonc` C module with a specific `LUA_CPATH`, not the shim) — keep it excluded here, with a comment saying where it runs instead.
2. **Shell specs:** replace the explicit `sh test/*_spec.sh` block with a glob of `test/*_spec.sh` (note: `*_spec.test.sh` files are a different suffix and stay owned by the `*.test.sh` discovery job — the glob must not pick them up).
3. Keep the one explicit non-`test/` invocation (`spike/ws-1845/e2e_test.sh`) as-is — it's a spike path, not a `test/` spec.
4. **Guard the exclusion (belt-and-suspenders):** add a tiny `*.test.sh` (auto-discovered, so it gates itself) that fails if any `test/*_spec.lua`/`test/*_spec.sh` on disk is neither picked up by `run_tests.sh`'s discovery nor on the documented EXCLUDE allowlist — so a future exclusion can't silently hide a spec, and the discovery globs can't be quietly reverted to a static list.

### TDD
Add the guard/coverage test FIRST and show it fail (drop a throwaway `test/zzz_canary_spec.lua` that the old static list wouldn't run, prove discovery now runs it, then remove the canary), then implement the globbing. Validate the full `sh test/run_tests.sh` stays green, and confirm on CI that the *OpenWrt Lua Tests* job still runs the full suite (865+ busted successes + all shell specs + ws e2e).

### Acceptance
- Dropping a new `test/foo_spec.lua` or `test/foo_spec.sh` into the tree makes it run under `sh test/run_tests.sh` (hence CI) with **no edit to any list**.
- A spec that exists on disk but is excluded from the busted run must be explicitly allowlisted with a reason, and the guard fails otherwise.
- `actionlint` + `shellcheck` clean; no assertion weakened.

Relates to #1870, #1875, #1023.

sameerparekh:
## What & why

`openwrt/test/run_tests.sh` hand-maintained **two** spec lists — an explicit `busted test/<a>_spec.lua …` invocation and an explicit `sh test/<x>_spec.sh` block. Adding a new `test/*_spec.lua` or `test/*_spec.sh` did nothing until someone also edited the list, so forgotten specs silently never gated — a green ✓ that proves nothing. This is the **root cause behind #1870** (the CI list and run_tests.sh drifted), and why #1845's `ws_frame_spec.lua` and #1870's `quic_spec`/`sni_spec`/etc. went ungated.

#1872 (merged) re-synced the static list and collapsed CI's *OpenWrt Lua Tests* job onto `run_tests.sh` — but the list is **still hand-maintained**, so the underlying "add a file ≠ it runs" gap remains. This PR closes that gap.

Fixes #1877.

## Changes

1. **Auto-discovery** — `run_tests.sh` now globs `test/*_spec.lua` (busted) and `test/*_spec.sh` (shell), sorted/deterministic, instead of two static lists. `"$@"` passthrough and `set -e` semantics preserved (a failing shell spec still fails the run). The `*_spec.test.sh` suffix is **not** matched by the `*_spec.sh` glob, so those stay owned by CI's existing "Discover and run *.test.sh" job.
2. **Single-source exclude allowlist** — `test/spec_exclude.txt` (data) + `test/spec_discovery.sh` (the parse), sourced by both `run_tests.sh` and the guard. Its only member is `jsonc_shim_spec.lua`: `run_tests.sh` puts the Lua `luci.jsonc` shim on `LUA_PATH`, so running that spec here would test the shim against itself — its real fidelity check runs in CI's *Contract Tests* job against the built `.so`.
3. **Coverage guard** — `test/spec_coverage_spec.test.sh` (a `*.test.sh`, auto-gated by CI's discovery job). Fails if any `test/*_spec.{lua,sh}` on disk is neither in `run_tests.sh`'s actual runlist (queried via a new `WH_LIST_SPECS=1` list mode — behavior, not text) nor documented in `spec_exclude.txt`. It also greps `run_tests.sh` for the glob tokens, so a silent revert to a static list is caught even before the next spec lands.

Since #1872 pointed CI's *OpenWrt Lua Tests* job at `sh test/run_tests.sh`, this auto-discovery is exercised directly by CI, and the guard self-gates via the *.test.sh job.

## TDD (red → green in history)

- **RED**: guard + canary `test/zzz_canary_spec.lua` against the still-static list — guard fails (canary not run; no glob).
- **GREEN**: glob discovery; canary removed; guard passes. Full suite: 865 successes / 1 pending, exit 0.

## Validation

- `sh test/run_tests.sh` → exit 0, 865 successes / 0 failures / 1 pending, all shell specs pass, ws e2e self-skips locally (runs in CI, which installs cqueues/luaossl).
- `shellcheck --severity=warning -x` clean on all shell files; guard passes under `sh` and `bash`; `actionlint` clean (ci.yml untouched).

### DIFF
--- a/openwrt/test/run_tests.sh
+++ b/openwrt/test/run_tests.sh
@@ -8,49 +8,29 @@
 set -e
 cd "$(dirname "$0")/.."
 
+# RED step for #1877: the spec lists are still hand-maintained (replaced by
+# auto-discovery in the green commit). zzz_canary_spec.lua is on disk but absent
+# here, so the coverage guard fails — that is the gap this issue closes.
+LUA_SPECS="test/conntrack_spec.lua test/render_spec.lua test/policy_spec.lua test/usage_spec.lua test/clock_spec.lua test/failover_spec.lua test/blocklists_spec.lua test/block_page_spec.lua test/contract_spec.lua test/selfheal_spec.lua test/version_spec.lua test/update_spec.lua test/metrics_spec.lua test/log_spec.lua test/dns_log_spec.lua test/dns_tail_sets_spec.lua test/sni_spec.lua test/sni_reassembly_spec.lua test/quic_spec.lua test/nft_drops_spec.lua test/nflog_spec.lua test/host_norm_spec.lua test/lan_warn_spec.lua test/eb_refresh_spec.lua test/static_ip_labels_spec.lua test/host_metrics_spec.lua test/ws_frame_spec.lua"
+
+SHELL_SPECS="test/init_spec.sh test/init_boot_spec.sh test/boot_skeleton_spec.sh test/update_spec.sh test/update_multi_pkg_spec.sh test/install_spec.sh test/lua_module_paths_spec.sh test/nft_log_dep_spec.sh test/agent_spec.sh test/rotate_dnsmasq_log_spec.sh test/nflog_tail_bounded_spec.sh test/init_sni_toggle_spec.sh"
+
+# List mode (WH_LIST_SPECS=1): print the specs this script would run, one per
+# line, and exit without running anything. spec_coverage_spec.test.sh reads this
+# to verify discovery is complete.
+if [ "${WH_LIST_SPECS:-}" = "1" ]; then
+  for f in $LUA_SPECS $SHELL_SPECS; do printf '%s\n' "$f"; done
+  exit 0
+fi
+
+# shellcheck disable=SC2086 # word-splitting LUA_SPECS into separate args is intentional
 LUA_PATH="./files/usr/lib/lua/?.lua;./files/usr/lib/lua/wifihaven/?.lua;./spike/ws-1845/?.lua;./test/shim/?.lua;./test/shim/?/init.lua;$(lua -e 'print(package.path)')" \
-  busted test/conntrack_spec.lua \
-         test/render_spec.lua \
-         test/policy_spec.lua \
-         test/usage_spec.lua \
-         test/clock_spec.lua \
-         test/failover_spec.lua \
-         test/blocklists_spec.lua \
-         test/block_page_spec.lua \
-         test/contract_spec.lua \
-         test/selfheal_spec.lua \
-         test/version_spec.lua \
-         test/update_spec.lua \
-         test/metrics_spec.lua \
-         test/log_spec.lua \
-         test/dns_log_spec.lua \
-         test/dns_tail_sets_spec.lua \
-         test/sni_spec.lua \
-         test/sni_reassembly_spec.lua \
-         test/quic_spec.lua \
-         test/nft_drops_spec.lua \
-         test/nflog_spec.lua \
-         test/host_norm_spec.lua \
-         test/lan_warn_spec.lua \
-         test/eb_refresh_spec.lua \
-         test/static_ip_labels_spec.lua \
-         test/host_metrics_spec.lua \
-         test/ws_frame_spec.lua \
-         "$@"
+  busted $LUA_SPECS "$@"
 
 echo ""
-sh test/init_spec.sh
-sh test/init_boot_spec.sh
-sh test/boot_skeleton_spec.sh
-sh test/update_spec.sh
-sh test/update_multi_pkg_spec.sh
-sh test/install_spec.sh
-sh test/lua_module_paths_spec.sh
-sh test/nft_log_dep_spec.sh
-sh test/agent_spec.sh
-sh test/rotate_dnsmasq_log_spec.sh
-sh test/nflog_tail_bounded_spec.sh
-sh test/init_sni_toggle_spec.sh
+for f in $SHELL_SPECS; do
+  sh "$f"
+done
 # ws-client e2e (#1845): runs ws://+wss:// round-trips through the real client;
 # self-skips (exit 0) when lua-cqueues/lua-luaossl aren't installed.
 sh spike/ws-1845/e2e_test.sh
--- a/openwrt/test/run_tests.sh
+++ b/openwrt/test/run_tests.sh
@@ -5,15 +5,36 @@
 # LUA_PATH is set so that:
 #   require("render")            →  files/usr/lib/lua/wifihaven/render.lua    (short form, used in tests)
 #   require("wifihaven.render")  →  files/usr/lib/lua/wifihaven/render.lua    (qualified form, used in modules)
+#
+# Specs are AUTO-DISCOVERED, not hand-listed (#1877): every test/*_spec.lua runs
+# under busted and every test/*_spec.sh runs as a shell spec, so dropping a new
+# spec into test/ gates it with no edit here. The only carve-out is the
+# documented exclude allowlist in test/spec_exclude.txt (today: jsonc_shim_spec,
+# which needs the real luci.jsonc C module and runs in CI's Contract Tests job).
+# spec_coverage_spec.test.sh enforces that this discovery stays complete.
+# Note: test/*_spec.test.sh files are a different suffix owned by CI's
+# "Discover and run *.test.sh" job — the *_spec.sh glob does not match them.
 set -e
 cd "$(dirname "$0")/.."
 
-# RED step for #1877: the spec lists are still hand-maintained (replaced by
-# auto-discovery in the green commit). zzz_canary_spec.lua is on disk but absent
-# here, so the coverage guard fails — that is the gap this issue closes.
-LUA_SPECS="test/conntrack_spec.lua test/render_spec.lua test/policy_spec.lua test/usage_spec.lua test/clock_spec.lua test/failover_spec.lua test/blocklists_spec.lua test/block_page_spec.lua test/contract_spec.lua test/selfheal_spec.lua test/version_spec.lua test/update_spec.lua test/metrics_spec.lua test/log_spec.lua test/dns_log_spec.lua test/dns_tail_sets_spec.lua test/sni_spec.lua test/sni_reassembly_spec.lua test/quic_spec.lua test/nft_drops_spec.lua test/nflog_spec.lua test/host_norm_spec.lua test/lan_warn_spec.lua test/eb_refresh_spec.lua test/static_ip_labels_spec.lua test/host_metrics_spec.lua test/ws_frame_spec.lua"
+# shellcheck source=test/spec_discovery.sh
+. ./test/spec_discovery.sh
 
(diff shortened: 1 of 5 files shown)


sameerparekh:

Reviewed the full merge-base diff (`git diff origin/main...HEAD`) against the 8-dimension checklist. Test-harness-only change: auto-discover OpenWrt specs in `run_tests.sh`, a single-source exclude allowlist, a coverage guard, and a stale-spec fix.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. `openwrt/test/run_tests.sh:18` (`is_excluded`) and `openwrt/test/spec_coverage_spec.test.sh:30` (`excludes`) each re-implement the same "strip comments/blanks from `spec_exclude.txt`" parse. The exclude **data** is correctly single-sourced in `spec_exclude.txt` (the part §1 cares about), so this is non-blocking — but a tiny sourced helper (`spec_discovery.sh` exposing one `wh_spec_excludes` fn) would COLLAPSE the parse idiom too. Optional.
2. `openwrt/test/nft_log_dep_spec.sh:76-83` — the guard now trusts that `build-{ipk,apk}.sh` consume `depends-list.sh` rather than grepping the build scripts directly. Verified this is not a coverage loss: `build_depends_sync_spec.test.sh:46-57` already enforces both builders call `depends-list.sh` and expand `${DEPENDS_LIST}` (no re-hardcoded literal). The new check correctly targets the single source. No action needed — noted for the record.

### Notes / verified
- **§1 single-source:** exclude allowlist single-sourced in `spec_exclude.txt`, read by both consumers. Not a finding.
- **§2 testing:** TDD red→green visible (`a58b839` RED guard+canary catches the canary *and* 3 real ungated specs; `64e76ae` GREEN globs + removes canary). No existing assertion weakened — the nft_log_dep change replaces an always-failing stale grep with a correct check of the same invariant. Guard asserts real behavior (queries `WH_LIST_SPECS=1` runlist) + belt-and-suspenders glob-token check against a static-list revert.
- **§7 scope/hygiene:** focused; `#1881` fix justified inline (blocks a green suite); three-dot diffs; no AI mentions in titles/bodies (Co-Authored-By trailer only); canary added+removed cleanly.
- **Correctness (shell):** `set -e`/`set -eu` interactions safe (all conditional uses are tested in `&&`/`||`/`if`); empty-glob guarded via `[ -e "$f" ] || continue`; `grep -qw` correctly disambiguates `kmod-nf-log` vs `kmod-nf-log6`; `*_spec.sh` glob does not match `*_spec.test.sh` (confirmed empirically — runlist excludes them and the guard); guard does not self-match.
- **§3–6, §8:** N/A (no router logic, wire shape, migration, runtime path/metric, or SQL/secrets).
- Local validation: `sh test/run_tests.sh` → exit 0, 865 successes / 0 failures / 1 pending, all shell specs pass, ws e2e self-skips. `shellcheck --severity=warning` clean; guard passes under `sh` and `bash`; `actionlint` clean (ci.yml untouched).

VERDICT: APPROVE
Auto-discovery replaces two hand-maintained spec lists with sorted globs + a single-source exclude allowlist, and a self-gating `*.test.sh` coverage guard closes the "add a file ≠ it runs" gap (#1877, root cause behind #1870) — immediately recovering 3 silently-ungated specs. Tests are TDD-visible, no assertion is weakened, and the change is correctly scoped. No blockers.

sameerparekh:
## Independent PR review — re-run (push since last review)

Delta since `dcd5562`: collapse the exclude-list parse into a shared sourced helper (`spec_discovery.sh`) + a follow-up fix. Re-checked prior findings and reviewed the incremental diff.

### Prior findings — status
|---|---|---|
| NIT 1 | parse idiom duplicated across `run_tests.sh` / guard | **ADDRESSED** — collapsed into `test/spec_discovery.sh` (`wh_spec_excludes` / `wh_spec_is_excluded`), sourced by both; data + parse now single-sourced. |
| NIT 2 | `nft_log_dep_spec.sh` no longer greps build scripts | **NOTED (unchanged, no action)** — confirmed not a coverage loss; `build_depends_sync_spec.test.sh:46-57` enforces the build-script→`depends-list.sh` consumption. |

### New findings (delta `dcd5562...2586ff03`)
BLOCKERS: (none)
SHOULD-FIX: (none)
NITS: (none)

**Resolved within this delta (transparency):** the collapse commit (`980fab7`) left a stale `$EXCLUDE_FILE` reference in the guard's section-2 stale-exclude message; under `set -eu` that branch would have aborted with an unbound-variable error instead of a clean FAIL (only reachable when a stale exclude exists, so the happy path hid it). Fixed in `2586ff03` and verified: with a synthetic stale exclude the guard now prints `FAIL: … lists 'nonexistent_spec.lua' but … does not exist` and exits 1. No open finding at HEAD.

### Re-validated at HEAD
- `sh test/run_tests.sh` → exit 0, 865 successes / 0 failures / 1 pending; all shell specs pass; ws e2e self-skips locally.
- Guard green under `sh` and `bash`; stale-exclude branch exercised and reports cleanly.
- `shellcheck --severity=warning -x` clean on all four shell files (source directives suppress SC1091); default-severity also clean.

VERDICT: APPROVE
The refactor cleanly single-sources the exclude allowlist (data + parse) with no behavior change; the one latent bug it introduced was caught by this on-push re-review and fixed. Suite stays green at 865/1-pending, the coverage guard self-gates via the `*.test.sh` job, and no blocker is open.

sameerparekh:
## Independent PR review — re-run (branch rebased onto current `main`)

The branch was rebased onto `main` after #1872 merged (which re-synced the static list, collapsed CI's *OpenWrt Lua Tests* job onto `run_tests.sh`, and independently fixed `nft_log_dep_spec.sh`). History was rewritten to a clean RED→GREEN pair; the redundant `nft_log_dep_spec.sh` change was dropped. Re-reviewed the full merge-base diff (`git diff origin/main...HEAD`) and statused prior findings.

### Prior findings — status
|---|---|---|
| NIT 1 | exclude-parse idiom duplicated across `run_tests.sh` / guard | **ADDRESSED** — single-sourced in `test/spec_discovery.sh` (`wh_spec_excludes`/`wh_spec_is_excluded`) from the first commit; data lives in `spec_exclude.txt`. |
| NIT 2 | `nft_log_dep_spec.sh` no longer greps build scripts | **OBSOLETE** — that change is dropped; `main` already carries the equivalent fix via #1872. Issue #1881 closed as resolved-by-#1872. |
| (delta bug from prior run) | stale `$EXCLUDE_FILE` ref under `set -eu` | **ADDRESSED** — both refs now use `$WH_SPEC_EXCLUDE_FILE` (`spec_coverage_spec.test.sh:42,48`); stale-exclude branch verified to report cleanly. |

### New findings (full current diff)
BLOCKERS: (none)
SHOULD-FIX: (none)
NITS: (none)

### Verified at HEAD
- **§1 single-source:** exclude allowlist (data + parse) single-sourced in `spec_exclude.txt` + `spec_discovery.sh`. The static→glob change removes the hand-maintained lists entirely. Not a finding.
- **§2 testing:** RED→GREEN visible (RED guard+canary catches the canary against the static list; GREEN globs + removes canary). No existing assertion weakened. Guard asserts real behavior (`WH_LIST_SPECS=1` runlist) + a static-list-revert grep.
- **§3 architecture / §7 scope:** test-harness only; no router logic, wire, migration, or runtime path. Focused 4-file diff; three-dot diffs; no AI mentions in titles/bodies (Co-Authored-By only).
- **Correctness (shell):** `set -e`/`set -eu` interactions safe; empty-glob guarded (`[ -e "$f" ] || continue`); `*_spec.sh` glob excludes `*_spec.test.sh` (confirmed); guard does not self-match; CI's *OpenWrt Lua Tests* job (post-#1872) runs `sh test/run_tests.sh`, so discovery is CI-exercised and the guard self-gates via the `*.test.sh` job.
- Local: `sh test/run_tests.sh` → exit 0, 865 successes / 0 failures / 1 pending; guard green under `sh` and `bash`; `shellcheck --severity=warning -x` clean; `actionlint` clean (ci.yml untouched).

VERDICT: APPROVE
Clean rebase onto post-#1872 `main`: auto-discovery replaces the two hand-maintained spec lists with sorted globs + a single-source exclude allowlist, and a self-gating coverage guard closes the "add a file ≠ it runs" gap (#1877). Tests are TDD-visible, no assertion weakened, redundant change dropped. No blocker.

PR #1887, opened 2026-06-23
feat(#1885): prove loki4j appender fail-open under load + drop metric + Grafana panel

### LINKED ISSUE #1885: log-export #2 (loki4j): prove fail-open under load + drop metric + Grafana panel
Second bit of #1831, depends on #1873 (loki4j appender — MERGED via #1878).

The appender is configured async/drop-on-backpressure; this sub-issue PROVES and instruments it so a Loki outage can never wedge the request path.

## Scope
- **Prove fail-open under load:** demonstrate (test/bench) that with Loki slow/unreachable, log volume is dropped (not blocked) and request latency is unaffected. Confirm the loki4j buffer/backpressure settings actually drop rather than block when full.
- **Drop metric:** surface a `*_dropped_total` (or loki4j's own emitted metrics) through AppMetrics/MetricGuard with bounded labels, per the instrument-new-functionality rule. So we can alert if logs are silently being dropped.
- **Grafana panel:** add/extend a panel under deploy/grafana/dashboards/ targeting the series actually emitted (grep api/src — don't author from a design catalog), per the metrics-need-a-dashboard rule. CI gate: grafana-terraform.

## Validation
Scala + Grafana TF. `mill __.test` + `mill scalafmtCheckAll`; grafana-terraform lint must pass. Relates to #1831.

sameerparekh:
Fixes #1885. Relates to #1831 (log export epic).

The #1873 loki4j appender ships API logs to Grafana Cloud Loki and is configured async / drop-on-backpressure. This sub-issue **proves** that fail-open behaviour under load and **instruments** the drops so a Loki outage silently shedding logs becomes alertable.

## Architectural grounding
loki4j is fail-open **by construction**: `AsyncBufferPipeline.append()` only enqueues to an in-memory buffer and returns immediately — it never does network I/O on the calling (request) thread. When the bounded `sendQueueMaxBytes` queue fills (Loki slow/unreachable, background sender backed up), the encoder flips `acceptNewEvents=false` and subsequent appends are **DROPPED**, never blocked. So request latency is structurally immune to a Loki outage.

### 1. Prove fail-open (`LokiAppenderFailOpenSpec`)
Drives a real `MeteredLoki4jAppender` at a `ServerSocket` that accepts the TCP connection but never responds (a black-hole Loki, worst case for a sync client), with a tiny send queue. Asserts:
- the appender **sheds** events (its drop counter climbs) — backpressure resolves to DROP, not block;
- **200k appends finish in <5s** while Loki is unreachable — a blocking/sync appender would stall on the per-batch timeout effectively forever. This is the regression guard against anyone wrapping it in a blocking appender or flipping the queue to block-on-full.

### 2. Drop metric (`loki_logs_dropped_total`)
- `MeteredLoki4jAppender`: a **zero-behaviour** subclass that re-exports loki4j's own `droppedEventsCount()` accessor. No Micrometer bridge — loki4j's Micrometer metrics publish to a different registry than the `zio.metrics` Prometheus one behind `GET /metrics`; the appender's own `AtomicLong` is the same number with none of the machinery. `logback.xml` swaps the LOKI appender `class=` to it (wire/log-path transparent).
- `LokiDropMetrics`: a `forkDaemon` poller (mirrors `DbPoolMetrics`/`RouterPresenceMetrics`) that diffs the appender's cumulative drop count each tick into `loki_logs_dropped_total` via `MetricGuard`. **Unlabelled** (appender-wide count; any per-mac/route label would breach the §4 cardinality firewall). No-ops on local/test where the appender is absent — same `GRAFANA_CLOUD_LOKI_URL` gate as the appender, no duplicated env check.
- `LokiDropMetricsSpec` pins the delta logic (accumulates positive deltas, ignores flat/non-increasing readings) end-to-end through the real Prometheus publisher.

### 3. Grafana panel
A "Loki log shipping drops" panel on `api-self-metrics.json` targeting the **actually-emitted** series (`sum(rate(loki_logs_dropped_total[5m]))`), thresholded to flag any sustained non-zero rate. Per the metrics-need-a-dashboard rule.

## Validation
- `mill api.test` — full module green (the two new specs + cardinality-guard + self-metrics suites pass).
- `mill scalafmtCheckAll` (scalafmt --check) clean.
- Dashboard JSON parses (`python3 -m json.tool`); only `infra/grafana` is fmt/validate-gated by `grafana-terraform`, untouched here.

### DIFF
--- a/api/resources/logback.xml
+++ b/api/resources/logback.xml
@@ -27,17 +27,26 @@
 
     Fail-open by construction (hard requirement — Loki being slow/down must NEVER
     wedge the request path):
-      - Loki4jAppender is asynchronous by design: the calling (request) thread
-        only enqueues; an internal background thread batches and sends. Do NOT
-        wrap it in an AsyncAppender — it is already async.
+      - The appender is asynchronous by design: the calling (request) thread only
+        enqueues; an internal background thread batches and sends. Do NOT wrap it
+        in an AsyncAppender — it is already async.
       - <batch> bounds the in-memory send queue with <sendQueueMaxBytes>. When
         the queue is full (Loki slow/unreachable and the background sender backs
         up), incoming events are DROPPED, never blocked — drop-on-backpressure.
       - <drainOnStop>false</drainOnStop> so JVM shutdown is never held waiting on
         Loki to drain.
       - <http><requestTimeoutMs> caps each send on the background thread.
-    (Load-proven backpressure + a drop metric + a Grafana panel are the #1831
-    sub-#2 follow-up, intentionally out of scope here.)
+    #1885 PROVED this fail-open behaviour under load (LokiAppenderFailOpenSpec:
+    with Loki stalled and a tiny send queue, the calling thread keeps appending at
+    full speed while the appender sheds events) and INSTRUMENTED the drops:
+
+      - The class below is MeteredLoki4jAppender, a zero-behaviour subclass of
+        loki4j's Loki4jAppender that re-exports the appender's own cumulative
+        dropped-events counter (the AtomicLong loki4j bumps every time the bounded
+        send queue rejects an event). LokiDropMetrics polls it into the
+        `loki_logs_dropped_total` Prometheus counter (GET /metrics), so a Loki
+        outage silently shedding logs is now alertable (Grafana: api-self-metrics).
+        Swapping the class is wire/log-path-transparent; it only adds an accessor.
 
     Label / cardinality model (critical — same bounded-cardinality rule as
     metrics): ONLY service / env / level are Loki STREAM LABELS. Every other MDC
@@ -49,7 +58,7 @@
   -->
   <if condition='isDefined("GRAFANA_CLOUD_LOKI_URL")'>
     <then>
-      <appender name="LOKI" class="com.github.loki4j.logback.Loki4jAppender">
+      <appender name="LOKI" class="com.github.loki4j.logback.MeteredLoki4jAppender">
         <!-- Stream labels: low-cardinality ONLY. Do NOT add mac/route/host here. -->
         <labels>
           service = wifihaven-api
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -14,6 +14,7 @@ import wifihaven.api.metrics.{
 }
 import wifihaven.api.notify.Notifier
 import wifihaven.api.observability.LoggingMiddleware
+import wifihaven.api.observability.LokiDropMetrics
 import wifihaven.api.policy.*
 import wifihaven.api.routes.*
 import wifihaven.api.usage.RetentionSweepJob
@@ -221,6 +222,11 @@ object Main extends ZIOAppDefault {
         routerRepoMetric  <- ZIO.service[RouterRepo]
         _                 <- RouterPresenceMetrics.loop(routerRepoMetric).forkDaemon
         _                 <- ZIO.logInfo("router-presence metrics fiber forked")
+        // #1885: poll the loki4j appender's fail-open drop counter into
+        // loki_logs_dropped_total. forkDaemon; no-ops on local/test where the
+        // LOKI appender isn't present (deployed-env gate lives in logback.xml).
+        _                 <- LokiDropMetrics.loop().forkDaemon
+        _                 <- ZIO.logInfo("loki-drop metrics fiber forked")
         _                 <- ZIO
           .logWarning(
             "WIFIHAVEN_SEED_TEST_BLOCKLISTS=1 set — seeding dev test_ads/test_social. " +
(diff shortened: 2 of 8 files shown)


sameerparekh:
**Independent PR review** (read-only, adversarial) — reviewed `1a08693`.

Scope: prove the #1873 loki4j appender is fail-open under load, surface its drop count as `loki_logs_dropped_total`, and add a Grafana panel. Reviewed the full merge-base diff against all 8 checklist dimensions.

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `api/src/observability/LokiDropMetrics.scala:66` — `emitDelta`'s `.when(delta > 0)` is redundant with the identical `delta > 0` guard already inside `AppMetrics.recordLokiDropped` (`api/src/metrics/Metrics.scala:467`). Harmless (defensive double-guard); could drop one. Not worth a respin on its own.
2. `api/src/observability/LokiDropMetrics.scala:74` — `loop` resolves the appender **once** at fiber start via `findAppender`. logback is configured lazily on first log line, which in practice happens long before this `forkDaemon` fires during `Main` startup (zio-logging SLF4J emits plenty earlier), so the appender is present in deployed envs. If it somehow attached *after* the poller started it'd no-op forever — acceptable given the ordering, but worth knowing.
3. `api/src/observability/MeteredLoki4jAppender.scala:7` — class deliberately lives in loki4j's own package to reach the package-private `droppedEventsCount()` accessor. Justified and documented (and the only no-Micrometer way to get the count); the maintenance risk is a future loki4j version sealing/modularizing the accessor — which would surface as a compile error caught by the existing build, so low risk.
4. `api/test/src/feature/LokiAppenderFailOpenSpec.scala:128` — phase-1 fill has a 20s safety cap and the test a 60s timeout; on a pathologically slow CI runner a drop might not register in time. The tiny send queue + stalled sender fills within ~hundreds of ms in practice, and a genuine non-drop is a real signal not a flake, so this is fine — flagging the timing assumption for visibility.

Dimension notes:
- **1 Single source of truth** ✓ — the drop count has ONE source (the appender's own `AtomicLong`); the poller diffs it, nothing re-derives it. The metric name appearing in allowlist + emit + dashboard is the established pattern, guard-spec-enforced.
- **2 Testing** ✓ — two net-new specs, no existing test weakened. Fail-open proof asserts both the drop AND <5s non-blocking under a black-hole Loki; delta logic covers flat + non-increasing readings. Real appender / real Prometheus publisher / `Ref` — no forbidden mocks.
- **3 Invariants** ✓ — no router-agent / DNS / snapshot / `/tmp` surface touched; `Clock` not needed (only wall-clock `nanoTime` for a test duration); appender presence is the deployed-env gate, no `sys.env` in Scala.
- **4 Wire** ✓ — no wire-visible change; logback class swap is internal, metric is additive on `/metrics`.
- **5 Migrations** ✓ — none.
- **6 Metrics+dashboard** ✓ — routed through `MetricGuard` with the `Allowed` entry added, unlabelled (bounded), and a checked-in panel on `api-self-metrics.json` targeting the emitted series.
- **7 Scope** ✓ — focused 8-file diff, no drive-bys, no dead code, no AI mentions in title/body (trailer only), three-dot diff.
- **8 Security** ✓ — no secrets / SQL / new routes.

VERDICT: APPROVE
Well-scoped fail-open proof + drop instrumentation that follows the metrics/dashboard and cardinality-firewall conventions exactly, with strong tests (the load-bearing <5s-non-blocking assertion is the right regression guard). Only minor nits, none merge-gating.

sameerparekh:
**Independent PR review (re-run)** — reviewed `c2f78f4`. Prior review covered `1a08693`; delta is the NIT-1 cleanup only (`api/src/observability/LokiDropMetrics.scala`, 1 file / +3 −2).

Prior-findings status:
| # | Finding | Status |
|---|---------|--------|
| NIT 1 | Redundant `.when(delta > 0)` in `emitDelta` | **ADDRESSED** — guard removed; `emitDelta` now calls `recordLokiDropped(cur - prev)` directly and the single guard lives in `recordLokiDropped` (`Metrics.scala:467`). `LokiDropMetricsSpec` (flat + non-increasing readings) still green. |
| NIT 2 | `findAppender` resolved once at `loop` start | NOT-ADDRESSED (by design; accepted — logback is configured before the daemon forks) |
| NIT 3 | Subclass in loki4j's package for the package-private accessor | NOT-ADDRESSED (by design; justified + documented) |
| NIT 4 | Fail-open spec phase-1 20s cap timing assumption | NOT-ADDRESSED (by design; generous bounds, a genuine non-drop is a real signal) |

New findings in the delta:
BLOCKERS — (none)
SHOULD-FIX — (none)
NITS — (none). The change is a pure simplification: removes the duplicate guard, behaviour unchanged (`recordLokiDropped` no-ops on non-positive delta). Comment accurately documents where the guard now lives.

VERDICT: APPROVE
Delta is a one-line review-nit cleanup with no behavioural change; specs remain green. No open blockers.

PR #1902, opened 2026-06-23
feat(#1842): marketing site MVP on apex/www + www/staging redirects

sameerparekh:
> **Update (scope change):** the apex `/blocked` router-compat shim has been **removed** at the operator's direction — existing routers were already re-pointed to `app.wifihaven.net` for their block page (per #1842), and new cloud installs default to it, so no router DNATs its block page at the apex anymore. The redirect ruleset now carries only the `www → app` and `staging → app-staging` 301s; the apex serves the marketing site for every path. `redirects_test.py` was reframed to validate those redirects and to assert the shim does not reappear. Stale shim-referencing comments merged in #1841 (`openwrt/install.sh`, `web/src/pages/BlockedPage.tsx`) were corrected. The sections below predate this change.

---

Fixes #1842. Relates to #1832 (host-rename epic). Final sub-issue of the cut-over (#1841 already repointed new installs to `app.wifihaven.net`).

## What this does

Two things: (a) stands up a marketing-site MVP on the apex/`www`, and (b) — the crucial part — adds the **router-compat `/blocked` shim** so already-deployed routers' block pages keep working after the host rename.

### Marketing site (`web-marketing/`)
- Static MVP `index.html` (no build step, no toolchain): hero positioning WifiHaven as **connection-layer parental controls, not DNS blocking**, a "how it works" section, a self-host vs cloud note, and a prominent **"Open the app → app.wifihaven.net"** CTA. On-brand (cream/slate/terracotta, Inter, brand assets vendored under `site/brand/`).
- New Cloudflare Pages project `wifihaven-www`.
- `.github/workflows/master-marketing.yml` — CI deploy job mirroring the existing SPA wrangler deploys; fires only on `web-marketing/**`.

### Cloudflare TF (`infra/cloudflare/main.tf`)
- New `wifihaven-www` Pages project; repoint the apex/`www` Pages domains + CNAMEs to it (the SPA already lives on `app.wifihaven.net`).
- New `cloudflare_ruleset.redirects` (zone-level `http_request_dynamic_redirect`):
  1. **`wifihaven.net/blocked*` → `app.wifihaven.net/blocked*` — 302, query-preserving, FIRST rule.**
  2. `www.wifihaven.net/*` → `app.wifihaven.net/:splat` (301).
  3. `staging.wifihaven.net/*` → `app-staging.wifihaven.net/:splat` (301).

## The compat shim — why and how it's correct

[...]

Issue #1842, opened 2026-06-22
Marketing site MVP on apex/www + redirects (incl. router-compat /blocked shim)


sameerparekh:
Sub-issue 4 of #1832 (see docs/design/app-host-rename.md §2.4/§2.6, §5). **Depends on #1839–#1841.**

Stand up the marketing site and move apex/www off the SPA:
- New Cloudflare Pages project `wifihaven-www` (recommend Astro static, or a single static index.html + Tailwind MVP). In-repo under `web-marketing/`.
- CI deploy job for it (mirror the existing wrangler pages deploy jobs).
- TF: repoint `cloudflare_pages_domain.apex`/`.www` + the apex/www CNAMEs to the marketing project.
- Redirect rules:
  - `www.wifihaven.net/* → app.wifihaven.net/:splat` (301)
  - `staging.wifihaven.net/* → app-staging.wifihaven.net/:splat` (301)
  - **MERGE-GATING:** `wifihaven.net/blocked* → app.wifihaven.net/blocked*` **query-preserving (?mac=&host=), 302**, ranked above the marketing catch-all. Required because `block_page_url` is a per-install UCI key that can't be pushed over the wire — pre-rename routers still DNAT to the apex.

MVP scope: hero (connection-layer parental controls, NOT DNS blocking), "how it works", self-host vs cloud note, prominent "Open the app → app.wifihaven.net" CTA. No auth/dynamic content.

Verify (critical): hit `https://wifihaven.net/blocked?mac=AA:BB:CC:DD:EE:FF&host=example.com` → lands on the app block page with reason rendered; apex root serves marketing.

Rollback: revert TF (apex CNAME back to the app project) — roll back the apex repoint and the /blocked shim removal together.

Relates to #1832.

sameerparekh:
## Update: `/blocked` apex compat-shim is no longer needed — drop it from scope

All cloud routers have now been migrated off the apex block-page host, so the query-preserving `wifihaven.net/blocked* → app.wifihaven.net/blocked*` redirect (design doc §1.2 / §2.4(2)) is **no longer required** and should be **dropped from this issue's scope**.

**Why the shim was needed** (per the design): `block_page_url` is a per-router UCI key that can't be pushed over the wire, so pre-rename routers would keep DNAT'ing blocked HTTP/80 to `wifihaven.net/blocked` forever — and once the apex became the marketing site, those block pages would break.

**Why it's now moot — the fleet is fully migrated:**
[...]

Issue #1832, opened 2026-06-22
Epic: website + app.wifihaven.net rename — move the SPA to its own host + marketing site (umbrella)


sameerparekh:
## Goal (operator priority #3, 2026-06-22)
Stand up the **website** and rename/move the cloud SPA to **app.wifihaven.net** (the app, distinct from a marketing/landing site on the apex/www). Today the cloud SPA is served from its current host via Cloudflare Pages (see spa-hosting-split.md: cloud staging/prod deploy the SPA to Cloudflare Pages).

## Scope — design/plan FIRST, this is cross-cutting
A plan pass must precede execution; it touches several surfaces that must move together:
1. **DNS / Cloudflare** — add `app.wifihaven.net` (and the marketing site host) to the Cloudflare zone; infra is declarative via Terraform on HCP (cloudflare-tf-backend.md) — do it in-repo, not the dashboard. Map app.wifihaven.net → the Cloudflare Pages project.
2. **OAuth / JWT** — audiences, allowed origins/redirect URIs, CORS, cookie domains must include the new host. Enumerate every place the current SPA origin is hard-coded (API CORS allowlist, JWT audience/issuer, any absolute links).
3. **SPA config** — base URL / API base, any absolute asset or callback URLs, PWA manifest, meta.
4. **Redirects** — old host → app.wifihaven.net (avoid breaking existing bookmarks/links); decide permanent vs temporary.
5. **Docs / links** — sweep repo docs, READMEs, block-page links, email/notification links to the new host.
6. **Marketing website** — separate decision: what stack/host for the public site (apex/www) vs the app. Define MVP scope.

## Deliverable of the first chip
A plan/RFC (doc PR or this-issue comment) enumerating every surface above with the concrete change for each, the cut-over sequence (DNS + CORS/JWT must land before redirect), rollback, and the implementation sub-issues to spawn. Flag anything risky (auth origin changes can lock users out). Do NOT cut over anything yet.

Note: distinct from the old familydns→wifihaven package renames (#363/#560/#561) — this is the SPA host + website.

sameerparekh:
## Plan landed — RFC + sub-issues

RFC doc PR: #1838 (`docs/design/app-host-rename.md`). Plan only; no cut-over.

**Dependency-ordered sub-issues:**
- [ ] #1839 — TF: add `app.*` Pages custom domains + DNS (additive) *(foundation)*
[...]

PR #1914, opened 2026-06-23
fix(#1842): drop cloudflare_ruleset.redirects (token scope + low value)

sameerparekh:
Relates to #1832. Follow-up to the merged #1902.

## What happened
On merge of #1902, **`master-cloudflare` (TF apply) failed** creating `cloudflare_ruleset.redirects`:

```
Error: error creating ruleset wifihaven redirects
Authentication error (10000)
```

Creating a Cloudflare **dynamic-redirect ruleset** requires a token scope (Zone → Dynamic Redirect → Edit) that the Pages+DNS deploy token doesn't carry. The DNS/Pages parts of the apply succeeded first — so the `wifihaven-www` project was created and apex/www were repointed to it — but the ruleset step then failed and aborted the apply. Separately, `master-marketing` raced ahead of the project creation and failed with `Project not found`.

Net effect: apex/www briefly **522'd** (pointing at an empty Pages project) until I re-ran the marketing deploy. **Live state is now healthy** — `wifihaven.net`, `www.wifihaven.net`, and `app.wifihaven.net` all return 200.

## The fix
The `www → app` / `staging → app-staging` redirects are **low-value**: apex and www both serving the marketing site is fine (the site has an "Open the app" CTA), and staging keeps serving the staging SPA. Rather than widen the Cloudflare token, **drop the ruleset**. `master-cloudflare` then applies cleanly with the existing token (the ruleset isn't in TF state — the create failed — so removing it from config is a no-op there; the rest of the apply already succeeded).

- Remove `cloudflare_ruleset.redirects` from `infra/cloudflare/main.tf`; reword the header + apex-domain comments.
- Docs: `infra/cloudflare/README.md`, `web-marketing/README.md` + `wrangler.toml`, `docs/deploy-cloud.md` (topology table → www/staging no longer redirect, apply note, verify curls), and an outcome note on the `app-host-rename.md` RFC.

This also follows the earlier removal of the `/blocked` shim (existing routers were re-pointed to `app.wifihaven.net`), so the zone now has **no** redirect ruleset at all.

## Validation
- `terraform fmt -check` + `terraform validate` clean.
- Live: apex/www/app all 200 (after the marketing redeploy that restored the apex).
- After merge, `master-cloudflare` should report a clean apply (no ruleset to create).

### DIFF
--- a/docs/deploy-cloud.md
+++ b/docs/deploy-cloud.md
@@ -23,21 +23,20 @@ Custom domain layout:
 | `app.wifihaven.net`            | Cloudflare Pages: `wifihaven` (canonical app host, #1832) |
 | `app-staging.wifihaven.net`    | Cloudflare Pages: `wifihaven-staging` (canonical staging app host) |
 | `wifihaven.net` (apex)         | Cloudflare Pages: `wifihaven-www` (marketing site, #1842) |
-| `www.wifihaven.net`            | Cloudflare Pages: `wifihaven-www` → **301-redirects to `app.wifihaven.net`** (#1842) |
-| `staging.wifihaven.net`        | **301-redirects to `app-staging.wifihaven.net`** (#1842) |
+| `www.wifihaven.net`            | Cloudflare Pages: `wifihaven-www` (marketing site, #1842) |
+| `staging.wifihaven.net`        | Cloudflare Pages: `wifihaven-staging` (staging SPA) |
 | `api.wifihaven.net`            | Render: `wifihaven-api-prod`      |
 | `api-staging.wifihaven.net`    | Render: `wifihaven-api-staging`   |
 
 `app.wifihaven.net` / `app-staging.wifihaven.net` are additional custom domains
 on the *same* Pages projects — the SPA bundle is identical, only the fronting
-host is canonical now. Post-#1842 the apex + `www` front the **marketing site**
-(`wifihaven-www`, source under `web-marketing/`), not the SPA. A zone-level
-dynamic-redirect ruleset (`cloudflare_ruleset.redirects` in
-`infra/cloudflare/main.tf`) sends `www`/`staging` → the app host (301,
-query-preserving) so old SPA bookmarks keep working. There is no apex
-`/blocked` router-compat shim — existing routers were already re-pointed to
-`app.wifihaven.net` for their block page (#1842), so the apex just serves
-marketing for every path. The API host (`api.wifihaven.net`) is unchanged.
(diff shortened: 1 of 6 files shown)


sameerparekh:
## Independent PR review — #1914 (drop cloudflare_ruleset.redirects)

Read-only pass over `git diff origin/main...HEAD` (TF + docs only). Remediates the `master-cloudflare` apply failure introduced by #1902.

**BLOCKERS** — (none)

**SHOULD-FIX** — (none)

**NITS**
1. The first-rollout ordering race between `master-marketing` (deploy) and `master-cloudflare` (project creation) that caused the `Project not found` failure is now moot (the `wifihaven-www` project exists), and there's no clean cross-workflow `needs:` to add. Worth keeping in mind if another Pages-project-on-first-merge ever lands, but nothing actionable here.

### Notes (why APPROVE)
- **Root cause correct.** The apply log shows `cloudflare_ruleset.redirects: Creating…` → `Error: Authentication error (10000)`, i.e. the token lacks the dynamic-redirect/rulesets scope. Dropping the resource is a valid fix that avoids a secret change; the redirects are genuinely low-value (apex+www both serving marketing is conventional; staging keeps its SPA).
- **State safety.** The create *failed*, so the ruleset never entered TF state and was never created in Cloudflare (auth error = API rejected the call, no partial object). Removing it from config is therefore a no-op against state — the next `master-cloudflare` apply should report no ruleset work, with the already-applied project/domain/record changes unchanged. (This is the PR's success criterion — confirm the post-merge apply is green.)
- **Diff hygiene.** Ruleset block fully removed; `cloudflare_record.api_prod` and the e2e fixtures intact; header + apex-domain comments reworded to match; `terraform fmt -check` + `validate` clean.
- **Docs consistency.** `infra/cloudflare/README.md` (resource list + state-list note), `web-marketing/README.md` + `wrangler.toml`, and `docs/deploy-cloud.md` (topology table now shows www/staging serving content rather than redirecting, apply note, verify curls) all updated; an "Outcome (2026-06)" note added to the `app-host-rename.md` RFC so the original plan isn't read as current state. No dangling references to `cloudflare_ruleset` / `www_to_app` / redirects_test remain.
- **No wire/policy/migration/metric surface** — pure Cloudflare infra + docs.

**VERDICT: APPROVE**

Correct, well-scoped remediation of the #1902 CD failure: drop the token-gated, low-value redirect ruleset and reconcile the docs. Live hosts are already healthy (apex/www/app all 200 after the marketing redeploy); this makes `master-cloudflare` green without a token change. No blockers.

PR #1920, opened 2026-06-24
fix(#1915): per-flow wh_drop LOG limiter — distinct blocked flows synthesize events again

### LINKED ISSUE #1915: bug(router): wh_drop 2/s throttle (#1864) starves nflog event synthesis → Gate 2 red ~1 day, blocking ALL router publishes
## Impact (critical path)
Master Router CD has been RED since ~2026-06-22 21:00, so **nothing has published since v0.3.13** (06-22 15:50). Every router fix merged after that is stuck unpublished — notably **#1865 (block-page never-dropped, PR #1868)** and **#1864 (nice-agent, PR #1867)** themselves. Operator-visible symptom: `::1:8443` still shows up blocked because the #1865 fix can't reach prod routers.

## Root cause — regression from #1864/#1867
#1867 tightened the nftables `wh_drop` LOG rate-limit from **10/s → 2/s** (burst 20) to stop kernel-ring-buffer flooding. But that same `wh_drop` kernel log is the SOURCE the agent's `nflog-tail` sidecar greps (`logread -f`) to synthesize **blocked/paused connection_events** (`wifihaven-agent:840–855`, #1122/#1126; dedup per `(mac,reason,dst)`). At 2/s, distinct `(mac,reason,dst)` drop events are rate-limited away BEFORE the agent's tail sees them → events never synthesized.

Gate 2 fake-mode failures (reproduce across runs 27983918102/#1845, 27991328611/#1846, 28040057595/#1841):
- `test_paused_mac_https_traffic_surfaces_as_nflog_event` — TimeoutError waiting for the nflog-synthesized Paused event.
- `test_ea_direct_requery_allowed_under_blocked_mac` — TimeoutError (likely same starvation of the blocked-flow event path).

The flat rate-limit conflates two purposes of the wh_drop log: **flood control** (suppress repeats) vs **event synthesis** (needs ≥1 line per distinct (mac,reason,dst)). A blunt 2/s throttle drops distinct events too.

## Fix direction
Reconcile both needs — **log each distinct `(mac, reason, dst)` at least once, suppress only REPEATS** (storms), instead of a flat global rate-limit:
- Preferred: nftables dynamic set keyed on `(ether saddr . ip[6] daddr . reason)` with a timeout — `update @wh_logged {...}` so the LOG fires only on first-seen per window (drop-rate-window). Distinct events always log once; retry storms are suppressed. Enforcement (`counter drop`) unchanged.
- Alternative: revert the harsh nft rate-limit and rely on the agent's EXISTING per-(mac,reason,dst) dedup + the self-truncating /tmp spool for flood control (the spool is bounded; dedup collapses storms in-agent) — i.e. move flood-control off the kernel log and into the agent where it doesn't starve synthesis.
- Whatever the approach: a distinct paused/blocked flow MUST surface as exactly one event; a retry storm must NOT flood. Add a busted test asserting the rendered rule logs distinct flows (not flat-throttled), and confirm Gate 2's paused/blocked-event tests pass.

## Validation / host
Router-agent + Gate 2 e2e (KVM). A macOS chip writes the render.lua + busted change and validates via CI (Gate 2 ~37min/run — be high-confidence); final reachability is Gate 2 on the self-hosted runner. NOTE the VM logs also show `udhcpc: no lease` networking noise — rule that in/out, but the deterministic cross-run reproduction + the exact event-starvation mechanism point at the throttle, not flake.

Fresh issue (do NOT reopen merged #1864/#1867). Relates to #1864, #1865 (blocked from shipping), #1826.

sameerparekh:
Fixes #1915. Relates to #1864, #1865.

## Problem (critical path — Master Router CD red ~1 day, blocking ALL router publishes incl. #1865)

#1864/#1867 tightened the nftables `wh_drop` LOG to a flat per-rule `limit rate 2/second burst 10` to stop the kernel ring-buffer flood. But that same `wh_drop` kernel log is the SOURCE the agent's `wifihaven-nflog-tail` sidecar greps (`logread -f`) to synthesize **blocked/paused `connection_event`s** (`wifihaven-agent:840–855`, #1122/#1126).

A flat per-rule limit is a **single token bucket SHARED across every flow** that hits the drop rule. Once a retry storm drains the bucket, the **next DISTINCT `(mac,dst)` flow's first packet is rate-limited away before the agent's tail ever reads it** → events never synthesize. Gate 2 `test_paused_mac_https_traffic_surfaces_as_nflog_event` and `test_ea_direct_requery_allowed_under_blocked_mac` time out, deterministically, across runs.

The flat limit conflated two purposes of the LOG: **flood control** (suppress REPEATS) vs **event synthesis** (needs ≥1 line per distinct event).

## Fix — per-element (per-flow) limiter (issue's PREFERRED option)

Replace the flat per-rule limit with a dynamic set keyed on `(ether saddr . ip[6] daddr)` with an **embedded `limit rate 1/minute burst 5`**:

```
… update @wh_drop_log4 { ether saddr . ip daddr limit rate 1/minute burst 5 packets } log prefix "wh_drop:<mac>:<reason> "
… counter drop comment "wh_drop:<mac>:<reason>"
```

- **Each distinct flow gets its own token bucket** → a fresh `(mac,dst)` element starts with a full burst, so the **first packet of every distinct flow always logs** → synthesis is never starved.
- A storm to the **same** `(mac,dst)` drains only that element's bucket → suppressed → ring buffer stays usable (the #1864 goal, now achieved per-flow instead of by a blunt global throttle).
- The embedded limit is a **match**: over budget the `update` evaluates false and the rule stops **before** the log (no verdict), falling through to the unchanged unconditional `counter drop`. **Enforcement and drop counts are exact** — only the LOG gating moved from per-rule to per-flow.

### Why `(mac, dst)` is the right key — it removes only duplicates, never real events

`(mac, dst_ip)` is exactly the granularity at which a distinct event exists. In `nflog.lua`, the synthesized event's `hname = lookup(dst_ip)` and `reason` come from the matched rule — **both are functions of `dst_ip`**: hname is a reverse lookup of it, and a given `(mac,dst_ip)` always hits the same first-matching drop rule (the rules don't match on L4 port), so the reason is deterministic. The agent's own dedup key `(mac, dst_ip, hname)` therefore collapses to `(mac, dst_ip)` — the same granularity my limiter uses. So every distinct flow logs its first packet; only repeats the agent would dedup anyway (60s window) are suppressed before they flood the ring buffer.

Whole-MAC drops are family-agnostic (no daddr predicate), so their LOG is split into a `meta nfproto ipv4`/`ipv6` pair to extract a `(mac,dst)` key per family; the single `counter drop` stays family-agnostic.

Why not the ALT (revert nft limit, rely on agent dedup + spool): the agent dedup runs *after* the kernel has already written every packet's LOG line, so it can't protect the **kernel** ring buffer — reverting re-introduces the #1864/#1826 flood. The per-element limiter protects both.

## Validation

- New failing-first busted specs (`render_spec.lua`, committed red then green): per-flow `(mac,dst)` limiter sets exist; the LOG is gated by `update @wh_drop_log…` and **never** a flat `limit rate N/second`; whole-MAC drops log both v4+v6 distinct flows; the drop verdict is still never gated by the limit. Updated the #1826/#1864 specs that pinned the old flat format.
- `busted test/render_spec.lua` → **201 passing**; `nflog_spec` → 21 passing. (Other specs' local errors are missing C libs `luci.jsonc`/`cqueues` on the lua5.5 host — env, not this change; CI runs lua5.1 with deps.)
- **`nft -c -f` clean (exit 0) on a real OpenWRT v1.1.6 / kernel 6.12 router** for both the ea-carve and whole-MAC `meta nfproto` rulesets — confirms the embedded-limit-in-set and `meta nfproto` syntax are kernel-valid.
- Final reachability is Gate 2 on the self-hosted KVM runner (can't run locally) — relying on this PR's CI.

## `udhcpc: no lease` VM noise — ruled out

Transient client boot-time DHCP lag. The harness already absorbs it: `_wait_for_client_lan` (`scripts/e2e/lib/vm.py`) retries the eth0 lease + DNS probe for up to 60s and only `log.warning`s on miss (never fails the run). The deterministic, cross-run failure is the event-synthesis `TimeoutError`, not a connectivity/DHCP error. Primary fix is the throttle.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -1080,12 +1080,14 @@ function M.nft(snapshot, opts)
   end
 
   -- #1122/#1126: each drop is rendered as a pair of rules sharing one predicate
-  -- (see #1826 / `emit_drop` below):
-  --   `<predicate> limit rate 2/second burst 10 packets log prefix "wh_drop:<mac>:<reason> "`
+  -- (see #1826/#1915 / `emit_drop` below):
+  --   `<predicate> update @wh_drop_log4 { ether saddr . ip daddr limit rate 1/minute burst 5 packets } log prefix "wh_drop:<mac>:<reason> "`
   --   `<predicate> counter drop comment "wh_drop:<mac>:<reason>"`
   -- (pre-#1826 this was a single `… log prefix … counter drop` rule; the LOG is
-  -- now on its own rate-limited rule so a retry storm can't flood the kernel
-  -- ring buffer, while the drop stays unconditional).
+  -- now on its own rule so a retry storm can't flood the kernel ring buffer,
+  -- while the drop stays unconditional. #1915 made the LOG gate a per-flow
+  -- (per-(mac,dst)) limiter set instead of a flat per-rule rate — see the
+  -- emit_drop comment below for why.)
   -- The `log prefix` form writes to the kernel ring buffer (the LOG backend),
   -- so the wifihaven-nflog-tail sidecar can read the dropped-packet records
   -- straight off `logread -f` on stock OpenWRT — no NFLOG netlink consumer
--- a/openwrt/files/usr/lib/lua/wifihaven/render.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/render.lua
@@ -1129,27 +1129,68 @@ function M.nft(snapshot, opts)
   -- shared `logread` ring buffer when several MACs retry-storm at once — the
   -- aggregate (10/s × N drop rules) evicted wifihaven-agent and dnsmasq lines
   -- from the ring buffer, leaving the router undiagnosable during an incident.
-  -- Throttle the LOG harder: 2/second with a burst of 10 still captures the
-  -- distinct flows the agent's nflog tail needs to label blocked-flow events
-  -- (the labels are deduped per (mac, reason, dst), so a retry storm to one
-  -- endpoint needs only a single sample), but cuts the steady-state ring-buffer
-  -- footprint ~5×. The burst absorbs the initial flurry of distinct hosts; the
-  -- low sustained rate keeps the ring buffer usable for everything else. The
-  -- `counter drop` rule is unchanged, so enforcement and drop counts are exact.
-  local DROP_LOG_LIMIT = "limit rate 2/second burst 10 packets"
-  local function log_suffix(mac, reason)
-    return string.format(" %s log prefix \"wh_drop:%s:%s \"",
-                         DROP_LOG_LIMIT, mac, reason)
+  -- That was first throttled harder (a flat 2/second per-rule limit), but —
+  --
+  -- #1915: a flat per-RULE `limit rate` is a single token bucket SHARED across
+  -- every flow that hits the rule. It conflates two needs the wh_drop LOG
+  -- serves: flood control (suppress retry STORMS) and event synthesis (the
+  -- agent's nflog tail needs ≥1 line per DISTINCT (mac,reason,dst) to emit a
+  -- blocked/paused connection_event). Once a storm drains the shared bucket,
+  -- the NEXT distinct flow's first packet is rate-limited away before the tail
+  -- reads it — so events silently stop synthesizing (Gate 2
+  -- test_paused_mac_https_traffic_surfaces_as_nflog_event hung; the regression
+  -- that took Master Router CD red ~2026-06-22 and blocked every router
+  -- publish, #1865 included).
+  --
+  -- Fix: gate the LOG with a PER-ELEMENT (per-flow) limiter instead — a
+  -- dynamic set keyed on (ether saddr . ip[6] daddr) with an embedded
+  -- `limit rate`. Each distinct (mac,dst) flow gets its OWN token bucket, so
+  -- the first packet of every distinct flow always logs (fresh bucket) and
+  -- synthesis is never starved; a storm to the SAME (mac,dst) drains only that
+  -- element's bucket and is suppressed, so the ring buffer stays usable. The
+  -- embedded limit is a match: when the bucket is empty the `update` expression
+  -- is false and the rule stops BEFORE the log (no verdict), falling through to
+  -- the unconditional `counter drop` on the next rule — enforcement and drop
+  -- counts are unchanged, exactly as the two-rule #1826 split guarantees. The
+  -- small burst headroom (5) is collapsed back to one event by the agent's own
+  -- per-(mac,reason,dst) dedup (nflog.new_dedup, 60s window). The two sets
+  -- (wh_drop_log4/6) are declared just above `chain wifihaven_block`.
+  --
+  -- Per-flow rate keyed on (mac,dst): 1/minute lines up with the agent's 60s
+  -- dedup window; burst 5 absorbs a couple of initial retransmits per flow
+  -- without flooding. Distinct flows are never throttled against each other.
+  local DROP_LOG_RATE = "limit rate 1/minute burst 5 packets"
+  -- LOG suffix for a family-constrained predicate ("ip" → v4 set, else v6).
+  local function log_suffix(family, mac, reason)
+    local set, key
+    if family == "ip6" then
+      set, key = "wh_drop_log6", "ether saddr . ip6 daddr"
+    else
+      set, key = "wh_drop_log4", "ether saddr . ip daddr"
+    end
+    return string.format(
+      " update @%s { %s %s } log prefix \"wh_drop:%s:%s \"",
+      set, key, DROP_LOG_RATE, mac, reason)
   end
   local function drop_suffix(mac, reason)
     return string.format(" counter drop comment \"wh_drop:%s:%s\"",
                          mac, reason)
   end
-  -- Emit the rate-limited log rule then the unconditional drop rule for one
-  -- predicate. `predicate` is everything up to (but excluding) the log/drop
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review — #1920 (first run)

Read-only adversarial pass over `git diff origin/main...HEAD` (render.lua + render_spec.lua) against the 8 checklist dimensions.

### BLOCKERS
(none)

### SHOULD-FIX
1. **`openwrt/files/usr/lib/lua/wifihaven/render.lua:1084`** — the canonical "each drop is rendered as a pair of rules" example comment still shows the **old flat form** `<predicate> limit rate 2/second burst 10 packets log prefix "wh_drop:…"`, now contradicted by the code this PR ships (which emits `<predicate> update @wh_drop_log4 { ether saddr . ip daddr limit rate 1/minute burst 5 packets } log prefix "wh_drop:…"`). It's the hot-path comment the next reader keys off — stale doc in the changed file. Update the example to the per-flow form.

### NITS
1. **render.lua (set decls `wh_drop_log4`/`wh_drop_log6`)** — the dynamic sets omit an explicit `size` cap, so they can grow with the number of distinct dropped `(mac,dst)` flows in the timeout window. Mitigated and **not** a new OOM risk: keyed on dropped-flow dsts only (a strict subset of all traffic) with a short `2m` timeout — strictly less exposure than the existing `mac_ip_tracking` set (all forwarded traffic, `6h` timeout, also no `size`) already accepted in prod. An explicit `size` would be defensive-in-depth only.
2. **PR body** carries the `🤖 Generated with Claude Code` footer; dimension 7 lists "no AI mentions in PR body." Per repo norm the authorship trailer is accepted, so non-blocking.

### Dimension notes (clean)
- **1 Single source of truth:** no policy quantity re-derived. The throttle key `(mac, dst)` is intentionally aligned with the agent's own event identity — `nflog.lua:257` derives `hname = lookup(dst_ip)` and `reason` from the matched rule, so `(mac,dst_ip,hname)` collapses to `(mac,dst_ip)`. The throttle therefore suppresses only what the agent's 60s dedup would discard anyway; no double-source.
- **2 Testing:** red→green visible (`a4d6eecc` failing test before `bf7a4d0b` impl). New per-flow behavior asserted positively (limiter sets exist; LOG gated by `update @wh_drop_log…`; whole-MAC logs both v4+v6; drop verdict still never gated). The edits to the #1826/#1864 specs **strengthen** rather than weaken — they replace the old flat-format pins with "never a flat per-second throttle," which is the bug being fixed (justified). No assertion weakened to hide a regression.
- **3 Architecture:** no decision logic added to the agent (only LOG gating); enforcement `counter drop` is byte-identical; no DNS-layer reasoning. No new `/tmp` writer (nft dynamic sets are kernel memory, timeout-bounded).
- **4 Wire contract:** the `wh_drop:<mac>:<reason> ` log prefix — the agent's parse contract — is unchanged, so an old agent against a new ruleset still synthesizes events. Additive/compatible.
- **5/6/8:** no migrations; no new metric needed (bugfix to an existing path, not a new one); no secrets/SQL/routes.
- **Validation corroborated:** `nft -c -f` exit 0 on real OpenWRT v1.1.6 for both the ea-carve and whole-MAC `meta nfproto` rulesets; all 14 `emit_drop` call sites pass the correct family (daddr-bearing → ip/ip6; the 2 family-agnostic whole-MAC calls → default `any`).

### VERDICT: APPROVE
Per-flow `(mac,dst)` LOG limiter cleanly fixes the #1864 flat-throttle regression that starved nflog event synthesis: first-seen of every distinct flow always logs (fresh element = full burst) while same-flow storms are suppressed, with enforcement untouched. No blockers; one stale hot-path comment (SHOULD-FIX) and two minor nits. Final reachability is Gate 2 KVM CI, which this PR relies on.

sameerparekh:
## Independent PR review — #1920 (re-run @ 516c1655)

### Prior-findings status
|---|---|---|
| SHOULD-FIX 1 | `render.lua:1084` stale flat-format example comment | **ADDRESSED** — updated to the per-flow `update @wh_drop_log4 { … limit rate 1/minute burst 5 packets } log prefix …` form (commit `516c1655`). |
| NIT 1 | dynamic sets omit explicit `size` cap | NOT-ADDRESSED (accepted — mitigated by precedent + 2m timeout; defensive-only) |
| NIT 2 | PR-body AI footer | NOT-ADDRESSED (accepted per repo norm) |

### Incremental delta (`git diff bf7a4d0b...516c1655`)
Comment-only change (6 lines) updating the hot-path example to match the emitted rule. No code/behavior change; `busted test/render_spec.lua` still 201 passing. No new findings.

### BLOCKERS
(none)

### VERDICT: APPROVE
The one SHOULD-FIX is addressed; the delta is documentation-only with no behavioral impact. Per-flow `(mac,dst)` LOG limiter fix stands as reviewed. No blockers. Final reachability remains Gate 2 KVM CI.

sameerparekh:
Gate 2 doesn't auto-run on PRs (it runs via `workflow_call` from Master Router CD on push to main). To validate the actual paused/blocked-event tests **before** merge, I manually dispatched **VM e2e (fake API — Gate 2)** against this branch (`516c1655`): https://github.com/wifihaven/wifihaven/actions/runs/28067529050

All PR-required checks (OpenWrt Lua Tests, Contract Tests, image build, CodeQL, etc.) are already green. Monitoring the Gate 2 run (~20–40min) through to confirm `test_paused_mac_https_traffic_surfaces_as_nflog_event` + `test_ea_direct_requery_allowed_under_blocked_mac` pass; will report back. Not merging / not arming auto-merge — operator's call.

Issue #1017, opened 2026-05-25
P0: prod traffic_reports empty after ~19:30 MDT 2026-05-24 — investigate


sameerparekh:
## Symptom

Traffic reports on prod (`wifihaven.net`) are empty after **~19:30 MDT today (2026-05-24)** — roughly 01:30 UTC 2026-05-25. Data flowed continuously up to that point and then stopped showing on the Traffic Usage page.

## Why this is critical

Prod telemetry has gone dark. Either:
- The router agent stopped reporting (agent crash, lost auth, lost connectivity, broken update),
- The API ingest started failing or silently rejecting,
- Rows are landing in the DB but a recent read-path change is filtering them out,
- A new aggregation path (rollups) broke,
- Or a TZ regression is mis-bucketing today's late-evening rows.

We need to diagnose the layer first, then fix.

## Recent prod-affecting merges to suspect

Today's deploy queue has shipped a lot. Suspect ranked roughly:
- **#1009 (#958)** blocklists v1 — touched PolicyService snapshot expansion (extraBlocked union). Adjacent to ingest? Probably not, but big change.
- **#970 (#969)** multicast filter — could potentially over-filter the read path. Check the predicate isn't catching non-multicast.
- **#1010-related work** if any partial fix landed mis-typed (chip is queued, may not have landed yet).
- Anything that touched `/api/usage/traffic` or `/api/usage/series` recently — #948 (paging), #918 (responsive), #936 (column filters), #943 (additive aggregation).

## Investigation method

This is a **prod live-debug** — read-only against `api.wifihaven.net` (admin password lives in operator memory at `prod_api_admin_password.md`; redact). Walk the layers in order:

### Layer 1: SPA / read path

- Curl `GET /api/usage/traffic?...` for a known household window straddling 19:30 MDT. Are rows returned?
- If rows are returned: the SPA is mis-rendering. Check the page query / filter state — has any recent merge broken the default `until=now` / cursor anchor?
- If no rows: descend.

### Layer 2: DB

- Find a way to query `traffic_reports` directly. On Render this is via the postgres dashboard or `render shell` access; operator's preference is probably to surface this from an admin endpoint if one exists. Check for `/api/admin/db-stats` or similar; if none, ask the operator for one-shot DB read access.
- Query: `SELECT count(*), max(period_start), min(period_start) FROM traffic_reports WHERE period_start > '2026-05-25T00:00:00Z'`.
[...]

PR #1927, opened 2026-06-24
feat(#1848): agent websocket sidecar (wifihaven-ws) + HTTP fallback

### LINKED ISSUE #1848: ws-transport: agent websocket sidecar (wifihaven-ws) + HTTP fallback (gated on spike)

sameerparekh:
Closes #1848. Sub-issue **C** of the websocket-transport epic — see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §3/§5. Builds on the spike-proven (#1845) `cqueues` + `luaossl` + hand-rolled RFC 6455 client and the #1846 server endpoint. Relates to #1023.

## What

A new **`wifihaven-ws`** procd sidecar — a separate process from the main `wifihaven-agent` (which owns the conntrack loop, **unchanged**). It owns the websocket event loop the agent deliberately lacks (§0.1):

- Opens `wss://…/api/router/ws` with the bearer token in the upgrade `Authorization` header; reconnects with **exponential backoff + jitter** (1s→60s cap, §5.1).
- **Heartbeat** ping/pong (§5.5); the proven `ws_client` answers server pings transparently.
- **Drains the agent's outbound usage/events** over a bounded tmpfs spool (the agent tees them to the spool when ws is healthy, else POSTs as before); **splits an oversized `UsageReport`** across multiple `usage` frames (the #1017 fix, §5.3).
- **Writes a pushed `policy` frame** to the snapshot file the agent applies (etag-guarded apply unchanged). *Dormant until the server-side push lands in #1849* — the code path is present + unit-tested.
- **HTTP fallback** (§3.1): a default-off UCI flag; on upgrade failure or a disconnect past `ws_fallback_after`, the agent resumes HTTP polling (it reads a ws-health sentinel the sidecar maintains).
- **Agent metrics** folded into the existing `/metrics` push via a tmpfs tally (like the dns/sni tails): `ws_connect_total{result}`, `ws_state`, `ws_fallback_total{result}`, `ws_frames_{sent,recv}_total{op}` — all bounded enums — plus a Grafana panel row.

## Default OFF / back-compat

Gated behind **`wifihaven.ws.enabled=0`** (the shipped default). With the flag off the init script never starts the sidecar, `ws_outbound` is a pure pass-through to `http_post`, and the metrics fold is a no-op — so **the main agent's HTTP poll/POST path is byte-for-byte unchanged**. Wire is additive (REST stays fully live; both sides ignore unknown fields). Note: the `hello`/`ready` handshake is **not** built (closed won't-do, §2 status / #1847) — the agent just upgrades and is ready to apply a pushed policy.

Enabling ws needs `cqueues` + `luaossl` on the router (opt-in; **not** a hard package DEPENDS, so default installs stay lean).

## Tests (TDD, red commit first)

New busted specs (pure, injected seams): `ws_backoff`, `ws_usage_split`, `ws_spool`, `ws_loop` (state machine vs a fake client), `ws_metrics`, `ws_outbound`. Shell specs: ws-toggle gating, init nice-count, rotation belt. API: `RouterWsAgentMetricsAllowlistSpec` pins the agent ws series in `MetricGuard.Allowed`. Full OpenWrt suite green (913 busted + all shell specs); API metrics specs green; scalafmt clean.

## plex.lan validation (real Lua 5.1 + cqueues/luaossl)

Validated on plex.lan (192.168.10.43) per the #1845 no-root recipe — not just the macOS 5.5 build, which masks 5.1/cqueues bugs:

```
== ws e2e (Lua 5.1.5) ==          # promoted, qualified-require ws_client
PASS  ws  (ws://127.0.0.1:24035)
PASS  wss (wss://localhost:24036)  # real TLS, self-signed
== ws e2e: ALL PASS ==

# production ws_loop driven against the echo server over a real socket:
[live] connect ok / sent usage / sent events / recv usage / recv events / sent ping
ok=true  sent_ops=usage,events,ping  recv_ops=usage,events  health_touches=5
ws_connect_total.ok=1  ws_state=0  RESULT: PASS

LUA5.1 PURE-MODULE SMOKE: PASS (ws_usage_split chunks=3, records=10; ws_outbound tee)
```


sameerparekh:
## Independent review — #1848 agent ws sidecar

Read-only adversarial pass over `git diff origin/main...HEAD` against the 8 checklist dimensions. First run for this PR. (Author also acting reviewer; findings surfaced during the pass were addressed in-branch and are noted below.)

### BLOCKERS
(none)

### SHOULD-FIX
1. **`deploy/grafana/dashboards/router-ws-transport.json` — full-file reindent (ADDRESSED).** The initial commit round-tripped the dashboard through `json.dump`, reformatting all ~870 lines and burying the real 4-panel change as an unrelated drive-by. Fixed in `fix(#1848): insert grafana ws panels as text` — the file is restored to the exporter formatting and only the 4 new panels + version bump now diff (250+/1-).
2. **Outbound delivery is best-effort-after-send, not at-least-once (ADDRESSED — scoped + documented).** `ws_loop`/`ws_spool` advance the spool offset on send *success* and the tee returns a synthetic 200, so a frame sent-then-lost (drop in the window, or an `ack reject`) is not retried — weaker than the HTTP path's retry-until-2xx. The server already sends `ack` (#1846) but it isn't used to gate spool removal and frames carry no `seq`. Acceptable for the **default-off** path (usage is best-effort per resilience §1; the default-on HTTP fallback carries the at-least-once guarantee), but must close before ws is armed in prod / the poll is deprecated. Softened the wording in `ws_loop.lua`/`ws_outbound.lua` to match reality and filed the ack-gated follow-up **#1928**.

### NITS
1. `openwrt/test/lua_module_paths_spec.sh:40` — excluding `ws_client` from the dev-host load check narrows that guard's coverage of bare-vs-qualified requires for that one file; mitigated by the spike `e2e_test.sh`, which loads it on a real Lua 5.1 + cqueues target (and was re-run on plex.lan this PR). Documented inline.
2. `openwrt/files/usr/sbin/wifihaven-ws` — `flush_metrics()` runs on every metric update (one tmpfs rewrite per frame). The file is tiny/bounded and frame rates are low (usage 60s, heartbeat 30s), so this is fine; could batch to the heartbeat tick if it ever shows up.
3. `openwrt/files/usr/sbin/wifihaven-ws` — the `while true do os.execute("sleep 3600") end` disabled-idle branch is unreachable in practice (the init script gates the instance on `ws.enabled`); it's defensive only.

### Dimension notes
- **1 Single-source-of-truth:** metric names/label-keys are mirrored across the sidecar tally, the agent fold, and the API allowlist — but that's inherent to the existing agent-metrics-fold pattern (dns/sni do the same), not new hand-sync. No re-derivation of policy/time/block-reason. ✓
- **2 Testing:** TDD red commit present (`test(#1848): failing specs…`); new behavior asserted across 6 new busted specs + 2 shell specs + an API allowlist pin; no existing assertions weakened (init nice-count 4→5 reflects the real new instance). ✓
- **3 Invariants:** router stays a dumb applier (no decision logic in the sidecar; pushed policy is just written to the snapshot file). Bounded /tmp: spool self-bounds at `ws_spool_max_bytes` + added to the copytruncate rotation cron; health sentinel + metrics tally are fixed-size atomic rewrites. ✓
- **4 Wire:** additive, default-off; REST path byte-for-byte unchanged with the flag off; ws carries REST payloads verbatim. ✓
- **5 Migrations:** none.
- **6 Metrics+dashboard:** 5 series added to `MetricGuard.Allowed` (bounded `result`/`op` only), pinned by `RouterWsAgentMetricsAllowlistSpec`, with 4 Grafana panels targeting the emitted series. ✓
- **7 Scope/hygiene:** focused; no AI mentions in commits/PR (Co-Authored-By trailer only); follow-up tracked as #1928, not a bare TODO. ✓
- **8 Security:** bearer token read from UCI into the upgrade `Authorization` header (existing pattern); TLS verify-peer in prod (`insecure=false`); no secrets, no SQL. ✓

### Validation
Full OpenWrt suite green (913 busted + all shell specs); API metrics specs green; scalafmt clean. Validated on **plex.lan (real Lua 5.1 + cqueues/luaossl)**: spike e2e ws://+wss:// round-trip PASS, and the production `ws_loop` driven against the echo server over a live socket PASS (connect → drain usage/events out → recv echo → heartbeat → clean drop).

VERDICT: APPROVE

Additive, default-off (`wifihaven.ws.enabled=0`) ws transport sidecar; the main agent HTTP path is unchanged with the flag off. No blockers; the two SHOULD-FIX items were addressed in-branch (Grafana diff minimized; delivery semantics accurately scoped + #1928 filed for ack-gated at-least-once). Validated on the real Lua 5.1 target per the #1845 recipe.

Issue #1512, opened 2026-06-06
perf: policy snapshot (GET /api/router/policy) is ~500ms — evaluate + optimize the build (batch queries / cache)


sameerparekh:
## Symptom
The **policy snapshot** (`GET /api/router/policy`) is slow — **~500ms**. The router ETag-polls this; 500ms per poll is expensive and scales poorly as profiles/devices/apps grow.

## Why this matters
The snapshot is on the router's polling hot path and is recomputed server-side (PolicyService) per request. 500ms suggests heavy per-request query work (per-MAC resolution, schedule expansion, blocklist/app host-set assembly, named-schedule joins after the recent #1482/#1490 changes) with no caching.

## Investigate / evaluate
1. **What's in the 500ms?** Instrument PolicyService snapshot assembly stages (or pull http_request_duration_seconds{route="/api/router/policy"} + db_query_duration_seconds). Attribute the time: DB queries vs in-memory composition.
2. **Per-MAC fan-out:** does it run queries per device/profile/app (N+1) when it could batch and resolve once? The snapshot is "Map[MacAddress, BlockRules] + global" — assemble with set-based queries.
3. **Recent additions:** named-schedule expansion (#1490), per-app schedule eval (#1379), global section (#1318), app host-set attribution (#1505) all add work to the build — make sure they're not each adding a query loop.
4. **Caching:** the snapshot is ETag'd — can the computed snapshot be cached and invalidated on policy change (profiles/devices/schedules/usage-block-state) rather than recomputed every poll? Time-dependent fields (schedule/time-limit block transitions) need care, but most of the build is static between changes.
5. **EXPLAIN** the heaviest queries against prod-like volume; add indexes where missing (per the #1261/#1254 lessons).

## Acceptance
- `GET /api/router/policy` p95 materially reduced (target well under 100ms for the current household), via batched queries and/or a computed-snapshot cache with correct invalidation. EXPLAIN-verified; correctness unchanged (same snapshot bytes for a given state).

Performance epic. Refs: #1318/#1379/#1490/#1505 (snapshot-build contributors), #798 (EXPLAIN audit), #389 (one-ETag tradeoff doc).

PR #1934, opened 2026-06-24
feat(#1849): PolicyService computed-snapshot cache + push-on-change

### LINKED ISSUE #1849: ws-transport: push-on-change + computed-snapshot cache in PolicyService (#1512)

sameerparekh:
## What

Implements the **computed-snapshot cache in `PolicyService`** (#1512) and **push-on-change** for the websocket transport (design [`§6.2`](docs/design/websocket-transport.md)). Sub-issue **E** of the ws-transport epic.

The snapshot used to be recomputed (~500 ms) on **every** `GET /api/router/policy` poll. Now:

- **The REST poll path reads the cache** — a `cache_hit` is served without rebuilding, so un-migrated HTTP agents stop paying the per-poll recompute too. #1512 improves for everyone the moment this lands, independent of transport.
- **On change, the snapshot is rebuilt once and pushed** to every connected ws router (one `policy` frame) — computed once per change, not once per poll per router.

**Behavior-preserving — same snapshot bytes for a given state.** The cache is opt-in: OFF for the `PolicyServiceLive.apply` test factory (so the ~40 existing snapshot specs keep building fresh every call), ON in the production layer.

## How it stays correct (the invalidation set)

A change is **exactly an ETag move** (the existing 200-vs-304 semantics), so nothing new defines "change". Two mechanisms keep the cache fresh:

1. **Explicit `invalidate`** on the policy-authoring mutation routes — pause/unpause, profile create/update/patch/delete, schedule attach + named-schedule edit/delete, device upsert/patch/delete, household settings (unmanaged policy / `blockEncryptedDns`), and `/api/time/extend` — co-located with the existing `TimeStatusCache` busts. Clears the cache (immediate REST freshness) and forks a background reconcile (push) so the mutating request isn't blocked on a rebuild.
2. **A reconcile ticker** (`policy.snapshotCacheRefreshSeconds`, default **5 s = the agent `policy_poll_interval`**) rebuilds and pushes-on-change, catching the **time/usage-dependent transitions** that have no mutation event — schedule boundaries, daily-limit/per-app-cap exhaustion. At 5 s it preserves the pre-cache per-poll freshness exactly while moving the build off the request path.

The push-on-change decision is keyed off a separate `lastPublishedEtag` (not the REST cache slot), so a REST poll racing in to repopulate the cache can never suppress a push.

`reevaluate` is the **single** rebuild-and-publish point — no second snapshot computation is introduced (AGENTS.md single-source-of-truth).

## Transport

- `RouterWsRegistry.publishPolicy` fans one `{op:"policy",payload:<snapshot>}` frame to every connected channel; `RouterWsRegistryLive` implements the new `PolicySnapshotPublisher` seam (kept in the `policy` package so `PolicyService` has no compile dependency on the transport layer).
- `RouterWsRoutes` pushes the current snapshot **on connect** (design §6.1) so a freshly-connected router has policy immediately.

## Metrics + dashboard

- `policy_snapshot_build_total{result=computed|cache_hit}` — proves the cache works (cache_hit should dominate; computed tracks the change/tick rate, not the poll rate).
- `router_ws_policy_push_total{result=ok|channel_closed}` — push fan-out health.

Both via `MetricGuard` (bounded enums, per `docs/process/instrumentation.md`), with two new panels on `deploy/grafana/dashboards/router-ws-transport.json`.

## Query perf

**No new SQL** — the cache reuses the existing snapshot-build queries, so there is no new query plan to EXPLAIN. The change *reduces* DB load (one build per tick instead of one per poll per router).

## Tests (TDD: red commit then green)

- `PolicySnapshotCacheSpec` (embedded Postgres, injected `Clock.TestClock`): cache_hit / stale-until-`invalidate`; `reevaluate` pushes once per ETag change (baseline + no re-push on unchanged); a **schedule boundary crossed with no DB write** is caught + pushed by the ticker mechanism.
- `RouterWsSpec`: a freshly-connected router is pushed the current snapshot on connect.
- Full `mill api.test` suite green.

Fixes #1849. Relates to #1023, #1512.

### DIFF
--- a/api/src/Config.scala
+++ b/api/src/Config.scala
@@ -112,7 +112,17 @@ case class CorsConfig(
 // Precursor to the DB-backed global profile in #937.
 case class PolicyConfig(
     uiAllowedHosts: String = "",
+    // #1849: how often the reconcile ticker rebuilds the cached snapshot and pushes it on an ETag
+    // move. Bounds the staleness of time/usage-dependent transitions (schedule edges, daily-limit
+    // exhaustion) on the cached REST poll path; defaulted to 5s to match the agent's
+    // `policy_poll_interval` so the cache preserves the pre-cache ~per-poll freshness exactly while
+    // moving the ~500ms build off the request path (#1512). Mutations invalidate immediately and do
+    // not wait for this tick.
+    snapshotCacheRefreshSeconds: Int = 5,
 ) {
+  val snapshotCacheRefreshInterval: zio.Duration =
+    zio.Duration.fromSeconds(snapshotCacheRefreshSeconds.toLong)
+
   val uiAllowedHostsParsed: List[Hostname] =
     uiAllowedHosts
       .split(",")
--- a/api/src/Config.scala
+++ b/api/src/Config.scala
@@ -120,8 +120,10 @@ case class PolicyConfig(
     // not wait for this tick.
     snapshotCacheRefreshSeconds: Int = 5,
 ) {
+  // Clamp to a 1s floor so a misconfigured `0`/negative can't turn `Schedule.fixed(Duration.Zero)`
+  // into a no-delay tight loop that pegs the DB with back-to-back ~500ms snapshot builds.
   val snapshotCacheRefreshInterval: zio.Duration =
-    zio.Duration.fromSeconds(snapshotCacheRefreshSeconds.toLong)
+    zio.Duration.fromSeconds(math.max(1, snapshotCacheRefreshSeconds).toLong)
 
   val uiAllowedHostsParsed: List[Hostname] =
     uiAllowedHosts
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -178,6 +178,10 @@ object Main extends ZIOAppDefault {
         // not one ~500ms recompute per poll per router. forkScoped (like the rollup loops) so it is
         // interrupted before the Hikari pool closes on shutdown. The first tick fires immediately;
         // `reevaluate` swallows+logs a transient build failure and keeps the last good cache.
+        // Idle cost: the tick rebuilds unconditionally even with zero connected routers / no recent
+        // poll — but for an always-polling household fleet that is ≈ the pre-cache per-poll load
+        // (which also ran every ~5s), so it is not a regression; gating on "any consumer present" is
+        // a possible later refinement.
         policyForPush <- ZIO.service[PolicyService]
         _ <- policyForPush.setPublisher(new wifihaven.api.policy.PolicySnapshotPublisher {
           def publish(snap: wifihaven.shared.PolicySnapshot): UIO[Unit] =
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -372,10 +372,12 @@ object Main extends ZIOAppDefault {
             // #1538: same cache instance TimeRoutes uses, so a schedule detach busts the
             // per-profile time-status entry instead of leaving a stale "paused for schedule".
             timeCache,
+            // #1849: bust the computed-snapshot cache on a profile/schedule mutation.
+            policy.invalidate,
           ) ++
-          ScheduleRoutes.routes(auth, namedSchedRepo) ++
-          HouseholdSettingsRoutes.routes(auth, hsRepo) ++
-          DeviceRoutes.routes(auth, deviceRepo, upRepo, profileRepo)
+          ScheduleRoutes.routes(auth, namedSchedRepo, policy.invalidate) ++
+          HouseholdSettingsRoutes.routes(auth, hsRepo, policy.invalidate) ++
+          DeviceRoutes.routes(auth, deviceRepo, upRepo, profileRepo, policy.invalidate)
 
       val statsRoutes: Routes[Any, Response] =
         TimeRoutes.routes(
@@ -391,6 +393,8 @@ object Main extends ZIOAppDefault {
           timeStatus,
           clock,
           timeCache,
+          // #1849: a +Time grant can lift a TimeLimit block, so bust the computed-snapshot cache.
+          policy.invalidate,
         ) ++
           LogRoutes.routes(auth, connRepo, upRepo) ++
           UsageRoutes.routes(
--- a/api/src/Main.scala
+++ b/api/src/Main.scala
@@ -182,6 +182,10 @@ object Main extends ZIOAppDefault {
         // poll — but for an always-polling household fleet that is ≈ the pre-cache per-poll load
         // (which also ran every ~5s), so it is not a regression; gating on "any consumer present" is
         // a possible later refinement.
+        // TODO(#1936): replace this fixed-interval ticker with a boundary-scheduled refresh — compute
+        // the next schedule-edge / daily-reset / limit-exhaustion instant from the data buildSnapshot
+        // already loads and sleep until then, so the ~2.5s build runs at real transitions instead of
+        // every tick.
         policyForPush <- ZIO.service[PolicyService]
         _ <- policyForPush.setPublisher(new wifihaven.api.policy.PolicySnapshotPublisher {
           def publish(snap: wifihaven.shared.PolicySnapshot): UIO[Unit] =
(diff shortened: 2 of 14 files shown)


sameerparekh:
## Independent review — PR #1934 (PolicyService computed-snapshot cache + push-on-change)

Reviewed the full merge-base diff (`git diff origin/main...HEAD`) against the 8 dimensions.

### BLOCKERS
1. **`api/src/metrics/Metrics.scala:514,520` — the two new metrics are silently rejected at runtime (dead series).** `policy_snapshot_build_total{result}` and `router_ws_policy_push_total{result}` are emitted through `MetricGuard.counter(...)` (correct), but neither name is added to `MetricGuard.Allowed`. `MetricGuard.check` does `Allowed.get(name) match { case None => reject("unknown_name") }` (`Metrics.scala:267-268`), so both series are dropped and counted as `metrics_rejected_total{reason=unknown_name}` — they never reach Prometheus. The two new Grafana panels added in this PR would show **no data**. `MetricCardinalityGuardSpec` did NOT catch this because its "every series is allowlisted" test scans only **bare `Metric.*`** emits (`BareEmit` regex), not `MetricGuard.counter` call sites — so the suite is green while the metrics are dead. This makes the §6 metric+dashboard requirement (the headline observability of the PR) unmet in practice. Fix: add `"policy_snapshot_build_total" -> Set("result")` and `"router_ws_policy_push_total" -> Set("result")` to `MetricGuard.Allowed` (the `result` key is already in `KnownLabelKeys`).

### SHOULD-FIX
1. **`api/src/Config.scala` — `snapshotCacheRefreshSeconds` has no lower bound.** A misconfig to `0` (or negative) makes `Schedule.fixed(Duration.Zero)` a tight, no-delay rebuild loop that pegs the DB with back-to-back ~500ms snapshot builds. Clamp the derived interval to a sane minimum (e.g. `max(1, snapshotCacheRefreshSeconds)`).
2. **PR body** contains a "🤖 Generated with Claude Code" line, which dimension 7 disallows in the PR title/body (the `Co-Authored-By` commit trailer is the allowed form). Remove it.

### NITS
1. **`api/src/Main.scala` (reconcile ticker)** rebuilds the snapshot every `snapshotCacheRefreshSeconds` unconditionally — even with zero connected ws routers and no recent REST poll, it runs a ~500ms DB build at idle. For an always-polling household fleet this is ≈ the pre-cache per-poll load (no regression), but it could later be gated on "any consumer present." Non-blocking; worth a comment acknowledging the idle cost.
2. **`PolicyService.invalidate`** forks a daemon `reevaluate` per mutation; a burst of mutations spawns concurrent rebuilds. Rare in the two-tier household model; a debounce/serialize could be a later refinement.

### Notes on the other dimensions (clean)
- **§1 single-source:** `reevaluate` is the single rebuild-and-publish point; the cache reuses the existing `buildSnapshot` queries (no second snapshot computation). The separate `lastPublishedEtag` is push-dedup state, not a re-derivation of the snapshot. Good.
- **§2 testing:** TDD red→green visible (red stub commit `test(#1849)…` then green). `PolicySnapshotCacheSpec` asserts cache_hit / stale-until-invalidate, push-once-per-etag-move, and a clock-driven schedule-boundary re-eval with injected `TestClock`; `RouterWsSpec` asserts connect-push. Cache is opt-in (off for `apply`) so existing snapshot specs still build fresh. No weakened/deleted assertions. (One gap that would have caught the BLOCKER: no test asserts the new metrics actually emit — consider one.)
- **§3 invariants:** no new snapshot wire field (the `policy` frame reuses the existing `PolicySnapshot`); `Clock` injected; ZIO effects, typed errors; not a router-agent `/tmp` write.
- **§4 wire:** additive — `{op:"policy",payload:<snapshot>}` is a frame the #1848 agent already handles; connect-push is additive and agents demux by op.
- **§5 migrations:** none. **No new SQL** (cache reuses existing build queries) → no EXPLAIN obligation; the change reduces DB load.
- **§8 security:** auth runs before the ws upgrade/connect-push; no secrets; no SQL interpolation (the frame JSON is built from `snap.toJson`, not SQL).

### VERDICT: REQUEST-CHANGES
Solid, behavior-preserving cache + push-on-change with good test coverage and a clean single-source design. One merge-gating BLOCKER: the two new metrics aren't registered in `MetricGuard.Allowed`, so they're silently dropped and the new dashboard panels would be empty — defeating the PR's observability goal. Fix the registration (+ the two SHOULD-FIXs) and this is an APPROVE.

sameerparekh:
## Independent review — re-run (PR #1934)

Re-run against the delta `git diff 760b289...100300c` (the review-fix push). Prior findings statused below, then the delta reviewed for new findings.

### Prior findings — status
|---|---|---|
| BLOCKER 1 | `policy_snapshot_build_total` / `router_ws_policy_push_total` not in `MetricGuard.Allowed` → silently rejected | **ADDRESSED** — both registered (`Metrics.scala:210-215`, `result` key). A new `GuardEmit` scan in `MetricCardinalityGuardSpec` now fails the build for any `MetricGuard.*("literal")` name missing from the allowlist (closes the test gap that hid this). |
| SHOULD-FIX 1 | `snapshotCacheRefreshSeconds` unbounded → `Schedule.fixed(0)` tight loop | **ADDRESSED** — clamped to a 1s floor (`Config.scala`, `math.max(1, …)`). |
| SHOULD-FIX 2 | AI-mention line in PR body | **ADDRESSED** — removed from the PR body (Co-Authored-By trailer retained, which is allowed). |
| NIT 1 | ticker idle-build cost | **ADDRESSED** — documented in `Main.scala` as ≈ pre-cache per-poll load, not a regression. |
| NIT 2 | `invalidate` forkDaemon per mutation | NOT-ADDRESSED — left as-is (acceptable in the household model; a later debounce is optional). |

### New findings in the delta
BLOCKERS: none.
SHOULD-FIX: none.
NITS: none. The fixes are minimal and on-point — registration entries, a 1-line clamp, one new allowlist-scan test mirroring the existing `BareEmit`/`AgentEmit` pattern, and a comment.

### VERDICT: APPROVE
The one BLOCKER (dead metrics) is fixed and now regression-guarded; both SHOULD-FIXs resolved. The cache + push-on-change is behavior-preserving, single-source (one rebuild-and-publish point, no second snapshot computation), additive on the wire, and metric+dashboard-complete. No open BLOCKER.

sameerparekh:
## Independent review — re-run (PR #1934)

Delta since last review is a **single comment-only commit** (`Main.scala`): a `TODO(#1936)` linking the fixed reconcile ticker to the boundary-scheduled-refresh follow-up. No code/behavior change.

### Prior findings — status
|---|---|---|
| BLOCKER 1 | metrics not in `MetricGuard.Allowed` | **ADDRESSED** (registered + `GuardEmit` regression test) — unchanged |
| SHOULD-FIX 1 | `snapshotCacheRefreshSeconds` unbounded | **ADDRESSED** (1s clamp) — unchanged |
| SHOULD-FIX 2 | AI-mention in PR body | **ADDRESSED** — unchanged |
| NIT 1 | ticker idle-build cost | **ADDRESSED** — now also linked to follow-up #1936 |
| NIT 2 | `invalidate` forkDaemon | NOT-ADDRESSED (acceptable) — unchanged |

### New findings in the delta
None — comment only. (The known ~2.5s-per-tick cost is now tracked in #1936 per the operator's merge-now + follow-up decision.)

### VERDICT: APPROVE
No open BLOCKER. Comment-only delta; the prior APPROVE stands.

PR #1933, opened 2026-06-24
fix(#1929): pin allow-path e2e probes to IPv4 — dual-stack leaf drifted to v6 redirect

sameerparekh:
## Summary
Master Router CD has been **red since 2026-06-22 13:45Z** (last green `1665d720`, v0.3.13). After #1920 fixed the #1915 starvation cause, Gate 2 still failed **deterministically** on a second test, so nothing publishes:

```
FAILED scenarios_fake/test_cname_direct_requery.py::test_ea_direct_requery_allowed_under_blocked_mac
  - TimeoutError: timed out waiting for allowed HTTP response for e2e-edge.wifihaven.net
```

This blocks the #1865 `::1:8443` fix, #1864, #1848 ws sidecar, and #1921 from ever shipping past v0.3.13.

## Diagnosis — test-harness probe drift, **NOT** an ea_ carve regression
The leaf `e2e-edge.wifihaven.net` is **dual-stack**: `A 192.0.2.10` **and** `AAAA 2001:db8::10`. The AAAA was added in #1677 (v6-attribution scenario), *after* this suite was authored against a v4-only leaf in #1360.

1. Plain `curl` happy-eyeballs **prefers the v6 address**, so the allow-path probe connects over IPv6.
2. The test populates and asserts only the **v4 `ea_`** set. The **v6 `ea6_`** carve set is **empty at curl-connect time** — curl's own AAAA lookup is the first v6 resolution, so dns-tail hasn't populated `ea6_` yet.
3. #1868 (#1865) switched the v6 block-page path from `dnat ip6 to ::1` (never delivered) to nft `redirect` (**does** deliver to the local uhttpd listener). So the v6 redirect now serves the **block page** for the uncarved v6 leaf → the allow assertion fails. *Pre-#1868 the same miss merely timed out → the test xfailed (green).*

### Why this is NOT a carve regression (ruled out on a real router)
- **Behavioral proof:** on green (`1665d720`) G2 xfailed = leaf **unreachable/None**, NOT a block page. The v4 DNAT target `127.0.0.1:8081` *does* deliver, so a broken v4 carve would have block-paged on green and red-gated it. It didn't.
- **Code proof:** `render.lua` carves both families — `ip daddr != @ea_…` (DNAT) and `ip6 daddr != @ea6_…` (redirect); `dns_tail_sets.maybe_populate_ea` branches on `r.family` and populates `ea6_` for AAAA replies.
[...]

Issue #1929, opened 2026-06-24
bug(router CD): Gate 2 test_ea_direct_requery_allowed_under_blocked_mac fails deterministically — ea_ carve-out HTTP times out / hits block page (likely #1868 regression) — blocks ALL router publishes incl. ::1:8443 fix


sameerparekh:
## Summary
Master Router CD has been **red since 2026-06-22 13:45Z** (last green sha `1665d720`, which published v0.3.13). #1915 (event-synthesis starvation) was **one** cause and is now fixed by #1920 (that test `PASSED [100%]`). But Gate 2 still fails **deterministically** (2 consecutive runs, pre- and post-#1920) on a **second, distinct** test, so nothing publishes — and the **#1865 `::1:8443` block-page fix stays unshipped on prod**.

```
FAILED scenarios_fake/test_cname_direct_requery.py::test_ea_direct_requery_allowed_under_blocked_mac
  - TimeoutError: timed out waiting for allowed HTTP response for e2e-edge.wifihaven.net
  → 1 failed, 45 passed, 1 skipped, 1 xfailed
```

## Mechanism
The test asserts the **ea_ allow-carve under a whole-MAC block** (`blocked=True` + `extraAllowed=[BRAND_HOST]`, the #1344 Math-Academy regression guard). Sequence:
1. **ea_ set membership PASSES** — the directly-requeried leaf IP (`192.0.2.10`) lands in `ea_<mac>_<brand>` (populator proof, line ~350). So DNS-side carve is intact.
2. **End-to-end HTTP FAILS** — `_xfail_if_leaf_unreachable(client)` (line 367) probes `http://LEAF_HOST/` with an 8s `http_get` and only xfails when `http_code is None`. It did **not** xfail here ⇒ the probe got a **non-None** code — almost certainly the **block page**. Then `wait_http_succeeds` (rejects block-page bodies) timed out at 120s ⇒ hard FAIL.

Getting the **block page** for an IP that is in the `ea_` allow-set means the per-MAC forward-chain carve-out (`ip daddr != @ea_<mac>_<brand>`) is **not actually carving the traffic** — it's being DNAT'd to the block page instead. The ea_ membership proves the set is populated; the HTTP step proves the **rule using that set isn't carving**.

## Leading hypothesis: #1868 regressed the carve
- The **test file is unchanged since 2026-06-04** (#1360) — the test didn't regress.
[...]

PR #1921, opened 2026-06-24
feat(#1911): enforce blockEncryptedDns in the router agent (NODATA relay/DoH + nft drop DoT/853 & :53-to-resolver-IPs)

sameerparekh:
Fixes #1911. Sub-issue of feature #1909, epic #1903.

Router agent enforces the additive, network-wide **`blockEncryptedDns`** toggle. Keeps the agent a dumb applier — all policy is the single snapshot flag; curated lists are baked into the agent.

## Wire contract (matches API PR #1916 / #1912)
- New **top-level** boolean `blockEncryptedDns` (sibling of `global`/`devices`/`profiles`/`blocklists`), additive, default `false`. This PR only **reads** it. Old agents ignore it; absent → false.
- Curated relay/DoH hostnames + public-resolver IPs are **baked into the agent** ([`encrypted_dns.lua`](openwrt/files/usr/lib/lua/wifihaven/encrypted_dns.lua)), not on the wire.

## Behaviour when `blockEncryptedDns == true` (two separable halves)
1. **NODATA half (dnsmasq):** `local=/<host>/` → a **negative** DNS answer (NXDOMAIN/NODATA, **not** a `0.0.0.0` sinkhole) for `mask.icloud.com`, `mask-h2.icloud.com`, `cloudflare-dns.com`, `dns.google`, `dns.quad9.net`, `dns.nextdns.io`, `dns.adguard-dns.com`, `doh.opendns.com`. This is Apple's documented clean way to disable iCloud Private Relay (a sinkhole does **not** disable it — the #1891 hang).
2. **Connection-layer half (nft):** a dedicated `wifihaven_encrypted_dns` forward chain dropping **DoT (TCP/853 to any IP)** + **DNS port 53 (UDP+TCP) to the curated resolver IPs only** (`1.1.1.1`,`1.0.0.1`,`8.8.8.8`,`8.8.4.4`,`9.9.9.9`,`149.112.112.112` +v6). **Deliberately NOT dropping `:443`/ICMP** to those IPs — they double as universal "am I online?" connectivity-check targets (#1909 port-scope refinement).

Absent/false → byte-identical no-op.

## Truth #1 exception
NODATA for these hostnames is the one sanctioned, narrow exception to Architectural Truth #1 (DNS-never-enforcement), scoped to bypass-disable *signaling*. Documented in [`docs/architecture.md`](docs/architecture.md) §0.1 and in code comments.

## Tests (TDD)
[...]

Issue #1911, opened 2026-06-23
feat(agent): enforce global.blockEncryptedDns (NXDOMAIN relay/DoH hosts + nft drop resolver IPs & DoT/853)


sameerparekh:
Sub-issue of #1909 (feature) / #1903 (epic). **Independent — no required landing order** (additive field + Lua agent ignores unknown keys + default-off).

Router agent enforces the additive `global.blockEncryptedDns` flag. When true:
- **dnsmasq NODATA/NXDOMAIN** for curated relay/DoH **hostnames**: `mask.icloud.com`, `mask-h2.icloud.com`, `cloudflare-dns.com`, `dns.google`, `dns.quad9.net`, NextDNS, AdGuard, OpenDNS. (Apple's clean Private-Relay disable signal.)
- **nftables drop** for curated resolver **IPs** (`1.1.1.1`/`1.0.0.1`/`8.8.8.8`/`8.8.4.4`/`9.9.9.9` +v6) and **DoT TCP/853** to any IP.
- Curated lists **baked into the agent** (small, stable; keeps the wire to a single bool).
- Flag absent/false → no-op.

Validate on a real router (Lua 5.1). Document the deliberate Truth-#1 exception (NXDOMAIN for these hostnames) in `docs/architecture.md`. Works as a no-op until the API emits the field, so it can ship before/with/after the API change.

Epic: DNS-Bypass & Attribution.

sameerparekh:
Implementation note (see #1909 blast-radius comment): the e2e MUST assert **opportunistic fallback** — after the flag is on, a device that previously used DoH still resolves via the LAN resolver and reaches the internet. Also keep the **NXDOMAIN-hostnames** half and the **drop-IPs** (1.1.1.1/8.8.8.8 — dual-use connectivity-check targets) half separable, and confirm OS captive-portal hostnames stay reachable (they're in global.extraAllowed).

sameerparekh:
## IMPORTANT refinement: port-scope the hardcoded-IP block (avoid connectivity-check collateral)

Do **NOT** drop all traffic to the public-resolver IPs. Many devices (e.g. the household's Skylight Calendar, which already flakes "internet not working / power-cycle" and touches these IPs) use `1.1.1.1`/`8.8.8.8` as their "am I online?" probe — a blanket drop makes them think they're permanently offline.

Block only the **DNS ports**, leaving connectivity probes (ICMP / :443) intact:
[...]

PR #1937, opened 2026-06-24
fix(#1935): make blockEncryptedDns e2e provable without external egress

### LINKED ISSUE #1935: bug(router CD): Gate 2 test_block_encrypted_dns_enforced precondition needs real egress to 8.8.8.8 — fails in isolated fake VM (blocks router publishes alongside #1929)
## Summary
Now that the agent boots again (#1931), Gate 2 runs the full fake-mode suite and shows **exactly 2 failures** (`2 failed, 47 passed, 1 skipped, 1 xfailed`): the ea_ probe drift (#1929, fix in PR #1933) **and this one**. Both must land for router CD to go green and publish past v0.3.13 (shipping #1865 ::1:8443, #1864, #1848 ws sidecar, #1921 enforcement).

```
FAILED scenarios_fake/test_encrypted_dns.py::test_block_encrypted_dns_enforced
  test_encrypted_dns.py:98: AssertionError: precondition: with the toggle OFF, DNS to 8.8.8.8 must reach it
  (rc=9 ';; communications error to 8.8.8.8#53: timed out\n;; no servers could be reached')
```

## Root cause — harness egress assumption, NOT an enforcement bug
The test (added by #1921 / #1911) opens with a **precondition** to prove the later :53 drop is caused by the toggle and not the environment:
```python
RESOLVER_IP = "8.8.8.8"; NORMAL_HOST = "example.com"
base = _dig_via(client, NORMAL_HOST, RESOLVER_IP)   # dig +short @8.8.8.8 example.com
assert _ipv4_answers(base.stdout), "precondition: with the toggle OFF, DNS to 8.8.8.8 must reach it"
```
This requires the client to reach the **real public 8.8.8.8 resolver over the internet**. The Gate 2 fake-mode KVM VM is **network-isolated** — the entire suite uses RFC 5737 TEST-NET origins and harness-served DNS *precisely because there is no external egress* (see test_cname_direct_requery's TEST-NET leaf rationale, #1360). So `dig @8.8.8.8` times out ("no servers could be reached") **regardless of the toggle**, and the precondition can never hold in CI.

Evidence it is NOT a mis-gated drop: with the toggle OFF the agent logs **no** encrypted-DNS enforcement application; the failure is the bare precondition, before any snapshot with `blockEncryptedDns=true` is served. The enforcement itself (NODATA for relay/DoH via dnsmasq-local; :53→resolver drop; DoT/853 drop) is fine on a real router that has egress — this is the same class as the #1930 upvalue bug: validated where real internet exists, only fails in the isolated runner.

## Fix direction
Make the encrypted-DNS e2e provable **without external egress**, consistent with how ea_/eb_/bl_ are proven:
- Use a **harness-reachable resolver IP** the VM can actually talk to (a local responder the fake harness serves on the curated RESOLVER_IP, or a TEST-NET IP the harness answers :53 on) instead of the real 8.8.8.8; OR
- Assert the **:53-to-resolver drop via the rendered nft ruleset membership** (origin-independent), and keep only the dnsmasq-local NODATA half (relay/DoH → no IP) and the LAN-resolver fallback as live checks — those need no egress.
- The toggle-OFF baseline must use the same harness-reachable path so it proves "toggle causes the drop" without depending on the public internet.

The fix chip MUST confirm the encrypted-DNS enforcement itself is correct on a real router (openwrt.lan / router.lan, which have egress) and only correct the **test's environment assumption** — do not weaken the enforcement or the regression guard.

## Acceptance
- `test_block_encrypted_dns_enforced` passes in the isolated Gate 2 fake VM (no dependency on reaching the real public internet).
- The NODATA, :53-drop, DoT/853-drop, and LAN-fallback behaviors of #1911 remain asserted.
- Combined with #1933, Gate 2 goes fully green; Master Router CD publishes past v0.3.13.

Relates to #1921 / #1911, #1929 / PR #1933, #1930, #1360. Found while watching the #1915 CD incident.

sameerparekh:
## Problem

`scenarios_fake/test_encrypted_dns.py::test_block_encrypted_dns_enforced` is one of the two remaining Gate-2 blockers keeping **Master Router CD red** and the OpenWRT release stuck at **v0.3.13** (so the #1865 `::1:8443` fix stays unshipped). The other blocker, #1929 / #1933, is already merged (`origin/main` is at it).

The test failed at its **precondition**, not in enforcement:

```python
base = _dig_via(client, NORMAL_HOST, "8.8.8.8")   # dig +short @8.8.8.8 example.com
assert _ipv4_answers(base.stdout)                  # times out → empty → FAIL
```

The Gate-2 fake-mode KVM VM is **network-isolated**. The router's dnsmasq is the only reachable resolver (its upstream is pinned through QEMU SLIRP — see `scenarios_fake/conftest.py`), and a **client-originated** packet to a public IP (`dig @8.8.8.8`, `curl https://1.1.1.1`) cannot leave the harness **regardless of the toggle**. So the precondition could never hold in CI — and it failed *before* any `blockEncryptedDns=true` snapshot was even served. This is a test-harness egress assumption, **not** an enforcement bug.

## Fix

Rework the suite to be **origin-independent**, exactly as the `ea_`/`eb_`/`bl_` suites prove their populators via nft membership (`test_cname_direct_requery.py`), keeping the regression guard intact (enforcement code untouched):

- **Drop + port-scope assertions** are made against the **rendered + loaded kernel ruleset** (the `wifihaven_encrypted_dns` chain): DoT/853 drop, the `:53→curated-resolver-IP` drop (v4 + v6, udp + tcp), and the `:53`-scope (`no dport 443`, no `icmp`) that keeps connectivity probes alive. Asserting the chain is **loaded** proves enforcement is live for any matching forwarded flow — without needing client→public egress.
- **Toggle-causation** is proven on that same observable: the chain is **ABSENT** with the toggle OFF and **PRESENT** (with the drop rules) when ON. Same no-egress observable on both sides, so it proves the toggle — not the environment — causes the drop.
- The two halves that need **no egress** stay **live checks**: dnsmasq **NODATA** for the curated relay/DoH hostnames, and the opportunistic **LAN-resolver fallback** (`example.com` still resolves after the toggle is on — the "did we break the network" assertion).

### All #1911 behaviours stay asserted

| Behaviour | Before | After |
|---|---|---|
| NODATA for relay/DoH | live (`dig`) | live (`dig`) — unchanged |
| `:53`→resolver drop | `dig @8.8.8.8` (needs egress) | live kernel ruleset membership |
| DoT/853 drop | *not asserted* | live kernel ruleset membership (**new coverage**) |
| `:443`/ICMP survive | `curl https://1.1.1.1` (needs egress) | rendered-ruleset property (no `dport 443`/`icmp`) |
| LAN-resolver fallback | live (`dig`) | live (`dig`) — unchanged |

The `:443` check matches the `dport 443` token specifically, since the live dump carries dynamic `counter packets N bytes N` values a bare `"443"` substring could spuriously hit.

## Real-router confirmation (before merge)

Validated the #1911 enforcement is correct independent of this test change:

1. **Render on the real Lua 5.1 target** (`api.lan`, no-root lua5.1): `render.nft`/`render.dnsmasq` emit the exact chain + DoT/853 + resolver:53 (v4+v6) + `local=/mask.icloud.com/`/`local=/dns.google/` NODATA lines; toggle OFF → byte-identical no-op. `_VERSION = Lua 5.1`.
2. **Rendered chain loads into a live OpenWRT kernel** (`openwrt.lan`): `nft -f` accepts it and re-renders identically — `dport 53`/`dport 853` present, **no `dport 443`**, comments (`wh_encrypted_dns:dot853` / `:resolver53`) preserved. (Throwaway table, deleted after.)
3. **dnsmasq half** (`openwrt.lan`, temp conf, reverted): a `local=/host/` directive returns a **negative answer (no Address)** while `example.com` still resolves via the LAN resolver — both with **no** client→public egress.

## Scope

Touches only the e2e Python file — no enforcement, no Lua, no wire change. Stays in the existing Gate-2 harness (no new docker-compose CI job).

Fixes #1935. Relates to #1911 / #1921 (enforcement), #1929 / #1933 (the other Gate-2 blocker).

### DIFF
--- a/scripts/e2e/scenarios_fake/test_encrypted_dns.py
+++ b/scripts/e2e/scenarios_fake/test_encrypted_dns.py
@@ -3,7 +3,9 @@
 VM-level end-to-end proof that a deployed agent enforces the additive top-level
 `snapshot.blockEncryptedDns` flag (feature #1909, epic #1903). The busted unit
 coverage (`openwrt/test/encrypted_dns_spec.lua`) proves the *render shape*; this
-proves the *enforced behaviour against live traffic on a real OpenWRT agent*.
+proves the *enforced behaviour against a real OpenWRT agent* — the chain is
+actually rendered AND loaded into the live kernel ruleset, and dnsmasq actually
+answers the curated hostnames locally.
 
 When the toggle is on the agent enforces two separable halves:
 
@@ -15,15 +17,33 @@ When the toggle is on the agent enforces two separable halves:
      to the curated public-resolver IPs ONLY — catches "system DNS hardcoded to
      8.8.8.8" and forces fallback to the LAN resolver.
 
-The two assertions that matter most (per the #1909 blast-radius + port-scope
+The behaviours this suite locks down (per the #1909 blast-radius + port-scope
 refinements):
+  - **NODATA for the curated relay/DoH hostnames** — live check via the LAN
+    resolver (dnsmasq answers locally; no egress).
+  - **:53→curated-resolver-IP drop** — asserted on the LIVE kernel ruleset.
+  - **DoT/853 drop** — asserted on the LIVE kernel ruleset.
   - **Opportunistic fallback survives** — after the toggle is on, a normal name
-    still resolves via the LAN resolver and the network is NOT broken. This is
-    the core "did we break the network" assertion.
-  - **Connectivity probes survive** — `:443`/ICMP to `1.1.1.1` still works,
-    because we port-scope the resolver-IP drop to `:53`. `1.1.1.1`/`8.8.8.8`
-    double as the world's most common "am I online?" targets; a blanket drop
-    would make such devices believe they are permanently offline.
+    still resolves via the LAN resolver and the network is NOT broken. THE core
+    "did we break the network" assertion (live check; no egress).
+  - **Connectivity probes survive** — the resolver-IP drop is `:53`-scoped, so
+    `:443`/ICMP to `1.1.1.1`/`8.8.8.8` is never dropped. Asserted as a property
+    of the rendered ruleset (no :443/icmp drop), not by dialing a public IP.
+
+── Why this suite is egress-independent (#1935) ─────────────────────────────
+The Gate-2 fake-mode VM is network-isolated: the router's dnsmasq is the only
+reachable resolver (its upstream is pinned through QEMU SLIRP — see
+scenarios_fake/conftest.py), and a *client*-originated packet to a public IP
+(`dig @8.8.8.8`, `curl https://1.1.1.1`) cannot leave the harness REGARDLESS of
+the toggle. The original precondition `dig @8.8.8.8 example.com` therefore timed
+out in CI and red-gated every router release even though the #1911 enforcement
+was correct. The drop / port-scope assertions are now made against the RENDERED
++ LOADED kernel ruleset — origin-independent, exactly as the ea_/eb_/bl_ suites
+prove their populators via nft set/rule membership (see
+test_cname_direct_requery.py). The toggle-causation is proven on that same
+observable: the `wifihaven_encrypted_dns` chain is ABSENT with the toggle off
+and PRESENT (with the drop rules) when it is on. The two halves that need no
+egress — dnsmasq NODATA and the LAN-resolver fallback — stay live checks.
 
 Reachability here is a connection-layer assertion, never "DNS resolved ⇒
 allowed". See docs/architecture.md §0.1 and the AGENTS.md anti-pattern callout.
@@ -32,7 +52,7 @@ from __future__ import annotations
 
 import pytest
 
-from lib.vm import client_exec, router_ssh
+from lib.vm import router_ssh
 from lib.wait import wait_until
 
 from ._observers import dig_ipv4_answers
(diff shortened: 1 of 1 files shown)


sameerparekh:
## Independent PR review — `test_encrypted_dns.py` egress rework (#1935)

Read-only adversarial pass over `git diff origin/main...HEAD` (one file:
`scripts/e2e/scenarios_fake/test_encrypted_dns.py`) against the 8 dimensions.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. `test_encrypted_dns.py:70-75` — the curated resolver IPs (`8.8.8.8`,
   `1.1.1.1`, `2606:4700:4700::1111`), the chain name, and the comment strings
   (`wh_encrypted_dns:dot853` / `:resolver53`) are hardcoded, duplicating
   `openwrt/.../encrypted_dns.lua`. This is intentional black-box pinning of the
   kernel-visible contract (the busted spec hardcodes the same constants), but if
   the curated lists change the e2e drifts silently. Acceptable; calling it out.
2. `test_encrypted_dns.py:_encrypted_dns_chain_lines` — relies on the nft dump
   putting the chain-closing `}` on its own line while the anonymous resolver
   sets (`{ … }`) stay balanced within a single rule line. That holds for
   `nft list table` output (verified against a live OpenWRT kernel), but it's a
   format assumption worth a comment — which the docstring already carries.
3. Baseline now asserts `example.com` resolves via the LAN resolver with the
   toggle OFF (`:188`). This adds a dependency on the router's SLIRP-pinned
   upstream — but it's the *same* dependency the cname suite already relies on,
   and strictly weaker than the old precondition (which dialed `8.8.8.8`
   directly). Net reduction in environmental coupling.

### Dimension notes
- **1 SSoT:** no logic re-derivation; constants are test expectations, not a
  computed quantity. Not a finding.
- **2 Testing integrity (critical):** the removed assertions (`dig @8.8.8.8`
  precondition, the `:53`-drop-via-`dig`, the `curl https://1.1.1.1` probe) were
  all **unreachable in CI** — the test died at the first (precondition) line, so
  in practice they asserted nothing. They are replaced by origin-independent
  checks on the **live kernel ruleset** (chain ABSENT toggle-OFF → PRESENT with
  drop rules toggle-ON) plus the two no-egress live checks (NODATA, LAN-resolver
  fallback). The PR body explains why the old precondition was wrong (harness
  isolation), satisfying the "test weakening must be justified" gate. **Coverage
  net-increases**: DoT/853 drop and the v6 resolver:53 drop are now asserted (the
  old test asserted neither). Asserting a loaded forward-hook drop rule is a
  genuine enforcement-state assertion, consistent with the blessed ea_/eb_/bl_
  nft-membership precedent for origin-independent cases.
- **3 Architecture:** DNS-never-enforces respected — blocking stays
  connection-layer (nft), the NODATA half is treated as the sanctioned #1911
  exception, no "resolved ⇒ allowed" reasoning, no router decision logic added.
- **4–6 Wire / migrations / metrics:** N/A (test-only, no wire/schema/metric).
- **7 Scope/hygiene:** focused one-file diff; removed-and-now-unused helpers
  (`_dig_via`, `_ipv4_answers`) and the `client_exec` import are cleanly deleted,
  no dead code; no AI mentions in title/body (Co-Authored-By trailer only).
- **8 Security:** N/A.

### Real-router corroboration (from PR body, independently plausible)
Render correct on real Lua 5.1; rendered chain loads into a live OpenWRT kernel
and re-renders identically (`dport 53`/`853`, no `dport 443`); dnsmasq
`local=/host/` returns a negative answer while `example.com` still resolves —
all consistent with the assertions this test now makes.

VERDICT: APPROVE
Test-only fix that removes a network-egress assumption which red-gated every
router release; the rework is origin-independent, net-increases #1911 coverage,
and weakens no assertion that was actually exercising behaviour in CI. No
blockers; the three NITs are non-gating.

sameerparekh:
Closing — the premise behind this PR was wrong.

I verified against the actual Gate-2 logs (and the runner host, api.lan, which IS the GitHub runner): the fake-mode VMs are **not** network-isolated. External DNS egress is **intermittent**, and that — not isolation — is what reddens Gate-2:

- **14:05 run**: `test_block_encrypted_dns_enforced` failed at the `dig @8.8.8.8` precondition (`no servers could be reached`) AND the router dnsmasq smoke check logged `nil for tiktok.com` AND the cname `test_eb_direct_requery_blocked` failed — all three = no external DNS at that moment.
- **00:47 run**: `test_block_encrypted_dns_enforced` **PASSED** (client reached 8.8.8.8 fine) — proving the precondition is not structurally impossible.
- Right now `dig @8.8.8.8` from the runner host works fine.

So the real root cause is flaky runner→public-resolver egress (the router VM pins dnsmasq upstream to 8.8.8.8/1.1.1.1 with `noresolv=1`), which also hits the cname tests and others. Rewriting only the encrypted-DNS assertions treats a symptom and wouldn't make Gate-2 reliably green. Re-running Master Router CD on the current main (which already has #1933) to confirm a clean pass.

Issue #1899, opened 2026-06-23
shared-hosts S4: most-permissive enforcement for shared hosts (allow-in, never block)


sameerparekh:
Depends on #1896 (S1b); parallel to S2/S3. See `docs/design/shared-host-allocation.md` (PR #1894), parent #1888.

Enforcement for shared hosts follows a single **most-permissive** rule — it does NOT depend on co-presence (that gates *attribution* only). Implement in `PolicyService.computeBlockRules` (`api/src/policy/PolicyService.scala:771`):

> A shared host follows the most permissive disposition of any app that lists it. App in ALLOW → shared host goes into `extraAllowed`. App in BLOCK → shared host is NOT put into `extraBlocked`.

- **Allow side — shared host added to `extraAllowed` unconditionally** when any app listing it resolves to `Allowed` enforcement for the profile (subject to `allowed_during_schedule_block`, #1679). No co-presence gate. Lets an Allowed-mode app function (e.g. Feeling Great carved around a Schedule block needs elevenlabs reachable). `extraAllowed` already beats every block path, so over-allowing is safe.
- **Block side — shared host NEVER enters a drop set.** Exclude from `appCapExhaustedHosts` (`:878`) and from a `Blocked`-mode app's `extraBlocked`. Blocking a shared backend on one app's block/cap-exhaustion would drop it for every other app — the #1636 collateral failure. Only distinctive hosts are block-eligible; the cap is still enforced through the distinctive hosts.
- **No new wire field/concept** — reuses the existing `extraAllowed`. Router stays a dumb applier (it can't express a co-presence-conditional allow; the unconditional most-permissive choice is the only safe one).

Enforcement decision table:

| App disposition | Distinctive | Shared |
|---|---|---|
| Allowed | extraAllowed | **extraAllowed** (unconditional) |
| Blocked (manual) | extraBlocked | **omitted** |
| TimeLimited, cap exhausted | extraBlocked | **omitted** |
| TimeLimited, under cap (exempt) | extraAllowed carve | **extraAllowed** if exempt-carve applies |

Tests pin: cap exhaustion / manual block never drops a shared host; Allowed-mode shared host carves into `extraAllowed`.

This is the snapshot-path restatement of the `_README.yml` shared-pool collateral rule ("over-allow safe, over-block collateral").

Issue #1897, opened 2026-06-23
shared-hosts S2: exclude shared hosts from per-app engaged-minutes stitch


sameerparekh:
Depends on #1896 (S1b). See `docs/design/shared-host-allocation.md` (PR #1894), parent #1888.

Make per-app engaged minutes (cap + daily-cap exemption) stitch **distinctive hosts only**. A shared host that overlaps a distinctive span is already counted in that span, so feeding it into the stitch is a no-op — excluding it is the clean, correct statement and guarantees shared hosts never inflate or extend an app's time.

- Plumb the per-`(app,host)` `shared` flag from `AppTimeLimitRepoLive.listForProfile` (`api/src/db/Repos.scala:1028`) into group construction.
- `Presence.appSecondsForProfile`/`appSpansForProfileWithDropCount` (`api/src/presence/Presence.scala:645`,`:708`) consume the distinctive-only host-set per app. Reuses the existing #715 stitch primitive — single-source-of-truth preserved, no second time-accounting impl.
- Tests pin: shared host inside a distinctive span → 0 added minutes; shared host with no distinctive activity → not credited to the app.

Blocks: S3 (reporting reuses the distinctive-span helper).

Issue #1896, opened 2026-06-23
shared-hosts S1b: template syntax shared_hosts: + parser + seeder + validation


sameerparekh:
Depends on #1895 (S1a migration). See `docs/design/shared-host-allocation.md` (PR #1894) and parent #1888.

Add the `shared_hosts:` template field and persist the flag. **No attribution behavior change yet** — flag is stored and ignored until S2/S3/S4.

- `AppTemplate.sharedHosts: List[Hostname]` (`api/src/AppTemplates.scala:46`), parse optional `shared_hosts:` (parallel to `hosts:`), default empty (backward compatible — existing templates unchanged).
- Seeder writes the `shared` flag through `AppRepo.setHosts`/`getHosts`; repo carries `shared` per `(app_id, host)` row.
- Document `shared_hosts:` in `api/resources/app_templates/_README.yml` (tie to the existing shared-pool collateral section).
- Directory-wide validation test enforces:
  1. every app has ≥1 distinctive host (`hosts:` non-empty — reject all-shared templates);
  2. global consistency — a host that is `shared` on any template must be `shared` on every template that lists it (else it re-introduces unconditional mis-credit).

Blocks: S2, S3, S4, S5.

Issue #1888, opened 2026-06-23
apps: shared hosts — credit a multi-app host to whichever app's DISTINCTIVE hosts are active (design)


sameerparekh:
## Problem
Some hosts are backends/CDNs shared by MANY apps — e.g. **elevenlabs** (TTS API), **launchdarkly** (feature flags). Today an app template's hosts are attributed to that app unconditionally, so putting a shared host in multiple app templates would mis-credit its traffic (counted for every app that lists it, or ambiguously).

## Desired behavior (operator)
A host can be marked **shared** and listed on MULTIPLE apps, but its traffic is **allocated to a given app only when that app's DISTINCTIVE (non-shared) hosts also have traffic** for the same device/window. So a shared host is credited to the app that's demonstrably in use (its distinctive hosts are active), and otherwise not credited to that app. Shared hosts thus coexist on many apps without double-counting.

## Design questions
- **Template syntax:** how to mark a host as `shared` in app_templates/*.yml (vs a distinctive host). What counts as 'distinctive'.
- **Allocation rule:** precise definition of 'the app's distinctive hosts are active' — same device, same time bucket/window? Co-presence within N minutes? What if two apps' distinctive hosts are both active and both list the shared host (tie-break / split)?
- **Where it lives:** this is an attribution/usage-allocation decision. Reconcile with the existing app-attribution work — note #715 proportionalMins is the canonical per-app usage surface and the byte-share/argmax-foreground heuristic was rejected (#842/#715 Path 1 won't-do). This must NOT resurrect that anti-pattern.
- **Enforcement vs reporting:** does 'shared' affect only usage attribution/reporting, or also allow/block (an Allowed-mode app listing a shared host)? Keep enforcement implications explicit.
- Apps are template-authored only (UI editing being removed, #1798) — no UI surface needed.

## Deliverable
A design doc (docs/design/) + dependency-ordered implementation sub-issues. Motivating consumers: the Feeling Great app (elevenlabs + launchdarkly are shared) — see the sibling app issue. Do NOT implement yet.

sameerparekh:
Design doc shipped: `docs/design/shared-host-allocation.md` (PR #1894). Dependency-ordered sub-issues filed: #1895 (S1a migration) → #1896 (S1b template syntax/parser/validation) → #1897 (S2 engaged-minutes excludes shared) / #1898 (S3 reporting co-presence allocation) / #1899 (S4 enforcement guardrails); #1889 is S5 (Feeling Great, first consumer).

PR #1938, opened 2026-06-24
feat(#1898): shared-host reporting allocation by co-presence + equal-split

### LINKED ISSUE #1898: shared-hosts S3: per-host reporting allocation (co-presence + equal-split + Other)
Depends on #1897 (S2). See `docs/design/shared-host-allocation.md` (PR #1894), parent #1888.

Attribute shared-host bytes/proportional-seconds in the usage-by-app reporting surface by **temporal co-presence**, replacing `minBy(appId)` for shared hosts only.

- In `buildUsageByApp` (`api/src/routes/UsageRoutes.scala:473`) and the by-app axis builder (`:601`), route shared-host rows through: overlap with each app's distinctive spans (from S2) → attribute to qualifying app(s) → **equal-split** among N qualifiers (deterministic, no winner-by-magnitude, NOT argmax) → no qualifier → existing `orphanHosts`/"Other" bucket (`:569`).
- Distinctive hosts keep the current unconditional `minBy(appId)` mapping.
- Tests pin the design doc's worked example (incl. split + Other fallback).
- Ships with an `AppMetrics` counter for shared-row attribution outcome (`attributed`/`split`/`other`, bounded labels) per `docs/process/instrumentation.md#instrument-new-functionality`, plus the consuming Grafana panel under `deploy/grafana/dashboards/` per `#metrics-need-a-dashboard`.

**Must NOT** resurrect the rejected byte-share/argmax-foreground heuristic (#842/#715 Path 1) — allocation is gated on the app's own distinctive presence, never on byte magnitude.

sameerparekh:
## What

Implements **S3** of the shared-host allocation design (parent #1888): the per-host reporting allocation. In the usage-by-app surface, **SHARED** hosts are now attributed by **temporal co-presence** instead of the unconditional `minBy(appId)` mapping that **distinctive** hosts keep.

Depends on S1b (#1896) and S2 (#1897), both merged.

## The rule (co-presence, NOT argmax)

A shared host's stitched session-seconds attribute to the app(s) whose **own DISTINCTIVE session overlaps** the span:

- **one qualifier** → all the span's seconds to that app (`attributed`)
- **≥2 qualifiers** → the seconds **split EQUALLY** among them (`split`) — deterministic, symmetric, **picks no winner by magnitude**
- **no qualifier** → the span surfaces in the existing `orphanHosts` / "Other" bucket (`other`)

This is explicitly **not** the rejected #842/#715 Path 1 argmax/byte-share heuristic — allocation is gated purely on an app's distinctive presence, never on byte magnitude. The worked example from `docs/design/shared-host-allocation.md` is pinned in tests (incl. the 50/50 split and the "Other" fallback).

## Single-source-of-truth

The co-presence test consumes the **S2 seam** `TimeStatusService.distinctiveSpansByApp` rather than re-deriving the distinctive partition — no second time-accounting impl. That seam is broadened to be **mode-agnostic** (`ProfileAppDispositions.appLabelDistinctiveHosts`) so an `Allowed`-mode app (e.g. Feeling Great) earns its shared backends too; the per-app engaged-minutes **cap** still reads the TimeLimited-only subset unchanged.

`Presence.allocateSharedHostSeconds` stitches the host the same way `proportionalHostSeconds` does, so the allocation **conserves** the host's proportional seconds exactly (equal-split remainder rides the lowest-`appId` qualifier).

## Surfaces touched

- **`buildUsageByApp`** — shared hosts routed through the co-presence allocation; distinctive hosts keep `minBy(appId)`.
- **`loadAppLookup`** (the by-app **time-series** axis) — shared hosts excluded from both `appOf` and `patternsBySlug`, so the per-app series span stays distinctive-only and reconciles with the S2 cap; a shared host surfaces there as a standalone host entry (per-bucket co-presence allocation is not applied to the time-series).
- **`AppHost`** gains the `app_hosts.shared` flag via `listAllHostMappings` (additive, defaulted) so shared hosts are classified catalog-wide.

## Instrumentation

`usage_shared_host_attribution_total{outcome=attributed|split|other}` — bounded 3-value label, allowlisted in the cardinality firewall — plus two consuming Grafana panels on the `data-quality-ingest` dashboard (rate-by-outcome + 24h "Other" share).

## Tests (TDD, red → green)

- `SharedHostAllocationSpec` (unit) — pins the worked example: co-presence attribution, equal-split, "Other" fallback, and seconds conservation.
- `UsageApiSpec` (feature, embedded Postgres, real route, no mocked repos) — Feeling Great + Speechify both listing `elevenlabs.io` as shared: the shared host splits to the co-present app and its uncovered span lands in "Other".

Full `mill api.test` green.

Fixes #1898. Relates to #1888, #1897. (Enforcement is S4 #1899 — not in scope here.)

### DIFF
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -3149,9 +3149,9 @@ class AppRepoLive(xa: Transactor[Task]) extends AppRepo {
     sql"SELECT version FROM app_hosts_version".query[Long].unique.transact(xa)
 
   def listAllHostMappings =
-    sql"SELECT app_id, host FROM app_hosts"
-      .query[(AppId, Hostname)]
-      .map { case (id, h) => AppHost(id, h) }
+    sql"SELECT app_id, host, shared FROM app_hosts"
+      .query[(AppId, Hostname, Boolean)]
+      .map { case (id, h, shared) => AppHost(id, h, shared) }
       .to[List]
       .transact(xa)
 
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -76,6 +76,10 @@ object MetricGuard {
       // 2-value enum (`in` agent→server / `out` server→agent); bounded, so it
       // satisfies the §4 cardinality firewall.
       "direction",
+      // #1898 — shared-host reporting attribution outcome for
+      // `usage_shared_host_attribution_total`. A fixed 3-value enum (attributed /
+      // split / other); bounded, so it satisfies the §4 cardinality firewall.
+      "outcome",
     )
 
   /**
@@ -135,6 +139,10 @@ object MetricGuard {
     // too aggressive (real sessions vanishing), a flat zero while phantom
     // inflation returns means it is too lax.
     "presence_app_sessions_dropped_total"       -> Set.empty[String],
+    // #1898 — outcome of attributing a SHARED host's stitched span in usage-by-app
+    // by temporal co-presence with apps' distinctive sessions. `outcome` is a
+    // bounded 3-value enum (attributed / split / other); no per-mac/host/app label.
+    "usage_shared_host_attribution_total"       -> Set("outcome"),
     // #1885 — log events the loki4j appender shed because its bounded send queue
     // (sendQueueMaxBytes) was full while Loki was slow/unreachable. The appender
     // is async/drop-on-backpressure by construction (fail-open: the request path
@@ -443,6 +451,25 @@ object AppMetrics {
           "presence_app_sessions_dropped_total",
           Map.empty,
           count.toLong,
+        ),
+      )
+      .unit
+
+  // ── #1898: shared-host reporting attribution outcome ─────────────────────────
+  // Emitted from UsageRoutes.buildUsageByApp, counted per shared-host stitched span:
+  // `attributed` (one co-present qualifier), `split` (≥2 qualifiers, bytes split
+  // equally), `other` (no qualifier → "Other"/orphan bucket). Bounded `outcome`
+  // label (exactly three values) — no per-mac/host/app cardinality. A sustained
+  // rise in `other` means shared backends are firing with no app demonstrably in
+  // use; a rise in `split` means several apps' distinctive sessions co-occur.
+
+  def recordSharedHostAttribution(outcome: String, count: Long): UIO[Unit] =
+    ZIO
+      .when(count > 0L)(
+        MetricGuard.counter(
+          "usage_shared_host_attribution_total",
+          Map("outcome" -> outcome),
+          count,
         ),
       )
       .unit
--- a/api/src/policy/ProfileAppDispositions.scala
+++ b/api/src/policy/ProfileAppDispositions.scala
@@ -76,6 +76,26 @@ final case class ProfileAppDispositions(
     capGroups.map(d => d.label -> d.distinctiveHosts)
 
   /**
+   * #1898 (shared-hosts S3): the (label → DISTINCTIVE hosts) pairs for EVERY assigned app,
+   * mode-agnostic. Same per-(app, host) `shared = false` partition as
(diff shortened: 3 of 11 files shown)


sameerparekh:
## Independent PR review — #1898 shared-host reporting allocation (S3)

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `api/src/presence/Presence.scala` (allocateSharedHostSeconds) — the shared host's per-device sessions are stitched here a second time, after `proportionalHostSeconds` already stitched every host in `buildUsageByApp`. The conservation test pins that the two agree, so it's correct, just mildly redundant work (one extra stitch per shared host per request). Fine to leave; a future refactor could thread the already-computed spans through.
2. `api/src/routes/UsageRoutes.scala` (computeUsageByApp, presence-seconds loop) — for a shared host that is *partially* attributed (some spans qualify, one span lands in "Other"), the host counts toward BOTH its app(s)' and the orphan presence-seconds for every bucket, not just the buckets whose span was unattributed. This is an over-credit on the **non-additive** presence-seconds surface only (the comment at line 463-465 documents presence is intentionally not additive across apps); the **proportional** seconds — the metric of record and the tested surface — are exact and conserved. Acceptable; noting for transparency.

**Dimension notes**
- **§1 SSOT — clean.** Co-presence reads the S2 seam `TimeStatusService.distinctiveSpansByApp` rather than re-deriving the distinctive partition; the seam broadening to mode-agnostic (`appLabelDistinctiveHosts`) keeps the per-app **cap** on the TimeLimited-only subset (`capGroupLabelDistinctiveHosts`) untouched. `allocateSharedHostSeconds` reuses `spanOf`/`stitch`, not a second time-accounting impl, and a conservation test pins `sum(alloc) == proportionalHostSeconds(host)`. No "keep in sync" comments; no display-vs-enforcement source split.
- **§2 Testing — red→green visible** (commit `79d8340` red test before `dbfbc31` impl). Feature test goes through the real HTTP route on embedded Postgres with no mocked repos; unit test pins the design doc's worked example incl. equal-split + "Other". Negative/edge (no-qualifier → Other) asserted. No existing assertions weakened.
- **§3 Architecture — N/A to enforcement.** Server-side reporting only; router/snapshot untouched; no DNS-as-enforcement reasoning; `Clock` not needed (pure projection over presence rows). Critically, this does NOT resurrect the rejected #842/#715 argmax/byte-share heuristic — allocation is gated purely on an app's own distinctive presence, equal-split picks no winner by magnitude.
- **§4 Wire — additive.** `AppHost` gains `shared: Boolean = false`; it is repo-internal (`listAllHostMappings`), not part of the policy snapshot, so no router-wire impact. Default keeps every legacy construction distinctive.
- **§5 Migrations — N/A.** `app_hosts.shared` already shipped in V60 (S1a); this PR only reads it.
- **§6 Metrics — present + dashboarded.** `usage_shared_host_attribution_total{outcome}` routed through `MetricGuard`, allowlisted (`MetricGuard.Allowed` + `KnownLabelKeys`), bounded 3-value enum, no per-mac/host/app label. Two consuming Grafana panels added to the already-tf-registered `data-quality-ingest` dashboard, querying the emitted series.
- **§7 Scope — focused.** No drive-by edits; design doc updated to mark S3 shipped; no dead code / bare TODOs; no AI mentions in title/body.
- **§8 Security — clean.** Parameterized Doobie SQL; existing `requireAuth` + `requireProfileReadAccess` on the route unchanged; no secrets.

VERDICT: APPROVE

S3 attributes shared-host reporting bytes/seconds by temporal co-presence with apps' distinctive sessions, splitting equally among co-present qualifiers and falling back to "Other" — reusing the S2 distinctive-span seam with no second time-accounting impl and no argmax/byte-share. Tests pin the worked example through the full stack; metric + dashboard ship in-PR. No blockers; two non-blocking nits on redundant stitching and a documented presence-seconds approximation.

PR #1940, opened 2026-06-24
feat(#1899): most-permissive enforcement for shared hosts (allow-in, never block)

### LINKED ISSUE #1899: shared-hosts S4: most-permissive enforcement for shared hosts (allow-in, never block)

sameerparekh:
## What

Shared-hosts **S4** — the last core slice of the shared-host work (parent #1888;
S1b #1896, S2 #1897, S3 #1898 merged). Implements the **most-permissive**
enforcement rule for shared hosts in `PolicyService` / `ProfileAppDispositions`.

A shared host follows the most permissive disposition of any app that lists it.
This is **independent of co-presence** — co-presence gates *attribution* (S2/S3)
only, never enforcement.

### Enforcement decision table

| App disposition | Distinctive host | Shared host |
|---|---|---|
| **Allowed** | `extraAllowed` | **`extraAllowed`** (unconditional) |
| **Blocked** (manual) | `extraBlocked` | **omitted** (never dropped) |
| **TimeLimited**, cap exhausted | `extraBlocked` | **omitted** (never dropped) |
| **TimeLimited**, under cap (exempt) | `extraAllowed` carve | `extraAllowed` if exempt-carve applies |

## Why

- **Over-allowing is safe.** `extraAllowed` already beats every block path at the
  router (an architectural invariant — admin allow overrides `@blocked_macs`
  too). Keeping a shared backend reachable breaks nothing and lets an Allowed-mode
  app actually work (e.g. Feeling Great carved around a Schedule block still needs
  `elevenlabs.io` for TTS).
- **Over-blocking is collateral.** Dropping a shared backend because *one* app is
  blocked / cap-exhausted would drop it for *every other* app on the device — the
  precise #1636 collateral failure. So a shared host **never** enters a drop set;
  only the app's **distinctive** hosts are block-eligible, and the cap is still
  enforced through them.

## How — no wire change

Reuses the existing `extraAllowed` / `extraBlocked` vocabulary; the router stays a
dumb applier (it can't express a co-presence-conditional allow, and per the
minimal-functional-shape rule we don't push one onto it).

- `ProfileAppDispositions.enforcement` now blocks only `distinctiveHosts` (both
  the `Blocked` base-mode and the `BlockedDuring` schedule-window branches). The
  allow side is unchanged — it keeps the full host-set.
- New `AppDayState.distinctiveHosts`, threaded from the single
  `ProfileAppDispositions.from` fold (`AppDisposition.distinctiveHosts`, #1897)
  through `groupAppLimits` → `appDayStatesFromMinutes`. `appCapExhaustedHosts`
  and the `/decision` fallback `timeLimitBlockFromState` read it, so a shared
  host is never per-app-cap-blocked and the two paths can't diverge (#1532).
- **Single-source-of-truth:** there is no second definition of "distinctive" — it
  comes from the same per-`(app, host)` `app_hosts.shared` partition S2/S3 use.

## Tests (TDD — red commit then green)

- `SharedHostEnforcementSpec` (unit): `enforcement` block side omits shared;
  `appCapExhaustedHosts` returns distinctive only; `computeBlockRules` never
  drops a shared host on cap exhaustion and carves an Allowed-mode shared host
  into `extraAllowed`.
- `PolicySnapshotAppsSpec` (feature, embedded Postgres): same rules end-to-end
  through the `app_hosts.shared` DB round-trip → snapshot — allowed app carves
  both host kinds; blocked app omits the shared host; cap-exhausted time-limited
  app blocks the distinctive host but never the shared backend; a shared host on
  a blocked *and* an allowed app ends up allowed.

Fixes #1899. Relates to #1888, #1898, #1636, #1679.

### DIFF
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -711,7 +711,10 @@ class PolicyServiceLive(
     // Time-limit blocks expire at the next household daily-reset Instant.
     val resetAt     = PolicyService.nextDailyResetAfter(settings, now).toString
     val appLimitHit = state.perApp.find { sd =>
-      sd.hosts.exists(hp => HostMatch.matchesPattern(hostname, hp)) &&
+      // #1899 (shared-hosts S4): match only the app's DISTINCTIVE hosts — a shared backend is never
+      // app-cap-blocked, so the /decision fallback agrees with the snapshot's `appCapExhaustedHosts`
+      // (which also reads `distinctiveHosts`). Keeps the two paths from diverging (#1532).
+      sd.distinctiveHosts.exists(hp => HostMatch.matchesPattern(hostname, hp)) &&
       // #1627: same Option-aware exhaustion check as `appCapExhaustedHosts`.
       sd.dailyLimitMinutes.exists(lim => sd.usedMinutes >= lim)
     }
@@ -1011,8 +1014,12 @@ object PolicyService {
       // #1627: `dailyLimitMinutes = None` means "no per-app cap configured" —
       // never exhausted. Only an app with `Some(lim)` and `usedMinutes >= lim`
       // contributes hosts to extraBlocked here.
+      // #1899 (shared-hosts S4): only the app's DISTINCTIVE hosts are block-eligible. A shared
+      // backend is NEVER dropped on cap exhaustion — dropping it because THIS app hit its cap would
+      // drop it for every OTHER app on the device (the #1636 collateral failure). The cap is still
+      // enforced through the distinctive hosts; `usedMinutes` already aggregates the whole host-set.
       case sd if sd.dailyLimitMinutes.exists(lim => sd.usedMinutes >= lim) =>
-        sd.hosts.map(Hostname.unsafe)
+        sd.distinctiveHosts.map(Hostname.unsafe)
     }.flatten
 
   /**
--- a/api/src/policy/ProfileAppDispositions.scala
+++ b/api/src/policy/ProfileAppDispositions.scala
@@ -126,6 +126,13 @@ final case class ProfileAppDispositions(
     val blocked = scala.collection.mutable.ListBuffer.empty[Hostname]
     perApp.foreach { d =>
       val hosts                      = d.hosts.map(Hostname.unsafe)
+      // #1899 (shared-hosts S4): the BLOCK side is restricted to an app's DISTINCTIVE hosts. A
+      // shared backend (e.g. `elevenlabs.io`) is listed on many apps; dropping it because THIS app
+      // is blocked would drop it for every OTHER app on the device — the #1636 collateral failure.
+      // The ALLOW side keeps the full host-set (`hosts`): over-allowing a shared backend is safe
+      // (extraAllowed already beats every block path at the router; design §"most-permissive"), and
+      // a shared host belonging to an allowed app must stay reachable for that app to work.
+      val distinctive                = d.distinctiveHosts.map(Hostname.unsafe)
       val pairs                      = schedWindows.getOrElse(d.assignmentId, Nil)
       val allowedActive              = pairs.exists { case (m, w) =>
         m == AppScheduleMode.AllowedDuring && PolicyService.windowActiveAt(w, now)
@@ -140,12 +147,12 @@ final case class ProfileAppDispositions(
         if (!(capExhausted && !d.exemptFromDaily) && !suppressedByScheduleToggle)
           allowed ++= hosts
       } else if (blockedActive) {
-        blocked ++= hosts
+        blocked ++= distinctive
       } else
         d.mode match {
           case AppMode.Allowed     =>
             if (!suppressedByScheduleToggle) allowed ++= hosts
-          case AppMode.Blocked     => blocked ++= hosts
+          case AppMode.Blocked     => blocked ++= distinctive
           case AppMode.TimeLimited => () // surfaces via the per-app cap path
         }
     }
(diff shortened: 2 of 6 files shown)


sameerparekh:
**Independent PR review** — shared-hosts S4 (#1899), most-permissive enforcement. Read-only pass over `git diff origin/main...HEAD` against the 8 checklist dimensions.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `api/src/routes/Routes.scala:1531` (out of diff) — the single-device per-app usage view still computes engaged minutes from `s.hosts` (full host-set) rather than the new `distinctiveHosts`. This is an *attribution/display* path, not enforcement, so it's correctly out of scope for #1899; but now that `AppDayState.distinctiveHosts` exists it would be the consistent source there too (matches the S2 stitch). Worth a follow-up under the S2/S3 attribution umbrella, not this PR.

**Dimension notes**
- **1 — Single source of truth (✓, the highest-priority check):** "distinctive" has exactly one definition — the per-`(app,host)` `app_hosts.shared` partition from `ProfileAppDispositions.from` (`AppDisposition.distinctiveHosts`, #1897). It is threaded, not re-derived, through `groupAppLimits` → `appDayStatesFromMinutes` → `AppDayState.distinctiveHosts`. Both block-side readers (`appCapExhaustedHosts` and the `/decision` fallback `timeLimitBlockFromState`) now read the same field, so the snapshot and `/decision` paths can't diverge (#1532) — the change actually *removes* a latent divergence rather than adding one.
- **2 — Testing (✓):** clean TDD red→green in history (test commit `test(#1899)…` precedes the impl commit). New behavior is asserted at both levels: `SharedHostEnforcementSpec` (pure: enforcement block-side, `appCapExhaustedHosts`, `computeBlockRules`, incl. the negative under-cap case) and `PolicySnapshotAppsSpec` (feature, embedded Postgres, real repos + `TestClock`, through the `app_hosts.shared` DB round-trip → snapshot, incl. the cross-app blocked+allowed case). No mocked repos/Clock.
- **3 — Architectural invariants (✓):** no new wire field — reuses `extraAllowed`/`extraBlocked`. `distinctiveHosts` lives on the internal `AppDayState` (verified: no `derives JsonCodec`, never serialized/persisted, single production constructor that always populates it, no `.copy`), so the `= Nil` default is never a production value — no silent under-block. `extraAllowed`-beats-block unchanged. No router logic added; no `/tmp` writes; Clock still injected.
- **4 — Wire compat (✓):** `AppDayState` and `groupAppLimits` are server-internal (`private[policy]`); the tuple-arity change is internal and the new field is additive with a default. Wire snapshot shape unchanged.
- **5 — Migrations (✓):** none.
- **6 — Metrics (✓):** a refinement of which host-set feeds an existing enforcement computation — not a new route/job/external call — so no new metric is warranted (consistent with S2/S3).
- **7 — Scope/hygiene (✓):** focused diff (3 src + 2 tests + 1 design-doc status line marking S4 shipped, matching the S1a/S3 precedent). No dead code, no debug logging, no AI mentions in title/body/commits.
- **8 — Security (✓):** no SQL, routes, or secrets touched.

**VERDICT: APPROVE**
Most-permissive shared-host enforcement: the block side is restricted to each app's distinctive hosts (so a shared backend is never dropped for one app's block/cap-exhaustion — the #1636 collateral hazard), while the allow side is unchanged. Correct, well-tested, single-source-of-truth-preserving, and no wire change. No blockers; the one nit is explicitly out of scope.

PR #1944, opened 2026-06-24
fix(#1935): point e2e harness DNS at the LAN resolver, not external DNS

### LINKED ISSUE #1935: bug(router CD): Gate 2 test_block_encrypted_dns_enforced precondition needs real egress to 8.8.8.8 — fails in isolated fake VM (blocks router publishes alongside #1929)

sameerparekh:
## Problem

Gate-2 (fake-API e2e) was intermittently red, blocking every router release at v0.3.13. Root cause: the harness pinned the e2e router VM's dnsmasq upstream to **public DNS (1.1.1.1 / 8.8.8.8)**.

That's wrong on two counts:

1. **#1911 blocks it by design.** The network-wide `blockEncryptedDns` toggle drops *forwarded LAN* traffic on `:53` to exactly those curated resolver IPs. The e2e VMs are LAN devices behind the household wifihaven router — so once that toggle is enabled on the gateway, the harness's own resolution would be killed by the very feature under test.
2. **It was flaky.** The public-DNS path (VM → SLIRP → host → gateway → ISP → 8.8.8.8) intermittently timed out — `dig @8.8.8.8` `no servers could be reached`, dnsmasq `smoke check … nil for tiktok.com` — red-gating releases even though the runner host itself resolved fine. The failure mode varied run-to-run (one run had the encrypted-DNS test *passing*), i.e. a flake, not isolation.

These took down *every* DNS-dependent test together (encrypted-DNS, cname direct-requery, v6 block page, the dnsmasq smoke check).

## Fix

**Point the router VM's dnsmasq at the LAN resolver** (the runner's gateway, `192.168.10.1`, env-overridable via `WH_ROUTER_LAN_RESOLVER`) instead of external public DNS:

- It **survives #1911**: a LAN device querying the gateway keeps resolving after the block — the gateway answers via its own upstream on the **output** hook, which the forward-hook drop never touches.
- It's a **shorter, reliable path** (directly-attached gateway vs. an ISP/internet round-trip through double-NAT).
- The `rebind_domain=wifihaven.net` carve-out (so dnsmasq doesn't discard the RFC 5737 TEST-NET leaf `192.0.2.10` as a rebinding attack) is now required at **both hops** — on the VM (this PR) and on the gateway (applied out-of-band; verified the gateway now returns `192.0.2.10`).

**Rework `test_block_encrypted_dns_enforced`** to assert only the **toggle-ON (blocked)** state and drop the toggle-OFF "8.8.8.8 reachable" baseline:

- The e2e client is nested **behind the e2e router VM behind the household router**. Once #1911 is live on that outer gateway, `8.8.8.8:53` is unreachable for *every* LAN device regardless of the inner toggle — so "reachable with the toggle off" is not a baseline the harness can rely on.
- That the toggle took effect is still confirmed: the rendered nft chain must load, and the NODATA half must return no IP for the relay/DoH hosts — **both toggle-on-only**.
- The flaky `:443` connectivity probe (dialed a public IP) is removed; **port-scope is proven at the render layer** by `openwrt/test/encrypted_dns_spec.lua`.

All retained #1911 behaviours: NODATA for relay/DoH, `:53`→resolver drop (when on), LAN-resolver fallback survives.

## Verification

- Confirmed on the runner host (api.lan) that `192.168.10.1` resolves normal names, and — after the gateway carve-out — returns the e2e cname leaf `192.0.2.10`.
- Confirmed the gateway has no `:53`/encrypted-dns drop today, so the switch is safe now and forward-compatible once #1911 is enabled there.
- `py_compile` clean; pytest markers (`encrypted_dns`, `smoke`) registered; no unused names.

## Scope

Touches only the e2e harness (`conftest.py`) and one e2e test. No wire/schema/enforcement change. Stays in the existing Gate-2 harness.

Relates to #1911 / #1921. Fixes #1935.

### DIFF
--- a/scripts/e2e/scenarios_fake/conftest.py
+++ b/scripts/e2e/scenarios_fake/conftest.py
@@ -56,6 +56,12 @@ ROUTER_IMAGE_PATH = os.environ.get("WH_ROUTER_IMAGE_PATH")
 KEEP_VMS = os.environ.get("E2E_VM_KEEP", "0") == "1"
 SKIP_VMS = os.environ.get("E2E_VM_SKIP_VMS", "0") == "1"
 
+# Upstream the router VM's dnsmasq forwards to — the runner's LAN resolver
+# (the household wifihaven gateway), NOT external public DNS. See the rationale
+# at its use site in `router_session` (#1935). Override via env if the
+# self-hosted runner sits on a different LAN.
+ROUTER_LAN_RESOLVER = os.environ.get("WH_ROUTER_LAN_RESOLVER", "192.168.10.1")
+
 BASE_SNAPSHOT = "e2e-base-fake"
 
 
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review — harness DNS → LAN resolver (#1935)

Read-only adversarial pass over `git diff origin/main...HEAD` (two files:
`scenarios_fake/conftest.py`, `scenarios_fake/test_encrypted_dns.py`) against the 8 dimensions.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. `conftest.py:62` — `ROUTER_LAN_RESOLVER` defaults to a LAN-specific literal
   (`192.168.10.1`). It's env-overridable (`WH_ROUTER_LAN_RESOLVER`) and documented,
   and the prior value (`8.8.8.8`) was also a hardcoded choice — but the harness is
   now coupled to the self-hosted runner's gateway. Acceptable for a self-hosted
   gate; flagging the coupling.
2. `conftest.py` (rationale block) — the cname leaf now resolves only because the
   **gateway** carries `rebind_domain=wifihaven.net`, applied out-of-band (not in
   repo). The comment documents it and the gateway state was verified, but it's an
   undeclared environmental precondition the suite depends on; a future
   reprovision of the gateway would silently re-break the cname tests. Consider
   making it declarative (follow-up). Non-blocking.
3. `test_encrypted_dns.py` (b) — with the toggle-OFF baseline gone, the
   "`dig @8.8.8.8` is blocked when on" assertion can pass *trivially* if 8.8.8.8 is
   independently unreachable. This is mitigated: `_wait_encrypted_dns_chain_loaded`
   and the NODATA half both succeed only when the toggle is on, so toggle-effect is
   still proven; and the scope reduction is operator-directed ("confirm blocked
   when on, don't confirm reachable when off") precisely because the nested outer
   router makes the OFF baseline invalid. Recorded transparently as an accepted
   tradeoff, not a defect.
4. PR body carries a "Generated with Claude Code" footer (added by the harness PR
   template). The checklist's hygiene item flags AI mentions in the PR body; the
   commit message itself only carries the allowed `Co-Authored-By` trailer.
   Non-blocking.

### Dimension notes
- **1 SSoT:** no re-derivation of a computed quantity; constants are config, not
  logic. The two-hop rebind carve-out is an environmental requirement, not
  duplicated code. Not a finding.
- **2 Testing integrity (critical):** this PR REMOVES two assertions from an
  existing test (the toggle-OFF `8.8.8.8`-reachable baseline and the `:443`
  connectivity probe). Both removals are explained and justified, clearing the
  "test weakening must say why the old assertion was wrong" bar: the OFF baseline
  asserted an *environment* property (8.8.8.8 reachable from a nested LAN client)
  that #1911 will permanently break on the outer gateway and that was the actual
  CI flake; the `:443` probe dialed a public IP (same flaky path) and port-scope
  is already proven at the render layer by `openwrt/test/encrypted_dns_spec.lua`.
  The feature is still gated: NODATA for relay/DoH, the `:53`-blocked assertion,
  the chain-load wait, and the LAN-resolver fallback all remain. Net is a
  justified, documented scope reduction, not a silent regression-hider.
- **3 Architecture:** DNS-never-enforces respected — the rationale correctly notes
  the gateway answers via the *output* hook (unblocked) while #1911 drops on the
  *forward* hook; no "resolved ⇒ allowed" reasoning; no agent/router decision logic
  touched. `os.environ` use is the e2e harness convention (the zio-config rule is
  for the Scala app), not a violation.
- **4–6 wire / migrations / metrics:** N/A (e2e harness only).
- **7 Scope/hygiene:** focused 2-file diff; removed `PROBE_IP` and the `:443`
  block cleanly, no dead code (`client_exec` still used via `_dig_via`); three-dot
  merge-base used.
- **8 Security:** no secrets; `192.168.10.1` is a private LAN IP, not a credential.

VERDICT: APPROVE
Repoints the e2e harness's dnsmasq from external public DNS (which #1911 is
designed to drop and which flaked in CI) to the LAN resolver, and narrows the
encrypted-DNS test to the toggle-ON blocked state since the nested outer router
makes the OFF baseline unreliable. No blockers; the test scope reduction is
explained, mitigated, and operator-directed.

Issue #1834, opened 2026-06-22
dashboard redesign 2/5: KPI strip above NOW + status-first restate (#821, #299)


sameerparekh:
Implementation chunk **2 of 5** of the `/dashboard` holistic redesign — see `docs/design/dashboard-redesign.md` §7 (umbrella #1148). Closes **#821** and applies **#299** to the KPI tiles.

### Scope (frontend only)
- Move the 4-tile KPI strip to render immediately under the `<h1>Dashboard`, **above** `<NowSection />` (today it renders after the entire NOW grid — `web/src/pages/DashboardPage.tsx:41-48`).
- Restate the tiles from cumulative-volume to status-first (design §7 decision Q1):
  - **Online now** — devices active in the last 5 min, derived from the `useDashboardNow` snapshot (no new query).
  - **Blocked now** — count of profiles currently `blocked` (paused / out-of-schedule / over-limit), from the same snapshot.
  - **Events (1h)** — `stats.totalHour` (renamed from "Queries (1h)", #299).
  - **Blocked (1h)** — `stats.blockedHour`.
- Drop the cumulative "today" tiles from the dashboard — daily volume is analytics and lives on `/usage/events` (design §2 non-goals).
- Keep `grid-cols-2 md:grid-cols-4` reflow.

### Acceptance
- All 4 KPI tiles visible without scrolling on a 1080p initial load; NOW renders below them.
- "Online now" / "Blocked now" derive from the NOW snapshot; no new endpoint.
- No tile reads "queries".
- Mobile wraps to 2×2.

### Dependencies
- Soft-orders after #1833 (terminology) only to avoid a copy-merge conflict; otherwise independent. No API change.

Parent: #1148. Design: `docs/design/dashboard-redesign.md` §7.

sameerparekh:
**Paused — blocked on websocket #1023.** Operator decision (2026-06-22): the `/dashboard` redesign is on hold until the websocket streaming transport (#1023, priority #1) lands. The redesigned live sections (NOW, KPI "online/blocked now", Most Recently Blocked) should source from the streaming push, not the current 10s poll, so building them on the polling path first would be throwaway. Moved out of In Progress → Blocked; resume after #1023.

sameerparekh:
[...]

Issue #747, opened 2026-05-21
dashboard show active bytes in/out


sameerparekh:
per-device/per-profile/per-app/per-host/overall maybe not all on dashboard, but in appropriate locations

sameerparekh:
Folded into the dashboard redesign umbrella #1148 as a design input. Hold piecemeal implementation until #1148's design pass dispositions this (fold / supersede / defer) — shipping these one-by-one risks conflicting changes (reorder vs. merge vs. drop).

PR #1947, opened 2026-06-24
docs(#1860): design SPA-side websocket infrastructure

### LINKED ISSUE #1860: Design: SPA-side websocket infrastructure (live browser updates, replace refetch polling)
## Design: SPA-side websocket infrastructure (live browser updates)

**Design chip — design only, no implementation.** Sibling to the router↔API websocket epic (#1023). That epic gives the API a *connection registry + push-on-change* core (see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §6, sub-issues #1846/#1849); this issue designs the **browser-facing** half: pushing live state to the WifiHaven SPA over a websocket instead of polling.

### Why
The SPA currently live-updates by **TanStack Query `refetchInterval` polling** — `web/src/api/queries.ts` polls the dashboard every 10 s, recent-blocked every 10 s, and several views every 30 s. That means:
- Up-to-10 s staleness on the dashboard / "now" views (a kid gets blocked, the parent's screen lags).
- A fixed poll tax per open tab regardless of whether anything changed — and it scales with concurrent admins/tabs, not with actual change rate.
- No "is my data live?" signal — a silently-dead poll looks identical to "nothing changed."

A SPA websocket lets the server **push** the handful of fast-moving facts (dashboard "now", recent blocked events, router-connected state, usage ticks) and lets React Query **invalidate/patch** the affected cached queries on receipt — collapsing the poll loops into one connection whose liveness is itself a signal.

### Produce — a design doc (`docs/design/spa-websocket.md`) covering
1. **Endpoint & protocol** — a browser-facing `wss://…/api/ws` (or `/api/admin/ws`), an `{op, payload}` envelope (mirror the router transport's framing discipline so the two share patterns, not code). Message types: server→SPA push (dashboard-now changed, new blocked event, router connected/disconnected, usage tick), client→server hello/subscribe. **No new read-model shapes** — payloads reuse the existing `/api/dashboard/now`, `/api/usage/*`, `/api/blocked` response bodies (same source-of-truth discipline as #1023).
2. **React Query integration** — the canonical pattern: does a push carry the new data (patch the cache directly) or just a "key X is stale" signal (trigger `queryClient.invalidateQueries`)? Recommend one. Keep polling as the fallback for un-websocketed views and offline reconnect (`refetchInterval` stays as belt-and-suspenders, paused while the socket is healthy).
3. **Auth** — the SPA authenticates with the **operator JWT** (`web/src/hooks/useAuth.tsx`, `web/src/api/client.ts`), Admin/ReadOnly via the existing middleware. Cover the upgrade-time auth (JWT can't ride a browser `WebSocket` `Authorization` header — specify the mechanism: ticket/subprotocol/cookie), token expiry mid-connection, and re-auth on refresh.
4. **Reuse vs. separate registry** — can the API's router-transport connection registry (#1846) be generalized to also hold SPA channels (keyed by session/user instead of router_id), or is a parallel `SpaWsRegistry` cleaner? Decide and justify. The **push-on-change** trigger (#1849) is the natural event source for both.
5. **Reliability** — reconnect/backoff in the browser, heartbeat/liveness (drives a real "live / reconnecting" indicator in the UI, replacing the silent-poll ambiguity — ties to `ApiUnreachableBanner.tsx`), backpressure, multi-tab behavior (one socket per tab vs. a shared worker).
6. **Metrics** — bounded-label server metrics for SPA connections (active gauge, frames, auth outcomes) per `docs/process/instrumentation.md`, + Grafana panel.
7. **Phased, independently-shippable rollout** — server SPA-ws endpoint → JWT handshake → one pilot view (dashboard "now") off polling → broaden → retire the redundant `refetchInterval`s. Polling stays the fallback throughout (no flag day).

### Relationship to #1023
- **Shares**: the API push-on-change core (#1849) and the registry pattern (#1846) — this should *consume* them, not rebuild a parallel change-detection path.
- **Differs**: browser `WebSocket` (no custom headers → different auth handshake than the router's bearer-on-upgrade), JWT (not router token) auth, React-Query cache integration instead of nftables/dnsmasq config apply, and a user/session key instead of `router_id`.
- **Cleanly independent**: the SPA path can be designed and built without blocking the router path, and vice-versa.

Relates to #1023. Cloud SPA hosting split (Cloudflare Pages vs. Render API) is a deployment consideration for the ws endpoint host — call it out in the doc (`docs/design/spa-websocket.md`).

sameerparekh:
## Summary

Two co-dependent design docs for the **browser-facing websocket** ([#1860](https://github.com/wifihaven/wifihaven/issues/1860)), the SPA half of the websocket epic ([#1023](https://github.com/wifihaven/wifihaven/issues/1023)). **Design only, no implementation.** Consumes the router transport's registry pattern ([#1846](https://github.com/wifihaven/wifihaven/issues/1846)) and push-on-change core ([#1849](https://github.com/wifihaven/wifihaven/issues/1849)) rather than rebuilding them.

> **Reframe (after operator review):** the socket is **not** a "poll/invalidate everything faster" transport. It exists to **stream the live surface** — live in/out bandwidth, NOW, just-blocked — while genuinely cacheable resources stay request/response. The doc now leads with **what data flows and why**.

### `docs/design/spa-websocket.md` — key decisions

- **§1 What flows over the socket** (the heart of the design) — a message catalog classified into three data classes:
  - **(1) push-native realtime streams** — **live throughput** (overall + per-profile in/out B/s, [#747](https://github.com/wifihaven/wifihaven/issues/747)). Genuinely new data: a *rate*, not a resource — no `GET` to cache (`DashboardNow*` carries no bytes). The clearest reason the socket must exist.
  - **(2) pushed live snapshots/appends** — **NOW** (`DashboardNow`) + **Most Recently Blocked**, reusing the existing REST bodies verbatim.
  - **(3) occasionally-changing resources** — profiles, devices, 24h rollups, time-status — **stay queries**, with at most a `stale` nudge.
- **One new shape: `ThroughputTick`** (overall + per-profile B/s) — justified because throughput has no cacheable REST form; per-device/host throughput stays off the dashboard.
- **Client integration per class** — push-data + `setQueryData` patch for the live surface (1/2); `invalidateQueries` for class (3). Replaces the earlier "invalidate everything." `refetchInterval` stays as the paused fallback.
- **Throughput aggregator (§5.3)** — derive overall/per-profile B/s **server-side, once** from the [#1023](https://github.com/wifihaven/wifihaven/issues/1023) router→API usage stream (a new lightweight ~2s `throughput` sample off the conntrack counters), pushed to the SPA.
- **Auth** — short-lived single-use ticket (`POST /api/ws/ticket`) over cookie/subprotocol; `jwtExp`-bounded mid-connection close + `reauth` seam.
- **Registry** — fork `SpaWsRegistry` (different key/payload/event vocab); share the **change sources** (widen `PolicySnapshotPublisher` to a hub + add `SpaEventHub`).
- **Reliability/Metrics/Hosting/Rollout** — backoff+jitter, live/reconnecting indicator, latest-wins throughput backpressure, socket-per-tab; bounded `spa_ws_*` metrics + `spa-ws.json`; Cloudflare-Pages-vs-Render split (ws → Render directly, `Origin` allowlist). Rollout pilot is the **realtime dashboard** (S0 router sample → S4 aggregator → S5 dashboard).

### `docs/design/dashboard-redesign.md` — Revision 3

- **Reverses rev-2 Q2:** live throughput (overall + per-profile in/out) is now a **primary NOW element**, not a deferred follow-up.
- Adds the **dashboard live data contract** (which sections stream vs. stay request/response) — the input to `spa-websocket.md` §1. Unblocks the stream-gated dashboard sub-issues (#1834/#1835).

## Scope

Design docs only. Implementation sub-issues (S0–S8, incl. the #1023 router throughput sample) are enumerated in the rollout section as follow-ups.

Fixes #1860. Relates to #1023, #1148, #747, #1846, #1849.

### DIFF
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -371,10 +371,12 @@ only, consistent with the rev-2 "glanceable, not analytics" principle.
 > existing read model **streamed**: overall/per-profile = `groupBy ∈ {∅,profile}`,
 > the averaging window = its `bucket`, and the client derives B/s from
 > bytes-over-bucket. No new shape — see [`spa-websocket.md` §1.3](spa-websocket.md)
-> (`trafficUsage` topic). The same stream powers the **Traffic Usage page** live,
-> not just the dashboard gauge. (Granularity floor is `1m`, the `traffic_reports`
-> bucket; a sub-minute "instant" gauge is an optional extra — `spa-websocket.md`
-> §5.3/§10 Q1.)
+> (`trafficUsage` topic). The same topic powers the **Traffic Usage page**, which
+> still loads its **history via the existing `GET`** and only takes the **live
+> edge** (the current bucket advancing) from the socket — the stream isn't a
+> replacement for the historical query. (Granularity floor is `1m`, the
+> `traffic_reports` bucket; a sub-minute "instant" gauge is an optional extra —
+> `spa-websocket.md` §5.3/§10 Q1.)
 
 ### 8.2 The dashboard's live data contract (what streams vs. what stays request/response)
 
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -375,8 +375,9 @@ only, consistent with the rev-2 "glanceable, not analytics" principle.
 > still loads its **history via the existing `GET`** and only takes the **live
 > edge** (the current bucket advancing) from the socket — the stream isn't a
 > replacement for the historical query. (Granularity floor is `1m`, the
-> `traffic_reports` bucket; a sub-minute "instant" gauge is an optional extra —
-> `spa-websocket.md` §5.3/§10 Q1.)
+> `traffic_reports` bucket; fully-realtime uses the `raw` bucket, which streams the
+> live edge at the router's usage-send cadence — no bespoke sample, no router
+> change. `spa-websocket.md` §5.3/§10 Q1.)
 
 ### 8.2 The dashboard's live data contract (what streams vs. what stays request/response)
 
@@ -419,12 +420,11 @@ source in first" step (§7.5) is now concrete:
 - **#747 bandwidth** is **no longer a deferred standalone** — it becomes part of
   the websocket dashboard pilot ([`spa-websocket.md` §9](spa-websocket.md)),
   streaming the existing `GET /api/usage/traffic` read model (`trafficUsage`
-  topic). It needs **no new backend data path** for `1m`-and-coarser bandwidth
-  (the API re-runs the existing traffic-usage query on usage ingest,
-  [`spa-websocket.md` §5.3](spa-websocket.md)); only the optional sub-minute
-  "instant" gauge depends on #1023 (a router conntrack sample). Earlier this note
-  said throughput was #1023-gated — with the read-model reuse, only the sub-minute
-  refinement is.
+  topic). It needs **no new backend data path and no router change**: the API
+  recomputes only the current bucket on usage ingest
+  ([`spa-websocket.md` §5.3](spa-websocket.md)), and fully-realtime uses the `raw`
+  bucket at the router's usage-send cadence. Earlier this note said throughput was
+  #1023-gated — it isn't; it just gets faster as #1023 streams usage.
 
 This revision records the decision; the wire/payload specifics live in
 `spa-websocket.md` (single source of truth for the protocol), and dashboard
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -247,12 +247,21 @@ implementation sub-issues §5 called for were never filed. This section is autho
 ```
 Dashboard (h1)
   Banners            — NewDevicesHint, AccessRequestsBanner (only when present)
-  KPI strip          — Online now · Blocked now · Events (1h) · Blocked (1h)      [above the fold]
-  NOW                — freshness pill; active cards (top-3 devices + expander); idle-collapse row
+  KPI strip          — Online now · Blocked now · Events (1h) · Blocked (1h) · Bandwidth ▲/▼   [above the fold]
+  NOW                — freshness pill; active cards (top-3 devices + expander, per-profile ▲/▼); idle-collapse row
   Most Recently Blocked  — blocked-only trailing-hour feed (#1338), View all → /usage/events
   Blocking activity (24h) — one merged ranked-host panel; one-line empty state
   (no inline log table)
 ```
+
+**Bandwidth tile (rev 3, §8.1).** The 5th KPI tile is the **overall household
+in/out** live gauge (▲ up / ▼ down), with a small **window selector**
+(`1m · 10m · 1h · raw`, §8.1) on the tile; the selected window drives both the
+overall tile and the per-profile ▲/▼ on the NOW cards (one bandwidth window for
+the page). It's a `trafficUsage` stream (§8.2), live like the other realtime
+elements. **Mobile:** the strip is now 5 tiles, so it reflows `grid-cols-2`
+(→ 3 rows) on mobile / `md:grid-cols-5` on desktop (supersedes the rev-1 §3
+`md:grid-cols-4` note).
 
 The two recency/blocking elements sit adjacent and are intentionally distinct: **Most Recently
 Blocked** answers "what just got dropped?" (live, un-aggregated, 1h); **Blocking activity (24h)**
@@ -355,7 +364,9 @@ Rev-2 Q2 de-scoped throughput (#747) as "backend-gated, its own follow-up." **Re
 3 reverses that.** The dashboard must show, updating live:
 
 - **Overall household in/out** — aggregate ▲ upload / ▼ download **rate** (B/s),
-  the headline "how much is the house using right now" gauge.
+  the headline "how much is the house using right now" gauge. **Placement
+  (decided 2026-06-25): the 5th KPI-strip tile** (§7.2), carrying the window
+  selector (`1m · 10m · 1h · raw`) that also governs the per-profile gauges.
 - **Per-profile in/out** — the same ▲▼ **rate** on each active NOW profile-card
   header (the rev-2 §3 wireframe already drew `▲2.4 ▼18` there; rev 3 makes it
   real and live, not deferred).
@@ -388,7 +399,7 @@ they ever diverge. Each dashboard section is classified by how it gets its data:
 
 | Dashboard section | Class | Source after #1860 |
 |---|---|---|
-| **Overall in/out bandwidth** (§8.1) | **streamed read model** | `trafficUsage` (`groupBy:∅`) — streams existing `TrafficUsageResponse`; no new shape |
+| **Overall in/out bandwidth** (§8.1, **5th KPI tile** w/ window selector) | **streamed read model** | `trafficUsage` (`groupBy:∅`) — streams existing `TrafficUsageResponse`; no new shape |
 | **Per-profile in/out bandwidth** (§8.1, on NOW card headers) | **streamed read model** | same `trafficUsage` topic, `groupBy:["profile"]` |
 | **NOW** active cards (who's online / watching) | **pushed live snapshot** | `DashboardNow` body pushed on change (reuses today's `/api/dashboard/now` shape) |
 | **Most Recently Blocked** feed (#1338) | **pushed live append** | `connectionEvents{blocked:true}` — reuses the `/api/logs` row shape; same topic streams the Connection Events page |
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -391,9 +391,18 @@ they ever diverge. Each dashboard section is classified by how it gets its data:
 | **Blocking activity (24h)** panel | request/response | rollup tables (#809); explicitly **not** streamed (rev-2 §7.5 already said so) |
 
 The rule the operator stated: **stream the things that need to move in realtime
-(throughput, NOW, just-blocked); leave everything that is genuinely a cacheable
-resource on request/response.** The websocket design (`spa-websocket.md`) is
-built around exactly this split, not a blanket "invalidate every query."
+(throughput, NOW, just-blocked, live time-usage); leave everything that is
+genuinely a cacheable resource on request/response.** The websocket design
+(`spa-websocket.md`) is built around exactly this split, not a blanket "invalidate
+every query."
+
+> **Live time-usage** (per-profile used/remaining + per-app minutes, counting up
+> in realtime) is also streamed — see [`spa-websocket.md` §1.2](spa-websocket.md)
+> (`timeStatus` / `appUsage`). Its primary home is `/profiles` (the screen-time
+> view), not a `/dashboard` section, so it isn't in the table above — but on the
+> dashboard it backs the "profiles over limit" KPI, which therefore updates live
+> off the same `timeStatus` push. It replaces today's adaptive refetch ladder
+> (`TIME_STATUS_REFETCH_LADDER`), which exists only because there was no push.
 
 ### 8.3 Effect on the locked plan (§7.4)
 
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -249,7 +249,9 @@ Dashboard (h1)
   Banners            — NewDevicesHint, AccessRequestsBanner (only when present)
   KPI strip          — Online now · Blocked now · Events (1h) · Blocked (1h) · Bandwidth ▲/▼   [above the fold]
   NOW                — freshness pill; active cards (top-3 devices + expander, per-profile ▲/▼); idle-collapse row
-  Most Recently Blocked  — blocked-only trailing-hour feed (#1338), View all → /usage/events
+  Recently Blocked   — TOP-LEVEL live panel: blocked-only, newest-first, device · host · reason;
+                       quick device filter; the "why isn't this working on my device?" diagnostic   [above the fold]
+  ── below the fold ──
   Blocking activity (24h) — one merged ranked-host panel; one-line empty state
   (no inline log table)
 ```
@@ -263,10 +265,22 @@ elements. **Mobile:** the strip is now 5 tiles, so it reflows `grid-cols-2`
 (→ 3 rows) on mobile / `md:grid-cols-5` on desktop (supersedes the rev-1 §3
 `md:grid-cols-4` note).
 
-The two recency/blocking elements sit adjacent and are intentionally distinct: **Most Recently
-Blocked** answers "what just got dropped?" (live, un-aggregated, 1h); **Blocking activity (24h)**
-answers "what's been getting blocked today?" (aggregated, ranked). Neither duplicates the other,
-and neither is the firehose.
+The two recency/blocking elements are intentionally distinct: **Recently Blocked** answers "what
+just got dropped?" (live, un-aggregated, 1h); **Blocking activity (24h)** answers "what's been
+getting blocked today?" (aggregated, ranked). Neither duplicates the other, and neither is the
+firehose.
+
+**Recently Blocked is a top-level, above-the-fold panel (rev 3, operator 2026-06-25).** It is the
+primary *diagnostic* surface: when someone says "X isn't working on my iPad," a parent glances here
+to see — within seconds, live — whether that device's traffic is being dropped and why, instead of
+digging through `/usage/events`. To serve that, each row shows **device · host · reason** (the
+`BlockReason`, e.g. Schedule / TimeLimit / category / host-block — `connectionEvents` carries it),
+and the panel offers a **quick filter by device** (and "View all →" deep-links to the device's
+`/usage/events`). It is the live `connectionEvents{blocked:true}` stream (§8.2) — newest-first,
+prepend-on-push, trailing ~1h — so a just-now block appears immediately. Promoting it above the fold
+(right under NOW) is deliberate: NOW answers "who's online," Recently Blocked answers "what's being
+blocked," and together they are the at-a-glance health/diagnosis pair. (Reachability is a
+connection-layer fact — these are real traffic-layer drops, not DNS events; DNS always resolves.)
 
 ### 7.3 Decisions on the §6 open questions
 
@@ -402,7 +416,7 @@ they ever diverge. Each dashboard section is classified by how it gets its data:
 | **Overall in/out bandwidth** (§8.1, **5th KPI tile** w/ window selector) | **streamed read model** | `trafficUsage` (`groupBy:∅`) — streams existing `TrafficUsageResponse`; no new shape |
 | **Per-profile in/out bandwidth** (§8.1, on NOW card headers) | **streamed read model** | same `trafficUsage` topic, `groupBy:["profile"]` |
 | **NOW** active cards (who's online / watching) | **pushed live snapshot** | `DashboardNow` body pushed on change (reuses today's `/api/dashboard/now` shape) |
-| **Most Recently Blocked** feed (#1338) | **pushed live append** | `connectionEvents{blocked:true}` — reuses the `/api/logs` row shape; same topic streams the Connection Events page |
+| **Recently Blocked** — top-level diagnostic panel (#1338) | **pushed live append** | `connectionEvents{blocked:true}` — reuses the `/api/logs` row shape; the quick device filter maps to the topic's `macs` param; same topic streams the Connection Events page |
 | **KPI: Online now / Blocked now** | **derived from the NOW push** | computed client-side off the pushed `DashboardNow`; no separate stream |
 | **KPI: Events (1h) / Blocked (1h)** | request/response | small enough to leave on a slow poll / invalidate; not realtime-critical |
 | **Blocking activity (24h)** panel | request/response | rollup tables (#809); explicitly **not** streamed (rev-2 §7.5 already said so) |
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -337,3 +337,78 @@ streaming-source wiring into chunks 2 and 3 (the live sections) as their first s
 layout changes.
 
 This note records the decision; no further dashboard PRs until #1023 lands.
+
+---
+
+## 8. Revision 3 (2026-06-24) — realtime throughput is a primary NOW element; live data contract
+
+Operator direction while finalizing the SPA-websocket design
+([#1860](https://github.com/wifihaven/wifihaven/issues/1860),
+[`docs/design/spa-websocket.md`](spa-websocket.md)): the websocket is **not** a
+"poll faster / invalidate faster" transport for the whole page — it exists to
+**stream the live surface**. That changes one rev-2 decision and pins the live
+data contract the two docs share.
+
+### 8.1 Reversal of Q2 — live throughput is in scope, as a primary NOW element
+
+Rev-2 Q2 de-scoped throughput (#747) as "backend-gated, its own follow-up." **Rev
+3 reverses that.** The dashboard must show, updating live:
+
+- **Overall household in/out** — aggregate ▲ upload / ▼ download **rate** (B/s),
+  the headline "how much is the house using right now" gauge.
+- **Per-profile in/out** — the same ▲▼ **rate** on each active NOW profile-card
+  header (the rev-2 §3 wireframe already drew `▲2.4 ▼18` there; rev 3 makes it
+  real and live, not deferred).
+
+This is the single most-requested live element and the main reason the dashboard
+is stream-gated. Per-**device** / per-**host** throughput stays **off** the
+dashboard (→ `/devices`, `/profiles`) — the dashboard shows overall + per-profile
+only, consistent with the rev-2 "glanceable, not analytics" principle.
+
+> Throughput is a **rate**, not a resource — there is no "current B/s" you can
+> `GET` and cache; it only exists as a stream. So unlike the NOW snapshot or the
+> blocked feed (which have REST bodies), live throughput is **push-native**: it
+> has no polling fallback beyond "no live gauge while the socket is down" (the
+> 24h volume on `/usage` covers the historical question). This is why it could
+> not ship before the stream — and why it is the clearest justification for the
+> websocket existing at all.
+
+### 8.2 The dashboard's live data contract (what streams vs. what stays request/response)
+
+This table is the **input to the websocket message catalog** in
+[`spa-websocket.md` §1](spa-websocket.md); the two must agree. Each dashboard
+section is classified by how it gets its data:
+
+| Dashboard section | Class | Source after #1860 |
+|---|---|---|
+| **Overall in/out throughput** (§8.1) | **push-native stream** | `throughput` tick — NEW push-only shape (no GET); §8.1 |
+| **Per-profile in/out throughput** (§8.1, on NOW card headers) | **push-native stream** | same `throughput` tick (per-profile entries) |
+| **NOW** active cards (who's online / watching) | **pushed live snapshot** | `DashboardNow` body pushed on change (reuses today's `/api/dashboard/now` shape) |
+| **Most Recently Blocked** feed (#1338) | **pushed live append** | each new blocked event pushed as it lands (reuses the `/api/logs?blocked=true` row shape) |
+| **KPI: Online now / Blocked now** | **derived from the NOW push** | computed client-side off the pushed `DashboardNow`; no separate stream |
+| **KPI: Events (1h) / Blocked (1h)** | request/response | small enough to leave on a slow poll / invalidate; not realtime-critical |
+| **Blocking activity (24h)** panel | request/response | rollup tables (#809); explicitly **not** streamed (rev-2 §7.5 already said so) |
+
+The rule the operator stated: **stream the things that need to move in realtime
+(throughput, NOW, just-blocked); leave everything that is genuinely a cacheable
+resource on request/response.** The websocket design (`spa-websocket.md`) is
+built around exactly this split, not a blanket "invalidate every query."
+
+### 8.3 Effect on the locked plan (§7.4)
+
+No sub-issue is added or removed here; the chunk-2/chunk-3 "fold the streaming
+source in first" step (§7.5) is now concrete:
+
+- **chunk 2/3** (NOW + KPI, #1834/#1835) consume the `DashboardNow` push and the
+  derived KPIs — as already planned.
+- **#747 throughput** is **no longer a deferred standalone** — it becomes the
+  realtime-throughput pilot of the websocket rollout
+  ([`spa-websocket.md` §9](spa-websocket.md)), since it is the push-native element
+  that most exercises the stream. It still needs the backend source (§8.1): the
+  API deriving per-profile + overall byte-rate from the #1023 router→API usage
+  stream, pushed to the SPA. That backend work is specified in
+  [`spa-websocket.md` §5](spa-websocket.md) and filed as a sub-issue there.
+
+This revision records the decision; the wire/payload specifics live in
+`spa-websocket.md` (single source of truth for the protocol), and dashboard
+implementation stays paused on #1860 per §7.5.
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -376,8 +376,9 @@ only, consistent with the rev-2 "glanceable, not analytics" principle.
 ### 8.2 The dashboard's live data contract (what streams vs. what stays request/response)
 
 This table is the **input to the websocket message catalog** in
-[`spa-websocket.md` §1](spa-websocket.md); the two must agree. Each dashboard
-section is classified by how it gets its data:
+[`spa-websocket.md` §1](spa-websocket.md); the two must agree, and
+[`spa-websocket.md` §1.2](spa-websocket.md) is the **authoritative `op` list** if
+they ever diverge. Each dashboard section is classified by how it gets its data:
 
 | Dashboard section | Class | Source after #1860 |
 |---|---|---|
--- a/docs/design/dashboard-redesign.md
+++ b/docs/design/dashboard-redesign.md
@@ -365,13 +365,16 @@ is stream-gated. Per-**device** / per-**host** throughput stays **off** the
 dashboard (→ `/devices`, `/profiles`) — the dashboard shows overall + per-profile
 only, consistent with the rev-2 "glanceable, not analytics" principle.
 
-> Throughput is a **rate**, not a resource — there is no "current B/s" you can
-> `GET` and cache; it only exists as a stream. So unlike the NOW snapshot or the
-> blocked feed (which have REST bodies), live throughput is **push-native**: it
-> has no polling fallback beyond "no live gauge while the socket is down" (the
-> 24h volume on `/usage` covers the historical question). This is why it could
-> not ship before the stream — and why it is the clearest justification for the
-> websocket existing at all.
+> Bandwidth is **not** new data — `traffic_reports` already holds directional
+> bytes, surfaced by `GET /api/usage/traffic` (`TrafficUsageResponse`,
+> `totalBytesIn`/`totalBytesOut` per group). So live bandwidth is just that
+> existing read model **streamed**: overall/per-profile = `groupBy ∈ {∅,profile}`,
+> the averaging window = its `bucket`, and the client derives B/s from
+> bytes-over-bucket. No new shape — see [`spa-websocket.md` §1.3](spa-websocket.md)
+> (`trafficUsage` topic). The same stream powers the **Traffic Usage page** live,
+> not just the dashboard gauge. (Granularity floor is `1m`, the `traffic_reports`
+> bucket; a sub-minute "instant" gauge is an optional extra — `spa-websocket.md`
+> §5.3/§10 Q1.)
 
 ### 8.2 The dashboard's live data contract (what streams vs. what stays request/response)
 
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review — `docs(#1860): design SPA-side websocket infrastructure`

Read-only adversarial pass over `git diff origin/main...HEAD` (one file, `docs/design/spa-websocket.md`, +620). This is a **design-only docs PR**, so dimensions 2 (testing), 5 (migrations), and most of 3/4/8 (runtime invariants) apply only insofar as the *design prescribes* code — checked on that basis. Cross-doc references and factual claims about the current codebase were verified against `api/src` and `web/src`.

### Dimension notes
- **1 — Single source of truth (highest priority):** ✅ The doc actively *enforces* SSOT rather than violating it: §0.3 forbids new read-model shapes, §2.1 recommends invalidate-by-default precisely so the `GET` stays the sole read-model producer, and §4.2 consumes #1849's change-event source (widening the existing `PolicySnapshotPublisher` sink + adding `SpaEventHub`) instead of building a parallel change-detection path. No re-derivation prescribed.
- **3 — Architectural invariants:** ✅ Browser-side only; touches no router/DNS/snapshot logic. Auth claims match `AuthService` (HS256 `{sub,role,iat,exp}`, roles admin/adult/child, 401→/login). Metrics are bounded enums only; `role` (3-value) and `topic`/`op`/`direction`/`result` are fixed enums — no per-user/session label (§6).
- **4 — Wire contract:** ✅ The `/api/ws` endpoint is purely additive; REST stays the fallback throughout (§2.4, §8). Unknown-op ignore+meter mirrors `RouterWsRoutes`.
- **6 — Metrics + dashboard:** ✅ New `spa_ws_*` families ship with a `spa-ws.json` Grafana panel in the same PR, authored against grepped series (§6.1) — matches the dashboard rule. (See NIT 1.)
- **7 — Scope/hygiene:** ✅ Focused single-file diff; all cross-doc/code links resolve; no dead content.
- **8 — Security:** ✅ Ticket auth reuses `requireAuth` (one JWT-validation impl); single-use short-TTL bounds the query-string exposure; Origin allowlist (§7) is a sound CSWSH defense for the cross-origin cloud case.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. **§6.1 dashboard registration** — the doc says the `spa-ws.json` PR is gated by `grafana-terraform` but doesn't mention that a *new* dashboard must also be added to the `local.dashboards` list in [`infra/grafana/main.tf:84`](https://github.com/wifihaven/wifihaven/blob/main/infra/grafana/main.tf#L84) (that's what the gate checks). Worth a one-line note in the sub-issue (S1) so the implementer doesn't miss it.
2. **§6 metric allow-list** — the new `spa_ws_*` series will need their `(name -> allowed keys)` entries added to `MetricGuard.Allowed` ([`api/src/metrics/Metrics.scala:91`](https://github.com/wifihaven/wifihaven/blob/main/api/src/metrics/Metrics.scala#L91)) or `MetricGuard` will reject them. The doc says "the same cardinality-firewall discipline `Metrics.scala` already enforces" — making the `Allowed` registration explicit (as the router-ws metrics did) would prevent a missed-entry surprise in S1.
3. **§2.2 `routers` topic → `['admin','routers']` key** — there is no TanStack Query hook with that key in `web/src/api/queries.ts` today (the routers list is a plain `api.routers.list()` call). The mapping is fine as a forward-looking design, but it's worth noting the `routers` topic implies *adding* a query (or a different live-status surface) rather than pausing an existing poll, since the other four topics map to existing polled hooks.

### VERDICT: APPROVE

A thorough, decision-bearing design doc that mirrors the quality and structure of `docs/design/websocket-transport.md`. It makes and justifies the contentious calls (ticket auth, fork-the-registry/share-the-event-source, invalidate-over-patch, socket-per-tab) rather than presenting a menu, consumes #1846/#1849 rather than rebuilding them, and lays out an additive, no-flag-day rollout. All findings are NITs (forward-looking implementation reminders for the sub-issues); none are merge-gating.

sameerparekh:
## Independent PR review (re-run) — delta `e67381e…1130d8e`

Statused prior findings against current code and reviewed the incremental delta (`git diff e67381e...HEAD`, +21/-1 in `docs/design/spa-websocket.md` — the three NIT clarifications).

### Prior findings status
| --- | --- | --- |
| NIT 1 | §6.1 — note new dashboard must join `local.dashboards` in `infra/grafana/main.tf` | **ADDRESSED** — §6.1 now states `spa-ws.json` must be added to the `local.dashboards` list in [`infra/grafana/main.tf`](https://github.com/wifihaven/wifihaven/blob/main/infra/grafana/main.tf), sibling to `router-ws-transport`. |
| NIT 2 | §6 — make `MetricGuard.Allowed` registration explicit | **ADDRESSED** — §6 adds an implementer note that each `spa_ws_*` series needs an `Allowed` entry in `Metrics.scala` or the emit is rejected. |
| NIT 3 | §2.2 — `routers` topic key is forward-looking (no such poll today) | **ADDRESSED** — §2.2 now calls out that 4/5 topics map to existing polled hooks and `routers` is the exception (plain call today), so S5 wraps it rather than retiring a poll. |

### New findings from the delta
(none — the delta is exactly the three clarifying notes; all links resolve, no new claims to verify, no code prescribed.)

### VERDICT: APPROVE

All three prior NITs addressed; the delta introduces no new findings. No BLOCKERs at any point in this review's history. Design doc is complete, decision-bearing, and consistent with the codebase it references.

sameerparekh:
## Independent PR review (re-run) — major reframe `1130d8e…94153b2`

The delta is a near-total rewrite of `spa-websocket.md` (+592/-543) plus a Revision 3 added to `dashboard-redesign.md`, in response to operator feedback that the doc didn't specify *what data flows* and shouldn't treat the socket as a blanket "invalidate everything." Reviewed the full rewrite against the 8 dimensions and statused the prior findings.

### Prior findings status
|---|---|---|
| NIT 1 | §6.1 — note new dashboard joins `local.dashboards` in `infra/grafana/main.tf` | **ADDRESSED** — carried into §7.1. |
| NIT 2 | §6 — make `MetricGuard.Allowed` registration explicit | **ADDRESSED** — implementer note retained in §7. |
| NIT 3 | §2.2 — `routers` topic key was forward-looking (no such poll today) | **ADDRESSED (by removal)** — the rewrite drops the `routers` topic; the cleaner three-class taxonomy (§1.1) puts router/profile/device state in class (3) `stale`, and the dashboard doesn't poll a routers query, so the stale claim is gone. |

### New review of the rewrite

**1 — Single source of truth (highest priority):** ✅ Materially *stronger* than the prior draft. §1.1's three-class split makes the read-model discipline explicit: existing bodies reused verbatim for `now`/`blocked` (class 2), the single new `ThroughputTick` justified because throughput is a rate with no cacheable REST form (verified: `DashboardNow*` in `shared/src/Models.scala` carries no bytes; `UsageRecord` carries `bytesIn/Out` but only period-batched). The rate is computed once server-side (§5.3), not per-tab. `now` pushes reuse the `DashboardNowRoutes` builder (one NOW implementation). No re-derivation.

**3 — Architectural invariants:** ✅ The router→API `throughput` sample (S0) is a dumb byte-counter emit off `conntrack.lua` — no decision logic added to the agent. It's a new *op* on the #1023 router ws, not a new field on the policy snapshot, so the minimal-snapshot-vocabulary rule is untouched. The `blocked` feed is connection-layer drops via `/api/logs?blocked=true` (traffic-layer, not DNS) — consistent with the architecture model.

**4 — Wire contract:** ✅ Both the SPA `/api/ws` ops and the router→API `throughput` sample are additive (new ops, ignore-unknown). S0 is correctly filed under #1023, not smuggled into the SPA endpoint.

**4-auth / 8-security:** ✅ Ticket auth unchanged and sound; §4.4 now explicitly role-filters class-(2) thick pushes (`now`/`blocked` body, per-profile throughput) before fan-out, reusing the `GET`'s filter — so the live surface inherits the same authz scoping. `Origin` allowlist for the cross-origin cloud upgrade retained.

**6 — Metrics:** ✅ `spa_ws_push_total{op,result}` `op` enum is bounded (throughput/now/blocked/stale); `MetricGuard.Allowed` + `main.tf` notes retained.

**7 — Scope:** ✅ The dashboard-redesign.md edit (Revision 3) rides in a #1860 PR, which is unusual but operator-directed and correct — the two docs are co-dependent (the dashboard's live-data contract is the *input* to the ws message catalog), and rev 3 cleanly cross-references spa-websocket.md as the protocol SoT. Commit message explains it.

### NEW findings from the delta

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. **Two tables describe the same live-data contract** — `dashboard-redesign.md` §8.2 (by dashboard-section) and `spa-websocket.md` §1.2 (by `op`). They overlap and could drift. Mitigated already (rev 3 names spa-websocket §1 the protocol SoT and says "the two must agree"), and the two views serve different readers, so this is acceptable — but worth a one-line "see §1.2 for the authoritative op list" pointer at the §8.2 table head to make the dependency direction unmissable.
2. **S0 expands #1023's scope** — the realtime `throughput` sample is filed as a #1023 sub-issue, which is the right home, but the #1023 epic doc (`websocket-transport.md`) doesn't yet mention it. A back-reference there (when S0 is filed) would keep the two epics' rollout tables in sync; not gating for a design doc.

### VERDICT: APPROVE

The reframe directly answers the operator's feedback: the doc now leads with *what data flows and why*, distinguishes push-native realtime streams (live throughput, the genuinely-new #747 data) from pushed live snapshots (NOW/blocked) from class-(3) resources that stay queries, and specifies the server-side throughput aggregator sourced from #1023. Single-source-of-truth and wire-contract discipline are intact; both NITs are non-gating polish. No BLOCKERs at any point in this review's history.

sameerparekh:
## Independent PR review (re-run) — delta `94153b2…0bc515c2`

Statused prior findings and reviewed the incremental delta (3 lines in `dashboard-redesign.md` §8.2).

### Prior findings status
|---|---|---|
| NIT 1 | Two tables describe the live-data contract; name §1.2 authoritative | **ADDRESSED** — §8.2 now states `spa-websocket.md` §1.2 is the authoritative `op` list if the two diverge. |
| NIT 2 | S0 expands #1023's scope; back-reference in `websocket-transport.md` | **DEFERRED (acknowledged)** — correctly contingent on S0 actually being filed under #1023; not actionable in this docs PR and not gating. |

### New findings from the delta
(none — the delta is exactly the one-line authoritative-pointer note.)

### VERDICT: APPROVE

Both substantive NITs resolved (NIT 1 fixed; NIT 2 is a future-filing reminder). No BLOCKERs at any point. The design is complete, decision-bearing, and now correctly framed around the realtime data the socket carries.

sameerparekh:
## Independent PR review (re-run) — live time-usage `0bc515c2…f001b21e`

Delta adds **live time-usage (per-profile + per-app)** as a streamed surface, per operator direction. Reviewed against the 8 dimensions; statused prior findings.

### Prior findings status
|---|---|---|
| NIT 1 | Two contract tables could drift; name §1.2 authoritative | **ADDRESSED** (unchanged; pointer present). |
| NIT 2 | S0 expands #1023; back-ref in `websocket-transport.md` when filed | **DEFERRED** (still contingent on S0 being filed). |

### New review of the delta

**1 — Single source of truth (highest priority):** ✅ The change is *correct on SSOT grounds specifically*. Live time-usage is sourced from the named SSOT primitives the review checklist itself calls out — `TimeStatusService.dayStateAllLive` (per-profile used/remaining, verified at `api/src/policy/TimeStatusService.scala:122/227`) and `Presence.appSecondsForProfile` (per-app, verified at `api/src/presence/Presence.scala:708`) — so the pushed minutes are computed in the one existing place, not re-derived for the socket. The push reuses the existing `/api/time/status` (`ProfileTimeStatus[]`) and `/api/profiles/{id}/usage-by-app` bodies verbatim (no new shape). `setQueryData` targets the real keys (`qk.timeStatusToday`, `qk.profileUsageByApp`, confirmed in `queries.ts`).

**Correct re-classification:** moving time-usage from class (3) "stays a query" to class (2) "pushed live snapshot" is the right call — it is exactly the surface whose continuous tick (minutes counting up) the adaptive `TIME_STATUS_REFETCH_LADDER` was hand-rolled to chase. The doc correctly keeps the ladder as the dormant disconnected fallback (§3.4) and slates it for retirement (S7), rather than deleting it now — consistent with the no-flag-day rule.

**3 — Architectural invariants:** ✅ Event sources are existing write sites (usage ingest, the #1849 time-boundary ticker, extension grant) — no new polling/diff loop, no decision logic relocated. The ticker→`timeStatus` mapping correctly handles the no-new-usage transitions (cap exhaustion, schedule boundary) that change remaining-minutes.

**6 — Metrics:** ✅ `timeStatus`/`appUsage` fold into the existing bounded `op` enum on `spa_ws_frames_total` / `spa_ws_push_total`; no new metric, no cardinality change.

**7 — Scope/consistency:** ✅ The two docs stay in sync — `dashboard-redesign.md` §8.2 notes time-usage is streamed (primary home /profiles, backs the dashboard "profiles over limit" KPI) and defers to spa-websocket §1.2. Rollout S6 split into S6a (time-usage push) / S6b (class-3 stale) cleanly; S0/S1–S5 unaffected.

### NEW findings from the delta

BLOCKERS — (none)

SHOULD-FIX — (none)

NITS
1. **`appUsage` cache key is window-parameterized** — `qk.profileUsageByApp(profileId, from, to)` is keyed by `(profileId, from, to)`, so a push must target the *currently-mounted* window's key to patch the right cache entry (the "today" window is the live one; past windows are immutable and need no push). Worth a half-sentence in §3.2/S6a so the implementer scopes the `appUsage` push to the live (today) window rather than fanning out across cached historical windows. Non-gating.

### VERDICT: APPROVE

Live time-usage is correctly modeled as a class-(2) pushed snapshot off the existing SSOT minute primitives, reusing existing bodies, with the adaptive poll ladder kept as fallback and slated for retirement. SSOT and wire-contract discipline intact; one non-gating NIT about scoping the per-app push to the live window. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — delta `f001b21e…c85de327`

Statused prior findings; reviewed the incremental delta (2 lines: §3.2 + S6a).

### Prior findings status
|---|---|---|
| NIT (live window) | `appUsage` push must target the live (today) window key, not all cached windows | **ADDRESSED** — §3.2 patches `qk.profileUsageByApp(profileId, todayFrom, todayTo)` with a comment that past windows are immutable; S6a restates it. |
| NIT 1 (tables) | name §1.2 authoritative | **ADDRESSED** (unchanged). |
| NIT 2 (S0 back-ref) | back-ref in `websocket-transport.md` when S0 filed | **DEFERRED** (contingent on filing). |

### New findings from the delta
(none — the delta is exactly the live-window scoping clarification.)

### VERDICT: APPROVE

All actionable findings resolved. Live time-usage (per-profile + per-app) is correctly a class-(2) pushed snapshot off the existing SSOT minute primitives, reusing existing bodies, scoped to the live window, with the adaptive poll ladder kept as fallback. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — subscriptions + choosable throughput windows `c85de327…7e4336a8`

Delta promotes subscriptions to a v1 mechanism and makes the throughput window (5s/1m/5m/10m) a subscription parameter, per operator direction. Reviewed against the 8 dimensions; statused prior findings.

### Prior findings status
|---|---|---|
| All prior NITs (tables-authoritative, appUsage live-window, S0 back-ref) | — | **No regression** — the authoritative-table pointer, live-window scoping, and S0/#1023 framing are all intact in the new text. The S0 back-ref NIT remains correctly deferred (contingent on S0 being filed). |

### New review of the delta

**1 — Single source of truth (highest priority):** ✅ The windowed-throughput design *strengthens* SSOT: §5.3 keeps the windowed-rate math (4 EWMAs over one sample stream) entirely server-side — "the client never averages raw samples itself … can't disagree tab-to-tab, and survives reconnect with no client-side history." That's the right call; pushing raw samples and letting each tab compute 5s/1m/5m/10m would have created N client-side re-derivations of the same rate. No new read-model beyond the `window` field on the existing `ThroughputTick`.

**3 — Architectural invariants / consistency:** ✅ Subscriptions are coherent end-to-end: protocol (§1.4 subscribe/unsubscribe + ack + reconnect-replay), registry state (§5.1 per-channel `{topic→params}`, two-gate fan-out), aggregator (§5.3 compute only subscribed windows at window-appropriate cadence), client (§3.1 subscribe/re-subscribe on selector change), reconnect (§6.1 replay). The "fan-out = subscribed AND authorized" framing keeps authz as a hard gate on top of the subscription narrowing — subscription can only *reduce* what a connection sees, never widen past its role. No stale "deferred seam" language remains (verified: the only "forward-compat seam" left is the unrelated `reauth` one).

**6 — Metrics (cardinality — the firewall):** ✅ Correctly handled the trap: `spa_ws_subscriptions_active` is labeled by `topic` (bounded enum) and the doc **explicitly excludes `window`/`profileId` from labels** ("window is ≤4 values but profileId is unbounded — keep params out of labels"). That's exactly right — a per-`profileId` label would be the per-entity cardinality violation the checklist forbids. `subscribe`/`unsubscribe` fold into the existing bounded `op` enum.

**4 — Wire contract / 7 — scope:** ✅ All additive (new ops + a `window` field). Rollout updated coherently (S1 gets the subscription model; S4 windowed aggregator; S5 the selector). §10 Q1 reframed to surface the one genuine product question (is {5s,1m,5m,10m} the right set?) with the additive escape hatch noted.

### NEW findings from the delta

BLOCKERS — (none)

SHOULD-FIX — (none)

NITS
1. **`timeStatus` lists no subscription param but is per-profile-filtered** (§1.2 says "all authorized profiles"). That's a fine default, but a `/profiles` page showing one profile still receives all of them. Not worth a param for v1 (the payload is small and role already scopes it), but a half-sentence noting "timeStatus is unparameterized in v1; a `profileId` filter is an additive future param like `appUsage`'s" would preempt the question. Non-gating.
2. **Subscription `ack` reuses the `ack` op shape** (`{op:"ack", topic, …}`) which differs slightly from the router path's `ack` (`{op, seq, status}`). Harmless (separate endpoints, separate codecs) but worth a one-word note that the SPA `ack` is keyed by `topic`, not `seq`, so a reader doesn't assume the router's shape. Non-gating.

### VERDICT: APPROVE

Subscriptions are now a clean, first-class, data-minimizing mechanism, and the choosable throughput windows fall out of it naturally (window = subscription param → only-subscribed-windows computed and pushed at window-appropriate cadence). SSOT (server-side windowed rate), cardinality discipline (params kept off labels), and the additive wire contract are all intact. Two non-gating NITs. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — delta `7e4336a8…a52057b8`

Statused prior findings; reviewed the incremental delta (2 lines: §1.2 + §1.4).

### Prior findings status
|---|---|---|
| NIT 1 | `timeStatus` unparameterized — note the future `profileId` filter seam | **ADDRESSED** — §1.2 now notes v1 sends all authorized profiles; a `profileId` filter is an additive future param. |
| NIT 2 | SPA `ack` keyed by `topic`, not `seq` like the router ack | **ADDRESSED** — §1.4 ack bullet now states the SPA ack is keyed by `topic` (vs the router's `seq`-keyed data-frame ack). |

### New findings from the delta
(none — exactly the two clarifications.)

### VERDICT: APPROVE

All findings resolved. The design now specifies the live data flowing over the socket, first-class subscriptions for data-minimization, and choosable throughput windows as a subscription parameter — with SSOT, cardinality, and additive-wire discipline intact. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — auth: ticket → scoped JWT cookie `a52057b8…91734dfe`

Delta switches the SPA ws upgrade auth from a query-string ticket to a tightly-scoped JWT cookie, per the operator's decision to authorize the pre-upgrade request directly. Reviewed against the 8 dimensions, with security (dim 8) the focus since this is an auth change.

### Prior findings status
|---|---|---|
| All prior NITs/decisions (tables-authoritative, live-window, timeStatus v1 scope, ack keying, S0 back-ref) | — | **No regression** — all intact; none touched by this delta. |

### New review of the delta — security (dimension 8), in depth

**Pre-upgrade auth + SSOT:** ✅ The cookie is verified at upgrade via the existing `AuthService.verify` (§4.2) — one JWT-validation path, no second auth surface (the checklist's dim-8 "new routes enforce auth via the existing middleware"). 401-no-101 on bad/missing/expired matches the router path. Stateless verification is a genuine improvement over the ticket: it removes the in-memory ticket store and its multi-instance caveat (§10 risk #3 correctly reframed — auth is now multi-instance-clean; only the registry/aggregator fan-out remains single-process).

**The same-registrable-domain reasoning is correct:** ✅ `app.wifihaven.net` and `api.wifihaven.net` share eTLD+1 `wifihaven.net`, so a `Domain=wifihaven.net` cookie is same-site/first-party for both topologies — the third-party-cookie blocking concern genuinely does not apply. The doc owns the earlier misjudgment explicitly, which is good for the design record.

**CSWSH / CSRF surface — handled correctly:** ✅ Two independent controls: `SameSite=Strict` (cookie won't ride a cross-site-initiated upgrade) **and** the `Origin` allowlist server-side backstop (§8). `Path=/api/ws` scopes the cookie off every REST call, so the bearer-on-header REST API keeps its CSRF immunity — no broad auth cookie is introduced. This is the right defense-in-depth for browser ws auth.

**No new persistent secret exposure:** ✅ The JWT already lives in `localStorage` (XSS-readable); a JS-set, path-scoped cookie set transiently around connect and cleared after is no worse. The doc is honest that `HttpOnly` isn't usable (JS sets it) and correctly judges that acceptable given the existing exposure. `Secure` (HTTPS-only) is specified, dev exempted as today.

**reauth correctness:** ✅ Correctly spotted that a cookie can't ride an *existing* ws frame, so `reauth` carries `{jwt}` in the frame payload on the open socket (not a cookie) — internally consistent, and a ws data frame isn't URL/log-exposed.

**Consistency sweep:** ✅ All dependents updated — §2.1 snippet (set-cookie-then-connect), §5.1 (no ticket store), §6.1 (re-set cookie on reconnect), §7 metric enum (cookie/Origin outcomes), §7.1 panel copy, §8 (SameSite+Origin pairing), §9 S2, §10 risk #3. No stale `ticket` references remain except in §4.1's options table where they correctly explain why the ticket was *not* chosen.

### Other dimensions
- **1 SSOT / 6 metrics / 4 wire:** unaffected and intact (auth metric relabeled within bounded enums; still additive).

### NEW findings from the delta

BLOCKERS — (none)

SHOULD-FIX — (none)

NITS
1. **`Secure` cookie on localhost dev.** §4.2 says "dev/localhost is exempted as today." Browsers refuse a `Secure` cookie over plain `http://localhost` — worth a half-line in S2 on the dev story (either `http://localhost` is treated as a secure context by browsers, which it is for cookies in modern Chrome/FF, or dev uses the same-origin self-hosted path). Minor implementation note, non-gating.
2. **Cookie clear-after-open is best-effort.** §4.2 clears the cookie right after the socket opens; if the tab is killed between set and open the cookie lingers until its (short) max-age. Worth specifying an explicit short `Max-Age` on the cookie (e.g. 60s) as the belt-and-suspenders bound, so a missed clear self-expires. Non-gating.

### VERDICT: APPROVE

The cookie mechanism is the correct call for this topology and is specified soundly: pre-upgrade verification via the existing `AuthService`, same-site first-party cookie (correct eTLD+1 reasoning), `SameSite=Strict` + `Origin` allowlist for CSWSH, `Path`-scoped off REST, stateless (multi-instance-clean). Two non-gating NITs (dev `Secure` story, explicit cookie `Max-Age`). No BLOCKERs at any point in this review's history.

sameerparekh:
## Independent PR review (re-run) — delta `91734dfe…f3886062`

Statused prior findings; reviewed the incremental delta (§4.2 cookie attributes, +7 lines).

### Prior findings status
|---|---|---|
| NIT 1 | dev `Secure`-on-localhost story | **ADDRESSED** — §4.2 notes `http://localhost` is a browser secure context so `Secure` cookies work in dev. |
| NIT 2 | explicit short cookie `Max-Age` as clear-after-open backstop | **ADDRESSED** — §4.2 adds `Max-Age` ~60 s with the self-expiry rationale. |

### New findings from the delta
(none — exactly the two cookie-attribute clarifications.)

### VERDICT: APPROVE

All findings resolved. SPA ws upgrade auth is a sound, fully-specified scoped JWT cookie: pre-upgrade verify via the existing `AuthService`, same-site first-party scoping, `SameSite=Strict` + `Origin` allowlist for CSWSH, `Path`-scoped off REST, `Secure` + short `Max-Age`, stateless/multi-instance-clean. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — reuse traffic/connection read models; drop ThroughputTick `f3886062…ed21dcf6`

Large delta: bandwidth is now the existing `GET /api/usage/traffic` read model **streamed** (`trafficUsage` topic) rather than an invented `ThroughputTick`; the `blocked` feed is generalized into a filtered `connectionEvents` topic that also powers the Connection Events page; and the fetch-vs-WebSocket auth distinction is clarified. Reviewed against the 8 dimensions; statused prior findings.

### Prior findings status
|---|---|---|
| All prior (cookie auth, window=subscription-param, live time-usage, tables-authoritative, etc.) | — | **No regression** — cookie auth (§4), subscriptions (§1.4), time-usage (§1.2), and the authoritative-table pointer all survive the rework intact. |

### New review of the delta

**1 — Single source of truth (highest priority):** ✅ This is a **net SSOT improvement** and the strongest version of the design so far. Dropping `ThroughputTick` in favor of streaming `TrafficUsageResponse` removes the only new read-model shape — verified the body already carries `totalBytesIn`/`totalBytesOut` per group with `groupBy ∈ {device,profile,…}` (`Models.scala`) and `bucket ∈ {raw,1m,10m,…}` (`web/src/types/api.ts`), so overall/per-profile/per-device bandwidth + window are all expressible from the existing endpoint. §5.3 re-runs the *existing* traffic-usage query (one query impl) rather than maintaining a parallel rate — no second producer of the bytes. The doc explicitly owns the earlier ThroughputTick misstep. `connectionEvents` likewise reuses the `/api/logs` row + filters. **Zero new `web/src/types/api.ts` shapes** — a stronger claim than the prior "one new shape."

**The "one stream, many consumers" generalization is correct and well-scoped:** ✅ Because each topic = endpoint + filter params, the dashboard gauge and the Traffic Usage page subscribe the *same* topic with different params (§0.3/§1.4/S6b); same for `connectionEvents` and the dashboard "recently blocked" (= `{blocked:true}`). This directly answers the operator's "use it on the traffic and connection pages too" and avoids a bespoke-per-surface design.

**Granularity honesty:** ✅ The doc correctly reconciles the earlier 5s/1m/5m/10m ask with the `traffic_reports` 1m floor: `{1m,10m,1h,…}` come free, `5m`/`10m` are additive re-aggregations, and a true 5s "instant" gauge is explicitly an **optional** S0 (router conntrack sample) rather than the primary path. Consequently live bandwidth is **no longer hard-gated on #1023** (only the sub-minute refinement is) — §10 Q2 flags the framing change. This is accurate and a good de-risking.

**3 — Architectural invariants:** ✅ The optional S0 sample is a dumb conntrack byte-counter emit (no agent decision logic); it's filed under #1023 and is genuinely optional now. Streaming connection-events stays traffic-layer (reuses `/api/logs`), consistent with the architecture model.

**8 — Security / 6 — metrics:** ✅ Cookie auth unchanged. Metric `op`/`topic` enums updated to `trafficUsage`/`connectionEvents` and remain bounded; the doc still explicitly keeps unbounded params (`profileId`, `groupBy`, `bucket`) **out of labels** — cardinality firewall respected.

**The JWT/fetch clarification is correct:** ✅ §0.2.1/§4.1 now state the header limitation is specific to the `WebSocket` constructor; `fetch` sets `Authorization` (verified `client.ts:41`). This resolves the conceptual confusion and is accurate.

**Consistency sweep:** ✅ §3.x renumbered cleanly after the §3.1/§3.2 merge (no dangling §3.4/§3.2 internal refs); §5/§6/§7/§9/§10 and dashboard-redesign §8 all updated to `trafficUsage`/`connectionEvents`. Remaining `ThroughputTick` mentions are all explicitly "we dropped this," and the `blocked`-feed mention is the intentional "generalized into connectionEvents" note.

### NEW findings from the delta

BLOCKERS — (none)

SHOULD-FIX — (none)

NITS
1. **Illustrative cache keys `trafficKey(params)` / `logsKey(filter)` (§3.1)** aren't existing `qk` factory entries (the current `qk` has `recentBlocked`/`dashboardNow`/`timeStatusToday` but no `traffic`/`logs`). The code comment says "the same query key the page reads," which is the right intent — just worth a word that these are the Traffic/Connection pages' own `useQuery` keys (to be unified into `qk`), so a reader doesn't grep for a factory entry that isn't there. Non-gating.
2. **`connectionEvents` ordering/dedup at scale.** Prepend-and-cap with dedup-by-id is fine for the dashboard's 20-row recent feed, but the Connection Events *page* can be filtered/paginated; streaming appends there interact with the page's cursor paging (#862). Worth a half-sentence in S6b that page-level streaming prepends only the live head (newest), leaving cursor-paged history to the existing request/response path. Non-gating.

### VERDICT: APPROVE

Reusing the existing traffic-usage and logs read models (instead of a bespoke throughput shape) is the right call and materially strengthens single-source-of-truth — zero new wire shapes, one query implementation, and the same topics now power the dashboard *and* the traffic/connection pages. Granularity is reconciled honestly (1m floor, optional sub-minute), and the fetch-vs-WebSocket auth point is corrected accurately. Two non-gating NITs (illustrative key naming, page-level stream/paging interaction). No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — delta `ed21dcf6…386179ed`

Statused prior findings; reviewed the incremental delta (2 clarifications: §3.1 illustrative keys, S6b head-vs-paging).

### Prior findings status
|---|---|---|
| NIT 1 | `trafficKey`/`logsKey` aren't existing `qk` entries | **ADDRESSED** — §3.1 now notes they're the pages' existing `useQuery` keys, to be lifted into `qk`. |
| NIT 2 | page-level streaming vs cursor paging (#862) | **ADDRESSED** — S6b now states streaming prepends only the live head; cursor-paged history stays request/response. |

### New findings from the delta
(none — exactly the two clarifications.)

### VERDICT: APPROVE

All findings resolved. The design streams the existing traffic-usage and logs read models (zero new wire shapes, one query implementation), powering the dashboard and the traffic/connection pages from the same topics; granularity is reconciled honestly and the fetch-vs-WebSocket auth point is correct. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — live-edge-only streaming `386179ed…afe851b6`

Delta clarifies that the `trafficUsage`/`connectionEvents` streams carry **only the live edge** (current bucket / new head rows); history (and cursor paging) stays on the existing `GET`, with the client merging the edge into the cached result. Reviewed against the 8 dimensions; statused prior findings.

### Prior findings status
|---|---|---|
| All prior (read-model reuse, illustrative keys, head-vs-paging, cookie auth, etc.) | — | **No regression** — this delta reinforces the head-vs-paging NIT (now applied consistently to `trafficUsage` too, not just `connectionEvents`). |

### New review of the delta

**1 — Single source of truth / correctness:** ✅ This is the correctness-tightening the design needed. Previously §5.3 said "re-run the subscribed query and push the refreshed `TrafficUsageResponse`," which ambiguously implied re-delivering the whole window; the delta scopes the push to the **current bucket only** and has the client **merge** it (`mergeHeadBucket` / `prependHead`) into the GET-loaded series. This is both correct (the `GET` remains the history authority — the stream can't silently diverge on old buckets) and efficient (a 24h×1h chart advances one bar/minute instead of re-sending 24 buckets). The §1.1 class-(1) definition, §1.2 payload column ("only the current/most-recent bucket"), §1.3, §3.1 (merge not replace), §5.3 (recompute only the head bucket), §9 S6b, and dashboard §8.1 are now internally consistent on this.

**3 — Architectural invariants:** ✅ No re-derivation introduced — the head-bucket recompute is the *same* traffic-usage query scoped to the current bucket, not a second code path. `connectionEvents` head-prepend with cursor-paged history on the `GET` correctly respects the existing #862 keyset paging (the stream never mutates older pages).

**Reconnect gap-fill:** ✅ §6.1 already refetches the live queries once on reconnect — since history is a `GET`, that refetch naturally fills any bucket/rows missed while disconnected before the live edge resumes; no separate gap-recovery machinery needed. Consistent with the new framing.

### NEW findings from the delta

BLOCKERS — (none)

SHOULD-FIX — (none)

NITS
1. **Bucket-rollover edge at the merge.** §3.1's `mergeHeadBucket` must key on the bucket's `windowStart` to decide replace-head vs append-new — worth naming that key explicitly (the `TrafficUsageAggregateRow.windowStart` field) so the implementer doesn't merge by array position. The doc already says "replace while accumulating, append when it rolls over"; naming the join key would make it unambiguous. Non-gating.

### VERDICT: APPROVE

Scoping the stream to the live edge (current bucket / head rows) while history stays on the `GET` is correct and efficient, and it's now consistent across §1–§9 + the dashboard doc. The `GET` stays the history authority (SSOT), paging is respected, and per-push cost is one bucket. One non-gating NIT (name the windowStart merge key). No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — delta `afe851b6…c29e2e0e`

Statused the prior finding; reviewed the incremental delta (1 line: §3.1 merge-key naming).

### Prior findings status
|---|---|---|
| NIT 1 | name the `windowStart` merge key for `mergeHeadBucket` | **ADDRESSED** — §3.1 now states the merge joins on `TrafficUsageAggregateRow.windowStart` (replace same-windowStart row, prepend on rollover), not array position. |

### New findings from the delta
(none — exactly the merge-key clarification.)

### VERDICT: APPROVE

All findings resolved. The design streams the live edge only (current bucket / head rows) over the existing traffic-usage and logs read models, with history + paging on the `GET` and the client merging by `windowStart` — zero new wire shapes, `GET` as the history authority, per-push cost of one bucket. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — resolve open questions; drop S0; file #1952 `c29e2e0e…f2d78dfa`

Delta records the operator's decisions on the five §10 open questions: 1m bucket floor + `raw` ingest-rate realtime (dropping the conntrack-sample S0), and multi-instance filed as its own issue (#1952). Reviewed against the 8 dimensions; statused prior findings.

### Prior findings status
|---|---|---|
| All prior (live-edge-only, read-model reuse, cookie auth, windowStart merge key, etc.) | — | **No regression.** |

### New review of the delta

**1 — SSOT / simplification:** ✅ Dropping the bespoke conntrack-sample (former S0) in favor of the existing `raw` `TrafficUsageBucket` pushed at the usage-send cadence removes a router-side data path entirely — fewer moving parts, still zero new shapes, realtime-ness now rides the one usage pipeline rather than a parallel sampler. The "1m floor + raw ingest-rate" framing is internally consistent across §1.3, §5.3, §9 (S0 row removed, S4 + critical path updated), §10 Q1/Q2, and dashboard §8. No leftover S0/conntrack-sample references remain except the §10 record that S0 was dropped.

**3 — Architectural invariants:** ✅ Net *fewer* invariant risks — no agent change at all now (the earlier S0 touched the router). Bandwidth realtime-ness is bounded by the existing usage-send cadence and improves automatically as #1023 streams usage; nothing is hard-gated on #1023.

**Multi-instance (#1952):** ✅ Promoting the single-process caveat to a filed issue is the right call — it's correctly scoped (auth is stateless and excluded; the registry + per-connection subscription set + SpaEventHub + trafficUsage aggregator are the state needing a story), names the options (sticky vs pub/sub vs hybrid), and notes subscription state rebuilds on reconnect so it needn't be replicated. §10 #3 references it cleanly instead of burying the limit. The doc's single-process assumption is now an explicit, tracked decision rather than an unowned risk.

**7 — scope/process:** ✅ #1952 filed with the right relations (#1860/#1846/#1849/#1023) and a board Epic set (matching #1023's "API Contract & Type Safety"). Consistent with the project-board standing rule.

### NEW findings from the delta

BLOCKERS — (none)

SHOULD-FIX — (none)

NITS
1. **`raw` bucket semantics worth one sentence.** `TrafficUsageResponse` for a `raw` subscription returns `rawRows` (per `(mac,host,period)`) rather than `aggregateRows` — the gauge for overall/per-profile would still need to sum those (or the server aggregates the raw period into the chosen groupBy). Worth a half-line clarifying that `raw` = "stream at ingest period granularity, still grouped per the subscription's groupBy," so `raw` isn't read as "ungrouped per-host rows on the dashboard gauge." Non-gating.

### VERDICT: APPROVE

The open questions are resolved coherently: 1m floor with a `raw` ingest-rate realtime mode (no bespoke sampler, no router change), and multi-instance correctly elevated to a tracked issue (#1952) as the first real horizontal-scaling requirement. The simplification strengthens the design. One non-gating NIT on `raw`-bucket grouping semantics. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — delta `f2d78dfa…a156c135`

Statused the prior finding; reviewed the incremental delta (1 sentence: §1.3 `raw`-bucket grouping).

### Prior findings status
|---|---|---|
| NIT 1 | clarify `raw`-bucket stream is grouped per subscription, not ungrouped per-host rows | **ADDRESSED** — §1.3 now states a `raw` subscription aggregates per the subscription's `groupBy` at ingest-period granularity (one overall/per-profile point per usage period). |

### New findings from the delta
(none — exactly the grouping clarification.)

### VERDICT: APPROVE

All findings resolved across the full review history. The SPA-websocket design is complete and internally consistent: streams existing read models (zero new shapes), live-edge-only over a `GET`-loaded history, cookie pre-upgrade auth, first-class subscriptions with the throughput window as a parameter, 1m floor + `raw` ingest-rate realtime (no router change), live time-usage, and multi-instance tracked as #1952. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — overall-bandwidth gauge → 5th KPI tile `10530e7d`

Delta places the overall household bandwidth gauge in the dashboard layout (`dashboard-redesign.md` §7.2/§8) — closing a self-flagged gap where §8.1 introduced the gauge but §7.2 didn't position it. Reviewed; statused prior findings.

### Prior findings status
|---|---|---|
| All prior | — | No regression; this delta is dashboard-doc layout only. |

### New review of the delta
- **Consistency:** ✅ The §7.2 KPI strip, the new bandwidth-tile paragraph, §8.1 placement decision, and the §8.2 contract row now agree (overall bandwidth = 5th KPI tile carrying the `1m·10m·1h·raw` window selector that also governs the per-profile NOW gauges). The self-flagged gap is closed.
- **Supersession handling:** ✅ Correctly did **not** edit the rev-1 §3 mobile note (kept verbatim as historical grounding); instead added the 5-tile reflow (`grid-cols-2` / `md:grid-cols-5`) to the authoritative §7.2 and explicitly noted it supersedes the rev-1 `md:grid-cols-4`. Consistent with the doc's stated "§7/§8 win where they disagree" convention.
- **Scope:** ✅ Dashboard-doc only; no change to `spa-websocket.md` (the protocol is unaffected — placement is pure UI).

### NEW findings
BLOCKERS — (none). SHOULD-FIX — (none).
NITS
1. One window selector now lives on a small KPI tile and governs both the tile and the per-profile NOW gauges — a reasonable single-control choice, but worth a line in the eventual implementation issue (#1834/#1835) that the selector's state is page-level (one window for the whole dashboard's bandwidth), not per-card. Non-gating; layout doc is fine as-is.

### VERDICT: APPROVE

Layout gap closed coherently: overall bandwidth is the 5th KPI tile with the window selector, per-profile stays on the NOW cards, mobile reflow updated, rev-1 text correctly left as historical. No BLOCKERs at any point.

sameerparekh:
## Independent PR review (re-run) — Recently Blocked → top-level diagnostic panel `b83aa6bd`

Delta promotes Recently Blocked to a top-level, above-the-fold diagnostic panel in the dashboard design (`dashboard-redesign.md` §7.2/§8.2). Reviewed; statused prior findings.

### Prior findings status
|---|---|---|
| All prior | — | No regression; dashboard-doc layout/framing only. |

### New review of the delta
- **Coherence:** ✅ §7.2 layout (Recently Blocked moved above the fold under NOW, fold marker before Blocking 24h), the diagnostic-rationale paragraph (device·host·reason + quick device filter + deep-link), and the §8.2 contract row all agree. The two recency/blocking elements stay correctly distinct (live 1h vs aggregated 24h).
- **Architecture invariant respected:** ✅ The note explicitly reiterates these are **traffic-layer drops, not DNS events** ("DNS always resolves") — consistent with the canonical model, and important here precisely because the panel is a diagnostic surface (a reader must not infer "resolved ⇒ reachable").
- **Protocol alignment:** ✅ The quick device filter maps to the existing `connectionEvents` `macs` subscription param (§1.2 in spa-websocket.md already lists `macs?`), and `BlockReason` per row comes from the existing `/api/logs` row shape — no new wire surface; the promotion is pure UI placement.
- **Scope:** ✅ Dashboard doc only; spa-websocket protocol unchanged.

### NEW findings
BLOCKERS — (none). SHOULD-FIX — (none).
NITS
1. The panel now shows `reason` (BlockReason) per row to serve diagnosis. Worth confirming in the impl issue that the `/api/logs` row the feed already uses surfaces the reason in a UI-friendly form (it carries `BlockReason`); purely an implementation detail, the design is correct. Non-gating.

### VERDICT: APPROVE

Recently Blocked is correctly elevated to a top-level above-the-fold diagnostic panel (device·host·reason, device filter, live `connectionEvents{blocked:true}`), paired with NOW as the at-a-glance health view, with the traffic-layer-not-DNS invariant called out. No new wire surface. No BLOCKERs at any point across the full review history.

PR #1948, opened 2026-06-24
fix(#1943): v6 block-page e2e flake — harden uhttpd-readiness race (not a #1933 regression)

sameerparekh:
## Classification: TEST FLAKE / TIMING — not a #1933 regression, not a real v6-delivery regression

`scenarios_fake/test_v6_block_page.py::test_whole_mac_block_serves_block_page_over_v6` reddened Gate-2 in the 14:05 run (sha `3da50362`, after #1933) and **passed** in the 03:07 run (sha `3f565d8f`, before #1933 — even while that run hit the #1935 egress flake).

### Diff review — #1933 provably cannot have regressed this test
- **render.lua: zero lines changed by #1933** (`git show 3da50362 -- '*render.lua'` is empty). The "no render.lua enforcement change" claim holds for the v6 path.
- **This test's code path is behaviorally unchanged.** It calls `http_get` directly (not `wait_block_page`/`wait_http_succeeds`), and #1933's shared-lib edits are purely additive: a new `ipv4_only` arg defaulting to `False` (the pre-#1933 behavior); `wait_block_page` only gained a comment. The test file itself was untouched.

### Real-router evidence the v6 delivery mechanism is sound
Validated on a real OpenWRT box (openwrt.lan, nft 1.1.6, kernel 6.12):
- current-source `render.lua` under the router's **real Lua 5.1** emits the paused whole-MAC v6 `redirect to :8081`/`:8443` + the `fib daddr type local` never-drop guard, and **no** `dnat ip6 to ::1` (the broken pre-#1868 path);
- the full rendered ruleset loads via `nft -c -f -` (exit 0);
- a `[::]` wildcard v6 listener (main uhttpd `:::80`) serves a v6 connection to the router's real br-lan GUA → **http_code 200** — the exact delivery half the #1865 `::1` bug broke. #1865's `[::]` bind keeps it sound.

### Root cause of the flake
The round-trip raced **uhttpd readiness**. PolicyApply (#341) bounces uhttpd (`setup-uhttpd-block-page.sh` re-runs), so the MAC can be in `blocked_macs` (all `_wait_mac_blocked` gates on) while `[::]:8081`/`[::]:8443` is mid-rebind; the #1868 redirect then lands on no listener and the probe times out opaquely — read as a "v6 delivery regression" in the degraded-network 14:05 window (the #1935 egress flake co-fired).

## Fix (test reliability only — no agent/render change; #1865 untouched)
[...]

PR #1946, opened 2026-06-24
test(#1939): Gate-2 e2e — ws policy snapshot push→apply through the real agent

sameerparekh:
## What & why

Closes the gap (#1939): the websocket transport's **server→agent push path was never exercised end-to-end through the real OpenWRT agent**. Existing coverage only hit parts — the API emits to a *test* client (`RouterWsSpec`), the agent ws machinery is unit-tested in isolation (busted `ws_*_spec.lua`), and ingest parity uses the stdlib `ws_send.py` *fake* client (Gate 3). None drove a server **push** through the real `wifihaven-ws` sidecar.

## Scope (operator decision: test-only, defer the wiring)

While implementing this I found the push→apply→**enforce** path isn't actually wired in the agent: the sidecar's `on_policy` only `save_snapshot`s the pushed frame to `/etc/wifihaven/policy.json`, and the running agent reads that file **only at startup** — its tick loop applies policy exclusively from the HTTP poll. So a pushed snapshot updates the on-disk cache but doesn't change live enforcement until the next poll/restart.

Per the operator's call, this PR is **test-only**: it proves the push **transport** reaches the real agent and the snapshot is **saved**, but does **not** assert a live enforcement flip. Agent-side apply-on-push wiring + an enforcement-flip assertion are tracked in **#1945** (filed, Epic: Router Agent & Release).

## Changes

**Fake-API ws push (prerequisite).** The Gate-2 fake had no ws server, so the agent (which talks to the fake) had nothing to upgrade against. Adds `GET /api/router/ws`:
- 401-before-101 auth at upgrade (mirrors the REST routes + the real ws route),
- first-policy-on-connect push (#1849),
- push-on-change fan-out from `POST /test/snapshot`.

The `{"op":"policy","payload":<snapshot>}` frame is byte-identical to `RouterWsRegistry.policyFrameText`, so the agent's `ws_loop` sees the same frame whether the push came from the real Scala API or the fake. Inbound agent frames (usage/events/metrics) are drained+discarded; control pings are auto-ponged. Adds `/test/ws_status` so a scenario can sync on "agent connected" and assert the server side of the push.

**Gate-2 scenario** (`scenarios_fake/test_ws_push_apply.py`, marker `ws_push`): enables the sidecar via UCI (`wifihaven.ws.enabled=1`, default-off everywhere else) and **freezes the HTTP poll** (`policy_poll_interval=3600`) so any `policy.json` change after connect is attributable to the push alone.
[...]

Issue #1939, opened 2026-06-24
test(e2e): no Gate-2 coverage for policy SNAPSHOT push→apply over websocket — real agent applying a server-pushed snapshot (+ push-on-change #1849)


sameerparekh:
## Gap
The websocket transport's **outbound push path is not exercised end-to-end through a real agent**. We have:
- ✅ ws upgrade auth parity (Gate-3 `scripts/e2e-router.sh`, `RouterWsSpec`)
- ✅ agent→API **ingest** over ws (usage/events/metrics demux parity, Gate-3) — but via the fake `scripts/e2e/lib/ws_send.py` stdlib client, **not the real agent**
- ✅ API **emits** policy-on-connect — `RouterWsSpec` "#1849: a freshly-connected router is pushed the current policy snapshot on connect" — but that asserts the **server sends** a frame to a test client; no real agent, no apply, no on-change, no enforcement
- ✅ agent ws machinery in isolation — busted `openwrt/test/ws_*_spec.lua` (frame/backoff/loop/spool/metrics/usage-split) + `init_ws_toggle_spec.sh`

**Missing:** an end-to-end test where the API/fake-API **pushes a `policy` snapshot frame over the websocket** and the **real OpenWRT agent's `wifihaven-ws` sidecar** receives it → writes the snapshot file → the main agent applies it (etag-guarded) → **enforcement reflects the new policy** — with NO HTTP poll. Both legs:
1. **policy-on-connect**: agent connects with ws enabled, gets the current snapshot pushed, applies it, enforces it (a blocked host drops / an allowed host carves) without ever polling `GET /api/router/policy`.
2. **push-on-change (#1849)**: with the agent connected, a policy mutation server-side pushes ONE updated `policy` frame; the agent re-applies and enforcement changes (e.g. newly-extraBlocked host starts dropping, or pause/unpause flips), faster than / independent of the poll interval.

Today Gate 2 runs the agent with the ws sidecar **default-off** (UCI `wifihaven.ws.enabled=0`, see `init_ws_toggle_spec.sh` / #1848), so this whole path is untested e2e.

## Placement (per docs/testing.md three Gates)
Belongs in the **existing Gate 2 harness** (real-router → fake-API), NOT a new docker-compose job. Concretely:
- Enable the agent ws sidecar in the scenario via the UCI toggle (default stays off elsewhere).
[...]

Issue #1930, opened 2026-06-24
bug(router CD, P0): agent crash-loops on Lua 5.1 — 'function at line 965 has more than 60 upvalues' (regression from #1921/#1911 blockEncryptedDns) — blocks ALL router publishes


sameerparekh:
## Summary — P0, supersedes #1929
The OpenWRT agent **crash-loops at boot on the real Lua 5.1 target**, so it never fetches policy and **every** Gate 2 fake-mode scenario errors in setup (`33 errors, 17 passed, aborted in 4 min`). Router CD is hard-red; nothing publishes past v0.3.13 (so #1865 ::1:8443, #1864, #1848 ws sidecar all stay stuck). This masks #1929 (the ea_ carve test can't even run).

```
daemon.err wifihaven-agent: /usr/bin/lua: /usr/sbin/wifihaven-agent:1293: function at line 965 has more than 60 upvalues
procd: Instance wifihaven::agent is in a crash loop, 6 crashes, 0 seconds since last crash
ERROR ... - TimeoutError: no policy fetch after id=0 in 180s   (×33)
```

## Root cause
**Lua 5.1 hard limit: a function may have at most 60 upvalues.** This is a **load-time** (compile) error — the whole `wifihaven-agent` script fails to load, so the agent can't start at all. The function defined at **line 965** (referenced from line 1293) now closes over **>60 upvalues**.

**Introduced by #1921 (#1911 blockEncryptedDns enforcement).** Evidence:
- Run `28067273534` (tree @ #1920, WITHOUT #1921) ran the full 35-min suite — agent booted fine.
- Run `28069941702` (tree @ #1924, WITH #1921 + #1927) crash-loops at load with the upvalue error.
- The ws sidecar (#1927) is NOT the cause: the log shows `ws sidecar disabled via uci (wifihaven.ws.enabled=0)` and it starts cleanly; the crash is in the **main agent** script.
- #1921 added blockEncryptedDns enforcement (curated relay/DoH host lists + resolver-IP/DoT drop logic) inline; the new locals captured by the big function at line 965 pushed it past 60 upvalues.

## Why it wasn't caught
[...]

PR #1950, opened 2026-06-25
fix(#1949): keep lua-openssl for quic.lua alongside luaossl (sni-tail openssl.cipher nil)

### LINKED ISSUE #1949: bug(router CD, P0): sni-tail crash-loops — quic.lua 'openssl.cipher' nil (lua-openssl vs luaossl collision from #1946 image change) — blocks release
## Summary — P0, the NEW Gate-2 blocker after the egress/v6 flakes were fixed
With #1944 (egress) and #1948 (v6 flake) merged, Gate-2 run `28131712580` (sha 0dd14b6d) got past those — but a **new** failure surfaced (`2 failed, 48 passed`). The release is still stuck at v0.3.13.

```
daemon.err wifihaven-sni-tail: /usr/bin/lua: /usr/lib/lua/wifihaven/quic.lua:602: attempt to index field 'cipher' (a nil value)
        stack traceback: quic.lua:602: in function 'openssl_crypto'
procd: Instance wifihaven::sni-tail is in a crash loop, 6 crashes
FAILED test_sni_attribution.py::test_sni_only_hostname_attribution  (TimeoutError — SNI is the only attribution source, and sni-tail is down)
FAILED test_cname_direct_requery.py::test_eb_direct_requery_blocked (TimeoutError — block-page; likely collateral of the degraded agent)
```
(The egress flake is fully gone — `no servers could be reached` count = 0 this run. This is NOT #1935/#1943.)

## Root cause — lua-openssl vs luaossl module collision
`quic.lua:602` is:
```lua
local aes_ecb = openssl.cipher.get("aes-128-ecb")
local aes_gcm = openssl.cipher.get("aes-128-gcm")
```
This is the **lua-openssl** API (the monolithic `require"openssl"` package, where `openssl.cipher` is a field). The error `attempt to index field 'cipher' (a nil value)` means `openssl.cipher` is nil — i.e. the `openssl` module loaded at runtime is NOT lua-openssl.

**#1946 (#1939 ws e2e) changed `scripts/vm/build-router-image.sh`** to add **luaossl** to the Gate-2 router image (the ws sidecar #1848 needs luaossl per the #1845 spike). luaossl is a DIFFERENT library with a DIFFERENT shape: `require"openssl.cipher"` as a submodule, NOT `openssl.cipher` on a monolithic table. Adding/shadowing it broke lua-openssl for quic.lua → `openssl.cipher` resolves nil → sni-tail crash-loops on the first QUIC packet.

Evidence it regressed in this tree: `test_sni_only_hostname_attribution` **passed** in the 03:07 (`3f565d8f`) and 14:05 (`3da50362`) runs; it only crash-loops now, with #1946's image change merged.

## Fix options (pick one — chip/operator decide)
1. **Image: provide BOTH bindings without collision** — ensure the router image keeps **lua-openssl** (for quic.lua / sni-tail) AND **luaossl** (for the ws sidecar) installed under their distinct module names so neither shadows the other. Verify `require"openssl"` (lua-openssl, monolithic) and `require"openssl.cipher"` (luaossl) both resolve. Likely the smallest change.
2. **Port quic.lua to luaossl** — rewrite the `openssl_crypto` AES-ECB/GCM usage (quic.lua ~595-640) to luaossl's `require"openssl.cipher"` API, so the agent standardizes on ONE openssl lib (luaossl, the one ws needs). Larger but removes the dual-dependency.

## Validation (mandatory)
- Reproduce + verify on the REAL Lua 5.1 target (`ssh api.lan` / plex.lan) and/or a real router (`ssh root@openwrt.lan`): `require"openssl"; openssl.cipher.get("aes-128-gcm")` must succeed, AND the ws sidecar's luaossl path must still load. The macOS build masks which openssl lib resolves — validate on-target.
- Add a busted/compile guard that loads quic.lua and asserts `openssl.cipher.get` is callable, so a future image change that drops lua-openssl fails fast (mirrors the #1930 luac compile-guard idea).
- Confirm Gate-2 `test_sni_only_hostname_attribution` + `test_eb_direct_requery_blocked` pass, Master Router CD greens, and a release publishes past v0.3.13.

Relates to #1946 / #1939, #1848 / #1845 (luaossl for ws), the sni-tail/quic SNI path. Found while watching the router CD incident.

sameerparekh:
### Root cause
After #1944 (egress) and #1948 (v6 flake) merged, Gate-2 surfaced a **new** crash:

```
daemon.err wifihaven-sni-tail: /usr/bin/lua: /usr/lib/lua/wifihaven/quic.lua:602: attempt to index field 'cipher' (a nil value)
procd: Instance wifihaven::sni-tail is in a crash loop, 6 crashes
```

`quic.lua:602` is `openssl.cipher.get("aes-128-ecb")` — the **lua-openssl** (zhaozg) monolithic API, where `openssl.cipher` is a field on one table.

**#1946** added **luaossl** to the Gate-2 router image for the ws sidecar (#1848/#1845). luaossl and lua-openssl both claim the `openssl` Lua **module name**:
- lua-openssl ships `/usr/lib/lua/openssl.so` (C → `package.cpath`)
- luaossl ships `/usr/lib/lua/openssl.lua` (Lua → `package.path`), a 25-byte `return require"_openssl"`

Lua's **Lua-file searcher runs before the C searcher**, so a bare `require"openssl"` now resolves to luaossl's top table — which has **no `.cipher`/`.hmac` field** → `openssl.cipher` is nil → sni-tail crash-loops on the first QUIC packet → SNI attribution dies (`test_sni_only_hostname_attribution`), and `test_eb_direct_requery_blocked` fails as collateral of the degraded agent.

This is the **second** time a bare `require"openssl"` killed sni-tail (first was #1717: lua-openssl missing from `.ipk` DEPENDS).

### Fix — Option 1 (keep both libs, resolve past the shadow)
`quic.resolve_openssl()` returns the lua-openssl module even when luaossl shadows the bare name: fast-path `require"openssl"` when unshadowed (default ws-off router), else force-load lua-openssl's `openssl.so` directly off `package.cpath`. sni-tail calls it instead of `require("openssl")`.

luaossl is left **fully intact** — its submodules (`require"openssl.cipher"/.rand/.ssl.context`, used by the ws sidecar's TLS) resolve independently of the bare `openssl` name, and the ws sidecar runs in a separate process anyway (it selects `ws_crypto.luaossl()`, never the monolithic backend).

**Why not Option 2 (port quic to luaossl):** luaossl is deliberately *not* a base DEPENDS — `openwrt/Makefile` keeps default (ws-off) installs lean and the operator installs luaossl out-of-band only when enabling ws. Porting quic to `require"openssl.cipher"` would crash sni-tail on every default prod router, which has only lua-openssl. quic stays on lua-openssl (always in DEPENDS, guarded by `build_depends_sync_spec` since #1717).

### Prod-image parity
The fix is in the **wifihaven package** (`quic.lua` + `wifihaven-sni-tail`), shipped identically to the Gate-2 image and to prod routers — **no `build-router-image.sh` change**. That is strictly better than an image-only change, which would fix the e2e VM but miss prod routers that install luaossl when an operator opts into ws.

### Guards (mirror the #1930 luac compile-guard: fail in seconds in unit CI, not 35 min into Gate 2)
- `quic_spec.lua`: busted `resolve_openssl` selection tests with injected fakes — deterministic in CI (CI installs `lua-luaossl`, the shadowing lib).
- `quic_openssl_resolve_spec.sh`: on-target end-to-end guard. Detects lua-openssl via `openssl.so` on cpath (**not** a bare require, which the shadow defeats — so it doesn't wrongly skip the exact scenario it guards), asserts `cipher.get("aes-128-gcm"/"aes-128-ecb")` is callable through `openssl_crypto`. Skips where lua-openssl is absent (unit CI, macOS); runs on the router / plex.lan.

### On-target validation (real Lua 5.1.5 on `openwrt.lan`, luaossl installed to reproduce the shadow)
- **Reproduced**: bare `require"openssl".cipher == nil` once luaossl present.
- **Fixed**: `resolve_openssl()` recovers lua-openssl; `cipher.get(gcm/ecb)` OK; `openssl_crypto` wired; **AES-128-ECB (FIPS-197) + HMAC-SHA256 (RFC 4231) known-answer vectors match**.
- **ws intact**: `openssl.cipher` / `openssl.rand` / `openssl.ssl.context` all still load.
- Test router restored to baseline (luaossl/cqueues removed) after validation.
- Full `openwrt/test/run_tests.sh` green locally.

Fixes #1949. Relates to #1946 / #1939 and #1848 / #1845.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/quic.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/quic.lua
@@ -32,6 +32,8 @@
 --   quic.parse_initial(eth_frame, crypto)
 --     → { dst_ip, src_mac, sni } | nil, reason
 --   quic.QUIC_V1_SALT  — HKDF salt for the v1 initial_secret derivation.
+--   quic.resolve_openssl() → the lua-openssl module, resolved even when luaossl
+--     shadows the bare `openssl` module name (#1949).
 --   quic.openssl_crypto(openssl) → crypto table backed by lua-openssl, for
 --     the sidecar to wire in.
 --
@@ -561,6 +563,75 @@ function M.parse_initial(eth_frame, crypto)
     return { dst_ip = dst_ip, src_mac = src_mac, sni = sni_str }, reason
   end
   return { dst_ip = dst_ip, src_mac = src_mac, sni = sni_str }
+end
+
+-- ---------------------------------------------------------------------------
+-- resolve_openssl([env]) — return the lua-openssl (zhaozg) monolithic module
+-- (the one whose table carries `.cipher` / `.hmac` directly), even when luaossl
+-- is also installed and shadows a bare require"openssl" (#1949).
+--
+-- Both feed packages claim the `openssl` module name:
+--   * lua-openssl (zhaozg) ships /usr/lib/lua/openssl.so   — found on cpath (C)
+--   * luaossl              ships /usr/lib/lua/openssl.lua   — found on path (Lua)
+-- Lua's Lua-file searcher (package.path) runs BEFORE the C searcher
+-- (package.cpath), so on any router that ALSO has luaossl installed (the ws
+-- sidecar's runtime dep #1848/#1845, and the Gate-2 e2e image which installs it
+-- explicitly) a bare require"openssl" returns luaossl's top table — which has no
+-- `.cipher`/`.hmac` field — and openssl_crypto below crashed sni-tail in a procd
+-- respawn loop (the QUIC SNI attribution path went dark). A default, ws-off
+-- router has only lua-openssl, so the fast path below returns it unchanged.
+--
+-- We do NOT touch DNS or remove luaossl: luaossl is required for the ws
+-- sidecar's TLS (require"openssl.ssl.context") and its submodules
+-- (require"openssl.cipher" etc.) resolve independently of which module the bare
+-- name `openssl` binds to, so they are unaffected. We simply force-load
+-- lua-openssl's openssl.so by path when the bare require is shadowed.
+--
+-- `env` is an injection seam for unit tests (require/loadlib/cpath/exists);
+-- production calls resolve_openssl() with no args.
+-- ---------------------------------------------------------------------------
+function M.resolve_openssl(env)
+  env = env or {}
+  local req     = env.require or require
+  local loadlib = env.loadlib or package.loadlib
+  local cpath   = env.cpath or package.cpath
+  local exists  = env.exists or function(p)
+    local fh = io.open(p, "rb")
+    if fh then fh:close(); return true end
+    return false
+  end
+
+  -- lua-openssl exposes the AES + HMAC primitives openssl_crypto needs as
+  -- fields on one table; luaossl's top module exposes neither.
+  local function is_lua_openssl(m)
+    return type(m) == "table" and m.cipher ~= nil and m.hmac ~= nil
+  end
+
+  -- Fast path: nothing shadows it (default ws-off router has only lua-openssl).
+  local ok, mod = pcall(req, "openssl")
+  if ok and is_lua_openssl(mod) then
+    return mod
+  end
+
+  -- Shadowed (luaossl won the bare name) or require failed: force-load
+  -- lua-openssl's monolithic C module directly off package.cpath.
+  for tmpl in cpath:gmatch("[^;]+") do
+    local file = tmpl:gsub("%?", "openssl")
+    if exists(file) then
+      local loader = loadlib(file, "luaopen_openssl")
+      if loader then
+        local okc, m = pcall(loader)
+        if okc and is_lua_openssl(m) then
+          return m
+        end
+      end
+    end
+  end
+
+  error("wifihaven(#1949): lua-openssl (monolithic require\"openssl\" exposing " ..
+    ".cipher/.hmac) not resolvable — luaossl shadows the 'openssl' module name " ..
+    "and no lua-openssl openssl.so was found on package.cpath. sni-tail needs " ..
+    "lua-openssl for QUIC Initial decryption.")
 end
 
 -- ---------------------------------------------------------------------------
--- a/openwrt/files/usr/sbin/wifihaven-sni-tail
+++ b/openwrt/files/usr/sbin/wifihaven-sni-tail
@@ -35,10 +35,15 @@
 -- quic.parse_initial. Dispatch is on the IPv4 protocol byte.
 
 local uci     = require("uci")
-local openssl = require("openssl")
 local sni     = require("wifihaven.sni")
 local reasm   = require("wifihaven.sni_reassembly")
 local quic    = require("wifihaven.quic")
+-- #1949: resolve lua-openssl explicitly (NOT a bare require"openssl"). When the
+-- ws sidecar's luaossl is also installed it ships its own openssl.lua, which the
+-- Lua-file searcher finds before lua-openssl's openssl.so and returns a table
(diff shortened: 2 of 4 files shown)


sameerparekh:
## Independent PR review — #1950 (first run)

Reviewed the full merge-base diff (`git diff origin/main...HEAD`, three-dot) against the 8 checklist dimensions. Read-only; verdict lives in this comment, not a GitHub review state.

**BLOCKERS**

(none)

**SHOULD-FIX**

(none)

**NITS**

1. `openwrt/test/quic_openssl_resolve_spec.sh:30-46` — the guard re-implements lua-openssl detection (`lua_openssl_installed()` scanning `package.cpath` for `openssl.so`) that mirrors `quic.resolve_openssl`'s own cpath scan (`quic.lua:592-604`). This duplication is **intentional and correct**: the guard must decide whether to skip *without* depending on the function under test, and a bare `require"openssl"` is exactly what the shadow defeats. Flagging only so a future reader doesn't "DRY" the two together and reintroduce the skip-under-shadow trap; the spec's comment already explains the rationale.
2. TDD red→green was demonstrated in-session (resolver tests fail with `attempt to call a nil value (field 'resolve_openssl')` before the fix, pass after) but landed in a single commit rather than a separate failing-test commit. Not worth rewriting history on a P0 hotfix; noting for the record.
3. `quic.lua:608` — `resolve_openssl` returns lua-openssl locally without overwriting `package.loaded.openssl` (which the failed fast-path `pcall(require,"openssl")` left pointing at luaossl). This is deliberate (no global side-effects; documented), and safe because sni-tail no longer does a bare `require"openssl"` anywhere else and `openssl_crypto` takes the module as a param. No action needed.

**Dimension notes**

- **1 Single-source-of-truth**: `resolve_openssl` is the one place resolving lua-openssl; no enforcement/display split, no "keep in sync." OK.
- **2 Testing**: `quic_spec.lua` is +100/-0 — no existing assertion weakened/deleted. New behavior is asserted across 5 busted cases (fast-path, shadow-recovery, skip-wrong-shape, require-errors, unresolvable→error) plus an on-target end-to-end guard. Bug-fix regression test fails without the fix (verified). No forbidden mocks.
- **3 Architecture**: explicitly does NOT touch DNS; no decision logic added to the agent; no new `/tmp` writes; `error()` on absent lua-openssl is correct fail-loud (DEPENDS + `build_depends_sync_spec` guarantee presence). OK.
- **4 Wire / 5 Migrations / 8 Security**: no wire/schema/route/SQL/secret changes; `loadlib` targets the fixed `"openssl"` name from system cpath (not attacker-controlled). OK.
- **6 Metrics**: startup module-resolution plumbing, not a new route/job/ingest path; failure crashes loudly and is caught by the guard + Gate 2. No metric warranted.
- **7 Scope**: focused 4-file diff, no dead code, no AI mentions in title/body/commit (only `Co-Authored-By` trailer), three-dot diffs used. OK.

**On-target validation** (reported in the PR body, independently credible): real Lua 5.1.5 on `openwrt.lan` with luaossl installed reproduced `require"openssl".cipher == nil`; `resolve_openssl()` recovered lua-openssl with AES-128-ECB (FIPS-197) + HMAC-SHA256 (RFC 4231) KATs matching; luaossl submodules still load. Prod parity holds because the fix ships in the `wifihaven` package (no image-build change).

VERDICT: APPROVE

A tightly-scoped P0 fix for the lua-openssl/luaossl `openssl` module-name collision that crash-looped sni-tail post-#1946. The resolver is correct, well-documented, fails loud when lua-openssl is genuinely absent, and leaves the ws sidecar's luaossl path untouched. No BLOCKERs or SHOULD-FIXs; the NITs are informational. On-target KAT validation on real Lua 5.1 is the right gate for this class of bug.

PR #1953, opened 2026-06-25
fix(#1951): ws-parity e2e tolerates #1849 policy push (match acks to sent frames)

### LINKED ISSUE #1951: bug(API/UI CD): live-e2e ws demux-parity check fails — #1849 push-on-change sends a 'policy' frame the #1846 'all acked' assertion doesn't tolerate
## Summary
Master **API/UI CD** has been red since ~2026-06-24 21:46 (the #1849 merge wave). The `Live e2e against staging stack` job runs `scripts/e2e-router.sh`, prints **"All e2e checks passed."**, then the **#1846 websocket demux-parity** section fails and the job exits 1:

```
▶ #1846: websocket transport demux parity with REST
  ✓ ws upgrade auth parity: missing/bad → 401, valid → 101
  ✗ ws frames not all acked ok: bad:
      [{"op":"policy","payload":{…snapshot…}},
       {"op":"ack","payload":{"op":"events","seq":1,"status":"ok"}},
       {"op":"ack","payload":{"op":"usage","seq":2,"status":"ok"}}]
```

## Root cause — test assumption broken by #1849, NOT a product bug
The #1846 ws-parity check (in `scripts/e2e-router.sh`) sends `events`/`usage`/`metrics` frames over one ws connection and asserts that **every reply frame is an `ack ok`**. That assumption held when #1846 landed. But **#1849 (push-on-change)** now makes the server **push an unsolicited `policy` frame** whenever the snapshot changes — and the e2e script performs policy mutations (create profile/device, etc.) earlier in the run, so by the time the ws frames are sent, a `snapshot_changed` push is delivered to the connected ws client and lands in the replies array alongside the acks. The check sees a non-`ack` (`op:policy`) reply and fails "not all acked ok".

The transport is behaving correctly: 2 acks (events seq1, usage seq2) + 1 server-initiated `policy` push. The check just predates #1849.

## Fix (test/harness only)
Update the #1846 ws-parity assertion in `scripts/e2e-router.sh` (and `scripts/e2e/lib/ws_send.py` if needed) to tolerate server-initiated `policy` push frames:
- Assert that **each SENT frame got a corresponding `ack ok`** (match acks to the seq numbers it sent), rather than "every reply is an ack".
- **Filter out / ignore `op==policy`** (and any other server→client push op) frames before the ack check — they are expected #1849 push-on-change traffic, not parity violations.
- Optionally add a positive assertion that a `policy` push IS received after a mutation (turn the surprise into coverage) — but the minimal fix is to stop failing on it.

Do NOT change the ws transport behavior — #1849's policy push is correct.

## Acceptance
- `scripts/e2e-router.sh` #1846 ws-parity section passes with #1849 push-on-change active (against the docker staging stack).
- Master API/UI CD goes green.
- (Bonus) the script asserts every sent frame is acked AND ignores/【optionally expects】 the policy push.

Relates to #1849 (push-on-change), #1846 (#1023 ws transport), #1939 (ws push→apply e2e). Found while investigating both CD pipelines red.

sameerparekh:
## Problem

Master **API/UI CD** has been red since the #1849 merge wave. The `Live e2e against staging stack` job runs `scripts/e2e-router.sh`, which passes its main checks and then fails the **#1846 ws demux-parity** section:

```
✗ ws frames not all acked ok: bad:
    [{"op":"policy","payload":{…snapshot…}},
     {"op":"ack","payload":{"op":"events","seq":1,"status":"ok"}},
     {"op":"ack","payload":{"op":"usage","seq":2,"status":"ok"}}]
```

## Root cause — test assumption broken by #1849, NOT a product bug

The #1846 parity check sent `events`/`usage`/`metrics` frames over one ws connection and asserted **every reply is an `ack ok`**. After #1849 the server pushes an unsolicited `{op:"policy"}` frame on connect (first-policy-on-connect, `RouterWsRoutes` `HandshakeComplete`) and on every snapshot change. An earlier policy mutation in the script makes a `policy` push land among the replies, so the check saw a non-`ack` and failed.

Worse, `ws_send.py` read **exactly one reply per sent frame**, so the interleaved push displaced the 1:1 mapping and the **third frame's ack was never read** — that's why the failing array shows only 2 acks.

The transport is correct: 3 acks + 1 server-initiated `policy` push. The assertion just predates #1849.

## Fix (test/harness only — match acks to sent frames)

- **`scripts/e2e/lib/ws_send.py`** — send all frames, then read until every **sent** frame has a *solicited* reply (`ack`/`pong`), collecting interleaved server pushes (`_PUSH_OPS = {"policy"}`) without counting them toward completion. The push can no longer displace a frame's reply or leave the last ack unread.
- **`scripts/e2e-router.sh`** — assert each **sent** frame got a matching `ack ok` (matched by `op`+`seq`) and **partition out** `op==policy` pushes before the check, instead of "every reply is an ack". Genuine parity is preserved: a missing or `reject` ack still fails. The policy push is surfaced as coverage (`ok:<count>`), **not** a gate — its arrival races frame dispatch, so it may land after the acks and the check must not depend on its timing.

**ws transport / server behavior is unchanged** — #1849's policy push is correct.

## Tests

New `scripts/e2e/lib/test_ws_send.py` drives `send_frames` against a minimal stdlib ws server that injects a `policy` push **before / between / after** the acks. It pins:
- every sent frame is acked regardless of where the push lands (the previously-dropped-ack case),
- the push is surfaced (not swallowed),
- no extra acks beyond the sent frames (parity not weakened).

Red→green committed separately. Verified the `e2e-router.sh` ack-matching branch across: push-first (the repro), no-push, missing ack, rejected ack, and ws_send error — only genuine breaks fail.

Fixes #1951. Relates to #1849 (push-on-change), #1846 / #1023 (ws transport), #1939 (ws push→apply e2e).

### DIFF
--- a/scripts/e2e-router.sh
+++ b/scripts/e2e-router.sh
@@ -364,10 +364,14 @@ WS_REPLIES=$(printf '%s' "$PARITY" \
 WS_PARITY=$(printf '%s\n%s' "$PARITY" "$WS_REPLIES" | _py "
 import json, sys
 sent = json.loads(sys.stdin.readline()).get('ws_frames', [])
-replies = json.loads(sys.stdin.readline())
+raw = sys.stdin.readline()
+if not raw.strip():
+    print('bad: ws_send produced no output (helper crashed before any reply)'); sys.exit()
+replies = json.loads(raw)
 if not isinstance(replies, list):
     print('bad: ws_send error: ' + json.dumps(replies)); sys.exit()
 # Server→client pushes are expected #1849 traffic — partition them from the acks.
+# NB: the push-op set here must mirror ws_send.py's _PUSH_OPS (only 'policy' today).
 pushes = [r for r in replies if isinstance(r, dict) and r.get('op') == 'policy']
 acks   = [r for r in replies if isinstance(r, dict) and r.get('op') == 'ack']
 def acked_ok(op, seq):
--- a/scripts/e2e-router.sh
+++ b/scripts/e2e-router.sh
@@ -348,22 +348,45 @@ for kind in events usage metrics; do
 done
 pass "REST usage/events/metrics accepted (200)"
 
-# ws leg — send all three frames over one connection; every reply must be ack ok.
+# ws leg — send all three frames over one connection; each SENT frame must get
+# a matching `ack ok` (matched by op+seq). The server ALSO pushes unsolicited
+# `{op:policy}` frames (#1849 first-policy-on-connect / push-on-change), and an
+# earlier policy mutation in this script makes one land among the replies; those
+# are expected push-on-change traffic, NOT a parity violation, so we partition
+# them out before the ack check rather than asserting "every reply is an ack"
+# (the pre-#1849 assumption that broke in #1951).
 # `|| true` so a ws_send.py failure doesn't abort under `set -e` before we can
 # surface it: on error the helper prints `{"error":…}` to stdout, which the
-# ACKS_OK parse below reports verbatim in the fail message.
+# WS_PARITY parse below reports verbatim in the fail message.
 WS_REPLIES=$(printf '%s' "$PARITY" \
   | _py "import json,sys; print(json.dumps(json.load(sys.stdin)['ws_frames']))" \
   | python3 "$WS_SEND" --url "$WS_URL" --token "$RTOK" || true)
-ACKS_OK=$(printf '%s' "$WS_REPLIES" | _py "
+WS_PARITY=$(printf '%s\n%s' "$PARITY" "$WS_REPLIES" | _py "
 import json, sys
-replies = json.load(sys.stdin)
-ok = (isinstance(replies, list) and len(replies) == 3
-      and all(r.get('op') == 'ack' and (r.get('payload') or {}).get('status') == 'ok' for r in replies))
-print('ok' if ok else 'bad: ' + json.dumps(replies))
+sent = json.loads(sys.stdin.readline()).get('ws_frames', [])
+replies = json.loads(sys.stdin.readline())
+if not isinstance(replies, list):
+    print('bad: ws_send error: ' + json.dumps(replies)); sys.exit()
+# Server→client pushes are expected #1849 traffic — partition them from the acks.
+pushes = [r for r in replies if isinstance(r, dict) and r.get('op') == 'policy']
+acks   = [r for r in replies if isinstance(r, dict) and r.get('op') == 'ack']
+def acked_ok(op, seq):
+    return any((a.get('payload') or {}).get('op') == op
+               and (a.get('payload') or {}).get('seq') == seq
+               and (a.get('payload') or {}).get('status') == 'ok'
+               for a in acks)
+missing = ['%s#%s' % (f['op'], f['seq']) for f in sent if not acked_ok(f['op'], f['seq'])]
+if missing:
+    print('bad: unacked ' + ','.join(missing) + ' in ' + json.dumps(replies))
+else:
+    # ok[:push-count] — surface whether the #1849 push was observed (coverage,
+    # not a gate: its arrival races frame dispatch so it may land after the acks).
+    print('ok:%d' % len(pushes))
 ")
-[ "$ACKS_OK" = ok ] || fail "ws frames not all acked ok: $ACKS_OK"
-pass "ws usage/events/metrics frames each acked ok"
+case "$WS_PARITY" in
+  ok:*) pass "ws usage/events/metrics frames each acked ok (op+seq); #1849 policy pushes observed: ${WS_PARITY#ok:}, ignored" ;;
+  *)    fail "ws frames not all acked ok: $WS_PARITY" ;;
+esac
 
 # (c) Deep parity — the connection events land identically whether posted over
 # REST or ws. Read both MACs via the public /api/logs and compare the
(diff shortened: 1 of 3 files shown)


sameerparekh:
## Independent PR review — `scripts/e2e-router.sh` ws-parity tolerates #1849 policy push

Read-only adversarial pass over `git diff origin/main...HEAD` (8 dimensions). Test-harness-only change; no product/wire/server behavior touched.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**

1. `scripts/e2e/lib/ws_send.py:44` + `scripts/e2e-router.sh:374` — the knowledge that `policy` is a server→client push op is encoded twice: `_PUSH_OPS = frozenset({"policy"})` in the client and `r.get('op') == 'policy'` in the shell filter. They're independent layers (transport read-completion vs. assertion partition), so this isn't a single-source-of-truth violation, but a future push op (e.g. a `#1847` `ready`) would need both updated. Low risk for a harness; a one-line comment cross-referencing the pair would suffice.

2. `scripts/e2e-router.sh:355-372` — if `WS_REPLIES` is empty (helper crashed before any stdout, swallowed by `|| true`), the second `sys.stdin.readline()` feeds `json.loads('')`, which raises, the snippet exits non-zero, and `WS_PARITY` is empty → the `*)` arm fails with an uninformative `ws frames not all acked ok:` (no detail). The `{"error":…}` path *is* handled cleanly (valid JSON, non-list → "bad: ws_send error"); only the truly-empty case is terse. Same failure mode existed pre-PR (`json.load` on empty), so it's not a regression — just unchanged roughness.

3. PR body carries the `🤖 Generated with Claude Code` footer, which the checklist §7 flags ("no AI mentions in PR body"). Non-blocking and noted only for completeness — it's harness-mandated and the operator's standing preference is that AI authorship needn't be hidden; commit messages and PR title are clean.

**Observations (not findings)**

- **Genuine parity preserved.** The new check still fails on a real break: a missing ack (`bad: unacked metrics#3`) or a rejected ack (`status != ok`) both produce `bad:` and `fail`. Verified across push-first / no-push / missing-ack / rejected-ack / ws_send-error inputs. The fix narrows tolerance to exactly server-initiated `op==policy` pushes, not "any non-ack reply."
- **`send_frames` behavior change is sound.** It now sends all frames, then drains until every *sent* frame has a *solicited* reply (`ack`/`pong`), counting `_PUSH_OPS` frames as collected-but-not-gating. This is what fixes the dropped-seq3 symptom (the old read-one-per-send misaligned on the interleaved push). `ping`→`pong` still counts as solicited, so the generic contract is intact; the bounded `sock.settimeout` means a never-completing server times out into `{"error":…}` rather than hanging.
- **TDD red→green visible** (commit `6e58a2db` failing test → `c00f85ac` fix). New `test_ws_send.py` drives the real `send_frames` against a stdlib ws server injecting the push **before / between / after** the acks, pinning: every sent frame acked regardless of push position (the previously-dropped-ack case), push surfaced not swallowed, no extra acks beyond sent frames (parity not weakened). Pure-transport unit test — appropriate level; no forbidden mocks (real sockets).
- Dimensions 3 (arch invariants), 4 (wire), 5 (migrations), 6 (metrics), 8 (security) — N/A; this is an e2e test client, no server/router/DB/route/secret surface.

**VERDICT: APPROVE**

Test-harness-only fix that makes the #1846 ws demux-parity check tolerate #1849's unsolicited `policy` push by matching acks to sent frames (op+seq) and partitioning out pushes, in both the shell assertion and the `ws_send.py` read loop. Genuine parity coverage is preserved (missing/rejected acks still fail) and TDD red→green is in history. No blockers; the NITs are harness-local and non-gating.

sameerparekh:
## Independent PR review — re-run (delta `c00f85ac...d174f90e`)

Re-ran on the NIT-fix push. Statused prior findings, reviewed the incremental delta for new findings.

### Prior findings status

|---|---|---|
| NIT 1 | `policy` push-op knowledge duplicated in `ws_send.py` `_PUSH_OPS` and the shell filter, no cross-link | **ADDRESSED** — reciprocal "must mirror" comments added in `ws_send.py:44-45` and `scripts/e2e-router.sh:370`. Still a deliberate two-layer split (not collapsible across a shell/python boundary), now drift-flagged on both sides. |
| NIT 2 | Empty `WS_REPLIES` → `json.loads('')` crash → uninformative empty `WS_PARITY` | **ADDRESSED** — `scripts/e2e-router.sh:367-369` now detects empty stdout and prints `bad: ws_send produced no output (helper crashed before any reply)` before parsing. Verified: empty→precise message, error-obj→"bad: ws_send error", happy+push→`ok:1`. |
| NIT 3 | PR body carries the Claude Code footer (checklist §7) | **NOT-ADDRESSED (intentional)** — harness-mandated; operator's standing preference is that AI authorship needn't be hidden. Non-gating, unchanged. |

### New findings in the delta

**BLOCKERS** — (none)
**SHOULD-FIX** — (none)
**NITS** — (none)

The delta is comment-only plus one guard clause (`if not raw.strip()`) that strictly precedes the existing parse; it adds no new branch to the happy/parity path and does not touch the transport. `bash -n` + `shellcheck --severity=warning` clean; the three `test_ws_send.py` cases still pass.

**VERDICT: APPROVE**

Both actionable NITs from the first pass are ADDRESSED with no new findings. The change remains a test-harness-only fix that tolerates #1849's `policy` push while preserving genuine parity (missing/rejected acks still fail). No blockers across either run.

PR #1956, opened 2026-06-25
fix(#1954): snapshot cache returns fresh on read-after-write (all mutation paths invalidate)

### LINKED ISSUE #1954: bug(API, P1): #1849 snapshot cache serves STALE snapshot — newly-created profile missing from /api/router/policy (Gate 1 / API-UI CD red)
## Summary — P1, blocks Master API/UI CD AND is a real enforcement-correctness regression
Master API/UI CD's **Gate 1 — API contract smoke (staging)** fails: after creating a profile + assigning a device, the next `/api/router/policy` snapshot **omits the just-created profile**. (The #1953 ws-parity fix cleared the earlier live-e2e failure, un-masking this one.)

```
✓ profile id=306 name=e2e-router-gh-…
✓ device mac=e2:e2:e2:8e:79:5b → profile 306
✓ usage posted (90 active seconds)
profile 306 missing from snapshot.profiles: 247,1,2,298,303,299,246,300,301,37,302,38,297
```

(The first Gate-1 attempt 502'd — that's the **known starter-tier cold-start** blip documented in render.yaml; the re-run got past it and hit this real failure.)

## Root cause — #1849 computed-snapshot cache invalidation gap (single instance)
- Staging is **`numInstances: 1`** (render.yaml:45) → NOT a cross-instance cache-coherence problem. It's a **single-instance** staleness.
- **Pre-#1849** the policy snapshot was **computed on every read** (always fresh). **#1849 (#1934)** added a computed-snapshot cache (`Ref[Option[(ETag, PolicySnapshot)]]` in `PolicyService`) invalidated "on any policy mutation." The Gate-1 sequence is create-profile → assign-device → **read snapshot**, and the read returns a snapshot that predates the create. So either:
  1. the **profile-create and/or device-create/assign (PUT)** mutation paths are **not wired into the cache invalidation** set, or
  2. invalidation marks the cache dirty but the **rebuild is async** (the #1849 re-evaluation ticker) and a read immediately after a mutation returns the stale cached value before the rebuild.
- Either way the #1849 cache broke the **create-then-read freshness invariant** the snapshot had pre-#1849.

## Why it matters (beyond CD)
The snapshot is what routers fetch via `GET /api/router/policy`. A stale snapshot after a policy change = **enforcement lag**: a router could keep serving an old policy (missing a new profile/device/block) until the cache happens to rebuild. #1849 is on prod now.

## What to determine + fix
1. Identify exactly which mutation paths invalidate the #1849 cache vs which don't (check `PolicyService` cache-invalidation wiring against ALL policy-affecting writes: profile create/update/delete, device create/assign/PUT/delete, schedule, blocklist, pause, manual block/allow, time-extension).
2. Fix so that **a read after ANY policy mutation returns a fresh snapshot** — invalidation must cover every mutation path AND the next read must rebuild synchronously (do NOT rely on the ticker for correctness; the ticker is only for time-dependent transitions). Preserve the #1849 perf win (cache hits when nothing changed) without sacrificing read-after-write freshness.
3. Add a **feature test**: create a profile (and a device assigned to it) and assert the VERY NEXT `/api/router/policy` / snapshot read includes it — i.e. the exact create-then-read invariant Gate 1 checks. Pin the same for device create/assign and delete.

Relates to #1849 (#1934 computed-snapshot cache), #1023. Found while investigating Master API/UI CD red (Gate 1 + Gate 3b).

sameerparekh:
## What

Fixes the **#1849 computed-snapshot cache** serving a **stale** snapshot after a policy mutation. Master API/UI CD **Gate 1** (`Router API smoke against staging`) failed: after creating a profile + assigning a device, the very next `GET /api/router/policy` snapshot **omitted the just-created profile** (`profile 306 missing from snapshot.profiles: …`). Staging is `numInstances: 1`, so this is single-instance staleness, not cache coherence.

## Root cause — an in-flight build clobbers the cleared cache

Every mutation route *does* call `invalidate` (audited: profile create/update/patch/delete, device upsert/patch/delete, schedule attach/edit, household settings incl. `blockEncryptedDns`, pause/unpause, manual block/allow, time-extension — all wired). So the missing-invalidation theory was **not** the bug.

The bug is the **async-rebuild race**. #1849's `invalidate` *cleared* the cache slot (`set(None)`) and let the next read rebuild, and `reevaluate`/the read path *unconditionally stored* whatever they built. That is correct for a quiescent create-then-read, but a `buildSnapshot` that began reading the DB **before** a mutation can finish **after** `invalidate` cleared the slot and repopulate it with **pre-mutation bytes**:

1. reconcile ticker's `reevaluate` reads the DB (profile 306 not yet created) — build is slow (~500 ms)
2. `POST /api/profiles` commits 306 → `invalidate` clears the slot
3. the in-flight build from (1) finishes and **stores its stale bytes** (no 306)
4. `GET /api/router/policy` → cache **HIT** on the stale entry → 306 missing ← Gate-1 failure

A stale snapshot served to a router is also real **enforcement lag**, not just a CD red.

## Fix — version-stamp the cache (choke-point, race-safe)

- A single monotonic `mutationVersion` is bumped once per mutation. `invalidate` now **bumps** the version instead of clearing.
- Every cache entry is **stamped** with the version observed *before* its DB read (`buildVersioned` / `reevaluate`). Because the bump happens-after the mutation's commit (routes call `invalidate` after the write), a stamp equal to the current version is guaranteed to reflect every committed mutation.
- The read path serves a `cache_hit` **only** when `builtAt == mutationVersion`; otherwise it rebuilds **synchronously**. `installSnapshot` refuses to overwrite a newer-version entry (CAS).
- A late stale build therefore lands tagged with the **old** version and is rejected at **both** install and read — read-after-write is fresh again. No reliance on the reconcile ticker for correctness (it remains only for time-dependent transitions: schedule boundaries, daily-limit exhaustion).

Single-source-of-truth preserved: `reevaluate` is still the sole rebuild-and-publish point; same snapshot bytes for a given state; no second snapshot computation.

### #1849 perf win preserved
Unchanged version ⇒ `cache_hit` with no rebuild (the #1512 win); 304 semantics unchanged; push-on-change still keys off `lastPublishedEtag` so a racing REST poll can't suppress a push.

## Tests (TDD: red commit, then green)

`PolicySnapshotCacheSpec` (embedded Postgres, injected `Clock.TestClock`):
- **Regression pin** — a `buildBarrier` test seam suspends an in-flight build; a create + `invalidate` + fresh rebuild land while it's parked; the stale build is released to install **last**; the next read must still include the new profile. Faithful-red commit (main's clear-on-invalidate behavior) fails **only** this test; green passes.
- create-then-read invariant: a newly created + invalidated profile (and its assigned device) is in the **very next** read — the exact Gate-1 sequence.
- delete-then-read: a deleted + invalidated profile disappears on the next read.
- existing cache_hit / invalidate / push-on-change / schedule-boundary specs still green.

Full `mill api.test` suite green (1109 tests).

Fixes #1954. Relates to #1849 / #1934 (computed-snapshot cache), #1023.

### DIFF
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -155,8 +155,9 @@ class PolicyServiceLive(
   // single-household model there is exactly one global snapshot, so one slot suffices (design §6.2
   // ticker note). Populated on first build, refreshed by `reevaluate`.
   //
-  // #1954: each entry is STAMPED with the `mutationVersion` observed at the START of its build
-  // (`builtAtVersion`, below). A reader trusts a cached entry only when its stamp is still the
+  // #1954: each entry is STAMPED with the `mutationVersion` observed at the START of its build (the
+  // leading `Long` of the tuple — the `gen` captured in `buildVersioned`/`reevaluate`). A reader
+  // trusts a cached entry only when its stamp is still the
   // current version; a stamp older than the current version means a mutation landed since the build
   // began, so the entry may predate that mutation and is treated as a miss (rebuild). This is what
   // closes the stale-read race that the old "clear-on-invalidate" alone could not: a `buildSnapshot`
--- a/api/src/policy/PolicyService.scala
+++ b/api/src/policy/PolicyService.scala
@@ -11,7 +11,7 @@ import zio.json.*
 
 import java.security.MessageDigest
 import java.time.{DayOfWeek, Instant, LocalDate}
-import java.util.concurrent.atomic.AtomicReference
+import java.util.concurrent.atomic.{AtomicLong, AtomicReference}
 
 trait PolicyService {
 
@@ -143,15 +143,37 @@ class PolicyServiceLive(
     // active; when false (test default) every `snapshot` rebuilds and `invalidate`/`reevaluate` are
     // no-ops. See the trait docs.
     cacheEnabled: Boolean = false,
+    // #1954: test-only barrier run at the END of every `buildSnapshot` (after the DB read, before the
+    // result is returned to the caller that installs it in the cache). Defaults to a no-op for
+    // production and the ~40 existing specs; PolicySnapshotCacheSpec wires a gate here to suspend an
+    // in-flight build deterministically and prove a stale build cannot clobber a fresher cache entry.
+    buildBarrier: UIO[Unit] = ZIO.unit,
 ) extends PolicyService {
 
-  // #1849: the single cached `(etag, snapshot)`. Process-local `AtomicReference` (matching the
-  // existing `lastSnapshotEtag` style in this class) so the `apply` factory stays a pure
-  // constructor. For the single-household model there is exactly one global snapshot, so one slot
-  // suffices (design §6.2 ticker note). Populated on first build, refreshed by `reevaluate`, cleared
-  // by `invalidate`.
-  private val snapshotCache: AtomicReference[Option[(ETag, PolicySnapshot)]] =
-    new AtomicReference(Option.empty[(ETag, PolicySnapshot)])
+  // #1849: the single cached snapshot. Process-local `AtomicReference` (matching the existing
+  // `lastSnapshotEtag` style in this class) so the `apply` factory stays a pure constructor. For the
+  // single-household model there is exactly one global snapshot, so one slot suffices (design §6.2
+  // ticker note). Populated on first build, refreshed by `reevaluate`.
+  //
+  // #1954: each entry is STAMPED with the `mutationVersion` observed at the START of its build
+  // (`builtAtVersion`, below). A reader trusts a cached entry only when its stamp is still the
+  // current version; a stamp older than the current version means a mutation landed since the build
+  // began, so the entry may predate that mutation and is treated as a miss (rebuild). This is what
+  // closes the stale-read race that the old "clear-on-invalidate" alone could not: a `buildSnapshot`
+  // that began reading the DB BEFORE a mutation could complete AFTER `invalidate` cleared the slot
+  // and repopulate it with pre-mutation bytes, so the very next read got a cache HIT on stale data
+  // (Gate-1: a just-created profile missing from `/api/router/policy`). With the version stamp, that
+  // late write lands tagged with the old version and every reader rejects it — the next read rebuilds
+  // synchronously and serves fresh bytes. See #1954.
+  private val snapshotCache: AtomicReference[Option[(Long, ETag, PolicySnapshot)]] =
+    new AtomicReference(Option.empty[(Long, ETag, PolicySnapshot)])
+
+  // #1954: monotonic counter bumped once per policy mutation (every `invalidate`). A build stamps the
+  // value it reads here BEFORE it touches the DB; a reader compares a cached stamp against the
+  // current value. Because the bump happens-after the mutation's DB commit (routes call `invalidate`
+  // after the write), any cached entry whose stamp equals the current version is guaranteed to
+  // reflect every committed mutation.
+  private val mutationVersion: AtomicLong = new AtomicLong(0L)
 
   // #1849: the push sink, installed at startup via `setPublisher`. Defaults to noop so a snapshot
   // rebuilt before the registry exists (or in a test) is simply not pushed.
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review — #1956 (first run)

Reviewed the full merge-base diff (`git diff origin/main...HEAD`): `PolicyService.scala` (+ the version-stamped cache) and `PolicySnapshotCacheSpec.scala`. Adversarial pass over all 8 dimensions.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `api/src/policy/PolicyService.scala:148-156` — the field comment refers to the stamp as "(`builtAtVersion`, below)", but there is no field/val by that name; the stamp is the tuple's first `Long` slot (the `gen` param). Slightly misleading cross-reference — consider "(the `gen` stamp, captured in `buildVersioned`/`reevaluate`)".
2. `api/src/policy/PolicyService.scala:151` (`buildBarrier`) — a test-only suspension seam threaded through the production `buildSnapshot` chain (`.zipLeft(buildBarrier)`, line ~519). It is defaulted to `ZIO.unit` (zero prod cost) and is consistent with the existing test-defaulted `cacheEnabled`/`uiAllowedHosts` params on the same constructor, so this is acceptable as-is; a repo decorator would avoid the prod-code seam but at much more boilerplate. Non-blocking.

**Dimension notes**
- **1 Single source of truth** ✓ — `reevaluate` stays the sole rebuild-and-publish point; `installSnapshot` is the single install primitive; no domain quantity re-derived; no new snapshot computation. The two `mutationVersion.get` captures (in `buildVersioned` and `reevaluate`) are each the natural pre-DB-read capture point, not duplicated logic.
- **2 Testing** ✓ — TDD red→green visible: the red commit keeps `main`'s clear-on-invalidate behavior and fails **only** the new race test; green passes all 7. The regression test deterministically reproduces the in-flight clobber via a gated build and asserts the new profile survives. create/assign + delete read-after-write invariants added. cache_hit path still asserted. No existing assertion weakened or skipped. No forbidden mocks; embedded Postgres + injected `TestClock`.
- **3 Architectural invariants** ✓ — no wire/snapshot-field change, no router logic, no `/tmp` writes, `Clock` injected, ZIO effects (the local `var`/`while` CAS in `installSnapshot` is pure and matches the class's existing `AtomicReference` style, no `throw`).
- **Correctness of the fix** ✓ — version bump happens-after the mutation's DB commit (routes sequence `write <- … ; invalidate`), so a stamp `== mutationVersion` provably reflects every committed mutation; readers reject any older stamp and rebuild synchronously; `installSnapshot` refuses to clobber a newer-version entry, and even when a stale build installs over `None`, the read-time `builtAt == version` check rejects it. The reconcile-ticker (same-version content refresh, e.g. schedule boundary) still works (`builtAt == version` → hit on refreshed content; pinned by the existing schedule-boundary test).
- **4 Wire compat** ✓ — internal cache only; no wire-visible shape touched.
- **5 Migrations** ✓ — none. **6 Metrics** ✓ — reuses `policy_snapshot_build_total{computed|cache_hit}`; no new path warranting a new series. **7 Scope** ✓ — two files, focused; no AI mentions in title/body/commits. **8 Security** ✓ — no SQL, no secrets, no new route.

**VERDICT: APPROVE**
Correctly diagnoses the Gate-1 failure as #1849's async-rebuild clobber (not a missing-invalidation gap) and fixes it with a version-stamped cache that restores read-after-write freshness while preserving the cache_hit/push-on-change perf win. Strong deterministic regression test with visible red→green. Only doc/seam nits remain.

sameerparekh:
## Independent PR review — #1956 (re-run @ 55924562)

**Prior findings status**

| # | Finding | Status |
|---|---------|--------|
| NIT 1 | misleading `builtAtVersion` doc reference (`PolicyService.scala:158`) | **ADDRESSED** — reworded to "the leading `Long` of the tuple — the `gen` captured in `buildVersioned`/`reevaluate`". |
| NIT 2 | test-only `buildBarrier` seam in prod `buildSnapshot` | **NOT-ADDRESSED** (accepted as-is — non-blocking; consistent with the existing `cacheEnabled` test-default precedent). |

**Incremental delta** (`git diff d3390890...HEAD`): a single comment-only edit in `PolicyService.scala` (the NIT-1 reword). No code/behavior/test change.

**New findings:** none.

**VERDICT: APPROVE**
Delta is documentation only and addresses the prior NIT. No BLOCKER ever open. Fix and regression test unchanged from the approved state.

Issue #1959, opened 2026-06-25
bug(agent): #1848 ws client does not reassemble fragmented frames — large #1849 policy push truncates


sameerparekh:
## Summary
The real OpenWRT agent's hand-rolled RFC 6455 ws client (#1848) does **NOT reassemble fragmented WebSocket messages**. When the API server's large #1849 `policy` snapshot push is split into a leading TEXT fragment (FIN=0) + CONTINUATION (0x0) frames, the agent returns only the **first fragment** — a truncated, unparseable JSON payload. Once apply-on-push (#1945) lands, a large snapshot would silently fail to apply (or apply a truncated/garbage snapshot). Surfaced while fixing the e2e harness twin of this bug in [#1958](https://github.com/wifihaven/wifihaven/issues/1958).

## Root cause
`openwrt/files/usr/lib/lua/wifihaven/ws_client.lua` `M:recv` (around line 157) calls `ws_frame.decode` once per frame and returns immediately:

```lua
else
  return frame.opcode, frame.payload       -- text/binary/continuation
end
```

It **never inspects `frame.fin`** and never accumulates continuation frames. The framing layer `ws_frame.lua` `decode` *does* correctly parse the FIN bit (`local fin = (b1 >= 0x80)`, returned as `frame.fin`) and knows `OP_CONTINUATION = 0x0` — but nothing consumes `fin`. So:

1. Frame 1: FIN=0, opcode=0x1 (TEXT), first chunk → `M:recv` returns it raw.
2. Frames 2..N (FIN=0/1, opcode=0x0 CONTINUATION) are left on the wire / returned as separate "messages".

The caller (`ws_loop.lua` `handle_inbound`) json-decodes the truncated first chunk, fails, and drops the push as `"unknown"`.

## Fix (separate PR, scoped to the agent)
In `M:recv`: when a TEXT/BINARY frame has `frame.fin == false`, accumulate `frame.opcode` + payload, then keep reading frames, appending `OP_CONTINUATION` (0x0) payloads, until a frame with `fin == true`; return the initial data opcode + concatenated payload. Handle interleaved control frames (ping/pong/close) per RFC 6455 §5.4 — they are never fragmented and must not be treated as continuation data, nor should their FIN bit end the data message. Must be validated on a real Lua-5.1 + packaged-cqueues/luaossl target per the #1845 spike recipe (macOS luarocks masks 5.1/cqueues bugs).

## Who fragments / at what size
[...]

Issue #1945, opened 2026-06-24
agent: apply a ws-pushed policy snapshot live (push→apply→enforce), not just save it


sameerparekh:
## Gap

The websocket transport's push path is wired server→agent for *delivery* but not for *enforcement*. The `wifihaven-ws` sidecar's `on_policy` (`openwrt/files/usr/sbin/wifihaven-ws`) only calls `policy.save_snapshot` — it writes the pushed snapshot to `/etc/wifihaven/policy.json` but does NOT apply it. The running `wifihaven-agent` reads that file **only at startup** (`policy.load_snapshot`); its tick loop applies policy exclusively from the **HTTP poll** (`policy.fetch` → `policy.apply` → `render.update_shared`). So today, with `wifihaven.ws.enabled=1`, a pushed `policy` frame updates the on-disk cache but enforcement (nft/dnsmasq) only changes on the next HTTP poll or an agent restart.

This contradicts the design intent — [`docs/design/websocket-transport.md`](docs/design/websocket-transport.md) §1.4 says "the agent applies it (etag-guarded)" and "push replaces the poll" — and means the push→apply→enforce path is a no-op for live enforcement.

## Why it's filed separately

Surfaced while implementing the Gate-2 ws push transport e2e (#1939). That issue was scoped (operator decision) to **test-only**: it proves the pushed frame reaches the real agent's sidecar and is *saved* (`policy.json` etag + `ws_frames_recv_total{op=policy}`), but deliberately does NOT assert a live enforcement flip from the push, because the wiring below doesn't exist yet. The Gate-2 scenario (`scripts/e2e/scenarios_fake/test_ws_push_apply.py`) is ready to be extended with an enforcement-flip assertion once this lands.

## Proposed work

Make the agent apply a ws-pushed snapshot live, etag-guarded, ws-gated, single enforcement owner (no two-writer race with the sidecar). Likely shape: an agent tick that, when ws is enabled, reads the on-disk snapshot and applies it via the existing `policy.apply` + `refresh_blocklists` + `render.update_shared` path when its `etag` differs from the currently-applied one — reusing the poll path's apply machinery. Mind the Lua 5.1 60-upvalue limit on `on_tick` (#1930) — factor the shared apply block into a helper rather than inlining.

[...]

PR #1960, opened 2026-06-25
fix(#1958): ws_send.py reassembles fragmented ws frames (large policy push)

### LINKED ISSUE #1958: bug(API/UI CD): ws_send.py truncates fragmented ws frames at 4096B — large #1849 policy push fails to parse (Gate 1 red); verify real agent reassembles too
## Summary
With #1956 (fix #1954, snapshot freshness) merged, Master API/UI CD Gate 1 got PAST the stale-snapshot failure (read-after-write now correct ✓) but hits a NEW ws-harness failure:

```
✓ ws upgrade auth parity: missing/bad → 401, valid → 101
✓ REST usage/events/metrics accepted (200)
✗ ws frames not all acked ok: ws_send error:
   {"error": "Expecting property name enclosed in double quotes: line 1 column 4097 (char 4096)"}
```

## Root cause — ws_send.py doesn't reassemble FRAGMENTED ws frames
`scripts/e2e/lib/ws_send.py` `_recv_frame()` reads exactly ONE WebSocket frame and **ignores the FIN bit (`b0 & 0x80`) and the continuation opcode (`0x0`)**. The truncation lands at **char 4096** — the `sock.recv(4096)` boundary — because the server splits the large #1849 `policy` push (the staging snapshot is >4KB: many profiles) into a TEXT fragment (FIN=0, ~4096B) + CONTINUATION frames. `_recv_frame` returns only the first fragment, and `json.loads` on the truncated payload fails at char 4096. (This is the same ws_send.py lineage as #1951/#1953; #1953 made it drain the policy push, which exposed the fragmentation gap now that the push is large.)

## Fix (harness)
In `_recv_frame` (or the caller), **reassemble fragmented messages**: keep reading frames while FIN=0, concatenating payloads of the initial data frame + subsequent CONTINUATION (`0x0`) frames, and only return the full message when FIN=1. Handle interleaved control frames (ping/pong/close) per RFC 6455 §5.4. Add a unit test in `scripts/e2e/lib/test_ws_send.py` for a multi-fragment text message > 4096B.

## ALSO verify (product-side — important)
If the **server** (zio-http/netty — or a Render/Cloudflare proxy) fragments large outbound ws frames, then the **real OpenWRT agent ws sidecar (#1848, hand-rolled RFC 6455)** must ALSO reassemble continuation frames — otherwise it CANNOT receive a large `policy` snapshot push in prod (silent truncation → bad apply). Determine:
1. WHO fragments (zio-http server write, or a proxy) and at what size (the 4096 boundary suggests a configured max-frame-size).
2. Does the #1848 Lua agent ws client reassemble FIN=0 + continuation frames? If NOT → that's a real product bug (file/fix separately under #1023; large snapshots would break push-apply once #1945 lands). If the agent already handles it (cqueues/luaossl per #1845), just fix ws_send.py.

## Acceptance
- `ws_send.py` reassembles fragmented messages; the >4KB policy push parses; Gate 1 ws-parity passes; Master API/UI CD goes green.
- A clear answer on whether the real agent reassembles fragments (and a follow-up filed if it doesn't).

Relates to #1951/#1953 (ws_send.py parity), #1849 (policy push), #1848 (#1845 agent ws client), #1023. Found while closing out API/UI CD.

sameerparekh:
## Problem
With #1956 merged, Master API/UI CD **Gate 1** got past the stale-snapshot failure but hit a new ws-harness failure:

```
✗ ws frames not all acked ok: ws_send error:
   {"error": "Expecting property name enclosed in double quotes: line 1 column 4097 (char 4096)"}
```

## Root cause
`scripts/e2e/lib/ws_send.py` `_recv_frame()` read exactly **one** WebSocket frame and **ignored the FIN bit (`b0 & 0x80`) and the continuation opcode (`0x0`)**. The large #1849 `policy` push (staging snapshot >4 KiB: many profiles) is split into a leading TEXT fragment (FIN=0, ~4096 B) + CONTINUATION frames. `_recv_frame` returned only the first fragment, so `json.loads` truncated mid-string at char 4096. (Same `ws_send.py` lineage as #1951/#1953 — #1953 made it drain the push, which is what exposed the fragmentation gap once the push got large.)

## Fix (harness)
- `_recv_frame` now returns the FIN bit: `(fin, opcode, payload)`.
- New `_recv_message` reassembles per **RFC 6455 §5.4**: concatenate the leading TEXT/BINARY frame + CONTINUATION (`0x0`) frames until a frame with FIN=1, returning the initial data opcode + joined payload. Interleaved **control frames** (ping/pong/close) between fragments are handled inline (ping/pong skipped, close → error) and their FIN bit is **not** mistaken for the data message's FIN.
- `send_frames` reads complete messages via `_recv_message`.
- **TDD**: `test_fragmented_policy_push_is_reassembled` (committed failing first) sends a ~12 KB policy push fragmented at the 4096 B boundary — both a clean split and one with an interleaved control PING — and asserts the reassembled message equals the original and that all three frame acks still land. The red commit reproduced the exact char-4096 truncation. All 4 ws tests (27 in `scripts/e2e/lib/`) green.

## Product-side investigation (per the issue)

**(a) Who fragments, and at what size?** The API server sends the snapshot as a **single, unfragmented** `TextWebSocketFrame` — `RouterWsRegistry.policyFrameText` → `WebSocketFrame.text(...)` (`api/src/routes/RouterWsRegistry.scala:66,122`). There is **no `maxFrameSize` / `WebSocketConfig` override** anywhere in `api/src` (Netty's ~64 KiB default applies; the only `Server.Config` tweak is the HTTP `disableRequestStreaming(4 MiB)` body cap, unrelated to ws). The failure is observed against **staging** (`wss://api-staging.wifihaven.net`, behind Render's edge), and the boundary is **~4096 B**, not 64 KiB — so the **edge/proxy re-fragments** large outbound ws frames at ~4 KiB. Net: fragmentation happens on the wire in prod regardless of the server writing a single frame, so any ws client MUST reassemble.

**(b) Does the real #1848 Lua agent reassemble?** **No — same bug.** `openwrt/files/usr/lib/lua/wifihaven/ws_client.lua` `M:recv` (~L157) returns `frame.opcode, frame.payload` directly for `text/binary/continuation` and **never inspects `frame.fin`** — even though the framing layer `ws_frame.lua` `decode` correctly parses FIN and knows `OP_CONTINUATION = 0x0`. A large `policy` push would be truncated to its first fragment and dropped as unparseable — a real product bug once apply-on-push (#1945) lands. Filed separately (kept out of this PR to keep it scoped to the harness): **#1959** under #1023. It needs the same RFC 6455 §5.4 reassembly, validated on a real Lua-5.1/cqueues target per the #1845 spike.

## Acceptance
- [x] `ws_send.py` reassembles fragmented messages; the >4 KiB policy push parses.
- [x] Unit test covers multi-fragment >4096 B message + interleaved control frame.
- [x] Clear answer on the real agent (does **not** reassemble) + follow-up filed (#1959).
- [ ] Gate 1 ws-parity passes / Master API/UI CD green (verified post-merge).

Fixes #1958
Relates to #1849, #1953, #1848, #1023, #1959

### DIFF
--- a/scripts/e2e/lib/test_ws_send.py
+++ b/scripts/e2e/lib/test_ws_send.py
@@ -32,8 +32,10 @@ sys.path.insert(0, str(Path(__file__).parent))
 from ws_send import send_frames  # noqa: E402
 
 _WS_GUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
+_OPCODE_CONTINUATION = 0x0
 _OPCODE_TEXT = 0x1
 _OPCODE_CLOSE = 0x8
+_OPCODE_PING = 0x9
 
 
 def _accept_key(client_key: str) -> str:
@@ -55,6 +57,46 @@ def _server_send_text(conn: socket.socket, text: str) -> None:
         header.append(127)
         header += struct.pack(">Q", n)
     conn.sendall(bytes(header) + payload)
+
+
+def _server_send_frame(conn: socket.socket, fin: bool, opcode: int, payload: bytes) -> None:
+    # Server→client frames are NOT masked (RFC 6455 §5.1). `fin`/`opcode` are
+    # explicit so a caller can emit a fragmented message: a leading data frame
+    # (FIN=0, opcode TEXT/BINARY) + continuation frames (opcode 0x0), the last
+    # with FIN=1.
+    n = len(payload)
+    b0 = (0x80 if fin else 0x00) | opcode
+    header = bytearray([b0])
+    if n < 126:
+        header.append(n)
+    elif n < 65536:
+        header.append(126)
+        header += struct.pack(">H", n)
+    else:
+        header.append(127)
+        header += struct.pack(">Q", n)
+    conn.sendall(bytes(header) + payload)
+
+
+def _server_send_fragmented_text(
+    conn: socket.socket, text: str, fragment_size: int, ping_after: int = -1
+) -> None:
+    """Send `text` as a fragmented TEXT message (RFC 6455 §5.4).
+
+    The payload is split into `fragment_size`-byte chunks: the first goes out as
+    a TEXT frame with FIN=0, the rest as CONTINUATION (0x0) frames, and only the
+    final frame carries FIN=1. If `ping_after >= 0`, an unfragmented control PING
+    is interleaved right after that fragment index — a control frame may appear
+    between fragments and must NOT be treated as continuation data.
+    """
+    payload = text.encode("utf-8")
+    chunks = [payload[i:i + fragment_size] for i in range(0, len(payload), fragment_size)]
+    last = len(chunks) - 1
+    for i, chunk in enumerate(chunks):
+        opcode = _OPCODE_TEXT if i == 0 else _OPCODE_CONTINUATION
+        _server_send_frame(conn, fin=(i == last), opcode=opcode, payload=chunk)
+        if i == ping_after:
+            _server_send_frame(conn, fin=True, opcode=_OPCODE_PING, payload=b"hb")
 
 
 def _server_recv_frame(conn: socket.socket, buf: bytearray) -> tuple[int, str]:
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review

Read-only adversarial pass over `git diff origin/main...HEAD` (2 files, harness-only) against the 8-dimension checklist.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `scripts/e2e/lib/ws_send.py:35` — `_OPCODE_BINARY` is introduced but never referenced in code (only named in the `_recv_message` docstring). `_recv_message` will happily reassemble/return a BINARY data message via the generic non-control path, so the constant is documentation rather than logic. Harmless; drop it or keep as a readability anchor.
2. `scripts/e2e/lib/ws_send.py:200` — an interleaved server PING encountered mid-reassembly is skipped without a PONG reply. This matches the pre-existing standalone-ping handling in `send_frames` (also `continue`, no pong) and is fine for a short-lived (≤15s) e2e client, so no change needed — noting only that the "respond to ping" semantics are intentionally absent on both paths.

**Dimension notes (why no blockers):**
- **1 Single source of truth:** reassembly lives in exactly one place (`_recv_message`); no prod-logic re-derivation. The `_OPCODE_*` duplication between `ws_send.py` and `test_ws_send.py` is the pre-existing stdlib-only-test pattern (the test deliberately hand-rolls its own framing), not a SSoT violation.
- **2 Testing integrity:** TDD red→green is visible in history — the failing test (`test(#1958)`) was committed first and reproduced the exact char-4096 truncation, then the fix turned it green. New behavior is positively asserted (`pushes[0] == expected` full-message equality + all three acks still land), and the boundary case (control PING interleaved between fragments, `ping_after=1`) is covered. No existing assertion weakened/skipped.
- **3 Architectural invariants:** harness-only; no router/agent/DNS/snapshot changes. Correctly does NOT touch the real agent — the same gap in the #1848 Lua client is filed separately as #1959 (kept out of scope, right call).
- **4 Wire contract:** no wire-shape change; this only makes the *client* tolerant of RFC 6455 §5.4 fragmentation (strictly more permissive). Reassembly correctly preserves the data message's FIN across an interleaved control frame (the `continue` deliberately does not advance `fin`) and rejects a non-continuation opcode mid-message.
- **5–8 Migrations / Metrics / Security:** N/A — no SQL, no prod code path, no routes, no secrets.

Verified locally: all 4 `test_ws_send.py` tests pass (27 green across `scripts/e2e/lib/`).

VERDICT: APPROVE
Harness-only fix making `ws_send.py` reassemble fragmented ws frames per RFC 6455 §5.4, unblocking Gate 1 ws-parity on the large #1849 policy push. Clean, well-scoped, TDD-backed, with the real-agent twin bug correctly split out to #1959. No blockers; two cosmetic nits.

sameerparekh:
## Independent PR review — re-run (delta `a3835448...e817b2fd`)

**Prior-findings status**

|---|---|---|
| NIT 1 | `ws_send.py` unused `_OPCODE_BINARY` constant | **ADDRESSED** — constant removed; `_recv_message` reassembles any non-control data opcode generically, no behavior change. |
| NIT 2 | interleaved mid-reassembly PING not PONG'd (intentional, matches existing path) | **NOT-ADDRESSED (by design)** — left as-is per the prior note; consistent with `send_frames`' standalone-ping handling, fine for the short-lived e2e client. |

**New findings (delta):** the only change is the one-line removal of the unused constant. No new logic, no new findings (BLOCKER / SHOULD-FIX / NIT: none). Tests re-run green (4/4 in `test_ws_send.py`).

VERDICT: APPROVE
Delta is a pure cleanup (drop dead constant) on top of the already-approved reassembly fix. No blockers, no open SHOULD-FIX; ready for merge-when-ready at the operator's discretion.

PR #1979, opened 2026-06-25
fix(#1959): agent ws client reassembles fragmented frames (large policy push)

### LINKED ISSUE #1959: bug(agent): #1848 ws client does not reassemble fragmented frames — large #1849 policy push truncates

sameerparekh:
## What

The OpenWRT agent's hand-rolled RFC 6455 ws client (`ws_client:recv`, #1848) did **not reassemble fragmented WebSocket messages** — it returned the *first* frame raw, never inspecting `frame.fin` or accumulating `CONTINUATION` frames. A large #1849 `policy` push that the edge re-fragments at ~4 KiB (observed against staging behind Render) therefore truncated to unparseable JSON, and `ws_loop.handle_inbound` dropped it as `"unknown"`. Once apply-on-push (#1945) lands, a large snapshot would silently fail to apply.

Fixes #1959.

## How

- **`ws_frame.reassembler()`** — a new **pure**, stateful accumulator (RFC 6455 §5.4): buffers a `TEXT`/`BINARY`(FIN=0) start frame + its `CONTINUATION` (0x0) frames into one message, returning the **start** opcode + concatenated payload; passes interleaved control frames (ping/pong/close) straight through (§5.4 permits them mid-message); and rejects the protocol violations (`unexpected_continuation`, `interleaved_data`, `fragmented_control`).
- **`ws_client:recv`** now drives the reassembler and answers control frames exactly as before. The reassembler persists across `recv()` calls, so a message split over multiple socket reads still completes.

Keeping the logic pure means it unit-tests on the dev host **and** runs identically under OpenWrt Lua 5.1 — the same split the rest of `ws_frame` uses. No new decision logic on the router; this is framing only.

## Tests (TDD — red commit `c3566e53`, green `4f48263b`)

- **Unit (`ws_frame_spec.lua`):** 11 new cases — continuation reassembly, three-fragment ordering, interleaved control, standalone control, >4 KiB many-fragment payload, second-message reset, and the three §5.4 protocol errors. Runs on macOS busted + PUC Lua 5.1.
- **On-target e2e (`spike/ws-1845`):** the echo server gained a `WS_ECHO_FRAGMENT` mode (splits its echo into §5.4 fragments with an interleaved ping); `poc_fragment.lua` asserts a 12 KiB payload round-trips byte-for-byte; `e2e_test.sh` runs it over `ws://` **and** `wss://`.

### Real-hardware validation (Lua 5.1 + packaged cqueues/luaossl, plex.lan)
```
== ws e2e (LUA=.../lua5.1) ==
PASS  ws       (ws://127.0.0.1:22288)
PASS  wss      (wss://localhost:22289)
PASS  frag-ws  (ws://127.0.0.1:22290)
PASS  frag-wss (wss://localhost:22291)
== ws e2e: ALL PASS ==
```
Pre-fix the client returns only the first 512 B fragment and the length check fails. Full openwrt suite: 947 busted + 44 shell, 0 failures.

## Wire contract

No wire change — purely the client's frame-decode path. Additive (`ws_frame.reassembler()`); both sides already ignore unknown fields.

Relates to #1023, #1848, #1849, #1945, #1958.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/ws_client.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/ws_client.lua
@@ -155,20 +155,35 @@ end
 -- the whole loop. Validated on a real Lua-5.1 + packaged-cqueues (epoll) target:
 -- without the clearerr the connection dies after a few idle heartbeat cycles.
 function M:recv(timeout)
+  -- Reassemble fragmented messages (RFC 6455 §5.4): a large server push (e.g. a
+  -- #1849 `policy` snapshot, which Render's edge re-fragments at ~4 KiB) arrives
+  -- as TEXT(FIN=0) + CONTINUATION frames. Returning the first frame raw (the
+  -- pre-#1959 behaviour) truncates the snapshot to unparseable JSON. The pure
+  -- reassembler (ws_frame) accumulates fragments and passes interleaved control
+  -- frames back for us to answer; it persists across recv() calls so a message
+  -- split over multiple socket reads still completes.
+  self.reasm = self.reasm or ws_frame.reassembler()
   while true do
     local frame, consumed = ws_frame.decode(self.rxbuf)
     if frame then
       self.rxbuf = self.rxbuf:sub(consumed + 1)
-      if frame.opcode == ws_frame.OP_PING then
-        self:send_pong(frame.payload)            -- heartbeat: keepalive
-      elseif frame.opcode == ws_frame.OP_PONG then
-        -- liveness ack; caller may track it. Loop for the next real frame.
-      elseif frame.opcode == ws_frame.OP_CLOSE then
+      local status, a, b = self.reasm:push(frame)
+      if status == "control" then
+        local opcode, payload = a, b
+        if opcode == ws_frame.OP_PING then
+          self:send_pong(payload)                -- heartbeat: keepalive
+        elseif opcode == ws_frame.OP_CLOSE then
+          self.closed = true
+          return nil, "closed"
+        end
+        -- OP_PONG: liveness ack; loop for the next frame.
+      elseif status == "message" then
+        return a, b                              -- (opcode, reassembled payload)
+      elseif status == "error" then
         self.closed = true
-        return nil, "closed"
-      else
-        return frame.opcode, frame.payload       -- text/binary/continuation
+        return nil, "protocol: " .. tostring(a)
       end
+      -- "more": a fragment was buffered; keep reading for the rest.
     elseif frame == false then
       self.closed = true
       return nil, "protocol: " .. tostring(consumed)
--- a/openwrt/files/usr/lib/lua/wifihaven/ws_frame.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/ws_frame.lua
@@ -174,4 +174,75 @@ function M.new_client_key(rand16_fn, base64_fn)
   return base64_fn(rand16_fn(16))
 end
 
+-- ── message reassembly (RFC 6455 §5.4) ───────────────────────────────────────
+--
+-- decode() returns ONE frame at a time; a single application message may be
+-- split across a leading data frame (TEXT/BINARY, FIN=0) plus CONTINUATION
+-- (0x0) frames, with control frames (ping/pong/close) optionally interleaved.
+-- The server (and intermediaries — Render's edge re-fragments large outbound
+-- frames at ~4 KiB, #1959) can split a large `policy` push this way, so the
+-- agent MUST reassemble or it truncates the snapshot to garbage.
+--
+-- This is the part of the cqueues client's recv loop that has no I/O, so it
+-- lives here (PURE, no cqueues) where it is unit-tested on the dev host AND
+-- runs identically under OpenWrt Lua 5.1 — the same split the rest of this
+-- module uses. The cqueues client (ws_client:recv) feeds each decoded frame to
+-- push() and acts on the verdict.
+--
+-- reassembler() → object with push(frame) → (status, a, b):
+--   "message", opcode, payload  — a complete application message (the START
+--                                 data opcode + concatenated payload). Returned
+--                                 for an unfragmented data frame, or the final
+--                                 CONTINUATION of a fragmented one.
+--   "control", opcode, payload  — a control frame (ping/pong/close); the caller
+--                                 handles it (pong/close) and keeps reading.
+--                                 Reassembly state is untouched (§5.4 allows
+--                                 control frames between message fragments).
+--   "more"                      — fragment accepted; need more frames.
+--   "error",  reason            — a protocol violation; the caller must drop the
+--                                 connection. reason ∈ { unexpected_continuation,
+--                                 interleaved_data, fragmented_control }.
+local Reassembler = {}
+Reassembler.__index = Reassembler
+
+function M.reassembler()
+  return setmetatable({ op = nil, parts = nil }, Reassembler)
+end
+
+function Reassembler:push(frame)
+  local opcode = frame.opcode
+
+  -- Control frames (0x8..0xF) are never part of a data message and MUST NOT be
+  -- fragmented (§5.5). They pass straight through; an in-progress fragmented
+  -- data message is unaffected.
+  if opcode >= 0x8 then
+    if not frame.fin then return "error", "fragmented_control" end
+    return "control", opcode, frame.payload
+  end
+
+  if opcode == M.OP_CONTINUATION then
+    if not self.op then return "error", "unexpected_continuation" end
+    self.parts[#self.parts + 1] = frame.payload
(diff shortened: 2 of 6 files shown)


sameerparekh:
## Independent PR review — `4f48263`

Read-only adversarial pass against the 8-dimension checklist on the full merge-base diff (`origin/main...HEAD`).

**BLOCKERS**
(none)

**SHOULD-FIX**
1. `openwrt/files/usr/lib/lua/wifihaven/ws_frame.lua:215-220` — the reassembler accumulates `self.parts` with no cap on total assembled size; it only resets on `FIN=1`. On a RAM-constrained OpenWrt router a buggy/compromised server (or an edge that never sets FIN) could stream unbounded continuation frames and OOM the agent. The channel is authenticated wss and the real snapshot is bounded, so this is not a BLOCKER — but given the router-agent resource-safety culture (bounded-tmp-writes etc.), consider a max-assembled-bytes guard that returns `"error", "message_too_large"` (→ `recv` drops the connection, same as the existing per-frame `frame_too_large` guard in `decode`). `decode`'s 64-bit-length guard bounds a *single* frame, not the sum across continuations.

**NITS**
1. `ws_frame.lua:189` — control frames are passed through without enforcing the RFC 6455 §5.5 ≤125-byte payload limit. A well-behaved server never violates it and the pre-#1959 code didn't check either, so this is purely defensive hardening.
2. `ws_client.lua:170-178` — a mid-stream reassembly `"error"` (protocol violation) now closes the connection with `"protocol: …"`, which surfaces only as a reconnect (no dedicated counter). The success path is already metered (`ws_frames_recv_total{op=policy}` via `handle_inbound`), and protocol errors are should-never-happen against a correct server, so a metric is optional — noting for observability completeness.
3. Behavioral note (not a defect): for reserved control opcodes `0xB–0xF` the new code ignores them and keeps reading, whereas the old code returned them raw as if a data message. The new behavior is strictly more correct.

**Dimension notes**
- **1 Single-source-of-truth:** ✅ improves it — reassembly was previously absent (the bug); now centralized in one pure `ws_frame.reassembler()`, `recv` is the sole caller. No re-derivation, no "keep in sync."
- **2 Testing:** ✅ TDD red (`c3566e53`) → green (`4f48263b`) visible. 11 new unit cases assert continuation reassembly, interleaved control, multi-fragment ordering, >4 KiB payload, second-message reset, and all three §5.4 protocol errors — including negative cases. No existing assertion weakened. On-target e2e added (`frag-ws`/`frag-wss`) and the maintainer reports all 4 cases PASS on real Lua 5.1 + packaged cqueues/luaossl (plex.lan); pre-fix the length check fails. Lua-5.1 syntax scan clean (no `\xHH`/`goto`/`::`/numeric underscores).
- **3 Architectural invariants:** ✅ pure framing only; no decision logic added to the router, no DNS reasoning, no snapshot wire field. Reassembly buffer is transient in-RAM (freed after `table.concat`), not a `/tmp` append-writer — bounded-tmp-writes rule N/A (see SHOULD-FIX for the RAM-accumulation angle).
- **4 Wire contract:** ✅ no shape change; the client now correctly handles a fragmentation the server/edge already emits. Additive.
- **5 Migrations / 6 Metrics / 8 Security:** N/A / see NIT 2 / no secrets·SQL·routes.
- **7 Scope:** ✅ focused (framing fix + its tests + on-target harness); no AI mentions in title/body (Co-Authored-By trailer is fine).

**VERDICT: APPROVE**
A correctly-scoped, well-tested RFC 6455 §5.4 reassembly fix that lifts the logic into a pure, dual-target-testable `ws_frame.reassembler()` and validates it on real Lua 5.1. No blockers; one optional resource-safety hardening (assembled-size cap) and minor observability/defensive nits. Required prerequisite for #1945 applying large pushes intact.

sameerparekh:
## Independent PR review (re-run) — `4265bc7`

Statused the prior review (`4f48263`) against current code, then reviewed the incremental delta (`4f48263...4265bc7`).

### Prior findings
|---|---|---|
| SHOULD-FIX 1 | Unbounded `self.parts` growth (no assembled-size cap) | **ADDRESSED** — `reassembler(max_bytes)` defaults to 1 MiB; continuation path sums `self.size` and returns `error,"message_too_large"` once over (`ws_frame.lua:223-225`), the start frame is checked too (`:238`), and `size` resets to 0 on completion (`:228`). New unit cases cover over-cap, first-frame-over-cap, and counter-reset. |
| NIT 1 | Control frame >125 B not rejected (§5.5) | **ADDRESSED** — `error,"oversized_control"` at `ws_frame.lua:215`, with a test. |
| NIT 2 | No metric on reassembly protocol-error | **NOT-ADDRESSED** — intentional; the success path is already metered (`ws_frames_recv_total{op}`) and a protocol error is a should-never-happen drop that surfaces as a reconnect. Optional, non-gating. |
| NIT 3 | (reserved-control behavior — not a defect) | N/A |

### New findings (delta)
**BLOCKERS** — (none)

**SHOULD-FIX** — (none)

**NITS**
1. `ws_frame.lua:236-241` — the cap is enforced on a fragmented *start* frame but an *unfragmented* single frame over `max_bytes` is returned without a size check. This is consistent and safe (a single frame is already bounded by one `decode()` read + the 64-bit `frame_too_large` guard; the cap exists to bound *accumulation* across continuations), so no change needed — noting only for completeness.

### Verdict
The delta is a clean, well-tested hardening pass that resolves the prior SHOULD-FIX and NIT 1 with no new BLOCKER. Re-validated on the real Lua-5.1 + cqueues target (plex.lan: all 4 e2e cases PASS); full openwrt busted suite green (now 32 ws_frame cases).

**VERDICT: APPROVE**
RFC 6455 §5.4 reassembly fix, now with an assembled-size cap and §5.5 control-length validation. Pure, dual-target-tested, hardware-validated. No blockers. Ready for operator merge-when-ready.

PR #1985, opened 2026-06-26
fix(#1972): attach Loki appender to root + surface send-error/auth failures

### LINKED ISSUE #1972: bug(#1831): 0 Loki ingest persists after GRAFANA_CLOUD_LOKI_* secrets set on Render
## Symptom
Operator set the `GRAFANA_CLOUD_LOKI_URL` / `GRAFANA_CLOUD_LOKI_USER` / `GRAFANA_CLOUD_LOKI_PASSWORD` secrets on the wifihaven-api-{staging,prod} Render web services, but Grafana Cloud Loki still shows **0 ingest**. So the earlier hypothesis (secrets unset = the gate) is now ruled out — the failure is elsewhere.

## Likely suspects to walk, observability-first (don't jump to code)
1. **Is the appender even trying?** Check `loki_logs_dropped_total` (the `MeteredLoki4jAppender` / `LokiDropMetrics` poller from #1885). If it's **climbing**, the appender IS attempting and failing → auth/URL/tenant/network. If it's **flat at 0** with 0 ingest, the appender likely isn't attached or isn't emitting at all.
2. **URL/auth shape.** Grafana Cloud Loki push wants `https://<loki-host>/loki/api/v1/push` with HTTP basic auth (user = numeric stack/instance id, password = a Grafana Cloud **access policy token** with `logs:write`). Verify the loki4j appender config maps the three secrets into the right URL + basic-auth header (not e.g. embedding creds in the URL, or a missing `/loki/api/v1/push` suffix, or a tenant/`X-Scope-OrgID` mismatch).
3. **Is the secret actually present in the running container?** `sync:false` Render secrets must be set on **each** service AND the service redeployed/restarted to pick them up — confirm via Render env + a fresh deploy, not just the dashboard showing them saved.
4. **Is the appender wired in logback at all** for the deployed profile, and gated on URL-presence the way #1873 intended? Confirm the gate fires when the URL is set.
5. **Token scope / 401 vs silent drop.** loki4j is fail-open — a 401/403 from a wrong token scope is swallowed; surface it (drop metric + a one-line WARN) so this isn't invisible next time.

## Definition of done
- Loki shows live ingest from staging (and prod) API logs.
- The failure mode that caused 0-ingest is made **observable** (metric and/or log) so a future recurrence is diagnosable, not silent.
- Root cause documented in the PR.

Relates to #1831, #1873, #1885.

sameerparekh:
## Summary

Fixes #1972 — Grafana Cloud Loki showed **0 ingest** from the API even after the
operator set the `GRAFANA_CLOUD_LOKI_*` secrets on the Render web services.
Observability-first investigation found **two independent root causes**, both
of which were silent (loki4j is fail-open by construction):

### Root cause A — the LOKI appender never attached (logback config bug)
`logback.xml` declared the root's `<appender-ref ref="LOKI">` inside an `<if>`
**nested within `<root>`**. logback **1.5.25** — the version loki4j 2.0.3
transitively forces onto the classpath, and what the deployed jar runs —
**rejects** an `<if>` nested in `<root>`/`<logger>`/`<appender>`
(`IfNestedWithinSecondPhaseElementSC`). The deploy logs show it exactly:

```
WARN  IfNestedWithinSecondPhaseElementSC - <if> elements cannot be nested within an <appender>, <logger> or <root> element
WARN  AppenderModelHandler - Appender named [LOKI] not referenced. Skipping further processing.
ERROR AppenderRefModelHandler - Failed to find appender named [LOKI]
```

So the appender was never attached and the API shipped **zero** logs to Loki.

**Fix:** declare a full `<root>` inside each branch of the **top-level** `<if>`
(LOKI+STDOUT when the secret is present, STDOUT-only otherwise). Every `<if>`
now sits at the top level, which logback allows on all versions, and LOKI
attaches deterministically. Verified end-to-end against logback 1.5.25 via the
real `JoranConfigurator`.

Also pinned `logbackVersion` to `1.5.25` (the version that already resolved/
shipped) so the declaration matches reality, removing the silent `1.5.8 -> 1.5.25`
bump and the "Found logback version ? / versions are different" startup noise.
The mismatch is also why the test classpath behaviour matched prod.

### Root cause B — the access-policy token lacks `logs:write` scope (operator action)
A live push from the **real staging creds** is rejected at the send boundary:

```
ERROR ... Loki responded with non-success status 401 ... {"error":"authentication error: invalid scope requested"}
```

The `GRAFANA_CLOUD_LOKI_PASSWORD` Grafana Cloud access-policy token does not have
`logs:write` for the logs instance. loki4j's `droppedEventsCount` only counts
**queue-full** drops on the way *in*, so a send-boundary 401 leaves
`loki_logs_dropped_total` flat at 0 — the invisible failure mode the issue's
suspect #5 named.

> **⚠️ Operator action required:** re-issue the access-policy token used for
> `GRAFANA_CLOUD_LOKI_PASSWORD` with **`logs:write`** scope on the logs instance,
> then redeploy `wifihaven-api-{staging,prod}`. The logback fix is necessary but
> not sufficient on its own — Loki will keep showing 0 ingest until the token is
> re-scoped.

## Observability (so neither failure mode is silent again)
- **WARN on misconfig:** when `GRAFANA_CLOUD_LOKI_URL` is set but no appender is
  attached, `LokiDropMetrics.loop` now logs a loud WARN instead of a silent debug
  (catches root cause A on any future regression).
- **`loki_logs_send_errors_total`:** a reset-resistant logback `StatusListener`
  counts loki4j send errors (401/403/4xx/5xx/timeout, which loki4j routes only to
  the StatusManager), polled by the existing `LokiDropMetrics` fiber. Catches root
  cause B. Ships with a consuming Grafana panel on `api-self-metrics.json`
  (new-metric-ships-with-its-dashboard rule); registered unlabelled in the
  cardinality firewall.

## Tests (TDD — red commit first)
- `LokiAppenderConfigSpec` now drives the **real `JoranConfigurator`** over the
  shipped `logback.xml` (loaded from the classpath, the same artifact the jar
  runs) with the secrets set via system properties, asserting the LOKI appender
  actually **attaches to root**. The prior DOM-shape assertions could not see the
  nesting bug. Red before the fix (`Failed to find appender named [LOKI]`), green
  after.
- `LokiDropMetricsSpec` pins the misconfig WARN, the `isLokiSendError` classifier,
  and the `SendErrorCountingListener`.

## Verification
- Full `mill api.test` green (one pre-existing unrelated flake, `MetricsExportSpec`,
  passes in isolation).
- Live push with real staging creds confirmed the 401 (root cause B) and that the
  restructured config no longer trips the nested-`<if>` rejection (root cause A).

### DIFF
--- a/api/resources/logback.xml
+++ b/api/resources/logback.xml
@@ -104,7 +104,31 @@
           <drainOnStop>false</drainOnStop>
         </batch>
       </appender>
+      <!--
+        #1972: the <root> MUST be declared HERE, inside this top-level <if>, and
+        must NOT carry a nested <if> of its own. logback (1.5.25 — the version
+        loki4j 2.0.3 forces onto the classpath and the deployed jar runs) REJECTS
+        an <if> nested within a <root>/<logger>/<appender> element
+        ("IfNestedWithinSecondPhaseElementSC"): the nested <appender-ref ref="LOKI">
+        was silently skipped and the LOKI appender never attached to root, so a
+        deployed API shipped ZERO logs to Grafana Cloud Loki even with the
+        GRAFANA_CLOUD_LOKI_* secrets set (0 ingest). Declaring a full <root> per
+        branch keeps every <if> at the top level (which logback DOES allow) and
+        attaches LOKI deterministically when the secret is present.
+        Pinned by LokiAppenderConfigSpec, which now loads this file through the
+        real JoranConfigurator at the shipped logback version.
+      -->
+      <root level="WARN">
+        <appender-ref ref="STDOUT" />
+        <appender-ref ref="LOKI" />
+      </root>
     </then>
+    <else>
+      <!-- Local dev / `mill __.test`: no Loki secret → console only. -->
+      <root level="WARN">
+        <appender-ref ref="STDOUT" />
+      </root>
+    </else>
   </if>
 
   <!--
@@ -113,15 +137,5 @@
     /api/router/* endpoints (mac, hostname, allowed/blocked, etc.). See
     docs/install-api.md § Debugging.
   -->
-  <root level="WARN">
-    <appender-ref ref="STDOUT" />
-    <!-- #1873: attach LOKI only where the secret is present (same gate as above). -->
-    <if condition='isDefined("GRAFANA_CLOUD_LOKI_URL")'>
-      <then>
-        <appender-ref ref="LOKI" />
-      </then>
-    </if>
-  </root>
-
   <logger name="wifihaven" level="${WIFIHAVEN_LOG_LEVEL:-INFO}" />
 </configuration>
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -152,6 +152,15 @@ object MetricGuard {
     // §4 cardinality firewall (service/env/level already ride the Loki stream
     // labels, not this Prometheus series).
     "loki_logs_dropped_total"                   -> Set.empty[String],
+    // #1972 — log batches the loki4j appender SENT to Grafana Cloud Loki that came
+    // back as an error (HTTP 401/403 wrong-scope token, 4xx/5xx, connection/timeout).
+    // Distinct from loki_logs_dropped_total, which only counts queue-full drops on
+    // the way IN: a misscoped access-policy token (logs:write missing) sheds every
+    // line at the SEND boundary, leaving drops flat at 0 — that auth failure was the
+    // invisible second half of #1972. loki4j is fail-open and routes these to the
+    // logback StatusManager only, so without this counter the loss is silent.
+    // Unlabelled (same cardinality firewall as the drop series).
+    "loki_logs_send_errors_total"               -> Set.empty[String],
     // §5.1 router-sourced, pushed via POST /api/router/metrics (#1205). Every one carries the
     // server-attached `router_id` + `installation_id` plus its own bounded enum label.
     "dnsmasq_restarts_total"                    -> Set("reason", "router_id", "installation_id"),
@@ -517,6 +526,15 @@ object AppMetrics {
     ZIO
       .when(delta > 0)(
         MetricGuard.counter("loki_logs_dropped_total", Map.empty, delta),
+      )
+      .unit
+
+  // #1972 — loki4j send-side errors (401/403/4xx/5xx/timeout) the appender reported
+  // to the logback StatusManager. Unlabelled; only emits when `delta > 0`.
+  def recordLokiSendErrors(delta: Long): UIO[Unit] =
+    ZIO
+      .when(delta > 0)(
+        MetricGuard.counter("loki_logs_send_errors_total", Map.empty, delta),
       )
       .unit
 
--- a/api/src/observability/LokiDropMetrics.scala
+++ b/api/src/observability/LokiDropMetrics.scala
@@ -78,7 +78,18 @@ object LokiDropMetrics {
   ): UIO[Unit] =
     findAppender.flatMap {
       case None      =>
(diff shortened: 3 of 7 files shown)


sameerparekh:
## Independent review — `fix(#1972)` Loki appender attachment + send-error observability

Read-only adversarial pass over `git diff origin/main...HEAD` (7 files) against the 8 checklist dimensions.

**BLOCKERS**
(none)

**SHOULD-FIX**
1. `api/src/observability/LokiDropMetrics.scala:58` — `isLokiSendError` counts *every* ERROR-level logback status whose origin is `com.github.loki4j.*`, not strictly send/HTTP failures. Post-start that's overwhelmingly send errors (so the metric is meaningful), but a one-off loki4j config/start ERROR would also tick `loki_logs_send_errors_total`. Acceptable as-is given the appender starts cleanly; worth a follow-up to narrow the match (e.g. message contains `non-success status`) if false positives ever show up. Not merge-gating.

**NITS**
1. `LokiDropMetrics.scala:141` — `lokiUrlConfigured` defaults to `sys.env.contains("GRAFANA_CLOUD_LOKI_URL")`, which brushes the "config via zio-config, not sys.env" convention. Here it's the right source: this secret is deliberately a logback-only gate (`isDefined(...)` in logback.xml), never a zio-config value, so reading the same env var keeps a single source for "is Loki configured." The param is by-name and only evaluated in the no-appender branch (not the hot path). Fine to leave.
2. `LokiDropMetrics.scala:123` — `installSendErrorListener` adds a StatusListener on each `loop()` call; `loop` is forked exactly once in `Main`, so no leak in practice. A guard would harden it against future double-invocation.
3. `deploy/grafana/dashboards/api-self-metrics.json:19` — new panel sits at `x:8,y:48` next to the drops panel (`x:0,y:48`); no overlap. Good.
4. PR body carries the harness-standard `🤖 Generated with Claude Code` footer; the `Co-Authored-By` trailers on commits are clean. Flagging only because checklist §7 mentions it — per project guidance the AI-authorship footer is expected, not something to scrub.

**Dimension notes**
- **1 SSOT:** `emitSendErrorDelta`/`recordLokiSendErrors` parallel the existing drop primitives against a *distinct* counter — no re-derivation of a domain quantity. Drops (queue-full, IN) vs send errors (rejection, OUT) are genuinely different facts, correctly kept separate.
- **2 Testing:** TDD red→green is visible in history (red `test(#1972)` commit `02abd91b` → green `fix` `f0ad3277`). The headline regression test now drives the **real `JoranConfigurator`** over the shipped `logback.xml` at the shipped logback version and asserts the LOKI appender actually attaches to root — it fails against the old config with the exact prod signature (`Failed to find appender named [LOKI]`). No existing assertion weakened/removed. Classifier + listener unit-tested.
- **3 Architecture:** no DNS-enforcement, router-agent, snapshot, or `/tmp`-write changes. ZIO effects throughout; listener's `AtomicLong` bridges the logback callback thread to the ZIO poller correctly (no metric mutation off-fiber).
- **6 Metrics+dashboard:** `loki_logs_send_errors_total` routed via `MetricGuard`, registered unlabelled in `MetricGuard.Allowed`, ships with a consuming Grafana panel. Compliant.
- **4/5/8:** no wire-shape change, no migration, no secrets in diff, no new routes/SQL.

**VERDICT: APPROVE**

Fixes the real root cause (logback `<if>` nested in `<root>` → LOKI appender never attached on logback 1.5.25 → 0 ingest) with a test that loads the actual config through Joran, and makes both silent failure modes observable (WARN on non-attachment; `loki_logs_send_errors_total` for the live 401/wrong-scope-token rejection the investigation surfaced). Note the PR's own call-out: the operator must re-issue `GRAFANA_CLOUD_LOKI_PASSWORD` with `logs:write` scope — the code fix alone won't make ingest non-zero. No blockers.

Issue #1989, opened 2026-06-26
ci/ops: validate #1988 JVM fix under smoke + green CD — bump staging to standard if it still OOMs


sameerparekh:
## Context
#1988 (closed #1984) bounded the staging JVM envelope in `render.yaml` (`-Xms192m -Xmx288m -XX:MaxMetaspaceSize=128m -XX:ReservedCodeCacheSize=64m -XX:MaxDirectMemorySize=64m -Dio.netty.maxDirectMemory=67108864 -XX:+UseSerialGC`; budget ~420-470MB). It MERGED at 02:10 and Render auto-redeployed staging (live 02:18). **But the fix is not yet proven and CD is still red:**

1. **The gates never re-ran.** #1988 was `render.yaml`-only, and master-api-ui.yml's changed-path detection skipped the build/deploy/gate jobs — there is NO master-api-ui run on the merge commit (e4e88b7b) and none queued. So Gate 1/3a/3b were never re-tested and the pipeline can't auto-green.
2. **Staging still OOM-killed at 02:21** () ~3 min after the 02:18 deploy — but with NO smoke running (no gate fired). So it's a single post-deploy/startup spike, untested against the real failure mode (OOM under the smoke burst).

## Work
1. **Check #1988 didn't mis-cap prod.** Confirm the JVM_HEAP_OPTS in `render.yaml` applies to the **staging** service only, not prod. Prod is `standard` (2GB) — if these 288m-heap limits leaked onto prod they'd needlessly throttle it. If so, scope them to staging.
2. **Force a real gate run against the fixed staging.** Since `render.yaml`-only changes don't trigger the gates, trigger master-api-ui (and master-router) deliberately — `workflow_dispatch` if supported, else a no-op commit touching an API path — so Gate 1 / Gate 3a / Gate 3b actually run the smoke against the redeployed staging. NOTE: a green master-api-ui run auto-promotes main to prod; main is just the merged render.yaml change, so that's acceptable — but be aware of it.
3. **Watch Render events during the smoke window** for `server_failed`/`oomKilled` on `srv-d8549fgjo89c73buvkf0` (token in orchestrator memory `render_api_token`). 
   - If NO oomKilled and the gates pass → the JVM tuning held. Done — report both pipelines green.
[...]

PR #1988, opened 2026-06-26
fix(#1984): bound full JVM memory envelope so staging API fits 512Mi under smoke

sameerparekh:
Closes #1984.

## Problem
`wifihaven-api-staging` (Render `srv-d8549fgjo89c73buvkf0`, plan **`starter` = 512Mi** hard cgroup cap) is **OOM-killed ~2 min after each deploy**, exactly when the (recently grown) CD smoke bursts load. Render `server_failed` events carry `"oomKilled": {"memoryLimit": "512Mi"}`. Each OOM = ~90-100s of 502s that no curl-retry window survives → recurring red on **Master API/UI CD (Gate 1, Gate 3b)** and **Master Router CD (Gate 3a)**.

This is the *second, distinct* failure after #1967's `/api/health` cold-start gate — that gate genuinely moved the smoke past where it used to die; the mid-run OOM is what remains.

## Root cause: only the heap was bounded
The staging container set `JVM_HEAP_OPTS=-Xms256m -Xmx384m` — heap capped, **everything else off-heap on JVM defaults**. Critically, `-XX:MaxDirectMemorySize` **defaults to ≈ `-Xmx`**, so Netty's pooled direct/off-heap buffers could grow *another ~384Mi on top of heap*. Metaspace and code cache were likewise unbounded.

```
worst-case heap (384) + Netty direct (~384) ≈ 768Mi  →  on a 512Mi container = guaranteed OOM under burst
```

## Fix — bound every region (staging only)
`render.yaml` `JVM_HEAP_OPTS` for `wifihaven-api-staging`:

| region | before | after |
|---|---|---|
| heap | `-Xmx384m` | `-Xms192m -Xmx288m` |
| metaspace | *(unbounded)* | `-XX:MaxMetaspaceSize=128m` |
| code cache | *(unbounded)* | `-XX:ReservedCodeCacheSize=64m` |
| **direct/Netty** | **≈ Xmx (~384m)** | `-XX:MaxDirectMemorySize=64m` + `-Dio.netty.maxDirectMemory=67108864` |
| GC | *(default)* | `-XX:+UseSerialGC` |

Direct memory is capped on **both** the JVM ceiling and Netty's own arena accounting (belt-and-suspenders). SerialGC is the lowest-footprint collector and is what the JVM picks anyway on starter's 0.5 CPU — set explicitly so it's deterministic regardless of container CPU detection.

### Memory math (512Mi budget)
```
heap      288
[...]

Issue #1984, opened 2026-06-26
ci/ops: staging API OOM-killed at 512Mi under smoke load — right-size JVM (bump plan only if needed)


sameerparekh:
## Root cause (diagnosed 2026-06-25)
The recurring Master API/UI CD (Gate 1, Gate 3b) and Master Router CD (Gate 3a) red on staging gates is **the staging API being OOM-killed**, not a cold-start flake and not a retry-budget problem. Render `server_failed` events on `wifihaven-api-staging` (`srv-d8549fgjo89c73buvkf0`) show `"oomKilled": {"memoryLimit": "512Mi"}` firing ~2 min after every deploy — exactly when the (recently grown) smoke bursts load. Each OOM = ~90-100s of 502s, which no curl-retry window survives. Staging is Render plan **`starter` (512Mi, 1 instance)**; prod is **`standard`**.

#1962/#1965/#1967 treated symptoms. #1967's `/api/health` cold-start gate genuinely helped (the smoke now runs far past where it died) — but the mid-run OOM remains. This is the second, distinct failure.

## Work
1. **Find the current JVM memory config** for the staging container — JAVA_OPTS / `-Xmx` / `-XX:MaxRAMPercentage`, the Dockerfile entrypoint, and any `render.yaml` envVars. Determine whether the JVM is container-memory-aware at all (if not, default max heap = 1/4 of the *host* RAM = far more than 512Mi → instant OOM).
2. **Right-size for 512Mi under load.** A ZIO/zio-http(Netty) app's footprint = heap + metaspace + **Netty direct/off-heap buffers** + thread stacks. The classic non-obvious OOM source here is **Netty direct memory** — cap it explicitly (`-XX:MaxDirectMemorySize` and/or `-Dio.netty.maxDirectMemory`) and set a conservative container-aware heap (e.g. `-XX:MaxRAMPercentage=50`), bounded metaspace, so heap+direct+metaspace+stacks < 512Mi with headroom. Apply via `render.yaml` env / Dockerfile (declarative).
3. **Prove it against the smoke.** Deploy to staging, run the smoke (or trigger the CD), confirm via the Render events API that NO `server_failed`/`oomKilled` fires during the smoke window. Render token in orchestrator memory `render_api_token`: `curl -H "Authorization: Bearer $KEY" https://api.render.com/v1/services/srv-d8549fgjo89c73buvkf0/events`.
[...]

PR #1991, opened 2026-06-26
ci(#1989): resolve base SHA on workflow_dispatch so CD gates can be triggered manually

sameerparekh:
Part of #1989 (follows #1984/#1988).

## Why
#1988 bounded the staging JVM envelope in `render.yaml` to stop the 512Mi OOM, but render.yaml-only changes aren't on the CD path filters, so **the gates never re-ran** and the fix is unvalidated. The natural way to force a gate run is `gh workflow run master-api-ui.yml --ref main` (both Master pipelines expose `workflow_dispatch`). I tried that — both runs **failed in CI** at `Migrations Immutability Check`, before any build/deploy/gate job:

```
.github/scripts/check-migrations.sh: line 12: 1: usage: check-migrations.sh <base-ref>
BASE_SHA:
```

## Root cause
`ci.yml`'s `migrations-immutable` job resolves a base SHA with a `case` over `github.event_name` covering only `pull_request` / `merge_group` / `push`. On `workflow_dispatch` the case falls through, `steps.base.outputs.sha` is empty, and `check-migrations.sh "$BASE_SHA"` aborts on its `${1:?usage}` guard.

This is self-inflicted: the `changes` job *already* forces every path output to `true` on `workflow_dispatch` (so all path-gated jobs, including this one, run) — the clear intent being that a manual dispatch exercises the full CD chain. That intent is defeated because the very job it enables can't resolve a base.

(The sibling `migration-isolation` job has the same `case` but is `if`-gated to `pull_request || merge_group`, so it correctly skips on dispatch and is unaffected. No change needed there.)

## Fix
Add a `workflow_dispatch` arm that uses `HEAD~1` — the parent of the dispatched commit, the same "what did this commit change" base as `push`. `fetch-depth: 0` on the job's checkout makes `HEAD~1` resolvable. On main, `HEAD~1...HEAD` over the migration dir is empty (no mutations ever land on main), so the check passes as intended.

## Validation path
`.github/workflows/ci.yml` is on **both** master-api-ui and master-router push-path filters, so **merging this PR triggers both CD pipelines via a real `push` event** (with a proper `github.event.before` → migrations check passes the normal way) — which runs Gate 1 / Gate 3a / Gate 3b against the redeployed staging API and validates the #1988 OOM fix end-to-end. Render staging (`srv-d8549fgjo89c73buvkf0`) has shown **no `oomKilled` since the 02:18 redeploy** (the one 02:21 spike was a startup transient with no smoke load); this merge runs the smoke against it for real.

Render Blueprint is untouched; `ruby -ryaml` parse + `actionlint` pass locally.

### Follow-up (not in this PR)
The deeper gap — a deploy-affecting `render.yaml` change doesn't run the deploy gates at all (render.yaml isn't on the CD path filters) — remains. Noting it here per #1989 item 4; worth a small follow-up to add `render.yaml` to the master-*.yml path filters so blueprint changes self-validate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -304,12 +304,20 @@ jobs:
 
       - name: Resolve base SHA
         id: base
-        # pull_request → PR base; merge_group → merge-base; push → before-SHA.
+        # pull_request → PR base; merge_group → merge-base; push → before-SHA;
+        # workflow_dispatch → parent of the dispatched commit (HEAD~1), the
+        # same "what did this commit change" base as push. Without this arm
+        # `sha` is empty on a manual dispatch and check-migrations.sh aborts
+        # with a usage error, reddening CD runs kicked off via
+        # `gh workflow run` (e.g. to validate a render.yaml-only deploy that
+        # the path filters otherwise skip). fetch-depth:0 above makes HEAD~1
+        # resolvable.
         run: |
           case "${{ github.event_name }}" in
-            pull_request) echo "sha=${{ github.event.pull_request.base.sha }}" >> "$GITHUB_OUTPUT" ;;
-            merge_group)  echo "sha=${{ github.event.merge_group.base_sha }}"  >> "$GITHUB_OUTPUT" ;;
-            push)         echo "sha=${{ github.event.before }}"                >> "$GITHUB_OUTPUT" ;;
+            pull_request)     echo "sha=${{ github.event.pull_request.base.sha }}" >> "$GITHUB_OUTPUT" ;;
+            merge_group)      echo "sha=${{ github.event.merge_group.base_sha }}"  >> "$GITHUB_OUTPUT" ;;
+            push)             echo "sha=${{ github.event.before }}"                >> "$GITHUB_OUTPUT" ;;
+            workflow_dispatch) echo "sha=$(git rev-parse HEAD~1)"                  >> "$GITHUB_OUTPUT" ;;
           esac
 
       - name: Reject changes to existing migration files


sameerparekh:
## Independent PR review

Read-only adversarial pass over `git diff origin/main...HEAD` (one file: `.github/workflows/ci.yml`).

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `.github/workflows/ci.yml:320` — `HEAD~1` covers only the *single* dispatched commit, whereas the `push` arm's `github.event.before` can span a multi-commit push. For the migrations-immutability check this is fine — `workflow_dispatch` only ever runs against already-merged `main`, where mutations can't exist, so a one-commit window is sufficient. Worth a one-line note only if the base is ever reused by a check where the full pushed range matters.
2. `.github/workflows/ci.yml:317-320` — the new `workflow_dispatch)` label is longer than the other arms, so the `echo` columns no longer line up perfectly. Purely cosmetic; bash is indifferent.

**Dimension coverage:** 1 SSOT — N/A (no business logic; the `case` arms are inherent, not duplicated logic). 2 Testing — workflow change, not TDD-able; no existing test weakened; validation is the merge itself firing the gates. 3 Architecture / 4 Wire — N/A. 5 Migrations — no migration added, so `check-migration-isolation` doesn't trigger; the immutability check on this PR resolves base from `pull_request.base.sha` and passes. 6 Metrics / 8 Security — N/A (no secrets; `git rev-parse HEAD~1` is deterministic, `fetch-depth: 0` guarantees it resolves). 7 Hygiene — focused single-file diff, no dead code, no AI mentions in title/body, three-dot diff.

**VERDICT: APPROVE**
Surgical, correct fix for a real CI hole: `workflow_dispatch` left `BASE_SHA` empty and crashed `check-migrations.sh`, blocking manual CD triggers. The added arm uses the same parent-commit base as `push`, guarded by the existing `fetch-depth: 0`. Merging also exercises the fix end-to-end (ci.yml is on both Master pipelines' push paths), running the staging gates that validate the #1988 OOM fix.

sameerparekh:
Closing per operator: the `workflow_dispatch) sha=$(git rev-parse HEAD~1)` base is wrong — it diffs only the **last commit**, so a manual dispatch CD run would under-check anything that changed across earlier ungated commits (we're not always validating just the tip). Not worth a narrower-than-push base.

Also no longer needed: the organic push run [28239722872](https://github.com/wifihaven/wifihaven/actions/runs/28239722872) (sha 3f8d478a, post-#1988-deploy) already exercised the staging gates and reproduced the failure — staging `oomKilled` at 13:17:41Z under the Gate 1 smoke (`curl (22) 502`). So the #1988 JVM tuning did **not** hold under smoke; the fix is a staging `plan: standard` bump (separate PR), which organic pushes will validate without any synthetic trigger.

Issue #1850, opened 2026-06-22
ws-transport: deprecate + remove the REST poll/ingest endpoints (operator-gated, LAST)


sameerparekh:
Part of the websocket-transport epic — see [`docs/design/websocket-transport.md`](../blob/main/docs/design/websocket-transport.md) §9 (sub-issue G), §3.

The **only** non-additive step in the epic. **Operator-gated — never armed automatically.** After the `router_transport{transport=ws}` rollout dashboard (sub-issue A/#1846) shows the fleet fully migrated to the websocket and the operator confirms the ws path healthy:

1. Add a one-release **deprecation log** to the REST poll path (`GET /api/router/policy`, and the `POST /api/router/usage|events|metrics` endpoints) — "deprecated, migrate to /api/router/ws".
2. Wait for the deprecation window per `docs/process/wire-contract.md` (fleet rolls forward).
3. Remove the REST poll/ingest endpoints in a **later** release.

Sequenced **last**; do not start until the operator gives the cutover go-ahead. Per `docs/pr-review-checklist.md#monitor-to-merged`, the cutover is the operator's explicit call.

Relates to #1023.

sameerparekh:
Gated by the 1-week prod ws soak **#2228** (cutover went live 2026-07-15). Do NOT start this removal until #2228's exit criteria pass (~2026-07-22). ws is live on prod with the HTTP poll as fallback; this issue removes that fallback and must wait for the soak.

PR #1994, opened 2026-06-26
feat(#1945): agent applies ws-pushed policy snapshots live (push→apply→enforce)

### LINKED ISSUE #1945: agent: apply a ws-pushed policy snapshot live (push→apply→enforce), not just save it

sameerparekh:
Closes #1945.

## Gap

The websocket push path was wired server→agent for **delivery** but not
**enforcement**. The `wifihaven-ws` sidecar's `on_policy` only called
`policy.save_snapshot` — it wrote the pushed snapshot to
`/etc/wifihaven/policy.json` but did **not** apply it. The running
`wifihaven-agent` reads that file only at startup; its tick loop applied policy
exclusively from the **HTTP poll**. So with `wifihaven.ws.enabled=1`, a pushed
`policy` frame updated the on-disk cache but enforcement (nft/dnsmasq) only
changed on the next poll or a restart — a no-op for live enforcement, and the
gating blocker for enabling the router↔API websocket in prod.

## Change

Add an **apply-on-push tick** to the agent's `on_tick` loop. When ws is enabled,
it re-reads the on-disk snapshot on a short cadence (`wifihaven.ws.apply_interval`,
default 2 s) and, **iff its `etag` differs** from the currently-applied one, runs
it through the **same** apply pipeline the poll uses
(`refresh_blocklists` → `policy.apply` → `render.update_shared`).

The shared apply+publish block is factored out of `on_tick` into
`apply_and_publish()` — reused by the poll-200, blocklist-refresh, and ws paths.
This is required by the **Lua 5.1 60-upvalue load limit** (#1930): inlining the
ws path would have pushed `on_tick` back over the limit. The refactor is a net
upvalue *reduction* for `on_tick`; verified by `loadfile` under real Lua 5.1.

### Architecture: single enforcement owner, no two-writer race
- The **sidecar** is the IN channel + sole **persister** (atomic `tmp→rename`);
  it never touches nft/dnsmasq.
- The **agent** is the sole **applier** of the enforcement plane.
- The **`etag`** is the one dedup key shared by the poll and the push, so a
  snapshot is applied **exactly once** even while the poll still runs (poll
  deprecation is #1850). The ws path does **not** re-save the file the sidecar
  wrote. A torn/missing read is retried next tick; atomic rename means the reader
  sees the old or the new complete file, never a partial one.
- `reason` for `dnsmasq_restarts_total{reason}` reuses the existing
  `policy_change` label — no new bounded label minted (no API allowlist /
  dashboard churn).

This follows AGENTS.md: the router stays a **dumb applier** — no decision logic
added; the ws path reuses the existing poll apply machinery verbatim.

## Tests

**Gate-2 e2e** (`scripts/e2e/scenarios_fake/test_ws_push_apply.py`) extended with
`test_ws_policy_push_applies_enforcement_live`: with the HTTP poll frozen at
3600 s, a pushed **paused** snapshot must install a per-MAC
`wh_drop:<mac>:Paused` forward-drop in nft — an observable that can only flip if
the running agent re-applied the pushed snapshot. Committed failing-first (red),
then green. Keyed on the `:Paused` drop rule rather than the `@blocked_macs` set
because a global allowlist (always present in real deployments, #1307/#1308)
routes blocked MACs to per-MAC rules carrying a `!= @global_allow` carve-out,
leaving `@blocked_macs` empty while the device is fully blocked.

## Real-hardware validation (mandatory — real Lua 5.1)

On **openwrt.lan** (Lua 5.1.5) pointed at the dev API (`api.lan:8080`), ws
enabled, poll frozen at 3600 s, snapshot pushed by simulating the sidecar's exact
atomic `save_snapshot` write (the router lacks `cqueues`, so the sidecar process
can't run there — but its only role in this path is that file write):

| step | on-disk etag | `wh_drop:<mac>:Paused` in nft |
|------|--------------|-------------------------------|
| startup (real dev-API policy) | sha256:56c2… | absent |
| push base (unpaused) | hwval-base-v1 | absent |
| **push paused** | hwval-paused-v2 | **PRESENT** |
| push base again (unpause) | hwval-base-v1 | absent |

Agent log per step: `ws apply: applied pushed snapshot etag=… (poll-independent)`.
The dev API never served the `hwval-*` etags, so the flip is attributable to the
ws push (on-disk read) **alone**. `on_tick` `loadfile`-compiles under real Lua 5.1
(60-upvalue limit satisfied). Router restored to its original state afterward.

### DIFF
--- a/docs/design/websocket-transport.md
+++ b/docs/design/websocket-transport.md
@@ -191,6 +191,22 @@ The ETag stays the snapshot's identity, but the *transfer* inverts:
 The `etag` field and `PolicyService.computeEtag` are unchanged — they remain the
 change-detection key the server uses to decide *whether* to push (§6.2).
 
+> **Implementation note (#1945).** "The agent applies it" is split across the two
+> agent processes to keep a *single enforcement owner*. The `wifihaven-ws` sidecar
+> (the only ws process) merely **persists** a pushed `policy` frame to
+> `/etc/wifihaven/policy.json` via the atomic `policy.save_snapshot` (tmp→rename);
+> it never touches nft/dnsmasq. The **main agent** owns enforcement: an
+> apply-on-push tick in its `on_tick` loop re-reads that file on a short cadence
+> (`wifihaven.ws.apply_interval`, default 2 s) and runs it through the *same*
+> apply pipeline the HTTP poll uses (`refresh_blocklists` → `policy.apply` →
+> `render.update_shared`) **iff the on-disk `etag` differs** from the
+> currently-applied one. The `etag` is the single dedup key shared by the poll
+> and the push, so even while the poll still runs (it is deprecated separately in
+> [#1850](https://github.com/wifihaven/wifihaven/issues/1850)) a snapshot is
+> applied exactly once and there is no two-writer race on the enforcement plane.
+> Before #1945 the sidecar saved the file but nothing re-applied it mid-run, so a
+> push was a no-op for live enforcement until the next poll/restart.
+
 ---
 
 ## 2. Capability negotiation (re-examining #376)
--- a/openwrt/files/usr/sbin/wifihaven-agent
+++ b/openwrt/files/usr/sbin/wifihaven-agent
@@ -122,6 +122,14 @@ end
 local ws_enabled        = uci_get_ws("enabled", "0")
 local ws_fallback_after = tonumber(uci_get_ws("fallback_after", "300"))
 local ws_spool_max      = tonumber(uci_get_ws("spool_max_bytes", "1048576"))
+-- #1945: apply-on-push cadence (seconds). When ws is enabled the sidecar writes
+-- pushed snapshots to /etc/wifihaven/policy.json (the IN channel); the running
+-- agent must APPLY them live, not only at startup. The on_tick loop re-reads the
+-- on-disk snapshot on this cadence and re-applies when its etag changed (a cheap
+-- read+compare otherwise). 2 s keeps push→enforce latency low while staying well
+-- above the cost of one snapshot parse (the HTTP poll already parses one every
+-- ~5 s in non-ws mode). No-op unless ws_enabled == "1".
+local ws_apply_int      = tonumber(uci_get_ws("apply_interval", "2"))
 
 log.init({ debug = debug_opt })
 -- #771: PKG_VERSION baked at install time into /usr/lib/wifihaven/VERSION.
@@ -681,6 +689,39 @@ do
   if f then f:write(block_page_url); f:close() end
 end
 
+-- ── Shared apply pipeline (#1945) ─────────────────────────────────────────────
+-- apply_and_publish(snap, reason) → applied:bool
+-- The single "push this snapshot onto the enforcement plane" sequence, factored
+-- out of on_tick so the HTTP-poll path, the blocklist-refresh path, AND the new
+-- ws apply-on-push path share ONE apply+publish block (single-source-of-truth)
+-- and so on_tick stays under the Lua 5.1 60-upvalue load limit (#1930): the heavy
+-- render / shared-table / metrics upvalues live HERE, not in the on_tick closure.
+--
+-- It does the enforcement-plane work only — capture-before-reset drop counters,
+-- policy.apply (dnsmasq conf + nft), then render.update_shared to rebuild the
+-- conntrack/usage-side per-MAC tables. It deliberately does NOT touch
+-- ts.current_etag / ts.last_snapshot / mark_poll_success / save_snapshot: those
+-- differ per trigger (the poll persists its fetch; the ws path must NOT re-save
+-- the file the sidecar already wrote), so each caller owns them. `reason` flows
+-- to record_apply for dnsmasq_restarts_total{reason}; the ws path reuses
+-- "policy_change" (it IS a policy change — no new bounded label is minted).
+local function apply_and_publish(snap, reason)
+  -- #1325: capture the about-to-reset enforcement-drop counters before apply
+  -- rebuilds the ruleset and zeroes them.
+  fold_enforcement_drops()
+  local at0 = clock.monotonic_seconds()
+  local applied = policy.apply(snap, write_file, run_cmd, log,
+                               { on_apply = record_apply(reason) })
+  metrics.observe(mreg, "policy_apply_duration_seconds", {},
+                  clock.monotonic_seconds() - at0)
+  if applied then
+    render.update_shared(snap, nft_sets, blocked_macs, blocked_reason,
+                         eb_hosts_by_mac, ea_hosts_by_mac, bl_hosts_by_mac,
+                         make_bl_member_iterator(snap))
+  end
+  return applied
+end
+
 -- ── Timer state ───────────────────────────────────────────────────────────────
 
 -- #1930: all mutable per-tick scheduler state lives in ONE table so the big
@@ -706,6 +747,7 @@ local ts = {
   last_metrics_run   = 0,
   last_blocklist_run  = 0,  -- #1412: category blocklist refresh cadence
   last_eb_refresh_run = 0,  -- #1658: eb_/bl_ ipset re-resolve cadence
+  last_ws_apply_run   = 0,  -- #1945: ws apply-on-push poll-of-disk cadence
   last_activity_sample = 0,
   last_nflog_run      = 0,  -- #1126: nflog drop-spool drain cadence
   -- Wall-clock anchor for usage report periodStart. Updated each time the usage
@@ -811,6 +853,7 @@ do
   ts.last_metrics_run       = mono
   ts.last_blocklist_run     = mono
   ts.last_eb_refresh_run    = mono
+  ts.last_ws_apply_run      = mono
   ts.last_activity_sample   = mono
   ts.last_nflog_run         = mono
 end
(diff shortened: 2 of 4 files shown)


sameerparekh:
## Independent review — first pass

### 1. Single source of truth ✅ (net improvement)
The change **removes** duplication: the apply+publish sequence that was inlined in the poll-200 path and the blocklist-refresh path is factored into `apply_and_publish()` (`wifihaven-agent:705`), now shared by poll / blocklist / ws paths. The ws-apply etag-guard (`wifihaven-agent:1140` `pushed.etag ~= ts.current_etag`) reuses the **same** `ts.current_etag` dedup key as the poll, so the two transports stay coherent (after a ws apply, the next poll's `If-None-Match` carries the ws etag → server 304, no redundant re-apply). No policy decision is re-derived on the agent.

### 2. Testing integrity ✅
- TDD red→green visible: `f1f12f79` (failing e2e) precedes `a217abbf` (impl).
- New behavior asserted on a concrete enforcement observable — a per-MAC `wh_drop:<mac>:Paused` forward-drop appearing in live nft with the HTTP poll frozen at 3600 s (`test_ws_push_apply.py:_paused_drop_rule_present` + `test_ws_policy_push_applies_enforcement_live`). The observable is robust to the global-allowlist branch: `render.lua:1295` emits the per-MAC `wh_drop:<mac>:<reason>` drop whether the MAC lands in `blocked_macs_list` (no global allow, Gate-2 fake) or `blocked_ea_macs` (global allow, real deploy) — verified against render, so it is not a fake-only false-pass.
- No existing assertion weakened; the save-only legs are unchanged. The `paused` kwarg on `_snapshot` defaults `False` (back-compat). New test rides the existing `ws_push` mark in the existing Gate-2 harness — no new docker-compose job.

### 3. Architectural invariants ✅
- **Router stays a dumb applier** — the ws-apply tick adds **no** decision logic; it reuses `refresh_blocklists` → `policy.apply` → `render.update_shared` verbatim. The etag compare is the existing "apply iff changed" guard, not new policy.
- **Single enforcement owner / no two-writer race** — the sidecar only persists (atomic `tmp→rename`, `wifihaven-ws:155`); the agent is the sole nft/dnsmasq applier. The ws path does **not** re-save the file (`wifihaven-agent:1144` comment). Atomic rename → reader sees old-or-new, never torn.
- **No new wire field** — `apply_interval` is a router-side UCI knob, not on the snapshot wire.
- **/tmp bounded** — the ws path *reads* `/etc/wifihaven/policy.json` (flash, whole-file) and writes nothing new under `/tmp`; downstream `policy.apply`/`refresh_blocklists` writes are the existing bounded whole-file/shard rewrites.
- **Lua 5.1 60-upvalue limit (#1930)** — the refactor is a net upvalue *reduction* for `on_tick` (render + 5 shared tables move into the helper; +4 for the ws path); `loadfile` under real Lua 5.1.5 confirmed in the PR body.

### 4. Wire contract ✅ — no wire shape change; additive UCI knob only.
### 5. Migrations ✅ — none.
### 8. Security ✅ — no secrets/SQL/new route; reads a root-owned file already in the trust boundary (startup load + poll save already trust it).

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `wifihaven-agent:1147` — a ws-triggered apply is not distinguishable in metrics from a poll-triggered one (both `record_apply("policy_change")`; `policy_apply_total{result}`). An operator infers ws-apply from `policy_apply_total` rising while `snapshot_poll_total` stays flat (and `ws_frames_recv_total{op=policy}` ticking). Acceptable — a dedicated `reason`/label would need a `MetricGuard.Allowed` + dashboard change; reusing the bounded label here is the right call. Worth a follow-up only if ws-apply visibility becomes an ops need.
2. `wifihaven-agent:1138` — the ws-apply tick is skipped while `ts.in_failover_render`. Correct for #1945 (the HTTP poll still runs at its normal cadence, so failover is driven by real poll failures and the failover chain owns the plane). But once #1850 deprecates/freezes the poll, a poll-failure→failover with a *healthy* ws socket would suppress live pushes — flag to revisit the failover-vs-ws precedence in #1850.
3. `wifihaven-agent:1147` — `policy.mark_poll_success(wall)` on a ws apply slightly overloads the "poll" name, but is semantically correct and load-bearing: it is the "last time we have fresh policy" timestamp, so marking it keeps a frozen-poll + healthy-ws router from tripping failover on `poll_age`.

VERDICT: APPROVE

Wires agent-side apply-on-push so a ws-pushed snapshot enforces live, etag-guarded, with the agent as the single enforcement owner and the apply machinery shared with the poll path (improving SSOT and staying under the Lua 5.1 upvalue limit). TDD red→green present; the enforcement-flip e2e is robust across the global-allowlist branch; validated on real Lua 5.1 hardware with the poll frozen. No blockers; three non-blocking observability/forward-looking notes.

sameerparekh:
## Independent review — re-run (incremental)

Prior review: `a217abbf`. Delta `a217abbf...d428b446` is **test-only** (`scripts/e2e/scenarios_fake/test_ws_push_apply.py`, +30/-3) — the agent and sidecar production code are unchanged.

### Prior-findings status
| # | Prior finding (NIT) | Status |
|---|---|---|
| 1 | ws-apply not distinguishable from poll-apply in metrics (`wifihaven-agent`) | **NOT-ADDRESSED** — unchanged; intentional (reuse bounded `policy_change` label, avoid `MetricGuard.Allowed`/dashboard churn). Non-blocking as noted. |
| 2 | ws-apply skipped during `in_failover_render`; revisit when #1850 freezes the poll | **NOT-ADDRESSED** — unchanged; correct for #1945 (poll runs at normal cadence in prod), flagged forward to #1850. |
| 3 | `mark_poll_success` on a ws apply overloads the "poll" name | **NOT-ADDRESSED** — unchanged; semantically correct (last-fresh-policy timestamp), keeps a frozen-poll+healthy-ws router out of failover. |

All three were NITs (non-merge-gating) and remain accurate design notes; no fix expected.

### New findings (delta)
The previous Gate-2 run failed: the live-apply test booted no client VM, so with no traffic the agent's conntrack-driven `on_tick` starved (the metrics/poll timers that would generate a flow are themselves *inside* `on_tick`, so it can't self-restart), and the pushed paused snapshot was never applied. Root cause is established agent architecture (`conntrack.lua:1144-1196` — `on_tick` runs per `conntrack -E NEW` line), not a regression in this PR. The fix nudges a fresh tracked flow (a short GET to the API the agent already reaches) each wait iteration so `on_tick` fires — exactly as live traffic does.

- **Testing integrity** ✅ — the nudge carries **no policy** (the HTTP poll stays frozen at 3600 s; a `/healthz` hit is not a snapshot fetch), so the enforcement flip stays attributable to the ws push. It does not weaken the assertion — it makes the agent's traffic-driven scheduler runnable, which a phantom-device scenario otherwise lacks. Verified on real hardware: 3 nudge curls → 3 `[NEW] tcp dport=8080` conntrack events (`conntrack -E -e NEW` is unfiltered, so a router-originated flow drives a tick).
- **Scope/hygiene** ✅ — single test file; documented rationale; no production code touched.

BLOCKERS
(none)

SHOULD-FIX
(none)

NITS
1. `test_ws_push_apply.py:_nudge_conntrack` — the nudge assumes the API is reachable from the router (true: the agent polls it) and that `/healthz` completes a TCP handshake (a 404 suffices; only connection-refused would not, and the fake is listening). Fine as-is; noting the dependency.

VERDICT: APPROVE

Delta is a test-only fix making the phantom-device live-apply scenario drive the agent's conntrack-gated scheduler so the apply-on-push tick runs; the enforcement-flip assertion is unchanged and remains push-attributable (poll frozen). Prior NITs stand as design notes; no blockers. Re-running Gate-2 to confirm green.

PR #1992, opened 2026-06-26
fix(#1989): bump staging API to plan: standard — 512Mi JVM tuning still OOMed under smoke

sameerparekh:
Closes #1989. Follows #1984/#1988.

## TL;DR
#1988's bounded JVM envelope did **not** hold staging under the CD smoke — it still OOM-killed at the `starter` 512 MB cap. Per the operator's pre-authorized fallback ("try JVM tuning, bump if needed"), this bumps **staging only** to `plan: standard` (2 GB) and restores prod-mirroring heap opts. Prod is untouched.

## Evidence the tuning was insufficient
The validation run was an **organic push** CD run, not a synthetic trigger: [run 28239722872](https://github.com/wifihaven/wifihaven/actions/runs/28239722872) (sha `3f8d478a`, deployed after #1988's env landed at 02:18Z).

- Render staging (`srv-d8549fgjo89c73buvkf0`) `server_failed` / **`oomKilled` `{"memoryLimit":"512Mi"}` at `2026-06-26T13:17:41Z`**, mid-Gate-1 (and again `14:12:33Z`) — *with* the full #1988 envelope live (heap 288m + metaspace 128m + codecache 64m + direct 64m + SerialGC).
- **Gate 1 — API contract smoke (staging)**: 8 curl retries → `curl: (22) ... error: 502`. Gate 3b (ipk + apk) red the same way. (Gate 3a on master-router likewise depends on staging.)

So honest tuning could not fit 512 MB under the grown smoke burst — exactly the bounded-OOM condition #1988's own note said would justify the bump.

## Change (staging only)
| | before (#1988) | after |
|---|---|---|
| `plan` | `starter` (512 MB) | **`standard` (2 GB)** |
| `JVM_HEAP_OPTS` | `-Xms192m -Xmx288m … MaxDirect=64m … SerialGC` | **`-Xms512m -Xmx1400m`** (mirrors prod) |

Dropping the cramped 512 MB-fit envelope and matching prod's heap profile makes staging memory-representative of prod, so the CD smoke runs against a prod-shaped heap.

## Scope / safety
- **Staging only.** `wifihaven-api-prod` (already `standard`, `-Xms512m -Xmx1400m`) is **not** touched — verified its block is byte-identical.
- render.yaml-only; no source. Render Blueprint Lint (`ruby -ryaml` parse) passes locally.

## Validation
[...]

PR #1919, opened 2026-06-23
Fixes #1911: enforce network-wide blockEncryptedDns (NXDOMAIN relay/DoH hosts + nft drop resolver IPs & DoT/853)

sameerparekh:
## What

Agent half of the household **"block encrypted DNS & relays"** toggle (#1909, epic #1903). Closes #1911.

The policy snapshot gains a single **additive top-level boolean `blockEncryptedDns`** (sibling of `global`/`devices`/`profiles`/`blocklists`, default `false`). This PR reads the flag and enforces; the API producer side is #1912 — they land in any order (additive field, agent ignores absent → off, every intermediate state is a no-op).

When `blockEncryptedDns == true`, the router enforces **network-wide (all managed MACs):**

1. **Negative DNS answer (NXDOMAIN), not a sinkhole**, for curated relay/DoH **hostnames** (`mask.icloud.com`, `mask-h2.icloud.com`, `cloudflare-dns.com`, `dns.google`, `dns.quad9.net`, `dns.nextdns.io`, `dns.adguard-dns.com`, `doh.opendns.com`) via dnsmasq `local=/<host>/`. For iCloud Private Relay this is Apple's **clean disable signal** — iOS falls back to direct, filterable connections **without** the client-side hang a silent connection-layer drop causes (the #1891 root cause). Verified on dnsmasq 2.91 that `local=/host/` returns a negative answer, **not** a `0.0.0.0` sinkhole.
2. **nftables forward-drop** for curated public-resolver **IPs** (`1.1.1.1`/`1.0.0.1`/`8.8.8.8`/`8.8.4.4`/`9.9.9.9`/`149.112.112.112` + v6) and for **DoT (TCP/853) to any IP**.

Curated lists are **baked into the agent** (`encrypted_dns.lua`), not shipped on the wire — the snapshot stays a single bool. Flag absent/false renders **byte-identically** to today.

## Architectural Truth #1 exception (documented)

This is the **one sanctioned** place the agent returns a negative DNS answer. It is **bypass-disable signalling**, not content blocking: the negative answer tells the client to stop tunnelling so the real connection-layer enforcement can see the traffic. Scoped to the curated list, gated on the household toggle. Documented in `docs/architecture.md` §0.1 ("Truth #1 — the `blockEncryptedDns` exception"). The curated-IP drops carry **no** `extraAllowed`/`@global_allow` carve-out (an allow entry must not re-open a bypass) and are **not** DNAT'd to the block page.

## Changes

[...]

Issue #1912, opened 2026-06-23
feat(api): blockEncryptedDns household setting + emit global.blockEncryptedDns in snapshot


sameerparekh:
Sub-issue of #1909 (feature) / #1903 (epic). **Independent — no required landing order** (additive field; agents ignore it until they enforce it).

Add the household-level "block encrypted DNS & relays" setting and emit it on the wire:
- **Schema-only PR**: migration adding the `household_settings` column (per AGENTS.md §migrations-back-compat).
- **Code PR**: `HouseholdSettings` field + repo read/write; emit additive `global.blockEncryptedDns: Boolean` in the snapshot `global` section (default false). Contract-fixture/back-compat tests.

Intra-dependency: the code PR depends on the migration PR (column must exist) — that's the only ordering, and it's internal to this issue. No ordering vs the agent (#1911) or UI (#1913).

Epic: DNS-Bypass & Attribution.

PR #1996, opened 2026-06-26
ci(#1995): make Gate 1 #1912 blockEncryptedDns smoke set/restore instead of asserting ambient default

### LINKED ISSUE #1995: ci(Gate 1): #1912 blockEncryptedDns smoke asserts ambient default=false against shared staging (household singleton left true) — make it set/restore
## Symptom
Master API/UI CD Gate 1 (API contract smoke against staging) fails at:
```
▶ #1912: blockEncryptedDns reflects the household setting
✗ #1912: expected default blockEncryptedDns=false, got: true
```
(run 28256444289, on standard-tier staging — NOT an OOM; staging deployed clean, Gate 3b passed, only this assertion failed. The earlier OOM was masking this exact step.)

## Root cause
`blockEncryptedDns` is a **household-singleton** setting. The code default is `false` (migration `V61__household_block_encrypted_dns.sql`: `block_encrypted_dns BOOLEAN NOT NULL DEFAULT FALSE`). But staging's household row currently persists `true` — almost certainly left enabled during the #1919/#1921 `blockEncryptedDns` enable/revert testing (the revert was done on the router, not necessarily on the staging API household). The Gate 1 smoke (`#1912`, in the staging "Router API smoke" script — Gate 1's e2e script, NOT the scenarios_fake Gate-2 tests) asserts the **ambient default is false** before exercising the round-trip. Asserting a default on a **mutable singleton in a shared, persistent staging DB** is inherently fragile and wrong.

## Fix (durable)
Make `#1912` self-contained and not dependent on ambient state:
- Read + remember the current `blockEncryptedDns`.
- PUT `false` → GET → assert `false`; PUT `true` → GET → assert `true` (this is what "reflects the household setting" actually means).
- Restore the original value in cleanup.
- Drop the "expected default=false" precondition assertion entirely (it can't hold on shared staging).

Optionally (immediate unblock, secondary): reset staging's household `block_encrypted_dns` to `false` via the API — but the durable fix above is what prevents recurrence.

## Definition of done
- Gate 1 passes on staging regardless of the household's starting `blockEncryptedDns` value, while still verifying the setting round-trips both ways.
- A fresh master-api-ui run is GREEN on Gate 1 (and Gate 3b) with no oomKilled.

Surfaced 2026-06-26 once the staging OOM fix (#1989/#1992) stopped masking this step. Relates to #1911/#1919/#1921.

sameerparekh:
Fixes #1995.

## Problem
Master API/UI CD **Gate 1** (API contract smoke against staging) fails at:
```
▶ #1912: blockEncryptedDns reflects the household setting
✗ #1912: expected default blockEncryptedDns=false, got: true
```
This was the last blocker to green CD — the staging OOM is fixed (#1992, staging now on Render standard), Gate 3b passes, no `oomKilled`; only this assertion failed once the OOM stopped masking the step (run 28256444289).

## Root cause
`blockEncryptedDns` is a **household-singleton** setting. The code default is `false` (`V61__household_block_encrypted_dns.sql`: `... DEFAULT FALSE`), but staging's household row persists `true` — left enabled during #1919/#1921 `blockEncryptedDns` testing. The `#1912` step in `scripts/e2e-router.sh` (Gate 1's staging smoke — *not* the `scenarios_fake` Gate-2 tests) asserted the **ambient default is `false`** before round-tripping. Asserting a default on a **mutable singleton in a shared, persistent staging DB** is inherently fragile and wrong.

## Fix (durable, test-isolation)
Make `#1912` self-contained — independent of the starting value:
- **Capture** the household's current `blockEncryptedDns` (assert only that it's a present bool — the field is non-optional on the wire).
- **Round-trip both ways** via PATCH → poll the wire: set `false` → assert `false`; set `true` → assert `true`. That round-trip is what "reflects the household setting" actually verifies.
- **Restore** the original value in `cleanup()` (same trap-fire pattern as the test profiles/devices/router), so the run is idempotent and leaves staging as-found.
- **Drop** the "expected default=false" precondition assertion entirely.

The PATCH/poll is factored into a small `set_bed <true|false>` helper. The existing `curl-retry.sh` override still wraps every call, so the new helper rides out transient staging 502s.

## Validation
- `bash -n scripts/e2e-router.sh` clean.
- Confirmed endpoint/JSON shape against `api/src/routes/Routes.scala`: `PATCH /api/household/settings` accepts `{"blockEncryptedDns": <bool>}`, and the flag surfaces top-level on `/api/router/policy`.
- `scripts/**` is on the `master-api-ui.yml` push-path filter and Gate 1 runs `scripts/e2e-router.sh`, so this change re-triggers Gate 1.

### DIFF
--- a/scripts/e2e-router.sh
+++ b/scripts/e2e-router.sh
@@ -91,6 +91,12 @@ pass "profile id=$PID name=$PROFILE_NAME"
 EXTRA_DEVICES=()
 EXTRA_PROFILES=()
 
+# Original household blockEncryptedDns, captured by the #1912 step before it
+# round-trips the singleton, so cleanup() can restore staging to as-we-found-it
+# regardless of the starting value (it is a shared, persistent DB — we must not
+# assume, or leave behind, a particular default). Empty until the step runs.
+BED_ORIG=""
+
 # Register cleanup so test data is removed even on failure.
 cleanup() {
   echo
@@ -104,6 +110,13 @@ cleanup() {
   for p in "${EXTRA_PROFILES[@]:-}"; do
     [ -n "$p" ] && curl -s -X DELETE "$BASE/api/profiles/$p" "${AUTH[@]}" >/dev/null 2>&1 || true
   done
+  # Restore the household blockEncryptedDns singleton to its pre-run value (#1912
+  # round-trips it). Skipped if the step never ran or it was already as-found.
+  if [ -n "$BED_ORIG" ]; then
+    curl -s -X PATCH "$BASE/api/household/settings" "${AUTH[@]}" \
+      -H 'content-type: application/json' \
+      -d "{\"blockEncryptedDns\":$BED_ORIG}" >/dev/null 2>&1 || true
+  fi
   pass "test profile + device + router removed"
 }
 trap cleanup EXIT
@@ -1286,11 +1299,17 @@ pass "#932: /api/usage/series returns topHosts + buckets arrays"
 # as the additive top-level boolean `blockEncryptedDns` (sibling of global /
 # devices / profiles / blocklists). The router agent (#1911) bakes the curated
 # relay/DoH host + resolver-IP lists itself; the API only ships the boolean.
-# Assert: default is false, flipping the household setting flips the wire flag,
-# and we reset it so the run is idempotent against persistent staging.
+#
+# What "reflects the household setting" means is the *round-trip*: flipping the
+# household setting flips the wire flag, both ways. We deliberately do NOT assert
+# an ambient default here — blockEncryptedDns is a household SINGLETON in a
+# shared, persistent staging DB, so its starting value is whatever a prior run /
+# manual test left (#1995 — it was left `true` after #1919/#1921 testing). We
+# capture the starting value, drive it to both states explicitly, and restore it
+# in cleanup() so the run is idempotent regardless of where it started.
 step "#1912: blockEncryptedDns reflects the household setting"
 
-read_bed() {  # → "true"/"false" from a fresh /api/router/policy read
+read_bed() {  # → "true"/"false"/"MISSING" from a fresh /api/router/policy read
   curl -fsS "${RAUTH[@]}" "$BASE/api/router/policy" >"$TMP/snap_bed.json"
   _py "
 import json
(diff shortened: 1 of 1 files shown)


sameerparekh:
## Independent PR review

Read-only adversarial pass over `git diff origin/main...HEAD` (one file, `scripts/e2e-router.sh`) against the 8-dimension checklist. This is a Gate 1 staging-smoke test-script change only — no Scala, no migration, no wire/metric surface.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `scripts/e2e-router.sh:113-114` — the cleanup comment says the restore is "Skipped … if it was already as-found," but the code restores whenever `BED_ORIG` is non-empty (it re-PATCHes the same value when already as-found). Harmless idempotent write; comment slightly overstates. Optional wording tweak.

**Dimension notes**
- **§2 Testing integrity** (the load-bearing dimension here): the removed `expected default blockEncryptedDns=false` assertion is a *justified* test change, not a hidden-regression weakening — the PR explains the old assertion was wrong (it pinned the ambient value of a mutable household **singleton** in a shared, persistent staging DB). The replacement is strictly stronger on the real contract: it still rejects a missing/non-bool wire field (`BED_ORIG` must be `true`/`false`) and now asserts the round-trip **both** ways (set→`false`→assert, set→`true`→assert) instead of relying on a fixed starting point. `fail()` does `exit 1`, so the EXIT trap fires and `cleanup()` restores `BED_ORIG` even on the failure path.
- **§1 Single source of truth**: the PATCH+poll is consolidated into one `set_bed` helper — reduces duplication vs. the prior two hand-rolled poll loops. No re-derivation.
- **§3/§4 Architecture/wire**: no DNS-as-enforcement reasoning, no wire-shape change; endpoint/JSON verified against `api/src/routes/Routes.scala` (`PATCH /api/household/settings` ⇄ top-level `blockEncryptedDns` on `/api/router/policy`).
- **§7 Hygiene**: focused single-file diff; `bash -n` clean; the existing `curl-retry.sh` override still wraps every call so `set_bed` rides out transient 502s; three-dot diff; no AI mentions in title/body.
- **§5/§6/§8**: no migration, no metric path, no new route/secret/SQL — N/A.

VERDICT: APPROVE
Rewrites the fragile Gate 1 `#1912` step so it captures→round-trips→restores the `blockEncryptedDns` household singleton instead of asserting an ambient default that cannot hold on shared staging. Correctly scoped, robustly cleaned up, and a stronger contract check than before. The single NIT is a comment nicety, non-blocking.

sameerparekh:

Re-run on push `878164d`. Delta from prior reviewed `58426ab` is a single comment-wording change in `cleanup()`; no logic change.

**Prior findings status**
|---|---|---|
| NIT 1 | `cleanup()` comment overstated the "skipped if already as-found" case | **ADDRESSED** — comment now states it skips only when `BED_ORIG` is unset and otherwise re-asserts (idempotent no-op). `scripts/e2e-router.sh:113-115` |

**New findings in delta**
(none)

VERDICT: APPROVE
Comment-only follow-up that resolves the sole prior NIT; no behavior change. The substantive fix (capture→round-trip both ways→restore the `blockEncryptedDns` household singleton, dropping the fragile ambient-default assertion) stands as reviewed. No open BLOCKERs.

PR #2000, opened 2026-06-27
ci(#1999): bump Gate 2 VM e2e timeout 50→75min (ipk exceeds cap after #1945)

### LINKED ISSUE #1999: ci: Gate 2 ipk VM e2e deterministically hits the 50-min timeout cap (apk fits) — bump e2e-vm-fake timeout / speed up
## Symptom
Master Router CD **Gate 2 — router fake-API e2e / VM e2e (fake API) — ipk** is cancelled at exactly **50 minutes**, twice in a row, while the **apk** variant of the same suite passes in ~35 min. Job timing: started 00:09:31Z → ended 00:59:54Z → `The operation was canceled.` The cap is `.github/workflows/e2e-vm-fake.yml:80: timeout-minutes: 50`.

This is NOT a transient flake (re-running just times out again, proven twice) and NOT the staging-OOM incident (that's resolved: Gate 3a green, Gate 2 apk green, no oomKilled). It is a deterministic timeout-budget problem on the ipk VM e2e.

## Likely cause
Router CD Gate 2 has historically run close to the cap (e.g. run 28177601079 took 49m44s). **#1945** (PR #1994) added a new Gate-2 enforcement-flip e2e (`test_ws_push_apply.py` push→apply→enforce assertion) — extra VM runtime that appears to have tipped the slower **ipk** variant past 50 min. (apk fits; ipk is consistently slower — packaging/boot overhead or runner variance.)

## Fix options (in this PR)
1. **Bump `timeout-minutes`** in `.github/workflows/e2e-vm-fake.yml` from 50 → ~75 (precedent: prior `ci/bump-vm-e2e-timeout` / `-120` branches). Fastest unblock.
2. **Investigate the apk-vs-ipk gap** (~15 min): is one test hanging/retrying only under ipk? a per-test slow spot? If a single test is pathologically slow under ipk, fix that instead of (or in addition to) the bump.
3. Optionally split/parallelize the VM e2e suite so neither variant approaches the cap.

Start with the bump to unblock Router CD, and note any apk/ipk-gap finding for follow-up if not fixed here.

## Definition of done
- Master Router CD Gate 2 (both apk + ipk) completes within its timeout and goes GREEN.
- Root cause of the ipk slowness noted (even if the fix is just a timeout bump).

Diagnosed 2026-06-26 during the post-OOM CD watch. Relates to #1945/#1994 (added the Gate-2 e2e). Distinct from the staging-OOM incident (#1984/#1989/#1992, resolved).

sameerparekh:
## Problem
Master Router CD **Gate 2 — router fake-API e2e** (`e2e-vm-fake.yml`) is deterministically RED: the **ipk** matrix arm is cancelled at **exactly 50 minutes**, twice in a row (job 00:09:31Z → 00:59:54Z → `The operation was canceled.`), while the **apk** arm of the same suite passes in ~35 min. This is the only thing keeping Master Router CD red — the staging-OOM incident (#1984/#1989/#1992) is resolved (Gate 3a green, Gate 2 apk green, no oomKilled).

Not a transient flake (proven twice), not OOM.

## Root cause
Gate 2 had already run as close as **49m44s** (run 28177601079) before this. **#1945** (PR #1994) added a new Gate-2 enforcement-flip scenario (`scripts/e2e/scenarios_fake/test_ws_push_apply.py`, push→apply→enforce). Each VM scenario carries a fixed ~2m snapshot-restore + boot cost; that extra runtime tipped the consistently-slower **ipk** arm past the 50m cap. ipk is slower for packaging/boot overhead + shared-host load on api.lan; apk still fits.

## Fix
Bump `timeout-minutes` 50 → 75 in `e2e-vm-fake.yml`. Both matrix arms (ipk + apk) share the single job-level cap (line 80), so this covers both. Verified there is **no second timeout**: the called scripts (`e2e-vm.sh`, `conftest.py`) only have per-fixture waits, and the `workflow_call` site in `master-router.yml` (`gate-2-router-fake-api`) sets no `timeout-minutes`. Precedent: prior 40→50m and 30→40m bumps as the scenario count grew.

This is the accepted unblock per #1999; the ipk slowness is uniform packaging/boot overhead rather than one pathological test (see issue diagnosis), so no per-test fix is bundled here.

Closes #1999

### DIFF
--- a/.github/workflows/e2e-vm-fake.yml
+++ b/.github/workflows/e2e-vm-fake.yml
@@ -77,7 +77,17 @@ jobs:
     # while the ipk arm passed at 32m30s. The pack itself is healthy (per-step
     # times unchanged); the cap just needs to grow with the scenario count.
     # 50m restores ~10m headroom on the slow arm without masking a regression.
-    timeout-minutes: 50
+    #
+    # Bumped 50→75m (#1999) after #1945 (PR #1994) added the
+    # test_ws_push_apply.py push→apply→enforce scenario: that extra ~2m
+    # snapshot-restore + boot + enforcement-flip cost tipped the slower ipk
+    # arm past the cap, which was cancelled at exactly 50m twice in a row
+    # (Master Router CD runs ~00:09→00:59Z) while the apk arm fit in ~35m.
+    # ipk is consistently slower (packaging/boot overhead + shared-host load
+    # on api.lan), and the pack had already run as close as 49m44s before this
+    # scenario landed. 75m re-establishes ~20m+ headroom on the slow arm
+    # without masking a regression in the scenario pack itself.
+    timeout-minutes: 75
     # #685: matrix over ipk-23.05 and apk-25.12 router images. fail-fast=false
     # so a regression in one flavor still reports the other arm's status
     # (needed to surface apk-only regressions like the #531 history case).


sameerparekh:
## Independent PR review

Reviewed the full merge-base diff (`git diff origin/main...HEAD`) — a single-line CI config change (`timeout-minutes: 50 → 75` in `.github/workflows/e2e-vm-fake.yml`) plus an explanatory comment block.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `.github/workflows/e2e-vm-fake.yml:80` — The bump applies to the whole `e2e-fake` job, so the fast **apk** arm now also tolerates up to 75m before cancellation. That's acceptable (fail-fast=false, and a real apk regression still surfaces as a test failure, not a hang), but it does widen the window before a genuinely-hung apk run is killed. No action needed; noting for the record.

**Verification performed**
- Confirmed the cap is the *only* timeout governing both arms: the matrix (`ipk`/`apk`) shares the single job-level `timeout-minutes` (line 80); the `workflow_call` site `gate-2-router-fake-api` in `master-router.yml` sets no `timeout-minutes`; the called scripts (`e2e-vm.sh`, `scenarios_fake/conftest.py`) carry only per-fixture waits (≤180s), none of which cap the whole suite. So the bump fully unblocks both arms.
- Confirmed the root cause from the cancelled run (28266671105): ipk job 00:09:31Z→00:59:54Z = 50m23s (cancelled at cap) vs apk 23:12:23Z→23:47:25Z = 35m02s (pass). Of the ipk time, ~49m19s is the single "Run e2e fake-mode suite" pytest step (all VM setup steps ~1m total), and it was cancelled mid-pack — consistent with uniform per-scenario overhead on the slower ipk arm after #1945 added a scenario, not one hanging test. Diagnosis matches the issue; the timeout bump is the accepted unblock.

**Checklist dimensions** — 1 SSOT: N/A (no logic). 2 Testing: N/A (CI config; no test files weakened). 3 Arch invariants: N/A. 4 Wire contract: N/A. 5 Migrations: N/A. 6 Metrics: N/A. 7 Scope/hygiene: focused single-file diff, comment cites #1999/#1945, no AI mentions in title/body. 8 Security: N/A.

VERDICT: APPROVE
Minimal, well-documented CI timeout bump that unblocks Master Router CD's Gate 2 ipk arm, which was deterministically cancelled at the 50m cap after #1945 added a scenario. The bump covers both matrix arms via the single job-level cap, and no secondary timeout limits the suite. Will be validated by the post-merge Master Router CD run (both arms passing within 75m).

PR #2002, opened 2026-06-27
fix(#2001): ws apply-on-push e2e — poll/startup save clobbers sidecar-pushed snapshot

### LINKED ISSUE #2001: bug(#1945): ws apply-on-push enforcement-flip e2e (test_ws_policy_push_applies_enforcement_live) flakes — both Gate 2 arms time out waiting for wh_drop:<mac>:Paused
## Symptom
Master Router CD Gate 2 fails on **both apk AND ipk** (run 28276547036) with the same failure:
```
FAILED scenarios_fake/test_ws_push_apply.py::test_ws_policy_push_applies_enforcement_live
  - TimeoutError: timed out waiting for nft gains a wh_drop:<mac>:Paused forward-drop via the ws push with the poll frozen — live apply-on-push (#1945)
```
Both arms, identical timeout. This is the new enforcement-flip e2e that #1945 (PR #1994) added. It is NOT the known cname/egress flake (that failed only on apk this run and passed/xfailed on ipk) and NOT a timeout-cap cancel (#2000 already bumped that 50→75; the job completed and the TEST failed). The internal path under test is API ws-push → agent apply-on-push → nft drop, with the HTTP poll frozen — no external egress involved.

## Why it's likely real (intermittent), not pure flake
- It failed on BOTH matrix arms with the same assertion (a one-arm egress flake wouldn't).
- It PASSED on the immediately prior run (28266671105, Gate 2 apk success), so the feature/test isn't categorically broken — it's intermittent.
- This run was abnormally slow (apk ~38min, ipk ~58min vs ~35min normal), which *could* mean the test's wait budget is marginal under VM load — OR there's a genuine race where the apply-on-push tick occasionally doesn't produce the drop.

## Investigate (determine which)
1. Reproduce: re-run Gate 2 a couple times and/or run the scenario locally on a real Lua 5.1 target. Does `test_ws_policy_push_applies_enforcement_live` fail repeatably or only under load?
2. **Feature race vs test fragility**: with the poll frozen, confirm the agent's apply-on-push tick (#1945, `apply_and_publish`/etag-dedup) reliably fires on the ws-pushed snapshot and produces the `wh_drop:<mac>:Paused` forward-drop. Check for: etag-dedup wrongly skipping the apply; the apply tick cadence being slower than the test's wait; a missed wakeup so the drop only lands on the next poll (which is frozen → never).
3. Fix accordingly:
   - If real race/latency in apply-on-push → fix the agent path (this is the #1945 feature working reliably, the whole point).
   - If apply is correct but the test's wait is too tight for a loaded VM → widen the wait budget / poll the nft set with backoff (keep it meaningful).
4. **This also covers #1945's still-outstanding openwrt.lan hardware validation** — validate the push→apply→enforce flip on the real openwrt.lan test router against api.lan, and confirm the go/no-go for enabling ws on the prod router.

## Definition of done
- `test_ws_policy_push_applies_enforcement_live` passes reliably on BOTH Gate 2 arms (Master Router CD green), with the root cause (race-fixed vs wait-widened) documented.
- openwrt.lan hardware validation of apply-on-push done + prod-enablement go/no-go delivered.

Follows #1945/#1994. Distinct from #1999/#2000 (timeout cap, fixed) and the cname egress flake.

sameerparekh:
## Summary
Fixes #2001 — `test_ws_policy_push_applies_enforcement_live` (#1945) intermittently timed out on **both Gate 2 arms across runs** waiting for the per-MAC `wh_drop:<mac>:Paused` forward-drop. This was the only thing keeping Master Router CD red.

## Root cause — a real apply-path race, not test fragility
Diagnosed from the failing run's captured agent log (run 28276547036, ipk):

```
ws: connected
wifihaven-ws: persisted pushed policy snapshot etag="...enforce-base-v1"   (on-connect push)
wifihaven-ws: persisted pushed policy snapshot etag="...enforce-paused-v2" (push-on-change → disk=PAUSED)
startup: applied initial policy etag="...enforce-base-v1"                   (startup save writes BASE over PAUSED)
```

When ws is the IN channel, **two writers touch `/etc/wifihaven/policy.json`**: the agent's poll/startup `save_snapshot` and the ws sidecar's atomic push-persist. The sidecar wrote BASE then PAUSED early, but the agent's still-in-flight startup fetch-apply (slow under load) then wrote **BASE over PAUSED**. With the poll frozen at 3600s, the apply-on-push tick forever saw `disk-etag == current_etag (BASE)` → never installed the drop → 180s timeout.

It's intermittent because under load the startup save lands *after* the push; on a fast VM it lands before (prior run passed). The test's `wait disk==BASE` is satisfied by the sidecar's *on-connect* BASE write, which opens the race window.

## Fix
**Agent (root cause):** `policy.save_snapshot` gains an optional compare-and-swap guard. With a `read_fn` supplied it re-reads the on-disk etag and **skips the write** (keeping the fresher file) iff that etag differs from both the etag we fetched against *and* the one we're about to write. Poll + startup call sites pass the etag they fetched against (`prev_etag` / `startup_disk_etag`). The apply-on-push tick (which never saves) then re-applies the preserved push. No behaviour change for ws-off / legacy callers (no `read_fn` → unconditional write).

**Defense-in-depth (e2e):** the conntrack `on_tick` nudge fires a small burst so conntrack's block-buffered stdout flushes promptly under VM load (no `stdbuf` on OpenWRT); the fake-scenario failure capture now dumps the on-disk `policy.json` etag + live `wh_drop` rules — the two facts whose absence made this hard to diagnose.

## TDD + validation
- New busted CAS cases in `policy_spec.lua` (red → green). Full openwrt busted: **888/0**.
- Validated on **real Lua 5.1.5** (openwrt.lan, per repo convention): agent + `policy.lua` `loadfile` OK; guard KAT + real-filesystem clobber-prevention pass; apply-on-push enforcement flip still ~4s with **0 misses across 11 base→paused cycles**.

## Notes
- Hardware (#1945's outstanding item): push→apply→enforce flip demonstrated live on openwrt.lan against api.lan with the poll frozen. Prod ws-enablement go/no-go in the issue thread.
- CI: the touched e2e + agent are on the Master Router CD push-path; `e2e-vm-fake.yml` exercises both arms on merge.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/policy.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/policy.lua
@@ -364,14 +364,39 @@ end
 
 local SNAPSHOT_PATH = "/etc/wifihaven/policy.json"
 
--- save_snapshot(snap, write_fn, rename_fn) → bool
+-- save_snapshot(snap, write_fn, rename_fn, log[, read_fn, expect_etag]) → bool
 --   write_fn(path, content)  → ok, err
 --   rename_fn(from_path, to) → ok[, err]
+--   read_fn(path)            → content|nil  (optional — enables the #2001 guard)
+--   expect_etag              → string|nil   (the etag we fetched against)
 -- Atomic: writes to <path>.tmp then renames over <path>. Skips the rename
 -- if the write fails, so a torn or partial write never replaces a good
 -- on-disk snapshot.
-function M.save_snapshot(snap, write_fn, rename_fn, log)
+--
+-- #2001 compare-and-swap guard: when ws push is the IN channel (#1849/#1945)
+-- the sidecar ALSO writes policy.json. A poll/startup save that fetched an
+-- OLDER snapshot must not clobber a NEWER one the sidecar persisted while the
+-- (slow, under-load) apply was in flight — otherwise the frozen-poll agent
+-- never re-applies the push and enforcement silently reverts. When a `read_fn`
+-- is supplied we re-read the on-disk etag just before the rename and SKIP the
+-- write (returning true — the fresher file is intentionally kept) iff the
+-- on-disk etag is a different, non-nil value than both `expect_etag` (what we
+-- fetched against) and the snapshot we're about to write. With no `read_fn`
+-- the behaviour is the unconditional write legacy callers/tests expect.
+function M.save_snapshot(snap, write_fn, rename_fn, log, read_fn, expect_etag)
   log = log or default_log()
+  if read_fn then
+    local on_disk = M.load_snapshot(read_fn, log)
+    local disk_etag = on_disk and on_disk.etag or nil
+    if disk_etag ~= nil
+       and disk_etag ~= expect_etag
+       and disk_etag ~= (snap and snap.etag) then
+      log.info("policy.save_snapshot: on-disk etag=%s differs from expected=%s "
+               .. "and from to-write=%s — keeping the fresher (ws-pushed) snapshot (#2001)",
+               tostring(disk_etag), tostring(expect_etag), tostring(snap and snap.etag))
+      return true
+    end
+  end
   local jsonc = require("luci.jsonc")
   local body = jsonc.stringify(snap)
   local tmp = SNAPSHOT_PATH .. ".tmp"
--- a/openwrt/files/usr/sbin/wifihaven-agent
+++ b/openwrt/files/usr/sbin/wifihaven-agent
@@ -792,9 +792,14 @@ end
 -- so enforcement resumes within seconds instead of waiting on the first
 -- successful poll. The boot-time default-deny skeleton (#308) stays in
 -- effect until apply() succeeds.
+-- #2001: the etag the agent last observed on flash, so the startup save below
+-- can detect a concurrent ws-sidecar push that arrived during the (slow) apply
+-- and avoid clobbering it. nil when there is no cached snapshot.
+local startup_disk_etag = nil
 do
   local cached = policy.load_snapshot(read_file, log)
   if cached then
+    startup_disk_etag = cached.etag
     local t0 = clock.monotonic_seconds()
     local applied = policy.apply(cached, write_file, run_cmd, log,
                                  { on_apply = record_apply("boot") })
@@ -836,7 +841,9 @@ do
       ts.current_etag = etag
       ts.last_snapshot = snap
       policy.mark_poll_success(os.time())
-      policy.save_snapshot(snap, write_file, rename_file, log)
+      -- #2001: don't clobber a fresher snapshot the ws sidecar persisted while
+      -- this initial apply was in flight (we fetched against `startup_disk_etag`).
+      policy.save_snapshot(snap, write_file, rename_file, log, read_file, startup_disk_etag)
       log.info("startup: applied initial policy etag=%s", tostring(etag))
     end
   else
@@ -1030,6 +1037,9 @@ conntrack.watch({
       metrics.observe(mreg, "snapshot_poll_duration_seconds", {},
                       clock.monotonic_seconds() - pt0)
       local fetch_ok
+      -- #2001: the etag we fetched against (If-None-Match); the save below uses
+      -- it to detect a concurrent ws-sidecar push and avoid clobbering it.
+      local prev_etag = ts.current_etag
       if snap then
         record_poll("200")
         -- 200: fetch + cache blocklists, attach hosts to snapshot for render.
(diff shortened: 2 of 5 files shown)


sameerparekh:

Read-only adversarial pass over `git diff origin/main...HEAD` (5 files: agent `save_snapshot` CAS guard + two call sites, `policy_spec.lua` cases, `conftest.py` failure capture, `test_ws_push_apply.py` nudge). Reviewed against all 8 checklist dimensions.

**BLOCKERS**
(none)

**SHOULD-FIX**
1. `scripts/e2e/scenarios_fake/conftest.py:323` — the diagnostic etag grep `'"etag":"[^"]*"'` truncates at the **escaped inner quote**: on disk the value is `"etag":"\"sha256:…\""` (an HTTP etag string carrying literal quotes), so `[^"]*` stops after `\` and only emits `"etag":"\"` — it never shows the actual BASE-vs-PAUSED value the capture exists to reveal. Use `[^,]*` (match to the field delimiter) or dump `head -c` of the file.
2. Process/TDD — the failing test and the implementation landed in **one commit** (`2158e10b`), so the red→green progression isn't visible in PR history as the autonomous-session TDD rule (checklist §2) wants. The red was demonstrated in-session (2 fail → 60 pass) and is described in the PR, but it should be split into a failing-test commit + an impl commit.

**NITS**
1. `policy.lua:373` — the CAS read+parse runs `load_snapshot` on every poll-path save (~60s) and once at startup; negligible vs the existing 2s ws-apply disk read, and harmless. No change needed.
2. `policy.lua` guard has a sub-millisecond TOCTOU between the disk re-read and the rename; documented in the PR. The prod poll (unfrozen) and the ws-apply tick converge if it ever loses that race, so acceptable — not worth a lock.
3. Optional: an agent `mreg` counter for "save skipped (concurrent push)" would aid fleet observability, but the `log.info` is adequate for this rare safety branch.

**Dimension notes (no findings):** SSoT — the etag comparison lives in one place; the guard (write-safety) and the ws-apply tick (apply dedup) are complementary, not duplicated. Router-dumb-applier — the guard is a file-write safety check, adds **no** policy decision logic. Wire contract — `save_snapshot`'s new params are internal Lua; on-disk JSON shape and the router↔API wire are unchanged; legacy/ws-off callers keep the unconditional-write path. Tmp-bounded — no new `/tmp` writer (policy.json is atomic whole-file on flash). Testing — 6 cases incl. the 2 clobber-skip (the bug, confirmed red without the fix) + write/idempotent/legacy/fresh edges; no mocks-of-record; no weakened existing assertions. Security — no secrets, no SQL, the test shell interpolates the router's own UCI `api_url`. Scope — focused, on-topic, no drive-bys.

**VERDICT: APPROVE**

Root cause is correctly identified and fixed: the poll/startup `save_snapshot` could clobber a fresher ws-sidecar-pushed snapshot, leaving a frozen-poll agent stuck on the stale etag — proven from the failing run's agent log and reproduced + fixed on real Lua 5.1.5. No BLOCKERs. Two SHOULD-FIX (a truncating diagnostic grep; bundle the TDD commit) are non-gating; recommend addressing before merge.

sameerparekh:
## Independent PR review — re-run (push since `2158e10b`)

The push restructured the single commit into a red `test(#2001)` commit (`e5fae60f`) + a green `fix(#2001)` commit (`6fcc2467`) and changed one diagnostic line. TREE delta vs the prior review = `conftest.py` only (4 +/1 -).

### Prior-findings status

|---|---|---|
| SHOULD-FIX 1 | `conftest.py` etag grep `[^"]*` truncated at the escaped inner quote | **ADDRESSED** — now `grep -o '"etag":[^,]*'`; verified on the live router it emits the full `"etag":"\"sha256:…\""` value. |
| SHOULD-FIX 2 | red→green not visible in history (single bundled commit) | **ADDRESSED** — split into `e5fae60f` (failing CAS cases) → `6fcc2467` (guard impl). The 2 clobber-skip cases are red at `e5fae60f` (policy.lua unchanged there), green at `6fcc2467`. |
| NIT 1 | CAS read+parse per save | unchanged — accepted (negligible). |
| NIT 2 | sub-ms TOCTOU | unchanged — accepted (documented; prod poll + ws-apply tick converge). |
| NIT 3 | optional save-skip counter | unchanged — accepted (`log.info` adequate). |

### New findings in the delta
(none) — the `conftest.py` change is a pure diagnostic-accuracy fix; the added comment is correct and the new pattern matches to the field delimiter as intended. No new BLOCKER / SHOULD-FIX / NIT.

**VERDICT: APPROVE**

Both prior SHOULD-FIX are resolved and the delta introduces no new issues. Root-cause fix (poll/startup `save_snapshot` no longer clobbers a fresher ws-pushed snapshot) stands, validated on real Lua 5.1.5. No open BLOCKERs — ready for operator merge-when-ready.

Issue #2005, opened 2026-06-27
feat(#1903): block Cloudflare IPv6 DoH on :443 (hardcoded-IP DoH bypass) under blockEncryptedDns


sameerparekh:
Parent epic: #1903 (filter-bypass prevention). Related: #1911 (blockEncryptedDns agent enforcement), #1906 (blockIpOnly backstop).

## Problem
A device hardcoding Cloudflare's resolver IPs and doing **DoH over :443** bypasses our filtering and shows as unattributed traffic. Observed in prod: IPv6 `2606:4700:4700::1001` (= Cloudflare `1.0.0.1`) showing as IPv6 traffic not assigned to any host.

The curated resolver set in `openwrt/files/usr/lib/lua/wifihaven/encrypted_dns.lua` already includes the Cloudflare v4 **and** v6 IPs, and `nft_rules()` emits drops for them — but **port 53 only** (plus DoT tcp/853 any-IP). DoH-by-IP on **:443** is not caught:
- The by-name NODATA list (`M.HOSTS`, `cloudflare-dns.com`) doesn't help a hardcoded IP.
- `blockIpOnly` (#1906) is the only current backstop, and it's a separate, broader toggle.

## Connectivity-check analysis (why :443 is safe to block here)
The resolver rules were kept :53-scoped because `1.1.1.1`/`8.8.8.8` double as common "am I online" targets. That rationale does **not** extend to :443:
- Captive-portal / NCSI detection (Apple/Android/Windows) uses dedicated **hostnames**, not Cloudflare resolver IPs.
- Devices that probe `1.1.1.1` for reachability use **ICMP / :53 / :80**, never :443.
- **DoH is essentially the only thing on :443 to a resolver IP** — i.e. the bypass itself, not collateral.
- For **IPv6** specifically, generic connectivity checks don't target the resolver's v6 literal at all, so a v6 :443 drop is near-pure signal.

## Decision (operator, 2026-06-27)
Scope to **Cloudflare IPv6 :443 only** — the observed leak, lowest possible collateral. Leave v4 :443 untouched for now (Cloudflare serves a real HTTPS site on `1.1.1.1`, a small residual risk if a niche client HTTPS-probes it; v4 can be a follow-up if wanted). Google/Quad9 untouched on :443 (connectivity-check targets).

## Scope
[...]

Issue #1903, opened 2026-06-23
epic: filter-bypass prevention (encrypted DNS, relays, VPNs, proxies)


sameerparekh:
## Umbrella: per-profile filter-bypass prevention

Split out of #1891. A device that bypasses the LAN resolver / tunnels its traffic defeats both **hostname attribution** and **filtering/enforcement**. iCloud Private Relay is one instance; there are several classes. This issue tracks the family and the cross-cutting design decisions; the concrete work lands in the child issues below.

### Why this matters (root cause from #1891)
The kid iPads run iCloud Private Relay, so Safari tunnels everything through `mask.icloud.com`. Under a whole-MAC block the relay ingress is dropped and the app hangs even though the allowed app's hosts are carved into `extraAllowed` — because the device never makes a *direct* connection to the allowed host (it tunnels through the relay). More broadly, whenever a profile is *not* fully blocked, an active relay/VPN/DoH lets the device bypass all WifiHaven filtering. See the #1891 investigation comment for the prod evidence.

### Taxonomy of bypass channels
1. **Encrypted DNS** — DoH (TCP/UDP 443 to public resolvers) and DoT (TCP 853). Cloudflare `1.1.1.1`/`cloudflare-dns.com`, Google `dns.google`/`8.8.8.8`, Quad9, NextDNS, AdGuard, OpenDNS, plus browser built-in DoH (Chrome/Firefox). (Existing issue: #572.)
2. **Relays** — Apple iCloud Private Relay (`mask.icloud.com`, `mask-h2.icloud.com`); Cloudflare WARP / 1.1.1.1 app.
3. **Consumer VPNs** — NordVPN, ExpressVPN, Mullvad, Proton, Surfshark, Google One VPN, Mozilla VPN, etc.
4. **Tor / public proxies / circumvention** — Tor directory authorities & bridges, Psiphon, public web proxies.
5. **Hardcoded-IP escape** — any of the above pinned to an IP, defeating DNS-based blocking. Backstopped by `blockIpOnly` (#425/#384/#574).

### Cross-cutting design decisions (resolve here before child impl)
[...]

PR #2008, opened 2026-06-27
feat(#2006): IPv6 support in static_ip_labels + Cloudflare/Google v6 DNS labels

### LINKED ISSUE #2006: feat(#1903): IPv6 support in static_ip_labels + Cloudflare/Google v6 DNS labels (better ip-only attribution)
Parent epic: #1903 (filter-bypass prevention). Sibling of #2005 (which BLOCKS Cloudflare v6 DoH). This issue is **attribution only** — it does NOT block anything.

## Problem
Unattributed ip-only traffic to IPv6 resolver IPs (observed: `2606:4700:4700::1001` = Cloudflare `1.0.0.1`) shows in the dashboard as a **bare IP literal** instead of a friendly label.

`openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua` is the manual last-resort IP-range → label map (the `17.0.0.0/8` → `apple-push` table, plus `1.1.1.1/1.0.0.1` → `cloudflare-dns`, `8.8.8.8/8.8.4.4` → `google-dns`). It already labels the **v4** Cloudflare DNS IPs — but the module is **v4-only**: `M.lookup()` explicitly punts on any colon/v6 literal (`if ip:find(":") then return nil`). The header even documents the intended extension: "To add a v6 entry later, extend `M._ranges` with `family = "v6"` rows and add an ipv6_to_uint128 path."

The call site already feeds v6 in: `conntrack.lua` `handle_flow` calls `static_ip_labels.lookup(flow.dst_ip)` for every flow regardless of family (conntrack.lua:779). So adding v6 support to the module makes v6 flows labelled automatically — **no call-site change needed**.

## Scope (all in static_ip_labels.lua + its spec)
1. Add IPv6 parsing + prefix matching to the module. Represent a v6 address as its 8×16-bit hextets (expand `::` compression) — each hextet ≤ 65535 fits a Lua 5.1 double, so NO 128-bit/bit32 dependency. Match a prefix by comparing the first `floor(prefix/16)` full hextets plus a modulo-masked partial hextet (≤16 bits, safe in 5.1). Host `/128` entries are the common case. Mirror the v4 "fail loud at module load on malformed/`/0`" behaviour.
2. Add v6 entries to `M._ranges` (with `family="v6"` or a parallel `M._ranges_v6`, whatever's cleanest given the existing v4 compile step):
   - `2606:4700:4700::1111/128`, `2606:4700:4700::1001/128` → `cloudflare-dns`
   - `2001:4860:4860::8888/128`, `2001:4860:4860::8844/128` → `google-dns` (symmetry with the existing v4 google-dns entries)
   Cite each in a comment beside the entry, per the module's existing convention.
3. `M.lookup` returns `(label, M.SOURCE)` for matching v6 IPs exactly as it does for v4 — same `source = "static-ip-range"` wire string, same HostId.Label shape on the API side.

## Architecture guardrails (from the module header — DO NOT violate)
- This map is **ATTRIBUTION ONLY** and MUST NOT participate in any drop predicate or carve-out. Do not touch enforcement. `match_hname` stays nil for label-typed attributions (existing conntrack.lua logic already handles this for v6 once lookup returns a label).
- Keep the map small and well-known-owner-only.

## Tests
- `openwrt/test/static_ip_labels_spec.lua`: add cases — a v6 IP inside each new entry returns the right `(label, "static-ip-range")`; a v6 IP NOT in the map returns nil; `::`-compressed and fully-expanded forms of the same address both match; a malformed v6 literal returns nil.
- **Real Lua 5.1 validation is mandatory** — the hextet math is exactly where 5.1 differs from macOS lua. Validate on openwrt.lan / plex.lan (5.1.5): module loads and `lookup("2606:4700:4700::1001")` returns `cloudflare-dns`.

## Out of scope
- Apple v6 push ranges (e.g. 2620:149::/32) — larger/messier; defer to a follow-up with prod evidence.
- Any enforcement/blocking (that's #2005).

sameerparekh:
Closes #2006. Parent epic: #1903 (filter-bypass prevention). **Attribution only — adds no blocking** (that's #2005).

## What
`static_ip_labels.lua` is the last-resort IP-range → label map (attribution only; MUST NOT participate in any drop/carve-out). It already labels the v4 Cloudflare/Google DNS IPs but was v4-only — `M.lookup()` punted on any colon/v6 literal, so v6 resolver flows with no SNI/DNS showed as bare IP literals in the dashboard.

This adds a v6 parse + prefix-match path to the module. No call-site change: `conntrack.lua handle_flow` already calls `static_ip_labels.lookup(flow.dst_ip)` for every flow including v6.

## How
- v6 address held as **8 × 16-bit hextets**, each ≤ 65535 → fits a Lua 5.1 double. **No 128-bit int / bit32 dependency** (neither exists on 5.1).
- `::` compression expanded (leading / trailing / embedded); a second `::`, dangling colon, non-hex or >16-bit token, IPv4-embedded form, or wrong group count → nil.
- Prefix match: compare the first `floor(prefix/16)` full hextets, then the partial hextet under a modulo mask (`h - h % 2^(16-rem)`). For the `/128` host entries here it's full-hextet equality. Mirrors the v4 `/0`-guard / skip-malformed compile behaviour (valid v6 prefix 1..128).
- New `M._ranges_v6` `/128` entries, each cited beside the row:
  - `2606:4700:4700::1111`, `::1001` → `cloudflare-dns`
  - `2001:4860:4860::8888`, `::8844` → `google-dns`
- `M.lookup` returns `(label, "static-ip-range")` for v6 matches — same wire `source` string and `HostId.Label` shape as v4. v4 path unchanged.

Out of scope (per #2006): Apple v6 push ranges; any enforcement/blocking.

## Tests (TDD — red commit then green)
`openwrt/test/static_ip_labels_spec.lua`: each new v6 entry returns its `(label, static-ip-range)` pair; a v6 IP not in the map → nil; `::`-compressed and fully-expanded forms of the same address both match; malformed v6 literals → nil; v4 cases unchanged.

**Real Lua 5.1 validation (mandatory):** module loads clean on a 5.1.5 OpenWRT target and `lookup("2606:4700:4700::1001")` → `cloudflare-dns`, `lookup("2001:4860:4860::8888")` → `google-dns`, `lookup("2620:fe::fe")` → nil.

### DIFF
--- a/openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/static_ip_labels.lua
@@ -54,6 +54,25 @@ M._ranges = {
   { cidr = "1.0.0.1/32",     label = "cloudflare-dns" },
 }
 
+-- IPv6 ranges. Each entry: { cidr, label }. Host /128 entries are the common
+-- case — clients that bypass the local resolver pin these literals.
+--
+-- 2606:4700:4700::1111, ::1001 — Cloudflare Public DNS (the v6 siblings of
+--                      1.1.1.1 / 1.0.0.1).
+--                      <https://developers.cloudflare.com/1.1.1.1/ip-addresses/>
+-- 2001:4860:4860::8888, ::8844 — Google Public DNS (the v6 siblings of
+--                      8.8.8.8 / 8.8.4.4).
+--                      <https://developers.google.com/speed/public-dns/docs/using>
+--
+-- Apple's v6 push ranges (e.g. 2620:149::/32) are intentionally NOT included
+-- yet — larger/messier, deferred to a follow-up with prod evidence (#2006).
+M._ranges_v6 = {
+  { cidr = "2606:4700:4700::1111/128", label = "cloudflare-dns" },
+  { cidr = "2606:4700:4700::1001/128", label = "cloudflare-dns" },
+  { cidr = "2001:4860:4860::8888/128", label = "google-dns"     },
+  { cidr = "2001:4860:4860::8844/128", label = "google-dns"     },
+}
+
 -- Parse "a.b.c.d" → 32-bit unsigned integer, or nil on malformed input.
 local function ipv4_to_uint(ip)
   if type(ip) ~= "string" then return nil end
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review — #2008 (IPv6 support in static_ip_labels)

Read-only adversarial pass over `git diff origin/main...HEAD` against the 8-dimension checklist. First run.

### BLOCKERS
(none)

### SHOULD-FIX
1. `openwrt/test/static_ip_labels_spec.lua:56-62` — The "intentionally small (operator-curated)" size-pin test only guards `#labels._ranges` (v4). `M._ranges_v6` has no equivalent cap, so a future PR can bloat the v6 last-resort map past the reviewable threshold without tripping any test — exactly the bloat the v4 pin exists to prevent. Mirror the pin for `_ranges_v6` (e.g. `assert.is_true(#labels._ranges_v6 <= 10)`).

### NITS
1. `openwrt/test/static_ip_labels_spec.lua:51` — The comment "Mixed-case hextets normalise too" is inaccurate: the asserted literal `2001:4860:4860::8888` contains no a–f hex letters, so it exercises no case-folding and merely duplicates the line above. Either drop the line or assert a label whose hextets actually contain letters (none of the four curated entries do, so dropping it is cleaner). `tonumber(tok,16)` does handle case correctly regardless.

### Dimension notes
- **1 Single-source-of-truth:** v6 path is genuinely distinct parsing, not a re-derivation of an existing primitive; `M.SOURCE` is reused (one wire string). The structural parallelism with the v4 compile/match loops is inherent to two address families, not drift-prone duplicated decision logic. OK.
- **2 Testing:** TDD red→green visible (test commit `c4c07c0e` precedes impl `fe20c5df`). The one edited existing assertion (`lookup("2001:4860:4860::8888")` was asserted nil) is correctly changed — that IP is now genuinely a `google-dns` match, so the old assertion became wrong; this is not regression-hiding. Compressed/expanded equivalence, outside-range, and malformed-literal cases all asserted. Validated on real Lua 5.1.5 per PR body.
- **3 Architecture:** Attribution-only; no enforcement / dnsmasq / render / conntrack touched. No `/tmp` writes. Hextet math uses doubles (no bit32/128-bit dep) — correct for Lua 5.1. OK.
- **4 Wire contract:** No wire shape change — reuses `source="static-ip-range"` / `HostId.Label`. Additive. OK.
- **5 Migrations:** none.
- **6 Metrics:** extends an existing lookup feeding existing connection_event emission; not a new route/job/ingest path — no new metric warranted. OK.
- **7 Scope/hygiene:** focused 2-file diff; no dead code; no AI mentions in title/body (Co-Authored-By trailer fine). OK.
- **8 Security:** no secrets, SQL, or routes. OK.

VERDICT: APPROVE
IPv6 attribution added cleanly to the last-resort label map with correct Lua-5.1-safe hextet math, validated on a real 5.1.5 target, behind a red→green TDD pair. No blockers. Two non-gating items: add a size-pin for `_ranges_v6` to preserve the curation guard, and fix/drop the inaccurate "mixed-case" test comment.

sameerparekh:
## Independent PR review — #2008 (re-run @ 81750f8)

Re-run on push `fe20c5d → 81750f8`. Statused prior findings, then reviewed the incremental delta.

### Prior findings status
|---|---|---|
| SHOULD-FIX 1 | `_ranges_v6` had no size-pin (bloat guard) | **ADDRESSED** — `static_ip_labels_spec.lua:88-90` now pins `#labels._ranges_v6 <= 10` mirroring the v4 guard. |
| NIT 1 | Inaccurate "mixed-case" test line (all-numeric literal) | **ADDRESSED** — the misleading duplicate assertion at the old `:51` was removed. |

### New findings (delta fe20c5d...81750f8)
BLOCKERS: (none)
SHOULD-FIX: (none)
NITS: (none)

The delta is exactly the two review fixes — a removed redundant assertion and an added `_ranges_v6` curation pin. Both are test-only, both pass locally (11/11) and on the real Lua 5.1.5 target; no enforcement/wire/architecture surface touched.

VERDICT: APPROVE
Both prior non-blocking findings ADDRESSED; no new findings. Ready for merge-when-ready (operator merge).

PR #2010, opened 2026-06-27
feat: add A-Z Animals app template

sameerparekh:
Adds the **a-z-animals.com** app to the app catalog per operator request.

## Change
- New template `api/resources/app_templates/a-z-animals.yml` — apex-only host-set (`a-z-animals.com`, covers all subdomains via HostMatch), favicon icon (`icon_type: url`).
- Registered the `a-z-animals` slug in `_index.yml`.
- Updated the `AppTemplatesSpec` catalog fixture to include the new slug.

## Notes
- Apex-only per the README's conservative-hosts + shared-pool collateral guidance — no off-domain CDN/vendor host needed.
- `AppTemplatesSpec` passes (37 tests): index↔file sync, slug format, uniqueness, distinctive-host rule, shared-host global consistency, favicon-not-emoji.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2004, opened 2026-06-27
bug(prod): SPA dashboard LIVE BANDWIDTH stuck at 0 B/s — trafficUsage ws push never delivers despite active traffic


sameerparekh:
## Symptom (prod, observed live in browser 2026-06-27 ~18:15Z)

On `https://app.wifihaven.net/dashboard` the **LIVE BANDWIDTH** panel shows `Overall ↓ 0 B/s · ↑ 0 B/s — "No live traffic right now"` even though there is heavy real traffic. (`wifihaven.net` now serves the marketing site post-#1832; the SPA is on `app.wifihaven.net`.)

## Evidence

Driven from the page context against `wss://api.wifihaven.net/api/ws` (operator JWT, exact dashboard subscription params):

- **Socket health is fine** — upgrade → `ready`; `subscribe` for `now`, `connectionEvents{blocked:true}`, `trafficUsage{groupBy:['profile'],bucket:'1m'}`, `timeStatus` all return `ack:ok`. The "Live" badge is genuinely green.
- **60-second frame capture while ~206 MB of real traffic flowed and blocks fired every ~1s:**

  | topic | pushes in 60s |
  |---|---|
  | `connectionEvents` | 2 (t=13s, 30s) |
  | `now` | 2 (t=14s, 31s) — only ever **paired with** connectionEvents |
  | **`trafficUsage`** | **0** |
  | `timeStatus` | 0 *(S6a not deployed — expected)* |

- **Ground truth — traffic IS being ingested:** `GET /api/usage/traffic?bucket=1m&groupBy=profile&from=…&to=…` (200) returns fresh `aggregateRows` for the 18:09 / 18:10 windows totaling **~206 MB in / ~14.6 MB out** in the last 10 min (one profile pulled 203 MB at 18:09). So `traffic_reports` is populating; the live-edge push is not delivering it.
- **Not an auth/Origin/cookie problem:** `WIFIHAVEN_WS_ALLOWED_ORIGINS` includes the apex; a no-cookie upgrade returns `401 missing wh_ws cookie` (route deployed, Origin accepted).
- **Code is deployed:** live prod API deploy finished 2026-06-27T03:23Z, after the S4 (#1997) and S5 (#1998) merges.
- **`pushTrafficUsage` is not throwing:** no `"spa ws push: trafficUsage recompute failed"` ERROR in 25 min of prod logs (`SpaPush.scala:166` would log it).

[...]

Issue #1974, opened 2026-06-25
ws-spa S6a: live time-usage push (timeStatus + appUsage) — retire the adaptive refetch ladder


sameerparekh:
**SPA-websocket implementation — step S6a of the §9 rollout** (`docs/design/spa-websocket.md` §1.2/§3.1/§5.2, design #1860; umbrella #1955). **Additive.**

**Live time-usage push** — lights up the live screen-time surface (per-profile + per-app, /profiles), collapsing the hand-rolled adaptive refetch ladder into a push.

### Scope (§9 S6a)
- **`timeStatus`** class-(2) push (per-profile used/remaining, `/api/time/status` body `ProfileTimeStatus[]`) via `TimeStatusService.dayStateAllLive` — wired to usage-credit / the **#1849 time-boundary ticker** (schedule boundary, cap exhaustion change remaining without new usage) / **`POST /api/time/extend` grant** (§5.2).
- **`appUsage`** class-(2) push (per-app minutes, `/api/profiles/{id}/usage-by-app`) via `Presence.appSecondsForProfile`, gated on `appUsage{profileId}` subscription.
- **SPA** patches the time-status + per-app caches (§3.1); the `appUsage` push targets only the **live (today) window** key — past windows are immutable. Pause the **`TIME_STATUS_REFETCH_LADDER`** when `wsLive` (don't delete it — S7).
- `spa_ws_push_total{op=timeStatus|appUsage}` coverage.

### Acceptance
- Crediting usage / a #1849 boundary tick / an extension grant pushes updated `timeStatus`; expanding a profile card subscribes `appUsage` and live-updates per-app minutes; the adaptive ladder goes dormant while `wsLive`.

Depends on: **S5 (#https://github.com/wifihaven/wifihaven/issues/1973)** (client) + **S3/S4 infra**. Relates to #1860, #1955.

sameerparekh:
Paused behind prod incident #2004 (LIVE BANDWIDTH trafficUsage push dead) — the orchestrator is shipping the #2004 fix first, then resuming S6a. One chip in flight at a time.

Issue #946, opened 2026-05-23
bug(web): +Time extension doesn't update UI — progress bar + 'time exceeded' still show original cap


sameerparekh:
## Bug

After adding screen time to a profile via the `+Time` action (header on the screen-time page, per PR #855), the UI still shows "time exceeded" and the progress bar appears to reflect the **original** daily limit instead of the extended one.

Expected: extending time on an exceeded profile should immediately move the progress bar to show the new cap, drop the "time exceeded" state, and reflect remaining time accordingly.

## Suspected location

- The "time exceeded" badge / banner on the screen-time profile card.
- The progress bar component that renders used-vs-cap for the day.
- Either the SPA isn't re-reading after `+Time` POSTs, OR the API response (`/api/time/status` or the per-profile time-status endpoint) is returning a cached cap that doesn't reflect the extension.

## Repro

1. Pick a profile near or over its daily limit (kid profile after school hours is easiest).
2. Use `+Time` to add (e.g.) 15 minutes.
3. Refresh or just observe the screen-time page UI: the "time exceeded" badge persists and the bar appears to still represent the original cap, not original + 15m.

## Where to look

- `web/src/pages/TimePage.tsx` (or wherever the screen-time page lives post-#777 / #832 collapse-profiles refactor).
- The `+Time` mutation handler — does it invalidate the relevant query cache (React Query / SWR / however the SPA does data fetching) after POSTing?
- The progress bar component — does it consume `cap` or compute from `remaining + used` (and is `cap` stale)?
- API side: `/api/time/status` — does it sum daily cap + extensions for the active day, or does it only read the static profile cap?
- The extension persistence layer — extensions probably live in a separate table (`time_extensions` or similar) keyed by `(profile_id, date)`. Confirm the time-status endpoint joins this in.

## Severity

Affects core "operator extends a kid's time" workflow. Currently the operator has no visual confirmation the extension worked. Workaround: navigate away and back, full reload, or trust the backend silently applied it.

## Acceptance

- Adding `+Time` to a profile immediately updates: the progress bar's max, the used-vs-remaining numbers, and removes the "exceeded" state if the new cap exceeds used.
- A regression test pins the behavior (component test that mocks the mutation + asserts the rendered cap updates).

PR #2014, opened 2026-06-27
fix(#2012): single-flight TimeStatusCache to stop /api/time/status CD timeout

### LINKED ISSUE #2012: incident: GET /api/time/status times out on staging since #1974 — Gate 3b red on main (3 consecutive runs)
## Symptom
**Master API/UI CD has been RED on `main` since #1974 merged (19:00, 2026-06-27)** — 3 consecutive failed runs, all on **Gate 3b (both apk + ipk arms)**:
- #1974 0fb25e04 (19:00) — failure
- #2004 eb5ede96 (21:14) — failure
- #2010 405eb0ec (22:03) — failure
Last green: #2001 993bd9cb (18:25), the run BEFORE #1974.

Gate 3b `gate3/test_smoke.py::test_gate3_smoke` fails with `TimeoutError: The read operation timed out`. The failing run proves the staging API is **up and serving** — earlier probes in the same test pass:
- block page hit for googlevideo.com ✓
- upstream OK for duolingo.com ✓
- `9 /api/logs rows attributed` ✓

The timeout fires on the **very next probe — the `/api/time/status` cross-check** (`admin.time_status()` at test_smoke.py:~92-97, right after the `/api/logs` assertion). So it's endpoint-specific: `/api/time/status` hangs/too-slow while other endpoints respond.

## Likely cause — #1974 (PR #2003, "SPA-ws S6a — live time-usage push")
#1974 is the first red and it directly reworked the timed-out endpoint:
- Refactored the GET `/api/time/status` builder into a shared `TimeStatusService.assembleProfileTimeStatus` (Routes.scala `buildProfileTimeStatus` → calls it; also used by SpaPush). The extraction *looks* pure (same `usedSecondsByMac` / `Presence.hostMinutes` / `proportionalHostMinutes`), so the GET's own compute shouldn't have gotten heavier — but verify.
- Added `spaEventBus.publish(SpaEvent.TimeStatusChanged)` on write sites: `RouterIngestService` (every usage-credit ingest), the #1849 reevaluate ticker in `Main`, and `POST /api/time/extend`.
- SpaPush rebuilds `ProfileTimeStatus[]` via `TimeStatusService.dayStateAllLive` (subscription-gated).

Hypotheses to check (observe before theorizing — use the incident-investigation skill, code LAST):
1. **The GET path itself** is slow/hangs at staging data scale (N profiles × {dayState, listPresenceRows, listForProfile, proportionalHostMinutes}) — did #1974 change the per-profile work or the number of profiles iterated?
2. **The new publish/ticker machinery wedges the API** — e.g. `spaEventBus.publish` blocking on a bounded bus with no/slow consumer, or the reevaluate ticker now rebuilding `dayStateAllLive` too often, starving the request pool. Ingest still worked (9 rows), so a hard ingest block is less likely, but resource starvation that lands on the heaviest endpoint (time/status is the documented hot path — see TimeUsedRollupJob.scala:31) is plausible.
3. Compounding from #2004 (eb5ede96, trafficUsage live-edge) — but failures started at #1974, before #2004, so #1974 is primary.

## Required first step: REPRODUCE on staging
Probe staging `GET /api/time/status` directly (admin auth; staging API + admin creds — see operator) and time it. Confirm hang-vs-slow and whether it's all-profiles or a specific profile. Rule out pure staging-sizing (project_staging_api_oom_512mi: staging was bumped to Render `standard` — but this is endpoint-specific, not OOM/502, so a code regression is far more likely). Compare against last-green code if needed.

## Fix + guard
- Fix the regression so `/api/time/status` returns promptly.
- Add a test that would have caught this (the existing SpaWsS6aSpec covers SSOT equivalence but not GET latency/hang). A feature test asserting the GET completes on representative multi-profile/presence data, and/or that the publish path can't block the request path.

## Notes
- Fresh issue per the no-reopening-merged-issues rule (do NOT reopen #1974).
- Main is red NOW; treat as priority.

sameerparekh:
## CD incident fix — closes #2012

**Master API/UI CD has been red on `main` since #1974** (3 consecutive Gate 3b failures: #1974, #2004, #2010) with `TimeoutError: The read operation timed out` on the `/api/time/status` cross-check.

### Root cause — a concurrency cliff, **not** a #1974 GET regression
The #1974 GET refactor (extraction of `TimeStatusService.assembleProfileTimeStatus`) is **byte-identical** — verified by diff. The real defect:

`TimeStatusCache.getOrLoad` had **no single-flight** (`getIfPresent` → miss → `load` → `put`). So N concurrent missers for the same `(profileId, date)` each independently run the **sequential O(N-profile) build** (`dayState` + `listPresenceRows` + `listForProfile` per profile) and **stampede the 8-connection pool** (`WIFIHAVEN_DB_POOL_SIZE=8`, render.yaml).

After `deploy-staging`, three jobs hit the **same** staging API **in parallel** on a cold 30s cache:
- `smoke-staging`
- Gate 1 `api-smoke-staging` → `scripts/e2e-tests.sh:179` hits `/api/time/status`
- Gate 3b `gate-3b-api-staging` → `test_smoke.py:97` hits `/api/time/status`

The concurrent calls stampede and exceed their read timeouts.

### Reproduced on staging (58 profiles)
| concurrency | latency |
|---|---|
| 1 cold call | ~4.8s |
| 3 concurrent cold | ~12.7s each (at the 15s gate edge) |
| 10 concurrent cold | **all > 30s** (curl timeout) |

(`/api/logs` baseline: ~0.3s.)

### Fix
Per-key single-flight via a `ConcurrentHashMap[Key, Promise]`: the caller that wins `putIfAbsent` owns the load; every other concurrent miss **awaits that one in-flight result and does no DB work**. The slot is always freed on completion — success, failure, **or interrupt** — so a failed/interrupted load can never wedge followers or cache a bad value.

This restores the **pre-#1974 single-build latency** (which was green: one ~5s build per 30s window shared across all gate callers, vs the multiplicative stampede that blew past the timeout). No parallelism is introduced — `api/src` is deliberately all-sequential for pool discipline. The #1974 consumer push path (`SpaPush.pushTimeStatus`) is subscription-gated (no ws subscriber in CI) and is left unchanged.

### Tests (TDD)
- `TimeStatusCacheSpec` gains a single-flight guard: N concurrent `getOrLoadDaily` for one key behind a shut gate (every caller misses concurrently) must invoke `load` exactly once. Committed **red first** (`count == 25`), green after the fix.
- Re-ran `TimeStatusCacheSpec` (6), `TimeApiSpec` (52), `TimeStatusServiceSpec` (36), `SpaWsS6aSpec` (6), `RouterIngestSpec` (9) — all pass. `mill __.compile` + scalafmt clean.

### DIFF
--- a/api/src/cache/TimeStatusCache.scala
+++ b/api/src/cache/TimeStatusCache.scala
@@ -6,6 +6,7 @@ import zio.*
 
 import java.time.LocalDate
 import java.time.Duration as JDuration
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.LongAdder
 
 import com.github.benmanes.caffeine.cache.{Cache, Caffeine}
@@ -141,6 +142,16 @@ object TimeStatusCache {
     private val hits   = new LongAdder
     private val misses = new LongAdder
 
+    // #2012: per-key single-flight. Without it, concurrent missers for the SAME (profileId, date)
+    // each independently run the O(N-profile) `load` and stampede the small DB pool — which timed
+    // out `/api/time/status` when CD Gate 1 + Gate 3b hit a freshly-deployed, cold-cache staging
+    // in parallel (reproduced: 1 cold call ~5s, 3 concurrent ~13s, 10 concurrent >30s). With it,
+    // the caller that wins `putIfAbsent` runs `load`; every other concurrent caller awaits that one
+    // in-flight Promise and does no DB work. The slot is always freed on completion (success,
+    // failure, OR interrupt) so a failed/interrupted load never wedges followers or caches a bad
+    // value — followers just see the same outcome and the next caller re-loads.
+    private val inFlight = new ConcurrentHashMap[Key, Promise[Throwable, AnyRef]]()
+
     def getOrLoadDaily(
         profileId: ProfileId,
         date: LocalDate,
@@ -173,10 +184,24 @@ object TimeStatusCache {
       val cache = if isTodayMode then todayCache else pastCache
       ZIO.succeed(Option(cache.getIfPresent(key))).flatMap {
         case Some(v) =>
-          ZIO.succeed(hits.increment()) *> ZIO.succeed(v.asInstanceOf[V])
+          ZIO.succeed(hits.increment()).as(v.asInstanceOf[V])
         case None    =>
-          ZIO.succeed(misses.increment()) *>
-            load.tap(v => ZIO.succeed(cache.put(key, v)))
+          // #2012 single-flight: register an in-flight Promise for this key. The caller that wins
+          // `putIfAbsent` owns the load; concurrent missers await its result instead of stampeding
+          // the DB pool with N redundant builds.
+          Promise.make[Throwable, AnyRef].flatMap { mine =>
+            Option(inFlight.putIfAbsent(key, mine)) match {
+              case Some(existing) =>
+                existing.await.map(_.asInstanceOf[V])
+              case None           =>
+                ZIO.succeed(misses.increment()) *>
+                  load
+                    .tap(v => ZIO.succeed(cache.put(key, v)))
+                    // Always free the slot and publish the outcome to followers — even on failure
+                    // or interrupt — so the in-flight entry can never wedge waiters or leak.
+                    .onExit(exit => ZIO.succeed(inFlight.remove(key, mine)) *> mine.done(exit))
+            }
+          }
       }
     }
 
--- a/api/src/cache/TimeStatusCache.scala
+++ b/api/src/cache/TimeStatusCache.scala
@@ -192,7 +192,9 @@ object TimeStatusCache {
           Promise.make[Throwable, AnyRef].flatMap { mine =>
             Option(inFlight.putIfAbsent(key, mine)) match {
               case Some(existing) =>
-                existing.await.map(_.asInstanceOf[V])
+                // Served off the in-flight load — no DB work — so it counts as a hit, keeping the
+                // periodic hit-rate stats representative of how much load the cache actually saved.
+                ZIO.succeed(hits.increment()) *> existing.await.map(_.asInstanceOf[V])
               case None           =>
                 ZIO.succeed(misses.increment()) *>
                   load
--- a/api/test/src/feature/TimeStatusCacheSpec.scala
+++ b/api/test/src/feature/TimeStatusCacheSpec.scala
@@ -281,6 +281,43 @@ object TimeStatusCacheSpec extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostg
         finalStats.hits == statsAfterGrant.hits + 1L,
       )
     },
+    test("concurrent misses for one key collapse to a single load (single-flight, #2012)") {
+      // #2012: the `/api/time/status` page is fetched concurrently by several callers at once
+      // (e.g. CD Gate 1 + Gate 3b hitting staging in parallel right after a deploy, cold cache).
+      // Without single-flight every concurrent miss independently runs the O(58-profile) build,
+      // stampeding the small DB pool until the endpoint times out. Assert that N concurrent
+      // `getOrLoadDaily` calls for the SAME key invoke `load` exactly once; the followers await
+      // the in-flight result instead of launching their own build.
+      //
+      // Deterministic: `gate` is held until every caller has passed the cache's miss decision, so
+      // no load completes (and nothing is `put`) during the window. Without single-flight all N
+      // callers therefore miss and run `load` (count == N); with it, exactly one runs `load` and
+      // the rest await the in-flight result (count == 1). The settle below only needs to let the
+      // forked fibers reach that decision — it never races a completion, because the gate is shut.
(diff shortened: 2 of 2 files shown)


sameerparekh:
## Independent PR review — `0e12468`

Read-only adversarial pass over `git diff origin/main...HEAD` (2 files, +65/−3) against the 8 checklist dimensions.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `api/src/cache/TimeStatusCache.scala:190-203` — single-flight *followers* (the `existing.await` branch) increment neither `hits` nor `misses`, so under concurrency the `logCacheStatsPeriodically` hit-rate/total under-counts real request volume. It's a log line only (not a metric series or behavior), but worth a one-word call-out if anyone reads those stats to size the cache. Counting a follower as a hit would be defensible.
2. `api/src/cache/TimeStatusCache.scala:183-188` — tiny benign window between `getIfPresent == None` and `putIfAbsent`: a concurrent loader can complete + populate Caffeine + clear its slot in that gap, so this caller redundantly reloads once. Single redundant load, never a stampede — not worth a re-check, noting for completeness.
3. `api/src/cache/TimeStatusCache.scala` — `invalidateAll` / `invalidateProfile` don't touch `inFlight`, so an in-flight load that started before an `invalidateProfile` (e.g. `POST /api/time/extend`) still `put`s its pre-grant value after the invalidate (≤30s self-healing staleness). This race is **pre-existing** (the old `load.tap(put)` had the same overlap) and is not worsened by single-flight — if anything fewer concurrent puts now occur. The sequential #946 extend-invalidation test still passes. Flagging only so it's on record; a generation counter would be over-engineering for a 30s self-heal.
4. `api/test/...TimeStatusCacheSpec.scala:312` — `arrived.get.repeatUntil(_ == n)` busy-polls a `Ref`; resolves instantly here so fine, but a `Promise`-based barrier would be cleaner.

### Dimension notes
- **1 SSOT (highest priority):** ✅ No re-derivation. The cache stays a generic `(key, load) ⇒ value` shim; `load` is unchanged. GET and ws-push still share `TimeStatusService.assembleProfileTimeStatus` — SSOT preserved, byte-identical bodies untouched.
- **2 Testing:** ✅ TDD red→green visible across two commits (`729e6efd` failing guard with `count == 25`, `0e12468` fix → green). Real cache, no mocked repo/Clock/Auth. The guard fails without the fix (verified) and pins the exact bug (concurrent missers ⇒ one load).
- **3 Architecture:** ✅ No DNS/router/snapshot/`extraAllowed` surface touched. No `java.time` now in the cache; the test's `ZIO.sleep` is correctly gated behind `@@ TestAspect.withLiveClock`. No `/tmp` writes. `onExit` correctly frees the slot + publishes the outcome on success, failure, **and** interrupt — no leaked/wedged in-flight entry.
- **4 Wire / 5 Migrations / 8 Security:** ✅ N/A — no wire shape, no SQL, no migration, no new route/secret.
- **6 Metrics:** No new route/job/poller — single-flight is an internal optimization of an existing cache, so no new metric series is required. (See NIT 1 on the existing internal stats.)
- **7 Scope/hygiene:** ✅ Focused 2-file diff; no dead code; commits + PR title/body carry no AI mentions (Co-Authored-By trailer only); three-dot merge-base diff used.

**VERDICT: APPROVE**

Single-flight on `TimeStatusCache` eliminates the reproduced `/api/time/status` concurrency cliff (10 concurrent cold: >30s → one shared ~5s build) that the parallel CD staging gates trigger, restoring the green pre-#1974 single-build latency. Correctness of the in-flight slot (always-freed on success/failure/interrupt, no leak, no SSOT divergence) holds up; only non-blocking NITs remain.

sameerparekh:
## Independent PR review (re-run) — `dc66244`

Statused prior findings against current code and reviewed the incremental delta `0e12468...dc66244` (single commit: count single-flight followers as hits).

### Prior findings
| # | Finding | Status |
|---|---|---|
| NIT 1 | followers uncounted in cache hit/miss stats | **ADDRESSED** — `TimeStatusCache.scala:194-196` now `hits.increment()` on the `existing.await` branch |
| NIT 2 | benign `getIfPresent==None`→`putIfAbsent` redundant-load window | NOT-ADDRESSED (by design — single redundant load, never a stampede; not worth a re-check) |
| NIT 3 | `invalidate*` doesn't clear `inFlight` (≤30s self-healing staleness) | NOT-ADDRESSED (pre-existing race, not worsened; generation counter would be over-engineering) |
| NIT 4 | test `repeatUntil` busy-poll barrier | NOT-ADDRESSED (resolves instantly; cosmetic) |

### New findings (delta)
(none) — the delta is a one-line stats increment with a clarifying comment. No SSOT/test/architecture/wire/security impact; sequential stats tests unaffected (single-flight only triggers under concurrency). `TimeStatusCacheSpec` still 6/6 green, scalafmt clean.

**VERDICT: APPROVE**

The only blocking-eligible dimension (SSOT, testing integrity) remains clean; the single addressed NIT improves stats fidelity without behavior change. Remaining NITs are by-design / cosmetic and non-gating.

Issue #1971, opened 2026-06-25
ws-spa S4: trafficUsage aggregator — live-edge push of the existing GET /api/usage/traffic


sameerparekh:
**SPA-websocket implementation — step S4 of the §9 rollout** (`docs/design/spa-websocket.md` §5.3, design #1860; umbrella #1955).

The **`trafficUsage` aggregator** — live bandwidth is **not new data**: it is the existing `GET /api/usage/traffic` query re-run on new usage and pushed as `TrafficUsageResponse` (§1.3). **No `ThroughputTick`, no new read-model shape, no router change.** **Additive.**

### Scope (§9 S4)
- On usage ingest (shared ingest handler), recompute **only the current/most-recent bucket** for each distinct `(groupBy, filter)` with ≥1 subscriber (§1.4) and push that one bucket as a `TrafficUsageResponse` (live edge); the client merges the head bucket (§3.1) — never re-stream history.
- **Cadence = bucket-appropriate, coalesced; latest-wins per `(groupBy,bucket,filter)`** (§6.3) — a slow client gets the freshest head, never a backlog. No subscriber → no query.
- **Granularity:** aggregated floor `1m` (`traffic_reports` minimum) → `{1m,10m,1h,…}` free; a **`raw`-bucket** subscription pushes at the usage-ingest cadence (fully-realtime, §10 Q1) — realtime-ness rides the existing usage path, **not gated on #1023** (gets faster as #1023 streams usage).
- `spa_ws_push_total{op=trafficUsage}` coverage; verify the re-run query's plan at prod scale if the access pattern changes (docs/process/query-perf.md — likely unchanged, it's the existing query scoped to the head bucket).

### Acceptance
- New usage ingest → subscribers to `trafficUsage{groupBy,filter}` receive only the current bucket; coalescing proven (burst of ingests → one freshest push per param-set); the streamed bucket equals what `GET /api/usage/traffic` returns for that window (SSOT). `raw` subscription pushes per ingest cycle.

Depends on: **S1 (#1968)**; lands alongside S2/S3. Relates to #747, #1023, #1860, #1955.

Issue #1464, opened 2026-06-05
presence-tuning: session-stitch primitive + profile interval-union daily cap


sameerparekh:
**Profile presence fix.** Spec: `docs/design/presence-tuning.md §5` item 2.

Replace `bucketSeconds = max(activeSeconds)` and the `period_start`-grid dedup in `totalSecondsByMac`/`dedupedTotalSeconds` (`api/src/presence/Presence.scala`) with the session-stitch model from the design §4:
- per-(device,app) session stitching on the idle gap; per-device union across apps; combine across devices via the profile's existing `crossDeviceOverlapMode` (`Sum` → add per-device totals; `Dedup` → union across devices — preserve BOTH paths).
- Read window bounds from `period_start`/`period_end`; **enforce N ≥ 2×R**. Keep the `legacy` path behind `presence_model`.
- **Wire the new knobs into the `time_used_daily` invalidation hook** — changing `presence_continuation_seconds`/`presence_model` must DELETE affected cached rollups, exactly like a heartbeat-setting change.

TDD pins (per §4.4): (a) a continuous sparse session reads its full span not the floor; (b) same day re-bucketed at R=10s vs 300s yields the same minutes (rate-independence §2d); (c) two concurrent apps in one minute = one minute (union); (d) editing a presence knob invalidates `time_used_daily` and the next rollup reflects it.

Depends on #1463. Tuning epic #1445; design PR #1449.

PR #2022, opened 2026-06-28
fix(#2016): clamp delayed-flush usage window so agent stalls don't inflate usedMinutes

### LINKED ISSUE #2016: incident: time usedMinutes massively over-counted (values >1440/day; app-mins > profile-mins) — kids wrongly past limits
## INCIDENT — usage over-counting (prod, observed 2026-06-28)

**Symptom:** kids' profiles show time/app usage far past their limits. Reported case: Gimkit shows ~29 min on **Octavius** (profile 7) against a **10-min** app limit; profile usedMins 51 vs 30-min daily limit.

### SIGNAL (read-only prod GET /api/time/status, all profiles, 2026-06-27)
| Profile | usedMins / dailyLimit | Σ hostUsage.usedMins (presence-bucket) |
|---|---|---|
| Octavius | **51** / 30 | **1** |
| Quintus | 46 / 30 | 3 |
| Prima | 42 / 30 | 6 |
| Kids | 32 / 30 | 7 |
| Sameer | **2491** / ∞ | 1113 |
| Rachel | 1520 / ∞ | 403 |
| Family | 872 / ∞ | 204 |

**Two independent proofs this is over-counting, not real usage:**
1. **Sameer usedMins = 2491 min = 41.5 h in one day** — physically impossible (max 1440/day).
2. **Prima `app:math-academy` usedMins = 45 but profile usedMins = 42** — an app's minutes cannot exceed the profile total. The per-app and per-profile minute computations have **diverged into inconsistent paths** (single-source-of-truth break).

Headline `usedMins` is wildly disconnected from `hostUsage` presence-bucket minutes (Octavius 51 vs 1). Kids cross their limits purely from the inflation.

**FAILURE CLASS:** data-correctness / over-counting in `usedMinutes` (and per-app `usedMins`). NOT an enforcement failure — `/api/logs` shows Octavius traffic is currently being blocked (reason `schedule`); enforcement is acting on the inflated counters (so kids are over-blocked, failing toward more-restrictive).

### ONSET / SUSPECTS (deploy timeline)
A wave of usage/time-accounting merges landed 06-26 → 06-28:
- #1971/#2007 (#2004) — trafficUsage **live-edge** aggregator + head-bucket anchoring
- #1974 (#2003) — SPA-ws S6a: refactored time-status into shared `TimeStatusService.assembleProfileTimeStatus` + `buildUsageByApp`; **last commit to `TimeStatusService.scala`**
- #1975 (#2011) — S6b traffic/connection streaming
- #2012 (#2014) — single-flight `TimeStatusCache`

Strongest leads:
- The **app-mins > profile-mins** inconsistency points at #1974's split of per-app vs per-profile assembly (`buildUsageByApp` vs `assembleProfileTimeStatus` / `dayState`).
- The **live-edge** path (#1971/#2004) is the prime suspect for cumulative double-counting (re-adding the same minute/period on each push/ingest tick) inflating beyond 1440. Note the GET path uses `dayState` (not `dayStateAllLive`), so confirm whether `dayState` itself is inflated or whether a cache/live write-back contaminates it (#2012 `TimeStatusCache`).

### REQUIRED INVESTIGATION (observe before theorize; SSOT)
1. Reproduce against prod-shaped data: compute `usedMinutes` for a profile and compare to distinct presence minute-buckets. Find where the count multiplies (per-device sum without wall-clock dedup? per-app double-count? live-edge cumulative add? cache write-back?).
2. Bisect the wave #1971→#2014 — the window is tight. Check whether `dayState` regressed or whether the live-edge/cache path writes inflated values that the GET then reads.
3. Restore single-source-of-truth: profile usedMinutes and per-app usedMins must come from ONE presence→minutes computation; an app's minutes must be ⊆ the profile's.

### FIX + GUARD
- Fix the over-count at its single source.
- Add a test pinning invariants: `usedMinutes ≤ 1440`; `Σ/any app usedMins ≤ profile usedMins`; usedMinutes reconciles with distinct presence buckets on representative multi-device/multi-app data.

### MITIGATION
If the operator wants immediate relief while root-causing: revert the correlated live-edge/time-usage PR (candidates above; CD redeploys). Lower urgency than usual because the inflation **over-restricts** (kids blocked too much), not under-restricts — but numbers are user-visibly wrong.

Fresh issue (no reopening merged #1974/#2004/#2012 per the no-reopen rule).

sameerparekh:
Fixes #2016. **Live prod incident** — kids over-blocked off massively inflated `usedMinutes`.

## TL;DR
The over-count is **not** a regression from the 06-26→06-28 wave (#1971/#1974/#2012). `spanOf`/`effectiveGap` (the #1464 session-stitch) last changed **2026-06-05**, and #1974's GET refactor reads `state` verbatim — reverting any wave PR would not help. The defect is **upstream of the server presence model**: the agent emits a usage window that reflects flush-stall latency, not monitored time, and the (correct-by-design) server model faithfully credits it.

## Root cause (observed read-only on prod)
`GET /api/time/heartbeat-explain/<octavius-ipad>` for 2026-06-27 shows the mechanism exactly — one device, rows with `active_seconds=10–20` but `period_seconds` up to **517s**:

```
17:44:43 host=www.gimkit.com         active=10 period=517 ...
17:44:43 1ofhdhmp.gimkitconnect.com  active=20 period=517 ...   (9 hosts share this 8.6-min window)
17:53:20 2zjlthhe.gimkitconnect.com  active=20 period=466 ...
18:01:06 www.gimkit.com              active=10 period=515 ...
```

- The agent's usage flush runs inside the conntrack `on_tick` callback, which fires **only on a `conntrack -E -e NEW` line**. A device on a long-lived connection (gimkit is a websocket game; streams behave the same) emits few NEW events, so `on_tick` — and the usage flush **and** the `activity_sample_int` sampler it carries — stalls for minutes.
- `period_start = last flush wall`, `period_end = now`, so the window balloons to cover the whole unmonitored gap. `active_seconds` is capped at `usage_report_interval` and only sampled while `on_tick` ran, so it stays small.
- The server's #1464 session-stitch (`spanOf` = full `[period_start, period_end]`, `effectiveGap = max(N, 2·max period_seconds)`) **by design** credits the entire window as continuous engaged time (rate-independence; pinned by unit tests `period=300 → 5 min`). Consecutive stalled buckets are wall-clock-contiguous, so they **chain** — turning ~20s of real activity into 8.6-min spans → Octavius headline **51 min** from ~1 min of bucket presence; Sameer **2491 min/day**.

On the two "invariant" signals in the issue: `app-mins(45) > profile-mins(42)` for Prima is the **documented** `exempt_from_daily` behaviour (`AppUsedRollupService` docstring §"per-app sum ⇄ profile total"), and `profile > 1440` is legitimate under `crossDeviceOverlapMode = Sum` (per-device totals added). Neither is a new SSOT break — the real, user-visible wrongness is the window-inflation above, which the fix targets at its single source.

## Fix — faithful, agent-side (operator-chosen)
Preserves the **entire** server presence model and all its tests. `usage.bounded_period_start(last_flush, now, max_window)` clamps the reported window to `usage_max_window_seconds` (new optional UCI knob, default **2× `usage_report_interval`** = 120s) so a delayed flush credits only the intended bucket, never the un-monitored stall gap. Accumulated **bytes are still POSTed in full** (real traffic is never lost); only the presence time-window is bounded. The default keeps the normal `on_tick` jitter band (~60–100s observed) whole while bounding a real stall.

## SSOT
The bound is one pure function (`bounded_period_start`); the loop calls it. No second presence path; the server computation is untouched.

## Tests / validation (TDD)
- `openwrt/test/usage_spec.lua` — failing spec committed first (`a52264a3`), then green: timely flush kept whole, stalled flush clamped, boundary, backward-clock guard.
- Full agent busted suite green (`usage` + `contract` + `render`: 287 pass, 2 pending = nft binary absent).
- **Validated on real Lua 5.1.5 (`openwrt.lan`)** per the no-mock-5.1 practice: `wifihaven-agent` + `usage.lua` parse, helper KAT passes. (No Scala touched, so `mill`/scalafmt N/A — the investigation redirected the fix from the API to the agent.)

## Rollout note
Relief reaches prod when routers pick up the next agent release via auto-update. If immediate API-side relief is wanted for not-yet-updated agents, the read-time clamp (option B in triage) can be added as a follow-up — say the word.

### DIFF
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -223,6 +223,13 @@ object MetricGuard {
     // usage reports flowing or are they stacking up?" without grepping the ring-buffer syslog.
     "usage_queue_depth"                         -> Set("router_id", "installation_id"),
     "usage_post_total"                          -> Set("result", "router_id", "installation_id"),
+    // #2016 — count of usage flushes whose reported window was clamped because the
+    // conntrack-driven on_tick stalled (sparse NEW events on long-lived connections),
+    // so period_start..period_end would otherwise span un-monitored time the server's
+    // #1464 session-stitch credits as presence. A rising rate means stalls are common
+    // and the `usage_max_window_seconds` bound is doing real work (the #2016 over-count
+    // mitigation). Unlabelled beyond the standard router identity pair.
+    "usage_window_clamped_total"                -> Set("router_id", "installation_id"),
     // Server-side ingest health for POST /api/router/metrics (#1205). Concrete, emitted now.
     "router_metrics_batches_total"              -> Set("status"),
     // #1846 — websocket router transport (server side). `router_ws_connections_active` is the live
--- a/deploy/grafana/dashboards/router-fleet.json
+++ b/deploy/grafana/dashboards/router-fleet.json
@@ -755,6 +755,35 @@
           "refId": "A"
         }
       ]
+    },
+    {
+      "datasource": { "type": "prometheus", "uid": "${datasource}" },
+      "title": "Usage window clamped (on_tick stalls) rate",
+      "description": "sum (rate(usage_window_clamped_total[5m])) — rate of usage flushes whose reported [period_start, period_end] window was clamped to usage_max_window_seconds because the conntrack-driven on_tick stalled (#2016). on_tick — and the usage flush + activity sampler it carries — fires only on a `conntrack -E -e NEW` line, so a device on a long-lived connection (a websocket game, a stream) emits few NEW events and the flush stalls; the un-monitored gap would otherwise be credited as presence by the server's #1464 session-stitch (the #2016 over-count). A non-zero rate means the bound is doing real work; a sustained HIGH rate means on_tick stalls are common on this fleet — consider raising usage_max_window_seconds only if the stalls reflect genuinely-coarse-but-monitored reporting, else it confirms the over-count mitigation is firing as intended.",
+      "type": "timeseries",
+      "gridPos": { "h": 8, "w": 12, "x": 12, "y": 88 },
+      "id": 24,
+      "fieldConfig": {
+        "defaults": {
+          "color": { "mode": "palette-classic" },
+          "unit": "short",
+          "min": 0,
+          "custom": { "drawStyle": "line", "fillOpacity": 20, "showPoints": "never", "lineWidth": 1 }
+        },
+        "overrides": []
+      },
+      "options": {
+        "legend": { "displayMode": "table", "placement": "bottom", "calcs": ["lastNotNull", "max"] },
+        "tooltip": { "mode": "multi", "sort": "desc" }
+      },
+      "targets": [
+        {
+          "datasource": { "type": "prometheus", "uid": "${datasource}" },
+          "expr": "sum (rate(usage_window_clamped_total{env=\"$env\"}[5m]))",
+          "legendFormat": "clamped/s",
+          "refId": "A"
+        }
+      ]
     }
   ],
   "refresh": "30s",
--- a/docs/router-tuning.md
+++ b/docs/router-tuning.md
@@ -45,6 +45,27 @@ How often (seconds) the agent scrapes nftables counters and POSTs
   `usage_report_interval` evenly. With defaults (10 / 60) every report
   carries 6 samples.
 
+### `usage_max_window_seconds` (default `2 × usage_report_interval`)
+
+Caps the wall-clock window a single usage POST may report as presence
+(`#2016`). The usage flush runs inside the conntrack `on_tick` callback, which
+fires only on a `conntrack -E -e NEW` line. A device sitting on a long-lived
+connection (a websocket game, a stream) emits few NEW events, so `on_tick` — and
+both the usage flush and the `activity_sample_int` sampler it carries — can
+stall for minutes. Without this cap, `period_start..period_end` then spans the
+whole unmonitored gap, and the server's session-stitch presence model (#1464)
+credits that **entire** window as continuous engaged time — the #2016 daily
+over-count (observed: `activeSeconds≈20` but `period=517s` read as ~8.6 min).
+
+- A delayed flush is clamped to report at most this many seconds of presence;
+  the **accumulated bytes are still POSTed in full** (real traffic is never
+  lost) — only the presence time-window is bounded.
+- **Lower** → tighter bound on stall inflation; risks slightly under-counting a
+  genuinely-coarse fleet whose `on_tick` legitimately fires less often than this
+  window. **Raise** → looser bound (more of a stall counts as presence).
+- **Couples with** `usage_report_interval`: keep it ≥ the interval. The default
+  (2×) keeps the normal `on_tick` jitter band whole while bounding a real stall.
+
 ### `activity_sample_int` (default `10`)
 
 How often (seconds) the agent samples per-MAC conntrack activity to compute
@@ -92,6 +113,7 @@ Force-flush the event buffer after this many seconds even if
 |---|---|---|---|
 | `policy_poll_interval` | 5 | 3–30 | — |
 | `usage_report_interval` | 60 | 30–300 | `activity_sample_int` (must divide evenly) |
+| `usage_max_window_seconds` | 2×interval | interval–600 | `usage_report_interval` (keep ≥ it) |
 | `activity_sample_int` | 10 | 5–30 | `usage_report_interval` |
 | `event_batch_size` | 50 | 10–200 | `event_flush_interval` |
 | `event_flush_interval` | 10 | 5–60 | `event_batch_size` |
--- a/openwrt/files/etc/config/wifihaven
+++ b/openwrt/files/etc/config/wifihaven
@@ -42,6 +42,17 @@ config wifihaven 'wifihaven'
 	# How often (seconds) to POST usage counters
 	option usage_report_interval '60'
 
+	# #2016: max reported usage window (seconds). The usage flush is driven by
+	# the conntrack on_tick callback, which only fires on a `conntrack -E -e NEW`
+	# line, so a device on a long-lived connection (a websocket game, a stream)
+	# can stall the flush for minutes. period_start..period_end would then span
+	# the whole unmonitored gap, and the server's session-stitch presence model
+	# (#1464) credits that entire window as engaged time — the #2016 over-count.
+	# A delayed flush is clamped to credit at most this many seconds of presence;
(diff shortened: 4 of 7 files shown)


sameerparekh:
## Independent review — PR #2022 (`fix(#2016)`)

Read-only adversarial pass over `git diff origin/main...HEAD` (5 files, agent-side Lua + config + docs + test). First run.

**BLOCKERS**
(none)

**SHOULD-FIX**
1. `openwrt/files/usr/sbin/wifihaven-agent:1312` — the new clamp path emits no metric (checklist §6). A bounded, label-free counter (e.g. `usage_window_clamped_total`, incremented when `ps_wall > ts.last_usage_wall_ts`) would give operators visibility into how often flushes stall and whether the default 2× window actually bites — i.e. it directly validates that this incident fix is working in prod and that the knob is tuned right. Cheap on the agent via the existing `metrics.inc_counter(mreg, …)` path; a full add also wants the server-side `MetricGuard` allow entry + a Grafana panel, so a follow-up issue is acceptable if that's too much for the hotfix.

**NITS**
1. `openwrt/files/usr/lib/lua/wifihaven/usage.lua:340` — only the pure helper is unit-tested; nothing pins that the loop at `wifihaven-agent:1312` actually *calls* it with `(last_usage_wall_ts, wall, usage_max_window)`. The `on_tick` loop isn't unit-testable without a harness, so this is acceptable, but a regression that silently dropped the call site would go uncaught. Consider a thin extraction or a comment cross-link.
2. `openwrt/files/usr/sbin/wifihaven-agent:90` — if an operator sets `usage_max_window_seconds` *below* `usage_report_interval`, normal (non-stalled) buckets get clamped → mild under-count. Docs say keep it ≥ interval, but there's no startup guard/warn (the agent already warns when `activity_sample_int` doesn't divide the interval — same spirit). Consider `math.max(usage_max_window, usage_int)` or a warn log.
3. PR body carries the "🤖 Generated with Claude Code" footer — checklist §7 flags AI mentions in the PR body (the `Co-Authored-By` trailer is the sanctioned form). Non-blocking, and per the repo's own convention this footer is the expected harness boilerplate; calling it out only for checklist completeness.

**Dimension notes:** §1 SSOT — clean: one pure `bounded_period_start`; server presence model untouched, no re-derivation. §2 testing — TDD red→green visible (`a52264a3` failing test → `3dd67303` impl), 4 edge cases incl. backward-clock; the stall case (gap 600 → clamped) pins the bug. §3 — no router *policy* decision added (this is reporting fidelity, like the existing `activeSeconds` cap), no `/tmp` writes. §4 wire — `period_start`/`period_end` shapes unchanged (value now `max(last_flush, now−window)`, same server interpretation); `usage_max_window_seconds` is a UCI key, not wire; next-bucket watermark correctly uses real `wall`, so no double-count or coverage gap beyond the intentionally-dropped stall slack. §5 migrations — none. §7 scope — focused. §8 security — no secrets/SQL/routes. Validated on real Lua 5.1.5 per the PR.

**VERDICT: APPROVE**
Faithful, well-scoped agent-side fix for the #2016 over-count: bounds a stalled usage flush so the server's (unchanged) #1464 session-stitch can't credit an unmonitored gap as presence. No blockers; the SHOULD-FIX (clamp-frequency metric) is the one thing I'd add to confirm the fix in prod, and the two NITs are hardening.

sameerparekh:
## Independent review — re-run @ `7d92e31`

Statused the prior review (`3dd6730`) and reviewed the incremental delta (`3dd6730...7d92e31`: agent counter + startup guard, `MetricGuard.Allowed` entry, Grafana panel).

### Prior findings
| # | Finding | Status |
|---|---|---|
| SHOULD-FIX 1 | clamp path emits no metric (§6) | **ADDRESSED** — `usage_window_clamped_total` emitted from the agent when `ps_wall > last_usage_wall_ts` (`wifihaven-agent:1326`), allowlisted with bounded labels `router_id`/`installation_id` (`Metrics.scala:226`), and a panel added to `router-fleet.json` (id 24). |
| NIT 1 | loop call-site not unit-tested | **NOT-ADDRESSED** (acceptable) — the `on_tick` loop still isn't harnessed; acknowledged as a known limitation, not worth a harness for the hotfix. |
| NIT 2 | no guard when `usage_max_window_seconds < interval` | **ADDRESSED** — startup warn + raise-to-interval (`wifihaven-agent:380-387`), mirroring the `activity_sample_int` divisor warn. |
| NIT 3 | "Generated with Claude Code" footer in PR body (§7) | **NOT-ADDRESSED** (intentional) — repo-standard harness boilerplate; non-blocking. |

### New findings (delta)
**BLOCKERS** (none)

**SHOULD-FIX** (none)

**NITS**
1. `deploy/grafana/dashboards/router-fleet.json:` new panel id 24 sits at `x:12,y:88` — the previously-empty half-row, no overlap; query `sum(rate(usage_window_clamped_total{env="$env"}[5m]))` targets the series actually emitted. Clean. (No finding — noted for completeness.)

**Dimension notes:** §1 SSOT — the counter is incremented from the single clamp site; no logic duplicated. §3/§6 cardinality — agent emits the counter label-less; the server folds in only `router_id`/`installation_id` (matches the allowlist `Set`, which the ingest's `subsetOf(allowed)` check accepts) — no per-mac/host/ip dimension. §2 — `RouterMetricsIngestSpec` (7) still green incl. the "unknown name dropped" assertion, so the additive allowlist entry doesn't weaken the cardinality firewall; `usage_spec` (76) green; agent re-validated on Lua 5.1.5. §7 — focused, no drive-by.

**VERDICT: APPROVE**
Delta cleanly closes the prior SHOULD-FIX (full metric: agent counter + allowlist + Grafana panel) and the misconfig NIT (startup guard). No new blockers; cardinality firewall intact. Ready for merge-when-ready at the operator's discretion.

sameerparekh:
Closing as **superseded** — per discussion this clamp papers over the symptom rather than fixing the cause, and it offers no rollout-timing advantage (it's agent-side, so like the durable fixes it only takes effect once routers auto-update). The root cause and its server-robustness gap are now tracked as:

- **#2024** — usage flush + activity sampler stall when the conntrack stream is idle (the actual stall: `on_tick` is gated on the blocking `conntrack -E -e NEW` read). Fixing this stops wide windows from ever forming.
- **#2025** — self-describing usage buckets: the agent emits real activity-window timestamps and the server computes presence from them, not the gross flush window (also carries the server-side regression tests so refactors can't reintroduce the over-count).

Once both land, the clamp here is redundant. The `usage_window_clamped_total` metric idea is noted in #2024 as a useful stall indicator. Reopen if interim relief is wanted before #2024/#2025 ship.

PR #2026, opened 2026-06-28
fix(#2024): drive on_tick on a wall-clock heartbeat so idle conntrack can't stall usage

### LINKED ISSUE #2024: Usage flush + activity sampler stall when conntrack stream is idle (root cause of #2016 over-count)
## Problem (root cause of #2016)

The agent's cooperative timers — **both the usage flush and the 10-second activity sampler** — run only inside `on_tick`, which is called once per `conntrack -E -e NEW` line in `openwrt/files/usr/lib/lua/wifihaven/conntrack.lua`:

```lua
while true do
  local line = handle:read("*l")   -- conntrack.lua:1145 — BLOCKS, no timeout
  if not line then break end
  ...
  if cfg.on_tick then cfg.on_tick() end   -- :1195 — sampler + usage flush live here
end
```

When the whole LAN is quiet except long-lived connections (a kid alone on a websocket game like gimkit; a single device streaming), **no `NEW` events arrive, so `on_tick` never fires**. Two failures compound in the resulting bucket:

1. **Window balloons.** `period_start = last flush`, `period_end = now`, so the next flush emits one bucket spanning the whole un-monitored gap (prod: `period_seconds = 517s`). The server's #1464 session-stitch credits that entire window as continuous presence → the #2016 over-count (51 min from ~1 min of real bucket presence).
2. **Sampling starves.** The activity sampler (`activity_sample_int`, default 10s) is also gated on `on_tick`, so during the quiet stretch it doesn't run → `active_seconds` under-samples (prod: `active=20` across a 517s window; the `hostUsage=1min` undercount).

So the bucket is simultaneously over-counted (window) and under-sampled (active_seconds) — we lack the evidence because sampling stalled.

## Fix

Make the loop wake on a **wall-clock cadence** even when the conntrack stream is idle, so `on_tick` (and therefore the sampler + usage flush + policy poll + blocklist refresh + nflog drain + metrics) fires every ~1–5s regardless of conntrack traffic. Candidate approaches (pick after a spike on the real target):

- Poll/select the conntrack `io.popen` fd with a timeout (luaposix `poll`, or wrap the fd in the **cqueues** the ws sidecar already uses) so the read returns on a timer tick with no line; call `on_tick()` on timeout.
- Or run the sampler + usage flush from a separate procd timer/service, decoupled from the conntrack reader entirely.

With reliable 10s sampling the window stays ~`usage_report_interval` and `active_seconds` is accurate, so the presence math is correct at the source.

## Constraints

- OpenWRT **Lua 5.1** target — **validate on real 5.1** (openwrt.lan / plex.lan), not the macOS build (3 real bugs were caught this way before; see repo memory).
- Single-fibered cooperative loop; do not break conntrack `NEW`-flow → `connection_attempt` event processing, the events retry-queue drain, or any existing `on_tick` consumer.
- Add an agent test (busted) proving `on_tick` fires on cadence when the conntrack reader yields no lines (inject a timed/mock reader). Add `docs/router-tuning.md` notes if a new knob appears.

## Relationship to other work

- Root cause of #2016. **Supersedes the agent-side window clamp in PR #2022** (that clamp only bounds the symptom and, being agent-side, has no rollout-timing advantage over this fix). Carry the `usage_window_clamped_total` metric from #2022 into this PR as a leading indicator of stalls if useful.
- Pairs with the self-describing-buckets work (sibling issue): once sampling is reliable, the activity envelope is accurate.

sameerparekh:
Fixes #2024. Root-cause fix for the #2016 `usedMinutes` over-count, and **supersedes the closed PR #2022** (which only clamped the symptom agent-side).

## Root cause
The agent's cooperative timers — the **usage flush** and the **10 s activity sampler**, plus the policy poll, nflog drain, and metrics push — all run inside `cfg.on_tick`, which `conntrack.watch()` drove **only when a `conntrack -E -e NEW` line arrived** (`conntrack.lua` read loop). On a quiet LAN — one device on a long-lived connection (a kid alone on a websocket game like gimkit; one device streaming) emits almost no NEW events — `on_tick` stalled for minutes. Two failures compounded in the resulting bucket:

1. **Window balloons.** `period_start = last flush`, `period_end = now`, so the flush emitted one bucket spanning the whole un-monitored gap (prod: `period_seconds = 517s`). The server's #1464 session-stitch (`Presence.spanOf`) credits that entire window as continuous presence → the over-count (Octavius headline **51 min** from ~1 min of real bucket presence).
2. **Sampling starves.** The `activity_sample_int` (10 s) sampler is also gated on `on_tick`, so `active_seconds` under-sampled (prod: `active=20` across the 517 s window).

## Fix
`watch()` multiplexes a **wall-clock heartbeat** into the conntrack popen stream: `conntrack -E -e NEW` is backgrounded and a shell loop echoes `M.TICK_SENTINEL` every `conntrack_tick_interval` seconds (default **1 s**). The read loop treats a sentinel as an inert tick — it skips flow parsing but **still** drives `batcher.tick()`, the events-queue drain, and `on_tick` — so every cooperative timer fires on a wall-clock cadence regardless of conntrack traffic. With reliable sampling the window stays ≈`usage_report_interval` and `active_seconds` is accurate, so the presence math is correct at the source.

Why this shape (no new package dependency):
- The main agent runs on **bare Lua 5.1 + busybox** — its `DEPENDS` has **no** luaposix / cqueues / nixio (cqueues ships only with the ws sidecar; nixio is only present where luci is installed). A poll/select approach would force a new, hard-to-validate package dep. The heartbeat needs only `io.popen` + `sh`, both guaranteed.
- The sentinel is `"\001wh_tick"` — a leading **0x01 (SOH)** byte that can never collide with a printable-ASCII conntrack line, so `parse_conntrack_line` rejects it.
- `kill -0 "$ct"` guards the heartbeat: when conntrack dies the loop exits and the pipe EOFs, so `watch()` breaks and procd respawns — the live heartbeat does **not** mask a conntrack crash.
- conntrack's argv stays **exactly** `conntrack -E -e NEW` (the `&`/redirect are shell syntax), so the #1716 orphan sweeper (matches that exact `/proc` cmdline at `ppid==1`) still reaps a conntrack left behind across a restart.

## Stall-indicator metric (carried from #2022)
`usage_window_stall_total` increments whenever a reported usage window still exceeds **2× `usage_report_interval`** — i.e. `on_tick` stalled anyway (heartbeat wedged / wall-clock jump). A healthy fleet holds it flat at 0; a climbing rate is the leading signal of the over-count regressing. Wired end-to-end: agent counter → server allowlist in `api/src/metrics/Metrics.scala` → new "Usage window stalls per router" panel on `deploy/grafana/dashboards/router-fleet.json`.

## Tests / validation (TDD)
- `openwrt/test/conntrack_tick_spec.lua` — **failing spec committed first** (`TICK_SENTINEL` + `cfg.open_reader` seam absent → red), then green. Injects a mock reader yielding heartbeat sentinels and **zero** flow lines; asserts `on_tick` fires once per heartbeat, no event is synthesized for a sentinel, and the per-conntrack-line cadence is preserved.
- Full agent busted suite green: **967 pass / 0 fail / 2 pending** (pending = `nft` binary absent on host, unchanged from baseline).
- `scalafmt --check` clean on `api/src/metrics/Metrics.scala`.
- **Validated on real Lua 5.1.5 (`openwrt.lan`)** per the no-mock-5.1 rule:
  - both `conntrack.lua` and `wifihaven-agent` parse via `loadfile` (the agent file is one macOS-`luac` falsely rejects);
  - the module loads with its real deps; the sentinel is rejected by `parse_conntrack_line` while a real flow still parses;
  - a **live idle `conntrack -E -e NEW`** heartbeat fired on cadence on hardware (5 ticks in ~4 s with the LAN quiet) — the 0x01 sentinel round-trips through `echo`/`read("*l")` intact.

## New knob
`conntrack_tick_interval` (default `1`, range `1–5`, must stay well below `activity_sample_int`). Added to the example UCI config and documented in `docs/router-tuning.md` (new section + summary-table row).

## Rollout
Relief reaches prod when routers pick up the next agent release via auto-update. Wire-shape unchanged (UCI knob is local agent config, not a snapshot field); the only API-visible addition is the allowlisted metric name.

### DIFF
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -223,6 +223,15 @@ object MetricGuard {
     // usage reports flowing or are they stacking up?" without grepping the ring-buffer syslog.
     "usage_queue_depth"                         -> Set("router_id", "installation_id"),
     "usage_post_total"                          -> Set("result", "router_id", "installation_id"),
+    // #2024 — idle-heartbeat stall indicator. The agent now drives its
+    // cooperative timers (usage flush, activity sampler, …) on a wall-clock
+    // heartbeat so on_tick fires even when conntrack is silent. This counter
+    // increments when a reported usage window still ran past 2× the configured
+    // usage_report_interval — i.e. on_tick stalled anyway (heartbeat wedged /
+    // clock jump) and the bucket spanned un-monitored time, the #2016
+    // over-count condition. A healthy fleet holds this flat at 0; a climbing
+    // rate is the leading signal of the over-count regressing.
+    "usage_window_stall_total"                  -> Set("router_id", "installation_id"),
     // Server-side ingest health for POST /api/router/metrics (#1205). Concrete, emitted now.
     "router_metrics_batches_total"              -> Set("status"),
     // #1846 — websocket router transport (server side). `router_ws_connections_active` is the live
--- a/deploy/grafana/dashboards/router-fleet.json
+++ b/deploy/grafana/dashboards/router-fleet.json
@@ -755,6 +755,48 @@
           "refId": "A"
         }
       ]
+    },
+    {
+      "datasource": { "type": "prometheus", "uid": "${datasource}" },
+      "title": "Usage window stalls per router (#2024)",
+      "description": "sum by (router_id) (rate(usage_window_stall_total[5m])) — per-router rate of usage buckets whose reported window ran past 2× usage_report_interval, i.e. on_tick stalled even though the #2024 idle heartbeat should keep it firing on a wall-clock cadence. A stalled bucket spans un-monitored time and the server's session-stitch credits the whole window as presence — the #2016 over-count (a kid alone on a long-lived websocket/stream read as tens of minutes from ~1 min of real activity). Steady-state value MUST be 0: a non-zero series means that router's heartbeat is wedged (or its wall clock jumped) and its usedMinutes are inflating right now.",
+      "type": "timeseries",
+      "gridPos": { "h": 8, "w": 12, "x": 12, "y": 88 },
+      "id": 24,
+      "fieldConfig": {
+        "defaults": {
+          "color": { "mode": "palette-classic" },
+          "unit": "short",
+          "min": 0,
+          "custom": {
+            "drawStyle": "line",
+            "fillOpacity": 30,
+            "showPoints": "always",
+            "lineWidth": 2,
+            "thresholdsStyle": { "mode": "line" }
+          },
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              { "color": "green", "value": null },
+              { "color": "red", "value": 0.0000001 }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "options": {
+        "legend": { "displayMode": "table", "placement": "bottom", "calcs": ["lastNotNull", "max"] },
+        "tooltip": { "mode": "multi", "sort": "desc" }
+      },
+      "targets": [
+        {
+          "datasource": { "type": "prometheus", "uid": "${datasource}" },
+          "expr": "sum by (router_id) (rate(usage_window_stall_total{env=\"$env\"}[5m]))",
+          "legendFormat": "{{router_id}}",
+          "refId": "A"
+        }
+      ]
     }
   ],
   "refresh": "30s",
--- a/docs/router-tuning.md
+++ b/docs/router-tuning.md
@@ -61,6 +61,36 @@ How often (seconds) the agent samples per-MAC conntrack activity to compute
   at startup if the values are inconsistent and continues with degraded
   accuracy.
 
+### `conntrack_tick_interval` (default `1`)
+
+How often (seconds) the agent's idle heartbeat drives `on_tick` — the
+cooperative dispatcher that runs the usage flush, the `activity_sample_int`
+sampler, the policy poll, the nflog drain, and the metrics push.
+
+Before #2024 those timers fired only when a `conntrack -E -e NEW` line
+arrived. On a quiet LAN — a single device on a long-lived connection (a kid
+alone on a websocket game, one device streaming) emits almost no NEW events —
+`on_tick` stalled for minutes: the usage window ballooned to span the whole
+un-monitored gap (the server's session-stitch then credited it all as
+presence → the #2016 over-count) and the activity sampler starved. The
+watcher now multiplexes a heartbeat line into the conntrack popen stream every
+`conntrack_tick_interval` seconds, so `on_tick` fires on a wall-clock cadence
+regardless of traffic.
+
+- **Lower** → tighter worst-case latency on every cooperative timer when the
+  LAN is idle; marginally more CPU (one extra `on_tick` per second is a few
+  microseconds of monotonic-diff comparisons that early-return — the agent is
+  niced below dnsmasq, #1864).
+- **Raise** → fewer idle wakes; the activity sampler and usage flush can run
+  up to `conntrack_tick_interval` seconds late, so keep this **well below**
+  `activity_sample_int` (a value ≥ `activity_sample_int` reintroduces sampling
+  starvation, the exact failure this knob fixes).
+- **Leading indicator**: the `usage_window_stall_total` metric increments
+  whenever a reported window still exceeds 2× `usage_report_interval` — i.e.
+  the heartbeat failed to keep `on_tick` alive. A healthy fleet holds it flat
+  at 0 (see the "Usage window stalls per router" panel on the router-fleet
+  Grafana dashboard).
+
 ### `event_batch_size` (default `50`)
 
 How many `connection_attempt` events the agent buffers before flushing to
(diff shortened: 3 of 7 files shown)


sameerparekh:
## Independent review — #2024 idle-heartbeat fix

Read-only adversarial pass over `git diff origin/main...HEAD` against the 8-dimension checklist. The core mechanism is sound and validated on real Lua 5.1.5 hardware (heartbeat fires on an idle `conntrack -E -e NEW`; sentinel round-trips through `echo`/`read("*l")`; orphan sweeper #1716 preserved via the exact `conntrack -E -e NEW` argv). One robustness BLOCKER on the new `conntrack_tick_interval` knob.

### BLOCKERS

1. **`openwrt/files/usr/lib/lua/wifihaven/conntrack.lua:1145` (and `:1163-1165`) — a fractional or zero `conntrack_tick_interval` crashes / busy-loops `watch()`.** `tick_int = cfg.tick_interval or 1` only guards `nil`. The value comes from `tonumber(uci_get("conntrack_tick_interval","1"))` (`wifihaven-agent:124`), so an operator setting e.g. `'2.5'` yields `2.5`, which is then fed to `string.format("%d", tick_int)` in the popen command builder. **On the Lua 5.1.5 target `string.format("%d", 2.5)` raises `bad argument #2 to 'format' (integer expected, got number)`** — an uncaught error in `watch()` (the foreground loop), so the agent exits and procd respawns into a crash loop (config unchanged) until the `respawn 3600 5 5` backoff trips. A value of `'0'` formats fine but produces `sleep 0` → a tight CPU-spin echoing sentinels and storming `on_tick`. This knob is uniquely exposed: every other cadence knob is only used in float-safe `>=` arithmetic, but this one feeds both `%d` and shell `sleep`. Fix: floor + clamp to a positive integer where it's consumed, e.g. `local tick_int = math.floor(tonumber(cfg.tick_interval) or 1); if tick_int < 1 then tick_int = 1 end`. (Verified the 5.1 crash on `openwrt.lan`.)

### SHOULD-FIX

(none)

### NITS

1. **`openwrt/files/usr/sbin/wifihaven-agent:1322-1326` — stall metric trips on a forward wall-clock jump.** `window_seconds = wall - ts.last_usage_wall_ts` guards the *backward* clock case (negative span < 2×interval), but a forward NTP correction > 2×`usage_report_interval` would spuriously increment `usage_window_stall_total` without an actual `on_tick` stall. Acceptable for a diagnostic counter (no enforcement impact), but worth a comment or a monotonic cross-check if false positives prove noisy.
2. **`openwrt/files/usr/sbin/wifihaven-agent:1320-1330` — the stall-metric threshold path is not unit-tested.** It's inline in the agent main loop (not a module fn), so it's awkward to cover without refactoring; the cadence fix itself is well-pinned by `conntrack_tick_spec.lua`. Low value, noted for completeness.

### Dimension notes (passing)

- **SSOT (1):** sentinel is one constant `M.TICK_SENTINEL` referenced by both the command builder and the loop compare; the test sources the same constant. No re-derivation of a decision; `window_seconds` is a fresh diagnostic, not a duplicated computation. ✓
- **Testing (2):** TDD red→green visible (failing `conntrack_tick_spec.lua` committed first, then impl). The cadence assertion pins exactly the bug (on_tick fires with zero flows). ✓
- **Architecture (3):** router stays a dumb applier (no decision logic added); no new snapshot field (knob is local UCI config); no new `/tmp` writer (sentinels are read-and-discarded, never spooled). ✓
- **Wire (4):** `usage_window_stall_total` is additive to the allowlist; old server drops unknown metric to `metrics_rejected_total`, old agent simply doesn't emit. ✓
- **Metrics+dashboard (6):** new counter → `Metrics.scala` allowlist + router-fleet Grafana panel; labels `router_id`/`installation_id` are the established bounded router-agent set (not in `ForbiddenKeys` = mac/device_id/domain). Agent emits `{}`; server attaches router labels, matching `usage_post_total`. ✓
- **Security (8):** `tonumber()` sanitizes the knob against shell injection (non-numeric → fallback); `%d`/integer is the only remaining gap, covered by the BLOCKER above. ✓

VERDICT: REQUEST-CHANGES

The heartbeat-multiplex fix correctly addresses the #2016 root cause at the agent and is validated on real 5.1 hardware; the metric/dashboard/docs are complete and additive. The one BLOCKER is a defensive-input gap: a fractional `conntrack_tick_interval` crashes the watcher loop on the 5.1 target — clamp the knob to a positive integer before it reaches `%d`/`sleep`.

sameerparekh:
## Independent review — re-run (delta `6cf73bd...b9fa3e8`)

### Prior-findings status

|---|---|---|
| BLOCKER 1 | Fractional/zero `conntrack_tick_interval` crashes/spins `watch()` (`%d` hard error on 5.1, `sleep 0` busy-loop) | **ADDRESSED** — new `M.sanitize_tick_interval` (`conntrack.lua:46-61`) floors + clamps to ≥1; `watch()` now calls it (`:1152`). Unit-pinned (`conntrack_tick_spec.lua:52-63`: 2.5→2, 0→1, -5→1, nil→1, "abc"→1) and re-validated on real Lua 5.1.5 (`openwrt.lan`): `string.format("%d", sanitize(2.5))=="2"`, `sanitize(0)==1`. |
| NIT 1 | Stall metric trips on a forward wall-clock jump | **ADDRESSED** — comment at `wifihaven-agent:1323-1328` documents it as an accepted false positive (best-effort diagnostic; a forward step genuinely inflates the window too). |
| NIT 2 | Stall-threshold path not unit-tested | NOT-ADDRESSED — non-gating; remains awkward to cover without refactoring the agent main loop. Acceptable. |

### New findings in the delta

BLOCKERS: (none)
SHOULD-FIX: (none)
NITS: (none) — the delta is the clamp helper, its unit test, and two comments; no new logic paths.

VERDICT: APPROVE

The #2024 heartbeat-multiplex fix resolves the #2016 over-count at its source and is validated on real 5.1 hardware; the BLOCKER (defensive input clamp on the new cadence knob) is fixed with a tested pure helper and introduces no new risk. The remaining NIT-2 is non-gating. Ready for merge-when-ready at the operator's discretion.

Issue #846, opened 2026-05-22
ui+api: Traffic Usage page — raw + aggregated views (#844)


sameerparekh:
Sub-issue of #844.

## Goal

New Traffic Usage page for raw + aggregated bytes/seconds views, with bucket-selector and group-by controls. Sibling to the Connection Events page.

## UI

- Bucket selector: **Raw (5m) / 1m / 10m / 1h / 12h / 1d / 1w**.
  - "Raw" maps to the underlying `traffic_reports` bucket (currently 5m).
  - "1m" is shown but disabled with a tooltip explaining it requires bumping the router upload cadence — track that as a separate router-blocked sub-issue when needed (do NOT change the router for this).
- Group-by selector (visible only when bucket ≠ Raw): **per-domain / per-apex / per-app**. Per-app gated on the apps track (#761-#769).
- Columns (Raw): mac/device, profile, host, bytes_in, bytes_out, seconds, period_start.
- Columns (Aggregated): group, window_start, total_bytes_in, total_bytes_out, total_seconds, distinct_devices.

## API

- Use `/api/usage/series` (per #813's granularity-by-window design) — extend it as needed to accept `groupBy=(domain|apex|app)` and additional aggregation columns.
- Pick storage source by bucket width: raw `traffic_reports` for short windows; `traffic_hourly` / `traffic_daily` rollups (per #809) for wide windows.
- Honor retention gating per #814 — return an empty range + 409 (or whatever the established shape is) if the requested window is outside retention.

## Constraints

- ZERO router-side changes.
- The "1m" bucket option is reserved for the future and surfaces disabled.
- Use household-local TZ for human-facing bucket alignment.

## Cross-refs

- #813 (granularity-by-window in /api/usage/series).
- #809 (rollup tables).
- #814 (retention gating in date picker).
- #715 (proportional time per host — once that's exposed, surface both presence-seconds and proportional-seconds in this page).
- #459 (logs page should include usage logs) — this sub-issue resolves the usage half; closes #459 when shipped.
- #783, #828, #829, #830 UX patterns.

## Acceptance

- Bucket switching across raw → 1w on a representative day works without slow queries.
- Aggregated rows match raw sum within rounding error.
- Per-app option visible but disabled.

sameerparekh:
Adjacent: #917 redefines the aggregation mental model on these pages (strictly additive group-by). Affects the same column headers as #865; coordinate.

sameerparekh:
Shipped via PR #853.

PR #2028, opened 2026-06-28
fix(#2018): raw bucket uses real report period + #2020 guardrail

### LINKED ISSUE #2018: bug(dashboard): 'raw' bandwidth bucket is a 5-min window, not the router's actual usage-send period — wrong/smoothed rate vs 1m
## Symptom

On the dashboard LIVE BANDWIDTH gauge, the **`raw`** bucket shows **different (lower, smoother) numbers than `1m`** and is no more real-time than `1m`. Operator expectation: **`raw` = whatever update granularity the router is actually sending** (one point per ingested usage period), i.e. the *finest / most real-time* view — not a coarse fixed window.

## Root cause — `raw` is hardcoded to a **5-minute** window

In `api/src/usage/UsageTraffic.scala` (pre-existing legacy bucket math, reused by the SPA-ws aggregator):

```scala
def floorTo(instant, Bucket.Raw, _) = instant.truncatedTo(MINUTES).minusSeconds(epochSecond % 300)  // 5-min boundary
def stepOf(Bucket.Raw)              = Duration.ofMinutes(5)
```

So `raw` is a **5-minute** bucket — *coarser* than `1m`, the opposite of what "raw" should be. The ws `trafficUsage` aggregator (`SpaPush.pushOneParamSet`, #1971/#2004) anchors the head on `floorTo(periodStart, raw)` → a 5-min window → aggregates ~5 min of bytes; the client divides by the ~300s span (`wsCache.bucketSeconds(raw)` reads the `[windowStart,windowEnd)` span), so the rate reads ~5× smoothed and diverges from `1m`. Confirmed live: a raw push carried a `windowStart→windowEnd` span of ~300s.

## Desired behavior (operator decision, 2026-06-27)

**`raw` = the router's actual usage-send period.** The ws aggregator already receives `UsageIngested(periodStart)` — for `raw`, the head window should be the *actual ingested period* `[periodStart, periodEnd)` (the router's report interval, ~60s today; sub-minute / toward real-time as #1023 streams usage), not floored to 5 min (or 1m). The client's span-based `bucketSeconds(raw, row)` then yields the correct instantaneous rate with no client change. This matches `docs/design/spa-websocket.md` §10 Q1 ("raw = at whatever cadence the router sends usage data").

## Scope / consistency

`floorTo`/`stepOf` for `raw` are shared by **(a)** `GET /api/usage/traffic`, **(b)** the Traffic Usage page, and **(c)** the ws aggregator — so the fix must keep all three consistent (raw windows = the real report periods). Verify the `raw` path of the GET still returns sensible per-period rows. Per-period boundaries come from the stored `traffic_reports` period (`period_start` + duration), not a synthetic floor.

## Acceptance
- Selecting `raw` on the dashboard shows a rate consistent in magnitude with `1m` (same underlying traffic), updating once per router usage-send period.
- The streamed `raw` head window equals the actual ingested period `[periodStart, periodEnd)`; `windowEnd - windowStart` = the report interval, not 300s.
- GET `/api/usage/traffic?bucket=raw` and the Traffic Usage page agree (SSOT).

Separate from #2017 (gauge GET-seed blank-on-load) and #2004 (head-anchor). Server-side (+ verify client rate). Relates to #1860, #1148, #747.

### LINKED ISSUE #2020: guardrail: usage-report period is single-sourced at the agent (usage_report_interval) — purge hardcoded 5-min API assumptions + prevent recurrence
## Why

The `raw`-bucket bug (#2018) happened because the usage **reporting period** — which is configurable and lives in exactly one place, the agent's `openwrt/files/etc/config/wifihaven` → `usage_report_interval` (default **60s**) — was **duplicated as a hardcoded API-side constant** (`UsageTraffic.floorTo(Raw) % 300`, `stepOf(Raw) = 5min`) in #846. When the agent cadence changed (300s → 60s, #529) the API copy silently drifted, and a false comment ("source rows are at 5-min boundaries") cemented the wrong mental model. This is a textbook single-source-of-truth violation (`docs/process/single-source-of-truth.md`).

This issue makes that class of error structurally hard to repeat.

## Scope

1. **Audit & purge stale 5-min / 300s assumptions about the usage period** across the API. The real per-report window is on every row as `periodStart`/`periodEnd` (driven by `usage_report_interval`); nothing in the API should assume a fixed 5-min (or any fixed) usage-report granularity. Targets (from a repo grep): `api/src/db/Repos.scala` (L62/L522/L600 "per 5-min" comments), `api/src/usage/RollupRepo.scala`, `api/src/usage/AppUsedRollupService.scala`, `api/src/usage/TimeUsedRollupJob.scala`, `api/src/routes/UsageRoutes.scala`, `api/src/usage/UsageTraffic.scala` (comments). Fix the comments to say "one row per agent usage-report period (`usage_report_interval`, ~60s; configurable)"; where logic genuinely assumes 5-min bucketing, make it period-agnostic (derive from the row span). Update test fixtures that seed synthetic `300`s buckets to either realistic 60s periods or clearly period-agnostic seeds (don't let them re-imply 5-min is the granularity).
2. **Pinning test** — assert the `raw`/period-derivation reads the row's real `[periodStart, periodEnd)` and would FAIL if anyone re-hardcodes a fixed window (e.g. seed a 37s and a 90s report period and assert the streamed/queried raw window equals each, and the derived B/s = bytes/span).
3. **Guardrail** — a cheap CI/grep check (sibling to the existing branch-diff/process checks) that flags new literal `300` / `% 300` / `ofMinutes(5)` / `* 300` in the usage/traffic bucket + rollup code paths with a pointer to "usage period is data-derived (`periodStart`/`periodEnd`), never hardcode it — see `usage_report_interval`." Keep it narrowly scoped to those files to avoid false positives; allow an explicit opt-out comment.
4. **Doc** — add this as a worked cautionary example in `docs/process/single-source-of-truth.md` (or the usage/architecture doc): the usage-report period is single-sourced at the agent (`usage_report_interval`) and carried per-row; the API must never hold its own copy.

## Acceptance
- No API code or comment asserts a fixed 5-min usage-report granularity (or, where a fixed bucket is intentional like 10m/1h aggregation, it's clearly the *display* bucket, not the *source* period).
- The pinning test fails if `raw`/period logic is re-hardcoded to a constant window.
- The guard rejects a newly-introduced hardcoded usage-period literal in the targeted paths.

Provenance + the original fix are in #2018. Relates to single-source-of-truth (`docs/process/single-source-of-truth.md`), #2018, #846, #529, #101.

sameerparekh:
## What & why

The dashboard **LIVE BANDWIDTH** gauge's `raw` bucket read **lower/smoother than `1m`** — the opposite of what "raw" should be. Root cause: the API hardcoded a **5-minute** `raw` window in `UsageTraffic` (`floorTo(Raw) … % 300`, `stepOf(Raw) = ofMinutes(5)`), coarser than `1m` and divorced from the real cadence.

**Provenance (git-confirmed):** `300` was the agent's *original* `usage_report_interval` default (#101, 2026-05-07). #529 moved it to **60s** (2026-05-17). Then #846 (2026-05-22) re-baked the dead `% 300` / `ofMinutes(5)` into the API — **five days after** the agent left 300s — with a false comment ("source rows are at UTC 5-min boundaries already"). The API copy silently drifted off the configurable cadence. The real per-report window already rides every `traffic_reports` row as `[periodStart, periodEnd)`.

This PR lands **#2018** (the fix) and **#2020** (prevent the whole class) together.

## Part A — #2018: raw uses the row's real period

- `floorTo(Raw)` no longer floors onto a synthetic grid — the row's real `periodStart` **is** the window start.
- `stepOf` now returns `Option[Duration]`; `Raw` → `None` (no fixed width).
- New `UsageTraffic.windowFor` SSOT: `raw` → the real `[periodStart, periodEnd)` report period; display buckets → the floored grid cell. Shared by the `GET /api/usage/traffic` aggregate path (`buildAggregate`) **and** the ws live-edge push (`SpaPush.pushOneParamSet`), so they can't disagree.
- `SpaEvent.UsageIngested` now carries `periodEnd` too, so the streamed `raw` head window equals the actual ingested period — `windowEnd − windowStart == the report interval`, not 300s. Client B/s (`bytes / span`, `wsCache.bucketSeconds(raw, row)`) comes out correct with **no client change**.
- Deleted the two false "5-minute boundary" comments.
- `GET ?bucket=raw` already emitted per-row `periodStart`/`periodEnd` via `buildRaw` (verified by `UsageApiSpec`) — GET, the Traffic Usage page, and the ws aggregator now all derive raw from the row's real period.

## Part B — #2020: purge + guard

- **Purged** stale "per 5-min" usage-period claims across `api/src` (Repos, RollupRepo, AppUsedRollupService, TimeUsedRollupJob, UsageRoutes, DashboardNowRoutes, DebugRoutes, Routes). Comments now say "usage-report period (`usage_report_interval`, ~60s, configurable)". Display-bucket aggregation (10m/1h/…) untouched — only the *source-period* assumption was wrong. (One comment even self-contradicted: "single 5-min agent bucket … each host row holds ~60s for that bucket".)
- **Pinning tests** (TDD, committed RED first): unit seeds 37s & 90s periods; ws feature ingests a 37s period — both assert the raw window equals the real period and derived B/s == bytes/span. They fail the moment a fixed window is re-hardcoded.
- **CI guard** `.github/scripts/check-usage-period-hardcode.sh` (+ `.test.sh`, wired into `ci.yml` as `usage-period-hardcode`, sibling to the migration checks) rejects new `% 300` / `* 300` / `ofMinutes(5)` / bare `300` literals in the usage/traffic bucket + rollup paths. Comment lines exempt; explicit opt-out marker `usage-period-ok`.
- **Doc** worked example added to `docs/process/single-source-of-truth.md`.

## Tests

- RED→GREEN: `test(#2018): RED …` commit fails on current code; the fix turns it green.
- `mill api.test.testOnly` green for `UsageTrafficSpec`, `SpaWsS4Spec`, `SpaWsSpec`, `UsageApiSpec`, `HouseholdPatchApiSpec`.
- `scalafmt --check` clean; guard `.test.sh` passes (9 cases) and the guard passes on this PR's own diff.

Closes #2018, closes #2020. Relates to #846, #529, #101, #1023, #1148, #2004.

### DIFF
--- a/.github/scripts/check-usage-period-hardcode.sh
+++ b/.github/scripts/check-usage-period-hardcode.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+# Guardrail (#2020): the usage-report PERIOD is single-sourced at the agent —
+# `openwrt/files/etc/config/wifihaven` → `usage_report_interval` (default 60s,
+# configurable) — and the real per-report window rides every `traffic_reports`
+# row as `period_start`/`period_end`. The API must NEVER hold its own copy of
+# that period. It once did: #846 hardcoded a 5-minute raw window
+# (`UsageTraffic.floorTo(Raw) % 300`, `stepOf(Raw) = ofMinutes(5)`) five days
+# after the agent moved off 300s (#529), and the API copy silently drifted —
+# the #2018 bug (the dashboard `raw` gauge read ~5x smoothed).
+#
+# This check fails the build if a PR ADDS a hardcoded usage-period literal
+# (`% 300`, `* 300`, `/ 300`, a bare `300`, or `ofMinutes(5)` / `ofSeconds(300)`)
+# to the narrow set of usage/traffic bucket + rollup code paths where such a
+# literal would be re-introducing a fixed period assumption. Derive the window
+# from the row span (`UsageTraffic.windowFor` / `period_start`/`period_end`),
+# never from a constant.
+#
+# Scope: only the files below, and only ADDED *code* lines (full-comment lines
+# are ignored, so it's fine to mention `% 300` / `ofMinutes(5)` in an
+# explanatory comment). Legitimate uses of one of these literals in a scoped
+# file (a genuine display-bucket constant, a page size, a timeout) can opt out
+# by putting the marker `usage-period-ok` on the same line.
+#
+# See docs/process/single-source-of-truth.md ("Usage-report period" worked
+# example) and AGENTS.md #single-source-of-truth.
+set -euo pipefail
+
+BASE="${1:?usage: check-usage-period-hardcode.sh <base-ref>}"
+
+# The bucket / window / rollup derivation paths. A hardcoded period literal in
+# any of these is almost certainly a re-introduced fixed-period assumption.
+SCOPED_FILES=(
+  "api/src/usage/UsageTraffic.scala"
+  "api/src/usage/UsageTrafficQuery.scala"
+  "api/src/usage/AppUsedRollupService.scala"
+  "api/src/usage/TimeUsedRollupJob.scala"
+  "api/src/usage/RollupJobs.scala"
+  "api/src/db/RollupRepo.scala"
+  "api/src/routes/SpaPush.scala"
+  "api/src/routes/UsageRoutes.scala"
+)
+
+# Patterns that smell like a hardcoded usage period.
+#   % 300 / * 300 / / 300  — arithmetic against a 300s grid
+#   ofMinutes(5) / ofSeconds(300) — the dead 5-min agent default
+#   a bare 300 token        — likely a period in seconds
+PERIOD_RE='(%[[:space:]]*300|\*[[:space:]]*300|/[[:space:]]*300|ofMinutes\([[:space:]]*5[[:space:]]*\)|ofSeconds\([[:space:]]*300[[:space:]]*\)|(^|[^0-9])300([^0-9]|$))'
+
+violations=0
+
+for f in "${SCOPED_FILES[@]}"; do
+  # Added lines only (three-dot merge-base diff, no context lines).
+  while IFS= read -r line; do
+    # `git diff --unified=0` prefixes added lines with a single '+'.
+    [[ "${line}" == +++* ]] && continue
+    [[ "${line}" != +* ]] && continue
+    code="${line:1}"
+
+    # Ignore explicit opt-outs.
+    if [[ "${code}" == *usage-period-ok* ]]; then
+      continue
+    fi
+
+    # Ignore full-comment lines (leading // or * or /*). Inline trailing
+    # comments on a real code line are still scanned — the code part counts.
+    trimmed="${code#"${code%%[![:space:]]*}"}"
+    if [[ "${trimmed}" == //* || "${trimmed}" == \** || "${trimmed}" == /\** ]]; then
+      continue
+    fi
+
+    if [[ "${code}" =~ ${PERIOD_RE} ]]; then
+      if [[ ${violations} -eq 0 ]]; then
+        echo "ERROR: new hardcoded usage-period literal in a usage/traffic path (#2020)."
+        echo "The usage-report period is data-derived (period_start/period_end, driven by the"
+        echo "agent's usage_report_interval). Derive the window from the row span"
+        echo "(UsageTraffic.windowFor), never from a constant. If this literal is a legitimate"
+        echo "display-bucket / page-size / timeout, add the marker 'usage-period-ok' on the line."
+        echo
+      fi
+      echo "  ${f}: +${code}"
+      violations=$((violations + 1))
+    fi
+  done < <(git diff --unified=0 "${BASE}...HEAD" -- "${f}")
+done
+
+if [[ ${violations} -gt 0 ]]; then
+  echo
+  echo "Found ${violations} hardcoded usage-period literal(s). See the message above."
+  exit 1
+fi
+
+echo "No new hardcoded usage-period literals in the scoped usage/traffic paths."
--- a/.github/scripts/check-usage-period-hardcode.test.sh
+++ b/.github/scripts/check-usage-period-hardcode.test.sh
@@ -1,0 +1,122 @@
+#!/usr/bin/env bash
+# Tests for check-usage-period-hardcode.sh (#2020).
+#
+# Rule under test: a PR may not ADD a hardcoded usage-period literal
+# (% 300 / * 300 / / 300 / bare 300 / ofMinutes(5) / ofSeconds(300)) to the
+# scoped usage/traffic bucket + rollup paths. Full-comment lines and lines
+# carrying the `usage-period-ok` marker are exempt; the same literal in an
+# UNSCOPED file is ignored.
+set -euo pipefail
+
+SCRIPT="$(cd "$(dirname "$0")" && pwd)/check-usage-period-hardcode.sh"
+[[ -x "${SCRIPT}" ]] || chmod +x "${SCRIPT}"
+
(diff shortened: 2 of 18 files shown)


sameerparekh:
## Independent review — PR #2028 (`fix(#2018) + #2020 guardrail`)

Read-only adversarial pass over the full merge-base diff (`origin/main...HEAD`, 18 files) against the 8 checklist dimensions.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. `api/src/usage/UsageTraffic.scala:308` — `maxPeriodEnd` is computed (a full `bucketRows` iteration) for every group, but `windowFor` only consumes it for `raw`; for display buckets the value is discarded. Negligible cost and the clarity is worth it — left as a note, not a request.
2. `.github/scripts/check-usage-period-hardcode.sh:60` — the bare-`300` rule (`(^|[^0-9])300([^0-9]|$)`) is broad; a future legitimate `300` constant landing in one of the 8 scoped files would need the `usage-period-ok` opt-out. That's the intended trade-off (scoped narrowly to bucket/rollup files, comment lines exempt, opt-out documented) — acceptable.

### Dimension notes
1. **SSOT (highest priority)** — the fix *centralizes*: new `UsageTraffic.windowFor` is the one place raw-vs-display window derivation lives, consumed by BOTH the GET aggregate path (`buildAggregate`) and the ws push (`SpaPush.pushOneParamSet`), so they cannot drift. `stepOf` returning `Option` forces callers off a constant. GET `?bucket=raw` (`buildRaw`) already emitted the row's real `periodStart`/`periodEnd`. This is the textbook resolution of the #2018 drift — strong.
2. **Testing** — RED commit (`e8fa1445`) precedes GREEN; pins are real (37s/90s unit periods, 37s ws ingest) and assert window == real period AND B/s == bytes/span, failing if a fixed window returns. The two edited assertions (the existing raw S4 test; the coalesce test) are *strengthened*, not weakened: the old "raw floors to the 5-min boundary" assertion encoded the bug and is correctly replaced with windowStart==periodStart && windowEnd==periodEnd; the coalesce test now asserts both period bounds survive. Justified per the checklist's "OLD assertion was actually wrong" carve-out. Plus a 9-case `.test.sh` for the guard.
3. **Architecture** — Clock still injected; ZIO effects, no `throw`; no router-agent or `/tmp` surface touched.
4. **Wire contract** — `SpaEvent` is a purely in-process bus type (no JSON codec, never serialized to router or SPA), so adding `periodEnd` to `UsageIngested` is not a wire change. `traffic_reports` row shape and the `TrafficUsageResponse` body are unchanged. No client change (the client already derives B/s from `[windowStart,windowEnd)`).
5. **Migrations** — none.
6. **Metrics** — no new route/job/poller; rides the existing `spa_ws_push_total{op="trafficUsage"}`. No new metric warranted.
7. **Scope/hygiene** — focused; comment purge stays within the #2020-enumerated files plus two clearly-false self-contradicting claims (DebugRoutes "5-min agent bucket … holds ~60s"); no AI mentions in title/body/commits beyond the allowed `Co-Authored-By` trailer; three-dot diffs throughout; CI guard wired into `ci.yml` aggregator + passes actionlint and on this PR's own diff.
8. **Security** — no secrets, no SQL string-interpolation, no new routes/auth surface.

VERDICT: APPROVE
The #2018 hardcoded 5-min raw window is removed and replaced with a data-derived `[periodStart,periodEnd)` window single-sourced through `windowFor`; #2020's pinning tests + CI grep-guard + SSOT doc make the regression structurally hard to reintroduce. TDD red→green is visible, the edited assertions are justified strengthenings, and there is no wire/migration/security exposure. No open BLOCKER.

PR #2029, opened 2026-06-28
feat(#2025): self-describing usage buckets — accurate screen-time from the real activity window

### LINKED ISSUE #2024: Usage flush + activity sampler stall when conntrack stream is idle (root cause of #2016 over-count)

### LINKED ISSUE #2025: Self-describing usage buckets: emit real activity-window timestamps; server computes presence from them (not the flush window)
## Problem (server robustness for #2016)

The server's #1464 session-stitch (`api/src/presence/Presence.scala` `spanOf`) treats each usage row's full `[period_start, period_end]` flush window as continuous engaged presence. That window is **"time since the last flush"**, not **"time the device was actually active"** — when a flush is delayed (see sibling stall-fix issue), the window balloons and the server over-counts (#2016: a 20s activity burst in a 517s window read as ~8.6 min). The server has no defense against a bad window because it trusts it blindly; a future agent bug or a refactor could re-introduce the inflation.

## Fix — buckets carry their real activity timing; the server computes from that

The agent's activity tracker (`openwrt/files/usr/lib/lua/wifihaven/usage.lua`, `tracker_sample`) already detects, per `(mac, dst_ip)`, which sample ticks saw byte growth. Record the **wall-clock of the first and last active sample** and emit it on each usage record as **additive, optional** `activeStart` / `activeEnd` (RFC3339).

Server side: `PresenceRow` gains an optional active window; `Presence.spanOf` uses `[activeStart, activeEnd]` when present (the real activity envelope), **falling back to `[period_start, period_end]`** when absent (old agents). This makes the over-count structurally impossible: a wide flush window with a 20s activity burst contributes a 20s span, not the whole window.

## Constraints

- **Additive wire only** (router↔API is a public contract; both sides tolerate absence — see AGENTS.md §wire-contract). No renaming/retyping existing fields. Decode path: `RouterIngestService` → `traffic_reports` (new nullable columns or derive at read) → `PresenceRow`.
- OpenWRT **Lua 5.1** — validate the tracker + `build_report` changes on real 5.1 (openwrt.lan).
- **Server regression tests** (this is the "improve testing so refactors can't break things" ask from #2016): a `PresenceSpec` / `TimeStatusServiceSpec` fixture mirroring the prod stall shape (wide `period_seconds`, tight activity envelope) asserting **bounded** usedMinutes; plus invariant pins — per-app `usedMins` ⊆ profile `usedMins` for a non-exempt app, Dedup-mode profile ≤ wall-clock, and rollup-path `dayState` ⇄ all-live `dayStateLive` equivalence — so a refactor (the #1974-class change) that silently alters the presence→minutes computation fails CI.
- Agent test (busted) for the tracker timestamps + `build_report` emission.

## Relationship to other work

- Server-robustness half of #2016. **Together with the stall-fix sibling issue, supersedes PR #2022's clamp.** Once both land, the clamp is redundant (stall fix prevents wide windows; this makes the server ignore the window anyway).

sameerparekh:
Resolves #2025 — the **server-robustness half of #2016**. Stacked on the schema-only migration #2027 (base will auto-retarget to `main` once #2027 merges).

## Problem
`Presence.spanOf` (the #1464 session-stitch) credits each usage row's full `[period_start, period_end]` **flush** window as continuous engaged presence. That window is "time since the last flush", not "time the device was active". When a flush stalls (the conntrack-gated agent loop on a quiet LAN — sibling #2024), the window balloons and the server over-counts (#2016: a ~20s burst in a 517s window read as ~8.6 min). The server had no defense — it trusted the window blindly.

## Fix — buckets carry their real activity timing; the server computes from that
**Agent** (`usage.lua` + `wifihaven-agent`): the activity tracker now records, per `(mac, dst_ip)`, the wall-clock of the **first and last** sample tick that saw byte growth (`tracker_sample` takes an optional `now`; the agent threads `wall`). `build_report` emits these as additive, optional `activeStart`/`activeEnd` (RFC3339).

**Wire/server**: additive only (router↔API contract — both sides tolerate absence). `UsageRecord` → `RouterIngestService` → `traffic_reports.active_start/active_end` (new nullable cols, #2027) → `PresenceRow`. `spanOf` uses `[activeStart, activeEnd]` when present, **falling back to `[period_start, period_end]`** when absent (old agents) — so old data is unaffected.

`effectiveGap` also derives `R` from `spanOf` (not `period_seconds`): otherwise the `2 × R` collapse guard would raise the idle gap to `2 × period_seconds` and **bridge** the idle stretches between bursts, re-chaining the inflation `spanOf` removed. Both halves together bound the credit regardless of flush cadence — the over-count is structurally impossible.

`aggregateForInsert` unions the envelopes of collapsed `(mac, host)` duplicates (earliest start, latest end); a malformed/absent wire timestamp degrades to the flush-window fallback, never a batch failure.

## Regression tests (the #2016 "refactors can't silently break the math" ask)
- **unit `spanOf`**: envelope used / fallback when absent / backwards-clamp to 0 / half-present → fallback
- **stall fixture** (wide flush window + tight activity envelope) → **bounded** usedMinutes, contrasted against the un-defended flush-window arm, pinned at three levels: unit (`PresenceSpec`), feature `dayStateLive` (`TimeStatusServiceSpec`), and the **real rollup path** `dayState ⇄ dayStateLive` equivalence (`TimeUsedRollupSpec`)
- **invariant pins**: per-app `usedMins` ⊆ profile `usedMins` for a non-exempt app; Dedup-mode profile ≤ elapsed wall-clock and ≤ 1440

TDD: the stall fixtures went red first and surfaced the `effectiveGap` half (the flush-window `R` was bridging the bursts); fixing it turned them green.

## Validation
- `mill api.test` green (full suite); `scalafmt --check` clean.
- Agent busted suite green; **validated on real Lua 5.1.5 (openwrt.lan)** per the no-mock-5.1 practice — module + agent parse, KAT covers tracker first/last recording, no-clock omission, reset, and `build_report` RFC3339 emission/omission.
- Contract golden `usage_report.json` regenerated (documents both populated + omitted shapes); `ContractGoldenSpec` round-trips it through the production decoder.

## Relationship to other work
Server-robustness half of #2016. Together with the agent stall fix #2024, **supersedes the closed PR #2022 clamp** (the stall fix prevents wide windows; this makes the server ignore the window anyway).

### DIFF
--- a/api/src/db/Repos.scala
+++ b/api/src/db/Repos.scala
@@ -471,6 +471,12 @@ case class TrafficReportInsert(
     // pre-#730 agents do not emit it; stored as NULL on traffic_reports.dest_ip
     // for those records.
     destIp: Option[IpAddress] = None,
+    // #2025: the real activity envelope (first/last growth-sample wall-clock).
+    // Optional because pre-#2025 agents do not emit it; stored as NULL on
+    // traffic_reports.active_start / active_end for those records, and
+    // Presence.spanOf falls back to the [period_start, period_end] flush window.
+    activeStart: Option[Instant] = None,
+    activeEnd: Option[Instant] = None,
 )
 
 case class BlockEventInsert(
@@ -1762,7 +1768,7 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
   def insertBatch(reports: List[TrafficReportInsert]) =
     DbMetrics.timed("traffic.insertBatch")(
       Update[TrafficReportInsert](
-        "INSERT INTO traffic_reports(router_id,mac,ip,host_type,host_value,date,period_start,period_end,active_seconds,bytes_in,bytes_out,dest_ip) VALUES(?,?,?,?,?,?,?,?,?,?,?,?) ON CONFLICT(router_id,period_start,mac,host_type,host_value) DO NOTHING",
+        "INSERT INTO traffic_reports(router_id,mac,ip,host_type,host_value,date,period_start,period_end,active_seconds,bytes_in,bytes_out,dest_ip,active_start,active_end) VALUES(?,?,?,?,?,?,?,?,?,?,?,?,?,?) ON CONFLICT(router_id,period_start,mac,host_type,host_value) DO NOTHING",
       ).updateMany(reports).transact(xa),
     )
   // TODO(#730): remove this read-side join once usage records carry dest_ip.
@@ -1825,7 +1831,19 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
       fromInstant: Instant,
       toInstant: Instant,
   ): Task[List[wifihaven.api.presence.PresenceRow]] = {
-    type Row = (MacAddress, LocalDate, Instant, HostId, Int, Long, Long, Instant, Instant)
+    type Row = (
+        MacAddress,
+        LocalDate,
+        Instant,
+        HostId,
+        Int,
+        Long,
+        Long,
+        Instant,
+        Instant,
+        Option[Instant],
+        Option[Instant],
+    )
     macs match {
       case Nil => ZIO.succeed(List.empty[wifihaven.api.presence.PresenceRow])
       case ms  =>
@@ -1841,7 +1859,8 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                            THEN 'fqdn' ELSE tr.host_type END,
                       COALESCE(CASE WHEN tr.host_type IN ('ipv4','ipv6') THEN ce.resolved_host_value END,
                                tr.host_value),
-                      tr.active_seconds, tr.bytes_in, tr.bytes_out, tr.period_start, tr.period_end
+                      tr.active_seconds, tr.bytes_in, tr.bytes_out, tr.period_start, tr.period_end,
+                      tr.active_start, tr.active_end
                FROM traffic_reports tr
                ${SqlFragments.resolvedHostLateral}
                WHERE tr.period_start >= $fromInstant AND tr.period_start < $toInstant
@@ -1849,9 +1868,10 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                  AND """ ++ Fragments.in(fr"tr.mac", nel)
         val cio =
           q.query[Row]
-            .map { case (m, d, ps, host, secs, bin, bout, pStart, pEnd) =>
+            .map { case (m, d, ps, host, secs, bin, bout, pStart, pEnd, aStart, aEnd) =>
               val periodSeconds = math.max(0L, pEnd.getEpochSecond - pStart.getEpochSecond).toInt
-              wifihaven.api.presence.PresenceRow(m, d, ps, host, secs, bin + bout, periodSeconds)
+              wifihaven.api.presence
+                .PresenceRow(m, d, ps, host, secs, bin + bout, periodSeconds, aStart, aEnd)
             }
             .to[List]
         // Bound each per-request read: a pathological window fails fast with a typed
@@ -1868,7 +1888,19 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
       to: LocalDate,
       since: Option[Instant] = None,
   ) = {
-    type Row = (MacAddress, LocalDate, Instant, HostId, Int, Long, Long, Instant, Instant)
+    type Row = (
+        MacAddress,
+        LocalDate,
+        Instant,
+        HostId,
+        Int,
+        Long,
+        Long,
+        Instant,
+        Instant,
+        Option[Instant],
+        Option[Instant],
+    )
     macs match {
       case Nil => ZIO.succeed(List.empty[wifihaven.api.presence.PresenceRow])
       case ms  =>
@@ -1879,7 +1911,8 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
                            THEN 'fqdn' ELSE tr.host_type END,
                       COALESCE(CASE WHEN tr.host_type IN ('ipv4','ipv6') THEN ce.resolved_host_value END,
                                tr.host_value),
-                      tr.active_seconds, tr.bytes_in, tr.bytes_out, tr.period_start, tr.period_end
+                      tr.active_seconds, tr.bytes_in, tr.bytes_out, tr.period_start, tr.period_end,
+                      tr.active_start, tr.active_end
                FROM traffic_reports tr
                ${SqlFragments.resolvedHostLateral}
                WHERE tr.date BETWEEN $from AND $to
@@ -1888,9 +1921,10 @@ class TrafficReportRepoLive(xa: Transactor[Task]) extends TrafficReportRepo {
             since.fold(fr"")(s => fr"AND tr.period_start >= $s")
         DbMetrics.timed("traffic.listPresenceRows")(
           q.query[Row]
-            .map { case (m, d, ps, host, secs, bin, bout, pStart, pEnd) =>
+            .map { case (m, d, ps, host, secs, bin, bout, pStart, pEnd, aStart, aEnd) =>
               val periodSeconds = math.max(0L, pEnd.getEpochSecond - pStart.getEpochSecond).toInt
-              wifihaven.api.presence.PresenceRow(m, d, ps, host, secs, bin + bout, periodSeconds)
+              wifihaven.api.presence
+                .PresenceRow(m, d, ps, host, secs, bin + bout, periodSeconds, aStart, aEnd)
             }
             .to[List]
             .transact(xa),
--- a/api/src/presence/Presence.scala
+++ b/api/src/presence/Presence.scala
@@ -22,6 +22,15 @@ case class PresenceRow(
     activeSeconds: Int,
     bytes: Long,
     periodSeconds: Int,
+    // #2025: the REAL activity envelope for this bucket — the wall-clock of the
+    // first and last sample tick the agent observed byte growth on, as opposed
+    // to the `[periodStart, periodStart + periodSeconds]` FLUSH window the
+    // envelope sits inside. Both present ⇒ [[spanOf]] uses them; either absent
+    // (pre-#2025 agents, NULL columns) ⇒ it falls back to the flush window, so
+    // the over-count defense degrades gracefully. See `traffic_reports`
+    // active_start / active_end (V62).
+    activeStart: Option[Instant] = None,
+    activeEnd: Option[Instant] = None,
 )
 
 /**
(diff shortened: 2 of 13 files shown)


sameerparekh:

Reviewed the code delta (PR base = the schema-only branch for #2027; this diff adds the wire field, agent emission, persist/read paths, presence math, tests, docs). Cited `file:line` against the 8-dimension checklist.

### BLOCKERS
(none)

### SHOULD-FIX
1. **`Presence.effectiveGap` change is global to every presence surface** (`api/src/presence/Presence.scala:138`) — deriving `R` from `spanOf` instead of `period_seconds` is correct and load-bearing (without it the `2×R` clamp re-bridges the bursts), but it silently changes the gap for *every* consumer (`patternSecondsByMac`, `proportionalHostSeconds`, `appSpansForProfile`, `allocateSharedHostSeconds`), not just the daily total. Mitigation present: for old/NULL-envelope rows `spanOf` returns the flush window so `R` is unchanged (prior behavior preserved), and the full `mill api.test` suite is green. Calling it out so the blast radius is explicit, not hidden — no change requested beyond the test evidence already provided.

### NITS
1. **TDD red→green not a separate commit** (`api/test/.../PresenceSpec.scala`, `TimeStatusServiceSpec.scala`, `TimeUsedRollupSpec.scala`) — the two `feat` commits each bundle tests + impl rather than a failing-test commit before the impl commit. The red→green did happen (the stall fixtures failed and surfaced the `effectiveGap` half before the fix landed; documented in the PR body), but it isn't visible as a standalone red commit in history. Tests are coupled to the new schema fields so a cleanly-compiling red commit would have needed the model stubs anyway.
2. **`minIso`/`maxIso` lexical compare** (`api/src/routes/RouterIngestService.scala:512`) — relies on both envelope timestamps being fixed-width RFC3339 `…Z`. That holds (single agent per batch, agent always formats via `os.date("!%Y-%m-%dT%H:%M:%SZ")`), and a malformed value degrades to the flush-window fallback at `parseInstantOpt`, so worst case is "fall back," never a wrong merge. Parsing to `Instant` before merging would be marginally more robust but is unnecessary here.
3. **No envelope-vs-fallback observability metric** — a counter of rows crediting the activity window vs falling back to the flush window would let an operator confirm the fix is taking effect as the fleet upgrades. Not required (no new route/job/external call; the fix is a structural correctness change), and the rollout is already observable via the #2016 headline numbers; flagging as optional follow-up.

- **SSOT (1):** no re-derivation — `spanOf`/`stitch`/`unionSeconds` stay the single primitive; the rollup and all-live paths both read through it (pinned by the new `dayState ⇄ dayStateLive` test). ✓
- **Wire contract (4):** additive optional `activeStart`/`activeEnd`; both sides tolerate absence; golden `usage_report.json` regenerated and round-tripped through the production decoder by `ContractGoldenSpec`. ✓
- **Architecture (3):** no router decision logic added (the agent only *records* timestamps); no DNS-enforcement reasoning; no new snapshot field. ✓
- **Migrations (5):** code PR adds no migration (`git diff <base>...HEAD -- api/resources/db/migration/` is empty); schema lives in #2027. The "PR base must be main" gate will fail transiently until #2027 merges and GitHub auto-retargets — expected for a stacked migration PR.
- **Testing (2):** new behavior asserted at unit + feature + real-rollup levels; no existing test weakened/deleted; no forbidden mocks; `Clock.TestClock` used. ✓
- **Lua 5.1 (3):** validated on real Lua 5.1.5 (openwrt.lan) per the no-mock-5.1 practice. ✓
- **Security (8):** parameterized Doobie throughout; no secrets; no new routes. ✓

VERDICT: APPROVE
Self-describing usage buckets: the agent emits the real activity envelope and the server credits `[activeStart, activeEnd]` (falling back to the flush window for old agents), with `effectiveGap` made envelope-aware so the `2×R` clamp can't re-bridge bursts — making the #2016 over-count structurally impossible. No blockers; the global `effectiveGap` change is the one thing to keep in view but is covered by the green full-suite run and the old-data fallback. Stacked on #2027 (migration); merge that first.

sameerparekh:
## Independent PR review — re-run (post-rebase)

**Delta since last review (`62ca15f9`):** pure rebase onto `origin/main` after migration PR #2027 squash-merged. The migration commit was dropped (already on `main`) and the two `feat` commits were replayed unchanged — `git diff 62ca15f9 HEAD` is **empty** (identical trees). PR base retargeted to `main`. No reviewed code changed.

### Prior findings — status
|---|---|---|
| SHOULD-FIX 1 | `effectiveGap` change global to all presence surfaces (`Presence.scala`) | UNCHANGED — still covered by old-data fallback + green full-suite run; no code change |
| NIT 1 | TDD red→green not a separate commit | UNCHANGED — history note only |
| NIT 2 | `minIso`/`maxIso` lexical compare (`RouterIngestService.scala`) | UNCHANGED — degrades to fallback, not a wrong merge |
| NIT 3 | No envelope-vs-fallback observability metric | UNCHANGED — optional follow-up |

### New findings (delta)
(none — rebase only; no content change)

### Migration isolation
Now that #2027 merged, V62 is on `main`; this PR adds **no** migration (`git diff origin/main...HEAD -- api/resources/db/migration/` is empty), and the base is `main`. The earlier transient "PR base must be main" failure is resolved by the retarget.

VERDICT: APPROVE
No blockers; the rebase introduced no changes. All substantive CI was green pre-rebase and is re-running on the new HEAD. Ready for merge-when-ready once checks settle (operator merges).

sameerparekh:
**Scope update (operator request):** this PR now also covers the **Traffic Usage / bandwidth** path so the same real activity window drives accurate bandwidth.

Added (3rd commit, `feat(#2025): compute accurate bandwidth from the real activity window`):
- raw `traffic_reports` rows carry `activeStart`/`activeEnd` into `TrafficUsageDbRow`; the aggregate's per-bucket active duration is the **UNION** of the rows' envelopes (concurrent multi-IP of one host don't double-count), flush-window fallback for old agents.
- `traffic_hourly`/`traffic_daily` rollups carry `active_window_seconds` (**new migration #2030**, stacked); the reroll writes `Σ (active_end-active_start)` (flush-window fallback) and the read tiers SUM it.
- `TrafficUsageRawRow` + `TrafficUsageAggregateRow` expose `activeWindowSeconds` (additive; SPA bandwidth math is a separate change). `totalSeconds` kept for back-compat.
- Union/sum reuses `Presence.unionSeconds` (SSOT with the screen-time path). New tests: raw-tier union (incl. overlap), flush-window fallback, rollup-tier sum, rollup reroll carry + fallback.

**Stacking:** base is now `claude/2025-rollup-active-window` (#2030, schema-only). The "PR base must be main" check fails transiently until #2030 merges and GitHub retargets this PR to `main` — same as the V62 dance. Merge order: **#2030 → #2029**.

`mill api.test` full suite green (incl. the #2028 raw-bucket merge); `scalafmt` clean.

sameerparekh:
## Independent PR review — re-run (bandwidth scope added)

**Delta since last review (`1a503f7d`):** one new commit `feat(#2025): compute accurate bandwidth from the real activity window` (+ a clean rebase onto current `main`, which absorbed #2028's `stepOf: Option[Duration]`/`windowFor` refactor — conflict resolved keeping #2028's version and adding `activeWindowSecondsFor`). Base retargeted to the V63 schema branch #2030. Reviewed the new commit against the 8 dimensions.

### Prior findings — status
|---|---|---|
| SHOULD-FIX 1 | `effectiveGap` global to all presence surfaces (`Presence.scala`) | UNCHANGED — covered by old-data fallback + green suite |
| NIT 1 | TDD red→green not a separate commit | UNCHANGED — new commit also bundles tests+impl (red→green was observed: the new bandwidth tests + the prior stall tests) |
| NIT 2 | `minIso`/`maxIso` lexical compare | UNCHANGED |
| NIT 3 | No envelope-vs-fallback observability metric | UNCHANGED — now also applies to the bandwidth path (still optional) |

### New findings (bandwidth delta)
**BLOCKERS** — none.

**SHOULD-FIX**
1. `RollupRepo.scala` reroll (`active_window_seconds = SUM(COALESCE(EXTRACT(EPOCH FROM active_end-active_start), EXTRACT(EPOCH FROM period_end-period_start)))`) — within a rolled `(mac, hostname, hour|day)` the source rows are sequential flush windows of one host (non-overlapping → the SUM equals the union), but a hostname that resolves to multiple dst-IPs concurrently active in the same bucket will have its durations **summed, not unioned**, slightly over-stating the denominator (→ under-stating bandwidth). This is the documented rollup-tier limitation (the raw tier unions exactly via `activeWindowSecondsFor`); acceptable for a bandwidth estimate and called out in the V63 migration doc. No change requested — flagged for visibility.

**NITS**
1. `RollupRepo.scala` reroll — a backwards envelope (`active_end < active_start`, clock skew) would contribute a negative term to the SUM. The agent's clock is monotonic within a bucket so this shouldn't occur; `GREATEST(EXTRACT(...), 0)` would harden it. Raw tier is already clamped (`activeInterval` + `Presence.Span.seconds.max(0)`).
2. `UsageTraffic.activeWindowSecondsFor` assumes a display bucket is single-tier (all raw → union, all rollup → sum); that holds because `UsageTrafficQuery.aggregate` fetches from exactly one tier and `asDbRows` always sets `activeWindowSeconds` on rollup rows. Documented in the method scaladoc.

- **SSOT (1):** the union/sum reuses `Presence.unionSeconds`/`Presence.Span` — same primitive as the screen-time path, no second implementation. `totalSeconds` (sampled) kept untouched for back-compat; `activeWindowSeconds` is the new accurate denominator. ✓
- **Wire contract (4):** `TrafficUsageRawRow` + `TrafficUsageAggregateRow` gain `activeWindowSeconds` additively (default `0L`); API→SPA only, old SPA ignores it. `RollupRow`/`TrafficUsageDbRow` additions are internal. ✓
- **Migrations (5):** code PR adds no migration (verified empty); schema is #2030. `active_window_seconds` is read alongside existing per-(mac,bucket) rollup SELECTs — no new access pattern/index. ✓
- **Testing (2):** new behavior asserted (raw union incl. overlap, flush-window fallback, rollup-tier sum, reroll carry + fallback); no existing assertion weakened; embedded Postgres for the rollup reroll. Full `mill api.test` green incl. the #2028 merge. ✓

VERDICT: APPROVE
Bandwidth now derives from the real activity window at every read tier — raw rows union their envelopes, rollups carry summed envelope seconds — so a flush stall (#2016) no longer dilutes the rate. No blockers; the rollup-tier sum-vs-union is a documented, bounded approximation. Stacked on #2030 (schema); merge #2030 → #2029.

sameerparekh:
## Independent PR review — re-run (NIT addressed)

**Delta since last review (`9a6d1fd2`):** amended the bandwidth commit to wrap the rollup `active_window_seconds` SUM in `GREATEST(…, 0)` (both hourly + daily reroll) — addressing prior NIT 1. No other change.

### Prior findings — status
|---|---|---|
| SHOULD-FIX 1 (`effectiveGap` global) | UNCHANGED — covered by old-data fallback + green suite |
| SHOULD-FIX 1 (rollup sum-not-union, cross-host) | UNCHANGED — documented bounded approximation, no change requested |
| NIT 1 (backwards-envelope negative term in reroll) | **ADDRESSED** — `GREATEST(COALESCE(…), 0)` now clamps each per-row term in both reroll SQLs, matching the raw tier |
| NIT 2 (single-tier assumption) | UNCHANGED — documented in `activeWindowSecondsFor` scaladoc |
| NIT 3 (no envelope-vs-fallback metric) | UNCHANGED — optional follow-up |

### New findings (delta)
(none — one-line SQL hardening; `RollupRepoSpec` active-window tests still green, full `mill api.test` green on rerun)

VERDICT: APPROVE
Clean re-run: the only change is the defensive `GREATEST(…,0)` clamp on the rollup active-window sum. No open blockers. Stacked on #2030 (schema) — merge #2030 → #2029.

sameerparekh:
**Scope reverted to screen-time only (operator decision).** Dropped the bandwidth commit and re-based onto `main`; base retargeted `main`. The bandwidth follow-up (rollup `active_window_seconds` + traffic-usage `activeWindowSeconds` plumbing) and its migration PR #2030 are removed/closed: for traffic over a period the rate is **bytes ÷ total time** (window width), not bytes ÷ active time, so that column was the wrong (and not cleanly deduplicated) denominator. This PR is back to the two screen-time commits (agent envelope emission + server `spanOf`/`effectiveGap` on the real activity window), on V62 (#2027, merged). Full `mill api.test` green; scalafmt clean.

sameerparekh:
## Independent PR review — re-run (bandwidth reverted)

**Delta since last review (`c9053c4b`):** the bandwidth commit was dropped and the two screen-time commits re-based onto `main`; base retargeted to `main`; migration PR #2030 closed. The current diff is **screen-time only** (no `UsageTraffic`/`RollupRepo`/`UsageTrafficQuery`/V63). Reviewed the now-removed surface + the screen-time code as it stands.

### Prior findings — status
|---|---|---|
| SHOULD-FIX (`effectiveGap` global to presence surfaces, `Presence.scala`) | UNCHANGED — screen-time code, still covered by old-data fallback + green suite |
| SHOULD-FIX (rollup sum-not-union, cross-host) | **REMOVED** — bandwidth/rollup code reverted |
| NIT (backwards-envelope clamp in reroll) | **REMOVED** — reverted with the rollup code |
| NIT (single-tier assumption in `activeWindowSecondsFor`) | **REMOVED** — reverted |
| NIT (`minIso`/`maxIso` lexical compare, `RouterIngestService.scala`) | UNCHANGED — degrades to flush-window fallback, never a wrong merge |
| NIT (no envelope-vs-fallback observability metric) | UNCHANGED — optional follow-up |

### New findings (delta)
(none — the delta is a clean removal of the bandwidth half; the screen-time code is byte-identical to the version reviewed at `1a503f7d` and is APPROVE)

### Current scope (screen-time only)
- Agent emits `activeStart`/`activeEnd` (real activity envelope) on usage records (`usage.lua`, `wifihaven-agent`); persisted to `traffic_reports.active_start/active_end` (V62, #2027 merged).
- `Presence.spanOf` uses `[activeStart, activeEnd]` when present (flush-window fallback for old agents); `effectiveGap` derives `R` from `spanOf` so the `2×R` clamp can't re-bridge bursts — the #2016 over-count is structurally bounded.
- Regression tests: unit `spanOf` (envelope/fallback/clamp/half-present), stall fixture (bounded usedMinutes) at unit + `dayStateLive` + rollup `dayState ⇄ dayStateLive`, invariants (per-app ⊆ profile non-exempt; Dedup ≤ wall-clock ≤ 1440); agent busted for tracker + emission. Contract golden round-trips the new optional wire fields.

VERDICT: APPROVE
Scope correctly reduced to the screen-time fix (active time is right for presence; bandwidth correctly uses bytes/total-time elsewhere and needed no change). No blockers. Base `main`, full `mill api.test` green, scalafmt clean — ready for merge-when-ready.

sameerparekh:
## Independent PR review — re-run (CI re-trigger, no content change)

**Delta since last review (`f3110ec3`):** a no-op `git commit --amend` to bump the SHA and fire a fresh `synchronize` event — the prior "PR base must be main" check had a stale **failed** run frozen from the pre-retarget force-push (re-runs replay the original event payload, so they kept failing even though the base is now `main`). The tree is **byte-identical** to `f3110ec3` (`git diff f3110ec3 HEAD` is empty); no code changed.

### Prior findings — status
All carried-forward findings unchanged from the `f3110ec3` review (screen-time-only scope):
- SHOULD-FIX `effectiveGap` global — UNCHANGED, covered.
- NIT `minIso`/`maxIso` lexical — UNCHANGED, degrades to fallback.
- NIT no envelope-vs-fallback metric — UNCHANGED, optional.
- (bandwidth findings remain REMOVED.)

### New findings
(none — pure SHA bump to clear the stale base-gate run)

VERDICT: APPROVE
Screen-time-only fix on `main`; the re-push only re-triggers CI so the base-gate evaluates against the correct base. No blockers — ready for merge-when-ready once the fresh checks go green.

PR #2030, opened 2026-06-28
chore(#2025): traffic_hourly/daily active_window_seconds (schema-only migration)

sameerparekh:
Schema-only migration for the #2025 bandwidth follow-up (accurate bandwidth from the real activity window).

## What
Adds a nullable `active_window_seconds BIGINT` to **`traffic_hourly`** and **`traffic_daily`**.

It stores, per `(router, mac, hostname, bucket)`, the summed **real** active-window seconds — `Σ (active_end - active_start)` over the bucket's source rows, falling back to the `[period_start, period_end]` flush window per row when the envelope is absent (pre-V62 rows / pre-#2025 agents).

## Why
V62 added `traffic_reports.active_start/active_end` (the real activity envelope). The Traffic Usage / live-bandwidth read path tiers its reads across raw `traffic_reports` + these two rollups; at the rollup tiers the only duration available is the **sampled** `active_seconds`, which under-samples during a flush stall (#2016) while the flush window over-states elapsed — so a bandwidth rate computed from either is wrong. This column gives the rollup tiers the same real active duration the raw tier already has, so `bytes ÷ active_window_seconds` is accurate everywhere.

## Isolation
Schema-only per AGENTS.md §migrations-back-compat. The follow-up **code PR** (#2029, stacked on this branch) adds the reroll write, the read, and the aggregate use.

## Cost
`traffic_hourly` / `traffic_daily` are derived growth tables. `ADD COLUMN ... BIGINT NULL` is metadata-only (no rewrite), safe on the startup critical path; the column backfills naturally as the rollup fiber re-rolls recent windows.

### DIFF
--- a/api/resources/db/migration/V63__traffic_rollups_active_window.sql
+++ b/api/resources/db/migration/V63__traffic_rollups_active_window.sql
@@ -1,0 +1,32 @@
+-- V63__traffic_rollups_active_window.sql
+-- #2025 (bandwidth follow-up): carry the real activity-window duration into the
+-- traffic_hourly / traffic_daily rollups so the Traffic Usage / live-bandwidth
+-- read path can compute accurate bandwidth (bytes ÷ real active duration) at the
+-- rollup tiers, not just the raw tier.
+--
+-- Background: V62 added traffic_reports.active_start/active_end (the real
+-- activity envelope per bucket). The screen-time path already uses it
+-- (Presence.spanOf). The Traffic Usage aggregate (UsageTrafficQuery) tiers its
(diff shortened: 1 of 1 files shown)


sameerparekh:
Closing — the bandwidth follow-up this migration supported is being reverted. For traffic over a period the correct rate is **total bytes ÷ total time** (the window width), not bytes ÷ active time, so the rollup `active_window_seconds` column isn't needed (and as discussed it wasn't a clean profile/app active time anyway). The screen-time fix (#2029, on V62) stands on its own. If we later want an *active transfer time* metric, it'll be a separate, properly-deduplicated computation — not a bandwidth denominator.

Issue #2034, opened 2026-06-28
ci(ops): Gate-2 fake-VM external-DNS egress flake (Mode 2 of #2033)


sameerparekh:
Split out from #2033 (Mode 2). #2033's PR delivered **Mode 1** (qemu client-VM boot resilience in `scripts/vm/client-up.sh`: bounded boot retry, early-crash detection, bridge-readiness wait, and console-log preservation for CI artifact upload). This issue tracks the remaining **Mode 2**.

## Problem (Mode 2)
Gate-2 fake-VM scenarios that depend on the router VM's dnsmasq upstream (`WH_ROUTER_LAN_RESOLVER`, default `192.168.10.1`) intermittently redden when that upstream can't resolve/egress:

- `test_eb_direct_requery_blocked` and siblings: `policy.apply: smoke check failed; dnsmasq may be serving a stale config — got "nil" for <host>` (openwrt agent `policy.lua:335`).
- blocklist `fetch failed (status=404)` (`blocklists.lua:200`).

This is intermittent *external egress* on the shared CD host, NOT test isolation (see memory `project_gate2_flaky_egress_not_isolation`; long-standing #1935; cf. the #1865 degraded-window guards in `test_v6_block_page.py`).

## Options (do NOT weaken what the test verifies — target only the flaky external dependency)
1. Make the egress-dependent scenarios deterministic: serve the required hosts / blocklist URLs from the in-process fake-API (`scripts/e2e/fake-api/`) or a local resolver, so resolution does not depend on host egress.
2. Add a bounded retry on the dnsmasq-upstream smoke (the `policy.apply` smoke check) before declaring failure.
3. Guard the egress-sensitive assertions so a degraded-egress window doesn't redden a sound enforcement path (the #1865 approach), while still failing on a genuine enforcement regression.

## Constraints
- KVM/VM e2e can't run on macOS — validate via the shell specs + reasoning; rely on CD (dispatch `e2e-vm-fake.yml`).
- Low priority (we re-run today) but a recurring tax.

Issue #2033, opened 2026-06-28
ci(ops): harden Gate-2 fake-VM flakiness — qemu client boot timeout (bridge-pool) + fake-VM external-DNS egress


sameerparekh:
## Problem
Master Router CD's **Gate 2 (fake-API VM e2e)** intermittently reddens on infra (not code), forcing manual re-runs. Two recurring, distinct modes observed across many runs (e.g. 28324121167, 28299729135):

1. **qemu client VM boot timeout** — `ERROR at setup of <test>`: `[client-up] '<name>' did not become reachable within 90s` → `scripts/vm/client-up.sh` rc=1. Hits a *different* arbitrary test each time (whichever runs when the VM won't boot).
2. **fake-VM external-DNS egress flake** — `test_eb_direct_requery_blocked` (and siblings) fail with `policy.apply: smoke check failed; dnsmasq may be serving a stale config — got "nil" for <host>` / blocklist `fetch failed (status=404)`. The router VM's dnsmasq upstream (`WH_ROUTER_LAN_RESOLVER`, default 192.168.10.1) intermittently can't resolve/egress. Long-standing (#1935; see also #1865 v6 block-page degraded-window guards).

## NOT a single-runner problem
All CD runners are on the **same host**; the only per-job difference is the **specific qemu bridge interface**, picked from a pool (`scripts/vm/lan-bridge-pool-bootstrap.sh`, `lan-bridge-up.sh`/`lan-bridge-down.sh`, `test-lan-bridge-pick.sh`). So mode (1) is almost certainly **bridge-pool contention / readiness / stale-cleanup races** between parallel jobs sharing the host — not faulty hardware on one runner. Earlier framing as "actions-runner-2 health" was wrong.

## Investigation + hardening (CI/ops; the bug is in the harness, not product code)
**Mode 1 — VM boot / bridge pool:**
- Audit bridge pick/allocation/teardown for races: can two concurrent jobs grab the same bridge? Is a bridge reused before its prior client is fully torn down (`client-down.sh`/EXIT trap)? Is `tap`/`bridge.conf` setup racy? Is the bridge guaranteed up+ready before `client-up.sh` attaches (it only checks `ip link show` exists at client-up.sh:75)?
- Add resilience: a **boot retry** in client-up.sh (one retry on the 90s timeout, fresh tap/bridge re-pick) and/or per-job bridge isolation that can't collide. Keep headroom sensible (clients normally come up ~5s).
[...]

PR #2035, opened 2026-06-28
ci(#2033): harden Gate-2 client-VM boot — retry + crash detection + console preservation

sameerparekh:
Fixes Mode 1 of #2033 (Gate-2 fake-VM flakiness). Mode 2 (external-DNS egress) split to #2034 to keep this PR focused and reviewable.

## Problem
Master Router CD's **Gate 2 (fake-API VM e2e)** intermittently reddens with `[client-up] '<name>' did not become reachable within 90s` → `scripts/vm/client-up.sh` rc=1, hitting an arbitrary scenario each time. All CD runners share one host; the bridge **pick/reservation** is already well-hardened (atomic `flock` + per-bridge reservation marker in `lib.sh::wh_pick_lan_bridge`, #891/#907/#1369), so the durable fix is **client-VM boot resilience**, not more bridge-pool plumbing.

Worse, a latent diagnostic bug made these failures un-debuggable: `client-up.sh`'s EXIT trap did `rm -rf "${RUN_DIR}"` on timeout — **deleting `console.log` before the CI artifact-upload step**. And a boot timeout is a pytest **setup** failure, which the `scenarios_fake/conftest.py` `pytest_runtest_makereport` hook (guarded on `call.when == "call"`) does not capture, and the `client_factory` finalizer's `client_down` then `rm -rf`s the run dir anyway. Net: mode-1 failures landed with **no console at all**.

## Changes (`scripts/vm/client-up.sh`)
- **One boot retry** (`WH_CLIENT_BOOT_ATTEMPTS=2`, override-able; per-attempt `WH_CLIENT_BOOT_TIMEOUT=90`): a fresh-overlay re-spawn on timeout turns a transient first-boot miss (DHCP not yet served, tap carrier lag, wedged boot) into a ~90s blip instead of a job failure. Normal ~5s path unchanged.
- **Early-crash detection**: if qemu exits during boot (e.g. failed tap attach), fail the attempt immediately instead of burning the full timeout; a non-zero qemu launch (transient `qemu-bridge-helper` failure under bridge contention) is a failed *attempt* (→ retry), not a `set -e` abort that bypasses the retry.
- **Bridge-readiness wait**: poll `IFF_UP` on `WH_LAN_BRIDGE` before attaching (the old check at line 75 only verified the netdev *existed*). Best-effort, never fatal; skips cleanly where `/sys` is absent.
- **console.log preservation**: the EXIT trap now removes only the disposable overlay + control socket (not the run dir), and on total boot failure the console(s) are copied to a sibling `.run/<id>/boot-failures/` dir that `client_down` (which only `rm -rf`s the per-`NAME` dir) does **not** touch — so the boot console is finally captured by the existing `scripts/vm/.run/**` upload glob. Per-attempt failed consoles are kept as `console.attemptN.log`.

## Tests (`scripts/vm/client-up_spec.sh`)
Counter-driven stubs (monotonic `date`, threshold `ssh`) replace the single-probe `date` stub so the spec can exercise multiple attempts. New/updated cases: console-preserved-on-timeout, `boot-failures/` copy, retry succeeds on 2nd attempt, early-crash detected. **9/9 pass; `shellcheck -x` clean** (only the pre-existing SC1091 source-resolution info).

## Validation
- `bash scripts/vm/client-up_spec.sh` → 9/9 pass (run on macOS; `/sys` + `/dev/kvm` absent paths exercised).
- `bash scripts/vm/test-lan-bridge-pick.sh` → still 6/6 (no bridge-pool behavior changed).
- **KVM/VM e2e can't run on macOS** — full Gate-2 validation is via CD. `e2e-vm-fake.yml` dispatched on this branch; re-running it a couple of times is the real proof it's no longer flaky.

### DIFF
--- a/scripts/vm/client-up.sh
+++ b/scripts/vm/client-up.sh
@@ -100,33 +100,52 @@ fi
 rm -rf "${RUN_DIR}"
 mkdir -p "${RUN_DIR}" "${WH_SOCK_DIR}"
 
-# EXIT trap: if the client never becomes reachable (or any failure between
-# here and the "client booted" sentinel), kill the qemu we just spawned and
-# remove the run directory so no orphan is left behind. Mirrors the pattern
-# from router-up.sh (#1225). Once the client is confirmed reachable, set
-# _client_booted=1 so the trap becomes a no-op on the success path.
-_client_booted=0
-# shellcheck disable=SC2329  # invoked via trap EXIT, not a direct call
-_cleanup_on_fail() {
-  [[ "${_client_booted}" == "1" ]] && return 0
-  local _pid=""
-  if [[ -f "${RUN_DIR}/qemu.pid" ]]; then
-    _pid="$(cat "${RUN_DIR}/qemu.pid" 2>/dev/null || true)"
-  fi
-  if [[ -n "${_pid}" ]] && kill -0 "${_pid}" 2>/dev/null; then
-    echo "[client-up] cleaning up orphan qemu (pid ${_pid})" >&2
-    kill -KILL "${_pid}" 2>/dev/null || true
-  fi
-  rm -rf "${RUN_DIR}"
-}
-trap _cleanup_on_fail EXIT
-
 if [[ -z "${SSH_PORT}" ]]; then
   SSH_PORT="${WH_CLIENT_SSH_PORT_BASE}"
 fi
 
 OVERLAY="${RUN_DIR}/overlay.qcow2"
-qemu-img create -f qcow2 -F qcow2 -b "${WH_CLIENT_BASE_IMG}" "${OVERLAY}" >/dev/null
+PIDFILE="${RUN_DIR}/qemu.pid"
+# QMP socket lives under WH_SOCK_DIR (short path) — see config.sh WH_SOCK_DIR
+# note for the AF_UNIX sun_path reason.
+QMP_SOCK="$(wh_client_qmp_sock "${NAME}")"
+
+# Per-attempt SSH-reachability budget (clients normally come up in ~5s) and the
+# number of boot attempts. One retry by default (#2033): on a loaded shared host
+# the occasional client never finishes its first boot (DHCP not yet served, tap
+# carrier lag, a wedged first boot), reddening an arbitrary scenario with
+# "did not become reachable". A single fresh-overlay retry turns that transient
+# into a ~90s blip instead of a job failure, while the normal ~5s path is
+# unchanged. Overridable for tests / unusual hosts.
+WH_CLIENT_BOOT_TIMEOUT="${WH_CLIENT_BOOT_TIMEOUT:-90}"
+WH_CLIENT_BOOT_ATTEMPTS="${WH_CLIENT_BOOT_ATTEMPTS:-2}"
+
+# EXIT trap: if the client never becomes reachable (or any failure between
+# here and the "client booted" sentinel), kill the qemu we last spawned and
+# drop the disposable overlay + control socket. Mirrors router-up.sh (#1225).
+# Crucially it does NOT remove ${RUN_DIR}: the serial console.log (and any
+# per-attempt console.attemptN.log) must survive so the CI "Upload failure
+# artifacts" step can capture them — a boot timeout was previously
+# undiagnosable because this trap rm -rf'd the whole run dir, deleting
+# console.log before upload (#2033). Leaving the dir behind is safe: the next
+# client-up start reclaims+wipes it (stale-reclaim above) and client-down.sh
+# removes it in teardown. Once the client is reachable, _client_booted=1 makes
+# the trap a no-op on the success path.
+_client_booted=0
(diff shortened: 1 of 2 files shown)


sameerparekh:
## Independent PR review — `174fc93`

Read-only adversarial pass against the 9-dimension checklist. Diff is two host-side harness files (`scripts/vm/client-up.sh`, `scripts/vm/client-up_spec.sh`); no product code, no wire/snapshot/router/SQL/metric surface touched, so dims 3–6 and 8 are largely N/A. Findings below.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. **`client-up.sh:178` — `seq 1 20` × `sleep 0.5` (=10s) bridge-readiness budget is an unsourced literal.** It's bounded and explicitly *never fatal* (the SSH loop is the real gate), so it doesn't drive behavior — but a one-line note on why 10s (vs. the 90s boot budget) would preempt a future "is this enough?" question. Dim 9, low severity.
2. **`client-up.sh:236` (inter-attempt cleanup) vs. `_cleanup_on_fail:135` — qemu-kill logic is duplicated** (read `PIDFILE` → `kill -0` → `kill -KILL`). This is deliberate belt-and-suspenders (loop reclaims the failed attempt before re-spawn; trap is the final net), matching the pre-existing pattern, so not a SSOT violation — noting only so a future edit keeps both in step.
3. **Worst-case wall time on a genuinely-broken setup doubles** (2 × `WH_CLIENT_BOOT_TIMEOUT` = 180s per client). Only hit on real failure (rare), and far inside the 75m gate cap — acceptable, but worth being aware the retry trades fail-fast for resilience.
4. **PR body carries a "🤖 Generated with Claude Code" footer** (dim 7 flags AI mentions in the body). Harness-mandated and per the operator's standing position the `Co-Authored-By` trailer / authorship is fine; flagging only because the checklist enumerates it. Commit message + PR title are clean.

### Dimension notes (verified, not findings)
- **Dim 2 (testing integrity).** The old `client-up_spec.sh` "(a) no orphan qemu.pid after timeout" assertion was **intentionally replaced**, not weakened to hide a regression: the new code deliberately stops `rm -rf`-ing `${RUN_DIR}` (to preserve `console.log`), so a dead pidfile legitimately remains. The real invariant — *no live orphan qemu* — is still asserted ("(a) qemu process killed"), and two stronger assertions were added (console preserved in-dir + under `boot-failures/`). The behavior change is explained in the PR body. The single-probe `date` stub was correctly reworked to a monotonic counter because the retry loop re-enters the wait (the old stub would have hung attempt 2). Justified.
- **Dim 2 (coverage).** Retry-success, early-crash detection, console-preservation, and `boot-failures/` copy are all asserted (9/9 pass). The dead-pidfile-leftover → fresh-boot path on the *next* invocation isn't directly tested but is covered by the unconditional `rm -rf "${RUN_DIR}"` at `client-up.sh:100`.
- **Dim 3 (`/tmp` bounded-writes).** N/A — `boot-failures/` is under the CI runner workspace `scripts/vm/.run/<id>/` (unique + wiped per job), not the OpenWRT router agent's tmpfs `/tmp`. Bounded by construction (one client name → one overwritten console set per job).
- **Dim 7 (scope).** Focused to the two intended files; no drive-by edits; TODO-free.

VERDICT: APPROVE

Durable, well-scoped Mode-1 fix: client-VM boot retry + early-crash detection + bridge-readiness wait, plus the genuinely valuable console-log preservation (boot timeouts were undiagnosable because the trap deleted `console.log` before upload, and a setup-phase failure escapes the conftest makereport hook). Spec coverage is real (9/9, shellcheck-clean) and the one existing-assertion change is intentional and explained. No blocking concerns; full Gate-2 proof is correctly deferred to CD (KVM can't run on macOS).

sameerparekh:
## Independent PR review (re-run) — `0587978`

Delta since `174fc93`: a 3-line clarifying comment on the bridge-readiness budget (`scripts/vm/client-up.sh:178`). No logic change.

### Prior findings — status
|---|---|---|
| NIT 1 | `seq 1 20 × sleep 0.5` (10s) readiness budget unsourced | **ADDRESSED** — `client-up.sh:178` now documents why ~10s vs. the 90s boot budget (carrier-lag absorber, not a bring-up wait). |
| NIT 2 | qemu-kill logic duplicated (loop vs. trap) | NOT-ADDRESSED — deliberate belt-and-suspenders; non-blocking, no action expected. |
| NIT 3 | worst-case wall time doubles on broken setup | NOT-ADDRESSED — accepted trade-off (resilience over fail-fast); well within the 75m cap. |
| NIT 4 | PR-body AI footer | NOT-ADDRESSED — operator-sanctioned; non-blocking. |

### New findings in delta
(none — comment-only change, shellcheck-clean, spec still 9/9.)

VERDICT: APPROVE

Comment-only follow-up that resolves the one actionable NIT. No new code paths, no new findings. Mergeable; full Gate-2 proof is via CD (e2e-vm-fake.yml dispatched on the branch).

Issue #2042, opened 2026-06-28
Replace wall-clock ZIO.sleep waits in feature tests with TestClock (flaky MetricsExportSpec + class)


sameerparekh:
## Problem

Several feature specs wait on **wall-clock `ZIO.sleep` under `TestAspect.withLiveClock`** to let a background fiber / Prometheus snapshot poller catch up, then assert. This violates our standing testing rule (`docs/process/testing.md`: *"Clock is always injected — use `Clock.TestClock` in tests"*) and is flaky by construction: the sleep is a race against real wall-time, so under heavy parallel CI load (14 workers) the awaited work hasn't landed when the assert runs.

This bit us concretely: `MetricsExportSpec` "*driving a rollup tick emits the rollup counter/histogram/gauge series*" **failed in CI for an unrelated PR** (#2037's branch) purely from contention — it passes in isolation. This is my own code (introduced in #1249 / #1243), owning it.

## The flaky pattern (MetricsExportSpec:78–121)

```scala
fiber <- TimeUsedRollupJob.loop(...).fork
_     <- ZIO.sleep(600.millis)   // hope the loop ran a tick AND the poller scraped
_     <- fiber.interrupt
_     <- ZIO.sleep(300.millis)   // hope the scrape landed
body  <- exposition
```

## Concrete deterministic fix (already scoped)

- Drop `@@ TestAspect.withLiveClock` and the `Clock.live` layer — let the spec run on the default `TestClock`.
- Drive the work **synchronously**: call `TimeUsedRollupJob.runOnce(runs, clock, now => oneTickForTest(...))` directly (the single-tick metric-emission site, already `private[api]` + `oneTickForTest` is exposed) instead of `.loop().fork`.
- Replace each `ZIO.sleep(...)` with `TestClock.adjust(pollInterval)` so the `connectors.prometheus.prometheusLayer` snapshot fiber scrapes deterministically, then read `exposition`.

The same shape (`withLiveClock` + `ZIO.sleep` to await a poller/fiber) recurs across these feature specs — fix the class, not just the one:

- `api/test/src/feature/MetricsExportSpec.scala` (the one that flaked)
- `api/test/src/feature/RouterMetricsIngestSpec.scala`
- `api/test/src/feature/ErrorBoundarySpec.scala`, `ErrorBoundaryRoutesSpec.scala`, `ErrorBoundaryStage2Spec.scala`
[...]

PR #2043, opened 2026-06-28
feat(#2037): policy poll dormant while ws link healthy

### LINKED ISSUE #2037: ws-transport: make the agent's HTTP policy poll dormant while ws is healthy (resume after ws_fallback_after)
## Problem

With ws enabled, the agent **still HTTP-polls `GET /api/router/policy` every 5 s, unconditionally.** The poll cadence (`policy_int = policy_poll_interval`, default 5 s, [`wifihaven-agent`] line ~79) and the `on_tick` poll gate (`if (mono - ts.last_policy_run) >= policy_int`, line ~1048) are **not** conditioned on ws being enabled or healthy. So a ws-enabled router runs the persistent socket *and* the 5 s poll in parallel.

This contradicts the design. `docs/design/websocket-transport.md` §3.1 specifies that on a healthy ws connect *"the main agent's HTTP poll timer **goes dormant**,"* resuming only after ws is down longer than `ws_fallback_after` (~5 min, "belt-and-suspenders"). The whole motivation of #1023 (§ lines 7/30) is to remove the **~500 ms snapshot recompute per poll** (#1512) — which keeps firing every 5 s per router as long as the poll runs. **So until the poll goes dormant, enabling ws delivers the latency win (sub-second push) but NOT the server-side load/scalability win.**

The apply-on-push work (#1945) intentionally left both transports running in parallel — etag dedup makes that safe (exactly-once apply, no two-writer race) — and the doc notes the poll *"still runs (deprecated separately in #1850)."*

## Scope — this is the agent-side dormancy, NOT the endpoint removal

This issue is the **small, agent-only, fully-reversible** change: make the policy-poll timer dormant while ws-health is fresh. It is distinct from #1850 ("deprecate + remove the REST poll/ingest **endpoints**" — server-side, capability-negotiation-gated via #376, irreversible, "LAST"). This one needs no wire change and can land well before #1850; it's the precursor that actually realizes the load win.

## Approach

Gate the policy-poll timer on ws-health freshness, mirroring the **existing outbound tee** (`ws_health_read` / `ws_fallback_after`, [`wifihaven-agent`] ~lines 465–490) that already switches usage/events between the ws spool and HTTP:
- ws enabled **AND** health sentinel fresh (mtime within `ws_fallback_after`) → **skip** the HTTP policy poll. The apply-on-push tick (`ws.apply_interval`, 2 s, reading `/etc/wifihaven/policy.json`) already keeps enforcement current from server pushes.
- health sentinel stale/absent past `ws_fallback_after` → **resume** the 5 s poll (today's behavior).
- Preserve failover: a resumed poll that *fails* must still trigger the failover render path (#422/#331) for closed-mode profiles.

## Watch-outs
- **Liveness**: the server touches `routers.last_seen_at` via the ws heartbeat too (§3.1), so dropping the poll while ws is up does **not** stale out liveness — confirm against `RouterWsRoutes`/`routerRepo.touch`.
- Reversible + no wire change → not #376-gated; safe to ship independently of #1850.

## Acceptance
- [ ] With ws healthy, the router emits **no** `GET /api/router/policy` (verify via server logs / a poll-suppressed metric).
- [ ] Policy changes still apply within `ws.apply_interval` via push while the poll is dormant.
- [ ] Kill ws (or block the socket) > `ws_fallback_after` → the 5 s poll **resumes**; restore ws → poll goes dormant again.
- [ ] Failover render still triggers on a failed poll during the fallback window.
- [ ] Router liveness (`last_seen_at`) stays fresh via the ws heartbeat while the poll is dormant.

## Gating
**Prerequisite for enabling ws on the prod gateway (#1023)** — enabling ws should stop the poll, not double it. Relates to #1850 (this is its agent-side precursor).

sameerparekh:
## What

Make the OpenWrt agent's HTTP policy poll **dormant while the websocket link is healthy**, so enabling ws delivers the server-side load win (no per-poll snapshot recompute, #1512), not just the push-latency win.

Closes #2037.

## Why

With ws enabled the agent ran the persistent push socket **and** still HTTP-polled `GET /api/router/policy` every 5 s, unconditionally. The poll cadence and the `on_tick` poll gate were never conditioned on ws health, so every ws router kept paying the ~500 ms snapshot recompute per poll. The design (`docs/design/websocket-transport.md` §3.1) specifies the poll timer goes **dormant** on a healthy ws connect and resumes only after the link is down past `ws_fallback_after` (default 300 s).

This is the **agent-side, fully-reversible, no-wire-change** precursor to #1850 (server-side endpoint removal, #376-gated).

## How

- Lifted the freshness predicate the outbound tee already uses into **`ws_outbound.is_healthy`** — a single source of truth so the poll-dormancy gate and the usage/events tee cannot drift.
- `wifihaven-agent` on_tick: when ws is enabled **and** the health sentinel is fresh, **skip** the HTTP poll. The apply-on-push tick (#1945, reads `/etc/wifihaven/policy.json`) keeps enforcement current from server pushes.
- Poll **resumes exactly as today** — including the #331/#422 failover render on a failed poll — when ws is disabled or the sentinel is stale/absent past `ws_fallback_after`. The failover edge is untouched (the resumed poll runs the same path).
- **Liveness verified**: the server touches `routers.last_seen_at` on every recognized ws frame incl. ping/pong heartbeat (`RouterWsRoutes.touch`, RouterWsRoutes.scala:148–159/223), so dropping the poll while ws is up does not stale liveness. No server code changed.

## Instrumentation

- New bounded-label counter `policy_poll_skipped_total{reason="ws_healthy"}` (one increment per suppressed poll) — allowlisted server-side (`MetricGuard.Allowed`) and pinned by `RouterWsAgentMetricsAllowlistSpec`.
- New Grafana panel on `router-ws-transport.json` ("Policy poll suppressed while ws healthy") — the agent-side complement of the existing computed-vs-cache_hit snapshot-build panel.

## Testing

- New busted spec `openwrt/test/ws_poll_dormant_spec.lua` (committed failing first, then green): (a) ws enabled + fresh sentinel ⇒ dormant / no HTTP fetch; (b) stale/absent sentinel past fallback ⇒ poll resumes; (c) a resumed poll that **fails** still trips the #331/#422 failover render (drives the **real** `policy.failover_transition`).
- **Validated on real Lua 5.1 (plex.lan)** per project rule (not just macOS): the agent monolith compiles within the **60-upvalue limit** (#1930); `is_healthy` + the failover-preservation gate model run green under `lua5.1`.
- Full busted suite **983/0**; `RouterWsAgentMetricsAllowlistSpec` green; `scalafmt --check` clean.

## Notes for reviewers

- The `else` body in the on_tick policy block is intentionally **not re-indented** to keep the diff minimal (~100 lines would otherwise churn); a `-- #2037: close the else` marker tags the added `end`. Lua doesn't care about the indentation; real-5.1 compile confirms the structure.
- A pre-existing flaky timing test (`MetricsExportSpec`, my code from #1249) surfaced during this work's full-suite run — filed as #2042 (deliberately **not** folded into this PR to keep it focused).

### DIFF
--- a/api/src/metrics/Metrics.scala
+++ b/api/src/metrics/Metrics.scala
@@ -177,6 +177,15 @@ object MetricGuard {
     "policy_apply_duration_seconds"             -> Set("router_id", "installation_id"),
     "snapshot_poll_total"                       -> Set("result", "router_id", "installation_id"),
     "snapshot_poll_duration_seconds"            -> Set("router_id", "installation_id"),
+    // #2037 — policy-poll dormancy. When the ws transport is enabled AND its
+    // health sentinel is fresh (within ws_fallback_after), the agent's 5 s HTTP
+    // `GET /api/router/policy` poll goes dormant — the push channel is the live
+    // transport, so the snapshot is no longer recomputed per poll (#1512). This
+    // counter increments once per suppressed poll so the server-side load win of
+    // enabling ws is directly visible (a climbing rate on a ws router with a flat
+    // snapshot_poll_total is the dormancy working). `reason` is a fixed 1-value
+    // enum today (`ws_healthy`) — bounded, no per-mac/host dimension.
+    "policy_poll_skipped_total"                 -> Set("reason", "router_id", "installation_id"),
     "agent_uptime_seconds"                      -> Set("router_id", "installation_id"),
     "agent_version"                             -> Set("version", "router_id", "installation_id"),
     "dns_queries_total"                         -> Set("result", "router_id", "installation_id"),
--- a/api/test/src/feature/RouterWsAgentMetricsAllowlistSpec.scala
+++ b/api/test/src/feature/RouterWsAgentMetricsAllowlistSpec.scala
@@ -39,6 +39,12 @@ object RouterWsAgentMetricsAllowlistSpec extends ZIOSpecDefault {
       hasEntry("ws_frames_sent_total", Set("op")) &&
       hasEntry("ws_frames_recv_total", Set("op"))
     },
+    // #2037 — policy-poll dormancy. The agent emits this once per HTTP poll it
+    // suppressed because the ws link was healthy; the firewall must permit the
+    // name + its bounded `reason` enum, else the whole push is rejected.
+    test("policy_poll_skipped_total carries reason (the dormancy gate enum)") {
+      hasEntry("policy_poll_skipped_total", Set("reason"))
+    },
     test("a forbidden per-mac label on a ws series is rejected") {
       val rejected = Metric.counter("metrics_rejected_total").tagged("reason", "forbidden_label")
       for {
--- a/deploy/grafana/dashboards/router-ws-transport.json
+++ b/deploy/grafana/dashboards/router-ws-transport.json
@@ -616,6 +616,63 @@
           "refId": "A"
         }
       ]
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "title": "Policy poll suppressed while ws healthy (#2037)",
+      "description": "sum by (reason) (rate(policy_poll_skipped_total[5m])) — the agent-side dormancy gate (design §3.1). When ws is enabled AND its health sentinel is fresh (within ws_fallback_after), the agent's 5 s HTTP GET /api/router/policy poll goes dormant (reason=ws_healthy) — the push channel is the live transport, so the server stops recomputing the snapshot per poll (#1512). This is the agent-side complement of the 'Policy snapshot builds: computed vs cache_hit' panel above: a climbing skip rate on ws routers alongside a flat snapshot_poll_total is the load win of enabling ws being realized. A skip rate that drops to 0 while ws_state stays 1 means the dormancy gate regressed (poll running despite a healthy link).",
+      "type": "timeseries",
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 0,
+        "y": 40
+      },
+      "id": 11,
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "unit": "cps",
+          "min": 0,
+          "custom": {
+            "drawStyle": "line",
+            "fillOpacity": 10,
+            "showPoints": "never",
+            "lineWidth": 1
+          }
+        },
+        "overrides": []
+      },
+      "options": {
+        "legend": {
+          "displayMode": "table",
+          "placement": "bottom",
+          "calcs": [
+            "lastNotNull",
+            "max"
+          ]
+        },
+        "tooltip": {
+          "mode": "multi",
+          "sort": "desc"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "sum by (reason) (rate(policy_poll_skipped_total{env=\"$env\"}[5m]))",
+          "legendFormat": "{{reason}}",
+          "refId": "A"
+        }
+      ]
     }
   ],
   "refresh": "30s",
--- a/openwrt/files/usr/lib/lua/wifihaven/ws_outbound.lua
+++ b/openwrt/files/usr/lib/lua/wifihaven/ws_outbound.lua
@@ -21,6 +21,27 @@
 
 local M = {}
 
+-- The SOLE freshness rule for "is the ws link live enough to be the transport":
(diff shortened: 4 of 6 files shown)


sameerparekh:
## Independent PR review

Read-only adversarial pass over `git diff origin/main...HEAD` (6 files, +244/−2) against the 9-dimension checklist. Reviewer is not assumed neutral on authorship — this code is ours; flagged accordingly.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `openwrt/files/usr/sbin/wifihaven-agent:1069–1177` — the real on_tick gate (the `if ws_link_healthy() then …skip+metric… else …poll… end` wiring) is **not directly asserted** by a test. The spec pins the extracted decision fn (`ws_outbound.is_healthy`) and a re-encoded `poll_tick` model that wires the *real* `is_healthy` + *real* `policy.failover_transition`, but the monolith's skip-vs-poll branch + metric emission rest on the real-Lua-5.1 compile check, not an assertion. This matches existing repo precedent (the agent monolith isn't require-able; `conntrack_tick_spec` likewise drives a *mock* on_tick), and the decision logic that could actually invert is extracted and covered — so the residual risk is the branch plumbing only. Acceptable as-is; noted for honesty.
2. `openwrt/files/usr/sbin/wifihaven-agent:1075–1170` — the `else` body (~100 lines, the existing poll path) is intentionally **not re-indented** to keep the diff minimal, with a `-- #2037: close the else` marker on the added `end`. Lua is whitespace-insensitive and the 5.1 compile confirms structure, but the de-indented block is harder to read. An alternative (extract the poll body into a `local function poll_policy()` and call it in the else) would read cleaner at the cost of a larger diff. Reviewer's call; left as-is.
3. `openwrt/files/usr/sbin/wifihaven-agent:502–508` vs `:485–490` — the `{enabled, health_read, now, fallback_after}` opts table is now constructed at two call sites (the outbound tee and `ws_link_healthy`). The *decision* is single-sourced in `ws_outbound.fresh`/`is_healthy` (the important thing), so this is only config-wiring duplication over the same upvalues — benign, but a tiny shared builder would remove it.

### Dimension notes
- **1 Single-source-of-truth (✓ improves it):** the freshness predicate is lifted into `ws_outbound.fresh`/`is_healthy` and the tee's old inline `not h or (now()-h) > fallback_after` is refactored to call it — the poll-dormancy gate and the tee now share one definition and cannot drift. Boundary preserved exactly (`<= fallback_after` ⇔ the tee's prior `> fallback_after` fallback edge), so the refactor is behavior-neutral for the tee.
- **2 Testing (✓):** TDD red→green visible (`6ba138e9` failing spec → `3532bb53` impl). New spec covers fresh/stale/absent/disabled/boundary for the gate, and asserts the #331/#422 failover still trips on a failed *resumed* poll via the real `policy.failover_transition`. Validated on real Lua 5.1 (plex.lan): agent compiles within the 60-upvalue limit (#1930); full busted suite 983/0.
- **3 Architecture (✓):** transport management, not policy decision — no schedule/time/category logic added to the agent; router stays a dumb applier. Liveness claim verified: `RouterWsRoutes.touch` updates `routers.last_seen_at` on every recognized ws frame incl. ping/pong (RouterWsRoutes.scala:148–159,223), so dropping the poll while ws is up does not stale liveness. No `/tmp` writer added.
- **4 Wire (✓):** no wire change; `policy_poll_skipped_total` is an additive metric series on the existing `/api/router/metrics` push. Reversible, not #376-gated. Distinct from #1850 (endpoint removal).
- **6 Metrics+dashboard (✓):** bounded-label counter `policy_poll_skipped_total{reason=ws_healthy}` (router_id/installation_id server-attached), allowlisted in `MetricGuard.Allowed` and pinned by `RouterWsAgentMetricsAllowlistSpec`; Grafana panel added to the already-registered `router-ws-transport.json`. No new dashboard → no `infra/grafana/main.tf` change needed.
- **7 Scope (✓):** focused 6-file diff; no AI mentions in commit/PR text (Co-Authored-By trailer only). A pre-existing flaky timing test surfaced during the full-suite run and was filed as #2042 — deliberately *not* folded in.
- **9 Unsourced facts (✓):** `ws_fallback_after` is the existing uci-sourced value (agent line 135), not a new literal; the `last_seen_at`/heartbeat and boundary-inclusive claims were verified against source, not asserted from comments.
- **5 Migrations / 8 Security:** N/A (no SQL, no migration, no new route, no secrets).

**VERDICT: APPROVE**

Agent-side, reversible, no-wire-change dormancy gate that realizes the #1023 load win; the freshness decision is single-sourced with the existing tee and the failover path is preserved and tested. No blockers; the only notes are a precedent-consistent test-coverage boundary on the monolith's branch plumbing and two cosmetic items.

Issue #2041, opened 2026-06-28
bug(dashboard): LIVE BANDWIDTH flashes '0 B/s' before data loads — show a loading state, not zero


sameerparekh:
## Symptom

The LIVE BANDWIDTH gauge **momentarily shows "0 B/s" / "No live traffic right now"** before the first data resolves (the GET-seed / first ws push), then fills in. The transient "0" reads as "there's no traffic" when it's really "data not loaded yet."

## Fix

Show a **loading / skeleton state** for the gauge until the first real bandwidth data is available, then render the value. Distinguish three states explicitly:
- **loading** — query in flight / no data yet → skeleton or "—", never "0 B/s".
- **live with data** → the rate.
- **genuinely idle** (data present, rate ~0) → may show "0 B/s" / "No live traffic right now" (this is the only case that should read as zero).

Frontend only (`web/src/components/dashboard/BandwidthGauges.tsx` + `useWsTrafficUsage` — there's likely an `isLoading`/`isPending`/`fetchStatus` signal from the seed `useQuery` and the `useWsTopicLive` indicator to gate on). Small UI polish; pairs with the functional window fix (the companion issue) but is independent.

Operator-requested 2026-06-28. Relates to #1148, #1837 (perceived-latency / skeletons), #2017.

Issue #2017, opened 2026-06-28
bug(dashboard): LIVE BANDWIDTH gauge blank for ~60-70s on load/bucket-switch — missing initial GET seed (ws-only series)


sameerparekh:
## Symptom

On `app.wifihaven.net/dashboard` the **LIVE BANDWIDTH** gauge shows nothing — `Overall ↓0 B/s ↑0 B/s` / *"No live traffic right now"* — for **up to ~60–70s after every page load and after every bucket switch**, for **all buckets including the 1m default**. It then populates once the first websocket push lands.

## Root cause — the gauge has no initial GET seed

The dashboard bandwidth gauge (`useWsTrafficUsage`, `web/src/hooks/useWs.tsx`) is **websocket-only**: its series lives purely in the React Query cache key `qk.trafficUsageLive({groupBy, bucket})`, written **only** by the `trafficUsage` push and read passively (`useCachedQueryData`). There is **no initial `GET /api/usage/traffic` fetch**. So on a fresh load (or any bucket change → new cache key), the series is empty until the **next usage-ingest push**, which only fires on the ingest tick (~up to ~70s). The panel shows "No live traffic right now" that whole window.

This contradicts the design (`docs/design/spa-websocket.md` §3.1 / §5.3): *"the page/gauge loads its window via the existing GET … the socket pushes only the live edge."* S5 (#1973) shipped the live-edge push but skipped the GET seed for the dashboard gauge.

## Evidence (verified live in-browser, post-#2004 deploy)

- The **server pushes valid data for every bucket** — direct ws probes got frames with real bytes for `raw` (50 MB/23 MB), `1m`, `10m` (54 MB/42 MB, window 01:50–02:00), `1h` (580 MB/98 MB), each with correct `windowStart`/`windowEnd`. So this is **not** a server bug and **not** per-bucket.
- The app's React Query cache (read directly via fiber walk) holds **only** the currently-selected bucket's `usage/traffic/live/<bucket>/profile` key, empty until a push merges in.
- After selecting a bucket and **waiting ~70s**, the gauge **does** populate (e.g. Raw → `Overall ↓194 B/s ↑239 B/s`; 1m → `↓14.1 KB/s ↑113.6 KB/s`). So every bucket works — it's purely first-paint latency.
- Reproduces deterministically: hard-reload `/dashboard` → 1m default shows "No live traffic right now" until the first push.

## Fix

[...]

PR #2046, opened 2026-06-28
fix(#2040,#2041): LIVE BANDWIDTH — most-recent COMPLETE bucket + aggregate raw + loading state

### LINKED ISSUE #2040: bug(dashboard): LIVE BANDWIDTH uses the in-progress bucket — 1m/raw blank, coarse buckets understated; show most-recent COMPLETE bucket + aggregate raw
## Symptom

On the dashboard LIVE BANDWIDTH gauge:
- **1m and raw are persistently empty.**
- **Coarser buckets (10m/1h/…) read low/wrong** — they show the *in-progress* bucket, which isn't a full bucket-width of traffic yet, so `bytes / bucketWidth` understates the rate.

## Root cause — the gauge uses the *in-progress* bucket, not the most-recent *complete* one

1. **Empty current window (1m/raw).** The #2017 GET-seed's `headWindow(bucket)` (`web/src/hooks/useWs.tsx`) requests `[floor(now, bucketWidth), now)` — the current bucket. Verified live (admin curl, server ~22:08Z):
   - `1m` `[floor(now),now)` → **0 rows**; `1m` `[now-5m,now)` → **6 rows** (data at 22:04). The data lands in the *prior* bucket (a ~60s report covers the previous minute — the #2004 insight), so the current-bucket seed is empty → blank until a push.
2. **Partial current window (10m+).** Even when the current bucket has data, it's only partway through `bucketWidth`, so `bytes / bucketWidth` is an understated average (operator: *"if we're looking at the current bucket that's not a full 10m worth of traffic"*).
3. **`raw` returns `rawRows`, not `aggregateRows`.** Verified: `GET /api/usage/traffic?bucket=raw&groupBy=profile&from=…&to=…` → `aggregateRows=[]`, `rawRows=200`. The gauge (`headBucketRows`) reads `aggregateRows`, so **raw never renders on the gauge regardless of window.** Per `docs/design/spa-websocket.md` §1.3, a `raw` subscription should be *aggregated per the subscription's groupBy at the ingest-period granularity* (one per-profile point per report period) — not ungrouped per-host `rawRows`.

## Fix direction

- **Show the most-recent COMPLETE bucket, not the in-progress one** — for every bucket. The complete bucket is a full `bucketWidth` of elapsed time, so `bytes / bucketWidth` is an accurate rate; for 1m/raw it's the prior minute/period (which has data), fixing the blank. Adjust `headWindow` (and the ws live-edge "head") to target the last complete bucket. (Alternative for the in-progress bucket — divide by *elapsed* seconds instead of full width — is noisier; the complete-bucket approach is preferred per the operator.)
- **Make `raw` aggregate per `groupBy`** so `GET ?bucket=raw&groupBy=profile` returns per-profile `aggregateRows` (one point per ingest period) consistent with the ws push and the design §1.3, instead of `rawRows` — so the gauge has rows to render.
- Keep GET + Traffic Usage page + ws aggregator consistent (SSOT — the #2018 `windowFor`/`buildAggregate` path).

## Verify (Chrome now allowed for app.wifihaven.net)
After the fix, the gauge shows non-zero, accurate rates for 1m, raw, and coarser buckets immediately on load (most-recent complete bucket), with no understatement.

Regression surface on #2017 (GET-seed) + #2018 (raw window) + #1973 (S5). Filed fresh. Relates to #1860, #1148, #747, #2004.

sameerparekh:
Fixes #2040 and #2041 — the dashboard LIVE BANDWIDTH gauge (`BandwidthGauges` + `useWsTrafficUsage`). Landed together per the #2040 note (same component).

## #2040 — the gauge used the *in-progress* bucket; `raw` never rendered

**Most-recent COMPLETE bucket, every bucket.** The #2017 GET-seed requested `[floor(now), now)` — the current bucket — and read the *newest* window. That window is **empty** for 1m/raw (a ~60s report lands in the *prior* period, the #2004 insight) and **partial/understated** for 10m+ (only part of `bucketWidth` has elapsed, so `bytes / bucketWidth` understates the rate).

- `completeHeadBucketRows(series, now)` selects the newest window whose full span has elapsed (`windowEnd <= now`), skipping the still-accumulating in-progress cell. `raw` rows are all closed ingest periods → the newest is shown.
- `headWindow` widens to `max(2×bucketWidth, 10min)` so the most-recent complete bucket is always present (fine buckets can sit several reports back; the current cell is empty/partial). The ws live-edge already anchors on the ingested `periodStart` (#2004) and the same selector now skips any in-progress bucket a push delivers — GET + ws + the #2018 `windowFor`/`buildAggregate` path stay consistent.

**Aggregate `raw` per `groupBy`.** `GET ?bucket=raw&groupBy=profile` returned `aggregateRows=[]` + per-host `rawRows`; the gauge reads `aggregateRows`, so raw never rendered. Now `raw` **with** a groupBy falls through to the aggregate path (one point per `(periodStart, group)` on the row's real `[periodStart, periodEnd)`, #2018), per `docs/design/spa-websocket.md` §1.3. `raw` with **no** groupBy (the Traffic Usage raw inspector) still returns `rawRows`.

## #2041 — "0 B/s" flash before data loads

`useWsTrafficUsage` now exposes `loading` (seed in flight, no data yet). The gauge renders a **skeleton** while loading and shows "0 B/s" / "No live traffic right now" **only** when data is present and the rate is genuinely ~0.

## Live evidence (admin curl vs prod, server ~22:25–22:45Z)

| query | before (prod, today) | after |
|---|---|---|
| `bucket=raw&groupBy=profile` | `aggregateRows=0`, `rawRows=200` | aggregated per-profile rows on the real period (local feature test) |
| `bucket=1m` current bucket | empty → gauge blank | wider window returns the complete `22:38` bucket; selector shows it |
| `bucket=10m` newest window | in-progress `[22:40,22:50)` → understated | complete `[22:30,22:40)`, full-width rate |

## Tests
- **api**: feature test — `raw&groupBy=profile` aggregates per profile on the row's real `[periodStart, periodEnd)` (RED→GREEN); full `mill api.test` green.
- **web** (node 22): `completeHeadBucketRows` complete-vs-in-progress selection; hook tests for complete-bucket render, full-`bucketWidth` 10m rate, raw-from-aggregated, and the loading→loaded transition; `BandwidthGauges` component test for skeleton vs idle vs paused. Lint + full vitest + build green.

### DIFF
--- a/api/src/routes/UsageRoutes.scala
+++ b/api/src/routes/UsageRoutes.scala
@@ -1096,7 +1096,14 @@ object UsageRoutes {
       // time bucket"). No implicit default.
       effectiveGroupBy = groupBySet
       resp <- bucket match {
-        case UsageTraffic.Bucket.Raw =>
+        // #2040: `raw` WITHOUT a groupBy is the page's per-host raw inspector → `rawRows`. `raw`
+        // WITH a groupBy (e.g. the dashboard gauge's `groupBy=profile`) falls through to the
+        // aggregate path below, which rolls each ingest period up per the groupBy at the row's REAL
+        // `[periodStart, periodEnd)` window (`buildAggregate` floors `raw` to the row itself, #2018)
+        // — one `aggregateRows` point per (period, group), per design §1.3. Without this guard a
+        // `raw&groupBy=profile` read returned per-host `rawRows` and empty `aggregateRows`, leaving
+        // the gauge (which reads `aggregateRows`) blank.
+        case UsageTraffic.Bucket.Raw if effectiveGroupBy.isEmpty =>
           for {
             rawCursor <- req.url.queryParam("cursor") match {
               case None    => ZIO.succeed(Option.empty[RawTrafficCursorKey])
@@ -1142,7 +1149,7 @@ object UsageRoutes {
             rawRowsTruncated = false,
             nextCursor = nextCur,
           )
-        case _                       =>
+        case _                                                   =>
           // Aggregated path: source table = finer of (bucket cap, window cost
           // preference) (#1262). The requested bucket is always honored; the
           // range can only steer the read toward finer/fresher data, never
--- a/api/test/src/feature/UsageApiSpec.scala
+++ b/api/test/src/feature/UsageApiSpec.scala
@@ -882,6 +882,106 @@ object UsageApiSpec extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostgres & C
           assertTrue(out.rawRows.forall(_.profileName.contains("Kids"))) &&
           assertTrue(out.rawRows.map(_.host.value).toSet == Set("youtube.com", "google.com"))
       },
+      // #2040: a `raw` request WITH a groupBy aggregates per the groupBy at the row's real ingest
+      // period (one point per (periodStart, group)) — it must NOT fall through to the ungrouped
+      // per-host `rawRows` shape (which left the dashboard LIVE BANDWIDTH gauge, reading
+      // `aggregateRows`, perpetually blank). The ungrouped raw inspector (no groupBy) still returns
+      // `rawRows` (covered by the test above).
+      test(
+        "raw view WITH groupBy=profile aggregates per profile on the row's real period (#2040)",
+      ) {
+        val today = TestClock.schoolDayAfternoon.toLocalDate
+        for {
+          _           <- cleanDb
+          profileRepo <- ZIO.service[ProfileRepo]
+          deviceRepo  <- ZIO.service[DeviceRepo]
+          kidsId      <- TestLayers.seedKidsProfile(profileRepo)
+          _           <- TestLayers.seedDevice(deviceRepo, testMac, "iPad", kidsId)
+          routerId    <- seedRouter
+          // Two hosts in the SAME ingest period (14:00, 300s) so they roll up into ONE per-profile
+          // point; a third host in a LATER period (14:05) makes a second point.
+          start1 = today.atStartOfDay(ZoneOffset.UTC).toInstant.plusSeconds(14 * 3600L)
+          start2 = start1.plusSeconds(300)
+          _  <- ZIO.serviceWithZIO[TrafficReportRepo](
+            _.insertBatch(
+              List(
+                TrafficReportInsert(
+                  routerId,
+                  MacAddress.unsafe(testMac),
+                  None,
+                  HostId.Fqdn(Hostname.unsafe("youtube.com")),
+                  today,
+                  start1,
+                  start1.plusSeconds(300),
+                  300,
+                  1000L,
+                  2000L,
+                ),
+                TrafficReportInsert(
+                  routerId,
+                  MacAddress.unsafe(testMac),
+                  None,
+                  HostId.Fqdn(Hostname.unsafe("google.com")),
+                  today,
+                  start1,
+                  start1.plusSeconds(300),
+                  300,
+                  500L,
+                  1500L,
+                ),
+                TrafficReportInsert(
+                  routerId,
+                  MacAddress.unsafe(testMac),
+                  None,
+                  HostId.Fqdn(Hostname.unsafe("netflix.com")),
+                  today,
+                  start2,
+                  start2.plusSeconds(300),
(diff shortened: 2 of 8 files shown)


sameerparekh:
## Independent PR review — #2040 + #2041 (LIVE BANDWIDTH gauge)

Read-only adversarial pass over `git diff origin/main...HEAD` against the 9-dimension checklist.

**BLOCKERS**
(none)

**SHOULD-FIX**
1. `web/src/hooks/useWs.tsx:336` — `SEED_LOOKBACK_MIN_MS = 10 * 60 * 1000` (dimension 9, the #2018 class). It's a new behavior-influencing duration not derived from a named/config source. Mitigating: unlike the #2018 `% 300` bug, this constant is a *tolerance* for the seed-window width, **not** a rate divisor — `completeHeadBucketRows` (selecting on `windowEnd <= now`) is the correctness guarantee, so a too-small lookback degrades gracefully to the loading/empty state until a ws push arrives and can never produce a *wrong* rate. The `2 × width` term derives from the sourced `bucketSeconds`. No frontend constant for the agent `usage_report_interval` exists to cite (the `/api/usage/config` payload exposes bucketTiers + horizons, not the report cadence), so 10 min is a defensible "~10 report intervals" headroom. Acceptable as-is given the inline rationale; ideal future improvement is to source the report cadence rather than hardcode the tolerance.

**NITS**
1. `web/src/api/wsCache.ts:78` — `completeHeadBucketRows` assumes the series is newest-first (`.find` returns the first complete window in array order). This holds (server orders `windowStart DESC`; `mergeHeadBucket` prepends, and `SpaPush.coalesce` is latest-wins) and is the *same* assumption the old `headBucketRows` made via `rows[0]`, so no regression — noting only that the invariant is now load-bearing for "complete-bucket" selection too.
2. TDD: the failing test and the implementation landed in a single commit rather than a visible red→green pair. RED→GREEN was demonstrated in-session (the api feature test failed on `aggregateRows=Nil`, then passed after the route guard) and the tests pin the bug, so this isn't a hidden-regression risk — just not split in history.
3. PR body carries a "🤖 Generated with Claude Code" line, which dimension 7 lists as an AI-mention hygiene item; non-blocking (the operator's documented stance is that this rule isn't theirs).

**Dimension notes (no findings)**
- **SSOT (dim 1):** the fix removes a divergence rather than adding one — `raw&groupBy` now flows through the shared `UsageTrafficQuery.aggregate`, so the GET body matches the `SpaPush` live-edge aggregator (both use `windowFor`); `completeHeadBucketRows` reads the server's `windowEnd` rather than recomputing widths.
- **Testing (dim 2):** no existing assertion weakened to hide a regression — the replaced `headBucketRows` test asserted the *buggy* "newest window" behavior; the new test asserts the corrected "skip in-progress" behavior. The `#2017` seed tests still assert rows render (their `windowEnd` is in the past). New behavior is asserted at unit (selection), hook (complete-bucket / full-width 10m rate / raw-from-aggregated / loading transition), component (skeleton vs idle vs paused), and api-feature (raw+groupBy aggregation on the real period) levels.
- **Wire (dim 4):** `TrafficUsageResponse` shape unchanged; the `raw&groupBy` *content* shift (rawRows→aggregateRows) is SPA-facing only (routers never call `/api/usage/traffic`), the sole SPA consumer of raw+groupBy is the gauge being fixed, and the Traffic Usage page sends raw with `groupBy: undefined` (unaffected). SPA deploys atomically with the API.
- **Scope/security (dim 7/8):** focused diff, no dead code (`headBucketRows` fully removed, no stale refs), reuses the parameterized aggregate path with auth unchanged, no secrets committed.
- **Verified live** (admin curl vs prod): `raw&groupBy=profile` returns `rawRows`/empty `aggregateRows`; the 1m current bucket is empty and the 10m current bucket in-progress — all three confirmed and addressed.

**VERDICT: APPROVE**
Correct, well-scoped fix for all three sub-bugs (in-progress→complete bucket, raw aggregation, loading state); it collapses a GET-vs-ws divergence rather than adding one, and tests pin each path. The only flagged item is a well-justified, non-rate-driving tolerance constant — not the dangerous #2018 class. No open blockers.

Issue #1975, opened 2026-06-25
ws-spa S6b: stream the Traffic Usage + Connection Events pages + class-(3) stale invalidation


sameerparekh:
**SPA-websocket implementation — step S6b of the §9 rollout** (`docs/design/spa-websocket.md` §1.4/§3, design #1860; umbrella #1955). **Additive, per-page.**

**Stream the Traffic Usage + Connection Events pages** — same topics as the dashboard, the pages' own filters. One stream, many consumers (§0.3).

### Scope (§9 S6b)
- Traffic Usage + Connection Events pages keep loading **history via their existing `GET`** (initial window + cursor paging, #862); they additionally subscribe `trafficUsage`/`connectionEvents` with **their current filters** for the **live edge only** — `trafficUsage` merges the head bucket, `connectionEvents` prepends new head rows (dedup by id); older/paged data never mutated.
- Pause each page's poll when `wsLive`.
- **Class-(3) `stale`** wiring for alerts / profiles / devices / schedules → `queryClient.invalidateQueries` on the mapped key via the existing `qk`/`useInvalidators`; the topic→key map is the one place the mapping lives (§3.2).

### Acceptance (web — `nvm use 22`)
- Both pages stream their live edge under their active filters while history/paging stays on the `GET`; changing filters re-subscribes; a `stale{profiles|devices|schedules|alerts}` from another tab invalidates only the mounted mapped query.

Depends on: **S5 (#https://github.com/wifihaven/wifihaven/issues/1973)**. Relates to #862, #1860, #1955.

Issue #862, opened 2026-05-22
ui: Traffic Usage + Connection Events — infinite scroll / paging for raw + aggregate


sameerparekh:
## Goal

Replace the current \"end-anchored fixed window\" UX on the Traffic Usage page (and eventually Connection Events) with infinite scroll. The operator wants to skim recent rows then scroll back through history without having to manipulate a date picker.

## Current shape

After [#846](https://github.com/wifihaven/wifihaven/issues/846), the page picks a fixed window based on the aggregation bucket (raw=100 rows, 10m=1 day, 1h=1 week, 1d=1 month, 1w=6 months). The end of the window is operator-controlled via a single \"end at\" datetime picker; the start is derived.

## What to build

- Lazy-load older rows as the operator scrolls down.
- Server gains a cursor / offset parameter on \`/api/usage/traffic\` (and the matching connection-events endpoint).
- Avoids the current 31-day on-the-fly cap by reading in chunks rather than one huge query.

## Acceptance

- Scrolling to the bottom of the table loads the next chunk seamlessly.
- Works for both raw and aggregated views.
- No more \"window too large\" 503s for the common ops case of \"keep scrolling back\".

## Related

- Surfaced during #846 audit follow-through. Replaces the bucket-default window heuristic when this lands.

PR #2049, opened 2026-06-28
fix(#2047): Traffic Usage page must not surface its own request cancellations

### LINKED ISSUE #2047: bug(prod, HIGH): Traffic Usage page unusable — aborts its own requests ('signal is aborted without reason'), stuck Loading, false 'can't reach API' banner (API is fine)
## Symptom (prod, app.wifihaven.net/usage — Traffic Usage page)

The page is **unusable across all buckets**:
- **raw** (default): table shows **"signal is aborted without reason"** + **"No rows in window."**
- **1m / 10m / …**: stuck on **"Loading…"**, never resolves.
- A red **"WifiHaven can't reach the API right now. Showing cached data where available."** banner is shown the whole time.

## It is NOT the API — it's client-side

Probed from the page's own context (its localStorage token):
- `GET /api/usage/traffic?bucket=1m&groupBy=profile&from=…&to=…` → **200, aggregateRows=10** (real data).
- `GET /api/dashboard/now` → **200**.

So the API is reachable and the token is valid. The page is **aborting its own requests** ("signal is aborted without reason" = an `AbortController`/`AbortSignal` abort) and surfacing the AbortError as (a) a table error, (b) a perpetual loading state, and (c) the global "can't reach the API" banner.

## Likely cause

Regression in the Traffic Usage page's data fetching, most likely the **S6b live-edge interaction** (#1975): `useWsTrafficUsageLiveEdge` calls `setAggRows` on each push / re-subscribes on param changes, re-rendering the page; if the paging GET (`#862`, an `AbortController` tied to an effect whose deps change identity each render) is aborted on every re-render, it never resolves → stuck loading + the AbortError leaks to the user and trips the `ApiUnreachableBanner`. (An `AbortError` from a cancelled request must be swallowed, never treated as "API unreachable".)

## Fix direction

- Stabilize the page's fetch/paging so live-edge-driven re-renders don't abort the in-flight request (stable effect deps / AbortController; don't re-create the request identity each render).
- Treat `AbortError` / "signal is aborted without reason" as a **cancellation**, not a user-facing error and not an API-unreachable condition (don't show it in the table, don't trip the banner).
- Verify on app.wifihaven.net: raw + 1m + 10m all load real rows, no spurious banner, no abort error.

Severity: HIGH — the whole Traffic Usage page is unusable in prod. Verified live 2026-06-28. Regression surface: #1975 (S6b live-edge) × #862 (paging). Relates to #1148, #1860.

sameerparekh:
Fixes #2047.

## Symptom (prod, app.wifihaven.net/usage)
The whole Traffic Usage page was unusable across all buckets:
- **raw** (default): table showed **"signal is aborted without reason"** + **"No rows in window."**
- **1m / 10m / …**: stuck on **"Loading…"**.
- A red **"WifiHaven can't reach the API right now"** banner throughout.

But the API is healthy: probed from the page's own context (its `localStorage` token),
`GET /api/usage/traffic?bucket=1m&groupBy=profile&from=…&to=…` → **200, aggregateRows=10**,
and `/api/dashboard/now` → 200. So this is purely **client-side**: the page surfaced its own
request cancellation as a user-facing error and a false API-unreachable state.

## Root cause (confirmed in code, not assumed)
The only `AbortController` in the SPA is the per-request 10s timeout in
[`client.ts`](web/src/api/client.ts). Its catch treated **every** abort identically:
`apiHealth.reportFailure('timeout')` (→ trips `ApiUnreachableBanner`) and re-threw the raw
`AbortError` DOMException. [`TrafficUsagePage.load`](web/src/pages/TrafficUsagePage.tsx) then
caught **any** thrown error — `AbortError` included — and rendered it via `ErrorBanner`. So any
abort of a Traffic Usage request (e.g. a request superseded by the live-edge/#1975 × paging/#862
interaction, or a transient timeout) shows the cryptic message in the table *and* lights the
red banner, even though the API is fine.

## Fix (request-lifecycle only — the API is healthy)
- **`client.req`**: distinguish a **caller cancellation** from the **10s timeout** abort. The
  page now passes a load-lifecycle `AbortSignal`; a caller-aborted request throws a typed
  `RequestCanceledError` and does **not** report an `apiHealth` failure. A genuine timeout still
  reports `'timeout'` → unreachable (unchanged, test-pinned).
- **`TrafficUsagePage`**: owns one `AbortController` per load-lifecycle, rotated by the reset
  effect on `filterKey` change and stored in a **ref** (created on filterKey change, **not** per
  render) so live-edge pushes (`setAggRows` → re-render) never abort the active request. A
  superseded load (bucket switch / filter change) is cancelled cleanly; `load()` swallows
  cancellation/`AbortError` — no table error, no banner, paging untouched — and a stale settle
  never clobbers the replacement load's spinner.

## Tests (TDD: red → green)
- page: a load rejected with `AbortError` must not render an error (was RED — showed
  "signal is aborted without reason").
- page: a superseded (bucket-switch) request cancels cleanly with no error.
- client: a caller-cancelled request throws a cancellation and does **not** trip `apiHealth`;
  a genuine 10s timeout still reports unreachable.

`npm run lint`, `npm test` (494 passed), and `npm run build` all green.

### DIFF
--- a/web/src/api/client.test.ts
+++ b/web/src/api/client.test.ts
@@ -3,7 +3,8 @@
 // `profileId=` param (parseMultiProfileIdParam splits on comma and only reads
 // the first occurrence), plus pass-through of date/tz/topN/groupBy.
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
-import { api } from './client'
+import { api, isCanceledError } from './client'
+import { apiHealth } from './apiHealth'
 import type { UsageSeriesBatchResponse } from '@/types/api'
 
 function okJson(body: unknown): Response {
@@ -106,5 +107,62 @@ describe('req opts every call out of the browser HTTP cache (#1299)', () => {
 
     const opts = fetchMock.mock.calls[0][1] as RequestInit
     expect(opts.cache).toBe('no-store')
+  })
+})
+
+// #2047 — a request the *caller* cancels (the Traffic Usage page superseding its
+// in-flight load on a bucket switch / filter change) must be distinguished from
+// the 10s timeout abort: a cancellation is NOT an API-health signal, so it must
+// never trip the unreachable banner, and it surfaces as a typed cancellation the
+// caller swallows rather than the cryptic "signal is aborted without reason".
+describe('req caller-cancellation is not an API-down signal (#2047)', () => {
+  beforeEach(() => {
+    localStorage.setItem('token', 'tok')
+    apiHealth.reset()
+  })
+  afterEach(() => {
+    vi.restoreAllMocks()
+    localStorage.clear()
+    apiHealth.reset()
+  })
+
+  it('forwards the caller signal and reports a cancellation, not unreachable', async () => {
+    // Real fetch semantics: a fetch given an already-aborted signal rejects with
+    // an AbortError DOMException.
+    const fetchMock = vi.fn().mockRejectedValue(
+      new DOMException('signal is aborted without reason', 'AbortError'),
+    )
+    vi.stubGlobal('fetch', fetchMock)
+
+    const controller = new AbortController()
+    controller.abort()
+
+    const p = api.usage.traffic({ bucket: '1m', signal: controller.signal })
+    await expect(p).rejects.toSatisfy(isCanceledError)
+    // The banner must stay clear — the caller cancelled, the API is fine.
+    expect(apiHealth.snapshot().unreachable).toBe(false)
+  })
+
+  it('still reports a genuine 10s timeout as unreachable (not a cancellation)', async () => {
+    vi.useFakeTimers()
+    // A fetch that never resolves until its own (timeout) signal aborts it.
+    const fetchMock = vi.fn(
+      (_url: string, opts: RequestInit) =>
+        new Promise<Response>((_res, rej) => {
+          opts.signal?.addEventListener('abort', () =>
+            rej(new DOMException('signal is aborted without reason', 'AbortError')),
+          )
+        }),
+    )
+    vi.stubGlobal('fetch', fetchMock)
+
+    const p = api.usage.traffic({ bucket: '1m' })
+    const assertion = expect(p).rejects.toThrow()
+    await vi.advanceTimersByTimeAsync(10_001)
+    await assertion
+    // No caller signal → the abort is the timeout → genuine unreachable.
+    expect(apiHealth.snapshot().unreachable).toBe(true)
+    expect(apiHealth.snapshot().reason).toBe('timeout')
+    vi.useRealTimers()
   })
 })
--- a/web/src/api/client.ts
+++ b/web/src/api/client.ts
@@ -30,18 +30,52 @@ function getToken(): string | null {
 // don't leak signal across calls.
 const REQUEST_TIMEOUT_MS = 10_000
 
+// #2047 — a request the *caller* cancelled (e.g. the Traffic Usage page
+// superseding its in-flight load on a bucket switch / filter change). This is
+// categorically NOT an API-health signal: the API is fine, the client just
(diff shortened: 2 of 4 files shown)


sameerparekh:

Reviewed the full merge-base diff (`git diff origin/main...HEAD`) against the 9 dimensions. Scope: SPA-only request-lifecycle fix — no server/router/wire/migration surface touched.

**BLOCKERS**
(none)

**SHOULD-FIX**
(none)

**NITS**
1. `web/src/pages/TrafficUsagePage.test.tsx:26` — the module-mock predicate `e.name === 'RequestCanceledError'` re-implements `isCanceledError` (the real one is `e instanceof RequestCanceledError`). Faithful today, but a behavioral mirror that could drift if the error class changes. Low risk for a test mock.
2. `web/src/api/client.test.ts:148` — the timeout test advances by the literal `10_001`; `REQUEST_TIMEOUT_MS` (10_000) isn't exported, so the magic offset is unavoidable here. Fine as-is.
3. `web/src/api/client.ts:99` + `TrafficUsagePage.tsx:171` — after this change a genuine 10s timeout shows ONLY the global `ApiUnreachableBanner` (already tripped by `reportFailure('timeout')`) plus the empty-state "No rows in window.", no in-table error text. This is the intended behavior (banner is the canonical signal; the cryptic "signal is aborted…" no longer leaks), and is test-pinned, but worth noting as a deliberate UX shift from the prior in-table error.

**Dimension notes**
- **1 SSOT** — `isCanceledError`/`RequestCanceledError` defined once in `client.ts` and reused by page + tests. No re-derivation. The page's extra `instanceof DOMException && name==='AbortError'` branch is a deliberate belt-and-suspenders for the timeout-abort that `req` rethrows after reporting — not a duplicated decision.
- **2 Testing** — TDD red→green visible in history (`test(#2047)` red commit precedes the `fix(#2047)` impl commit). New behavior asserted (no error / no banner / clean supersede), and the negative case (genuine timeout STILL reports `unreachable`+`timeout`) is pinned so the cancel/timeout split can't silently collapse. No forbidden mocks (`apiHealth` is real; mocking the api-client module is the established page-test pattern).
- **3 Invariants** — no DNS/router/snapshot/Clock surface. `REQUEST_TIMEOUT_MS` unchanged.
- **4 Wire** — `signal` is a client-only param passed to `req`; verified it is NOT added to the `/usage/traffic` query string. No wire-shape change.
- **5 Migrations / 8 Security** — none / none (no SQL, secrets, or new routes).
- **6 Metrics** — bug fix to existing client request handling, not a new server route/job/poller; the WifiHaven metric convention is server-side `AppMetrics`. N/A for SPA code.
- **7 Scope** — 4 related files, focused; comments cite #2047; no AI mentions in title/body (Co-Authored-By trailer + generated-with footer are allowed).
- **9 Magic constants** — no new behavior-driving literal; the 10s timeout is pre-existing and untouched.

**Root-cause check** — the controller is created only in the reset effect (keyed on `filterKey`), stored in a ref, so live-edge `setAggRows`→re-render cannot recreate/abort it; `load`'s deps are exactly `filterKey`'s components, so its identity is stable across pushes. This is the correct fix for the #1975×#862 interaction the issue names. The `finally` guard (`abortRef.current === controller`) correctly prevents a superseded load's late settle from clearing its replacement's spinner. Listener add/remove is balanced via `cleanup()` on both fetch paths.

VERDICT: APPROVE
SPA-only request-lifecycle fix: caller cancellations are now distinguished from the 10s timeout abort, so a superseded/aborted Traffic Usage request no longer surfaces "signal is aborted without reason" as a table error or falsely trips the unreachable banner. Correct root-cause fix, well test-pinned (including the timeout-preservation negative), no wire/schema/security exposure. No blockers.

PR #2050, opened 2026-06-28
test(#2042): replace wall-clock ZIO.sleep waits with deterministic TestClock

### LINKED ISSUE #2042: Replace wall-clock ZIO.sleep waits in feature tests with TestClock (flaky MetricsExportSpec + class)

sameerparekh:
## Problem

Several feature specs waited on a wall-clock `ZIO.sleep` under `TestAspect.withLiveClock` to let a background fiber / Prometheus snapshot poller catch up, then asserted. The sleep races real wall-time, so under heavy parallel CI load (14 workers) the awaited work hadn't landed when the assert ran. `MetricsExportSpec` flaked on an unrelated PR for exactly this; it passes in isolation. This violates `docs/process/testing.md` ("Clock is always injected — use `Clock.TestClock` in tests"). Introduced in #1249/#1243 — our own code.

## Root cause (the part the issue's scope didn't yet know)

For the metrics-scrape specs, simply swapping `ZIO.sleep` → `TestClock.adjust(pollInterval)` is **not** sufficient when the prometheus layer lives in the shared `bootstrap`. `TestClock.adjust` → `run` → `awaitSuspended` only freezes fibers tracked in `TestAnnotation.fibers` **for the current test**. A bootstrap-scope publisher fiber is in the shared scope and invisible to `awaitSuspended`, so `adjust` completes the fiber's sleep promise but returns *before* it republishes — only a 10ms live window covers it, which loses under parallel load. (I verified this against the zio-test `TestClock` source.)

The fix: provide the prometheus layer **per-test** via suite-level `provideSomeLayer` (non-shared → fresh instance per test), so the snapshot fiber is a supervised descendant of the test fiber and `adjust` deterministically awaits it. `MetricsExportSpec` already provided it per-test and so was never affected — that's the working pattern this aligns the rest to.

## Changes

- **MetricsExportSpec**: drop `withLiveClock` + `Clock.live`; drive the rollup tick synchronously via `runOnce`/`oneTickForTest` instead of forking the loop; both sleeps → `tickPublisher`.
- **RouterMetricsIngestSpec, MetricsSelfMetricsSpec, ErrorBoundary{,Routes,Stage2}Spec**: move the prometheus layer from `bootstrap` to per-test `provideSomeLayer`; each scrape-await sleep → `tickPublisher` (waits for the snapshot fiber to park on the TestClock, then advances exactly one poll interval).
- **MetricsApiSpec**: scrape-await sleep → `tickPublisher`; drop two *dead* sleeps + `withLiveClock` from the auth/disabled tests (they never depended on the exposition).
- **TimeStatusCacheSpec**: the single-flight test's sleep was unnecessary — registration is an atomic `putIfAbsent` and the winner `cache.put`s before freeing its slot, so `count == 1` is invariant for any interleaving. Removed sleep + `withLiveClock`; runs on `TestClock`.
- **SpaWsS4Spec, SpaWsS6aSpec**: the remaining `ZIO.sleep(wait)` awaits a real push frame over an actual Netty WS socket against a live server — not a timer the TestClock can advance. **Kept** under the suite's `withLiveClock` with an inline comment explaining why TestClock won't serve (the issue's documented escape hatch).

## Acceptance

- [x] No feature spec uses `ZIO.sleep` to *wait for* a background fiber / poller / cache — those waits now use `tickPublisher` (= park-then-`TestClock.adjust`). The only `ZIO.sleep`s left are the two real-socket WS settles.
- [x] `MetricsExportSpec` runs on `TestClock` (no `withLiveClock`, no wall sleeps), deterministic under parallel load.
- [x] Remaining legitimate `withLiveClock` uses documented inline.
- [x] Full `mill api.test` green across **two** repeated parallel runs + targeted reruns of the previously-flaky specs.

Closes #2042

### DIFF
--- a/api/test/src/feature/ErrorBoundaryRoutesSpec.scala
+++ b/api/test/src/feature/ErrorBoundaryRoutesSpec.scala
@@ -31,13 +31,25 @@ import zio.test.*
  *      Boundary integration: those migrated error responses are now LOGGED (4xx → WARN) and METERED
  *      (`api_errors_total{route,status}`) in one place.
  */
-object ErrorBoundaryRoutesSpec
-    extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostgres & Clock & PrometheusPublisher] {
+object ErrorBoundaryRoutesSpec extends ZIOSpec[TestDatabase.AllRepos & EmbeddedPostgres & Clock] {
+
+  // Poll interval for the Prometheus snapshot fiber; tests advance the zio TestClock by exactly
+  // this much to drive one deterministic scrape.
+  private val pollInterval = 100.millis
+
+  // #2042: deterministically fire ONE scrape of the shared Prometheus publisher fiber. The key is
+  // waiting until that background fiber has actually re-parked on the TestClock (its `Schedule.fixed`
+  // sleep is registered) BEFORE advancing — under parallel CI load a bare `adjust` can slip past a
+  // momentarily mid-flight and never trigger the scrape. The publisher layer is provided PER TEST
+  // (suite-level `provideSomeLayer` below) so its snapshot fiber is a supervised descendant of the
+  // test fiber — only then does `TestClock.adjust` await it (a bootstrap-scope fiber is invisible to
+  // its `awaitSuspended`). Once parked, advancing one interval fires exactly one registry snapshot.
+  private val tickPublisher: UIO[Unit] =
+    zio.test.TestClock.sleeps.repeatUntil(_.nonEmpty) *> zio.test.TestClock.adjust(pollInterval)
 
   override val bootstrap =
     TestDatabase.layer ++
-      TestLayers.withClock(TestClock.schoolDayAfternoon) ++
-      MetricsRuntime.prometheus(100.millis)
+      TestLayers.withClock(TestClock.schoolDayAfternoon)
 
   private val jwtCfg = JwtConfig(secret = "test-secret-at-least-32-chars!!", expiryHours = 1)
 
@@ -102,7 +114,7 @@ object ErrorBoundaryRoutesSpec
         tok     <- auth.login("kid", "childpass").map(_.token.value)
         resp    <- get(rs, "/api/me", tok)
         body    <- bodyText(resp)
-        _       <- ZIO.sleep(700.millis)
+        _       <- tickPublisher
         scrapeB <- scrape.catchAll(r => bodyText(r))
         logs    <- ZTestLogger.logOutput
       } yield assertTrue(resp.status == Status.Forbidden) &&
@@ -112,7 +124,7 @@ object ErrorBoundaryRoutesSpec
         .provideSome[TestDatabase.AllRepos & EmbeddedPostgres & Clock & PrometheusPublisher](
           ZTestLogger.default,
         )
-    } @@ TestAspect.withLiveClock,
+    },
     test("login with bad credentials is 401 'Invalid credentials', logged + metered") {
       (for {
         _       <- cleanDb
@@ -123,7 +135,7 @@ object ErrorBoundaryRoutesSpec
             .addHeader(Header.ContentType(MediaType.application.json)),
         )
         body    <- bodyText(resp)
-        _       <- ZIO.sleep(700.millis)
+        _       <- tickPublisher
         scrapeB <- scrape.catchAll(r => bodyText(r))
         logs    <- ZTestLogger.logOutput
       } yield assertTrue(resp.status == Status.Unauthorized) &&
@@ -133,7 +145,7 @@ object ErrorBoundaryRoutesSpec
         .provideSome[TestDatabase.AllRepos & EmbeddedPostgres & Clock & PrometheusPublisher](
           ZTestLogger.default,
         )
-    } @@ TestAspect.withLiveClock,
+    },
     test("GET /api/profiles/{unknown} is 404 'Profile not found', logged + metered") {
       (for {
         _       <- cleanDb
@@ -142,7 +154,7 @@ object ErrorBoundaryRoutesSpec
         tok     <- auth.login("admin", "changeme").map(_.token.value)
         resp    <- get(rs, "/api/profiles/99999", tok)
         body    <- bodyText(resp)
-        _       <- ZIO.sleep(700.millis)
+        _       <- tickPublisher
         scrapeB <- scrape.catchAll(r => bodyText(r))
         logs    <- ZTestLogger.logOutput
       } yield assertTrue(resp.status == Status.NotFound) &&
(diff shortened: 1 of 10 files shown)


sameerparekh:

Test-hardening PR: replaces wall-clock `ZIO.sleep` waits with deterministic `TestClock` mechanics across 10 feature specs. Reviewed the full merge-base diff (`git diff origin/main...HEAD`, 10 files) against the 9 dimensions.

### BLOCKERS
(none)

### SHOULD-FIX
(none)

### NITS
1. `api/test/src/feature/MetricsApiSpec.scala:22` — the `tickPublisher` comment calls it the "**shared** Prometheus publisher fiber," but `MetricsApiSpec` (like `MetricsExportSpec`) provides the layer **per-test** via `.provide(...)` inside each test, so the fiber isn't shared. Cosmetic wording mismatch with the (accurate) per-test framing used in the other specs' comments.
2. PR body ends with a "🤖 Generated with Claude Code" footer — dimension 7 calls out "no AI mentions in the PR body." Non-gating and consistent with this repo's standing PR-footer convention + the `Co-Authored-By` trailer (which the checklist explicitly allows); flagging only for completeness.

### Dimension notes
- **(1) Single source of truth** — no re-derivation introduced. The rollup test now drives the metric-emission path through the existing `TimeUsedRollupJob.runOnce`/`oneTickForTest` primitives instead of forking `.loop()` — same code path, no copied logic.
- **(2) Testing integrity** — the critical dimension here, since existing tests are edited. **No assertion was weakened, deleted, or retrofitted**: every `assertTrue(...)` is byte-identical pre/post; only the *wait mechanism* changed (blind `ZIO.sleep` → park-then-`TestClock.adjust`, or synchronous tick) and `withLiveClock` was dropped where no longer needed. `TimeStatusCacheSpec`'s single-flight assertions (`count == 1`, `results.length == n`, `forall(_ == sample)`) are unchanged; the removed sleep was provably unnecessary (atomic `ConcurrentHashMap.putIfAbsent` + winner `cache.put` before slot-free ⇒ `count == 1` is interleaving-invariant — verified against `TimeStatusCache.getOrLoad`).
- **Root-cause correctness** — the per-test `provideSomeLayer` move (vs. `bootstrap`) is load-bearing: `TestClock.adjust` → `run` → `awaitSuspended` only freezes fibers in `TestAnnotation.fibers` for the current test, so a bootstrap-scope publisher fiber is invisible to it and a bare `adjust` races the scrape. Confirmed against the zio-test 2.1.9 `TestClock` source. The non-shared `provideSomeLayer` makes the snapshot fiber a supervised test descendant. Validated empirically: full `mill api.test` green across **two** repeated 14-worker parallel runs plus targeted reruns of the previously-flaky specs.
- **(3) Clock injected** — strictly improved: specs now run on `TestClock` instead of `withLiveClock`.
- **(9) Unsourced facts / magic constants** — `pollInterval = 100.millis` is a self-contained test cadence (same value as the prior bootstrap literal, now referenced by both the layer and the `adjust`), **not** a re-hardcoded single-sourced prod value (#2018 class). The "how TestClock works" comments are accurate (traced to the zio-test source).
- **(4) wire, (5) migrations, (6) metrics, (8) security** — N/A; test-only diff, no source/wire/schema/route changes.
- **(7) Scope** — focused: 10 related feature specs, no drive-by edits. The two remaining `ZIO.sleep(wait)` calls (`SpaWsS4Spec`, `SpaWsS6aSpec`) await a real push frame over a live Netty WS socket and are correctly **kept** under `withLiveClock` with an inline justification — `TestClock` can't drive Netty's real-time event loop.

### VERDICT: APPROVE
A clean, well-scoped test-hardening change that removes the flaky `ZIO.sleep` waits with the correct deterministic mechanics, preserving every assertion. The fragile bit (TestClock awaiting the publisher fiber) is correctly handled by moving the layer per-test, validated by repeated green parallel runs. No blockers; two NITs only.

sameerparekh:

Re-run on push `fce3c42` (prior review covered `8c799e8`). Delta is comment-only.

### Prior findings status
|---|---|---|
| NIT 1 | `MetricsApiSpec` `tickPublisher` comment said "shared" fiber, but the layer is per-test | **ADDRESSED** — comment now reads "provided per-test via `.provide` below, so it is a supervised test descendant" (`MetricsApiSpec.scala:22`). |
| NIT 2 | PR body "Generated with Claude Code" footer vs. dimension-7 "no AI mentions" | **NOT-ADDRESSED** (intentional) — non-gating; consistent with this repo's PR-footer convention and the allowed `Co-Authored-By` trailer. |

### New findings from the delta
(none — the only change since `8c799e8` is the NIT-1 comment correction; no code/assertion changes.)

### VERDICT: APPROVE
NIT-1 addressed; the incremental delta is a documentation-only fix with no behavioral change. No blockers across either run.

PR #2052, opened 2026-06-29
feat(#2048): stream the Traffic Usage page's raw-rows view (prepend new rows + Live badge)

### LINKED ISSUE #2048: enh(dashboard): stream the Traffic Usage page's raw-rows view (prepend new rows + Live badge) — default raw view is GET-only
## Ask

The Traffic Usage page **defaults to the `raw` view** but `raw` is excluded from live streaming (`useWsTrafficUsageLiveEdge`: `enabled = until == null && bucket !== 'raw'`), so the default page shows no `<LiveBadge>` and never subscribes to `trafficUsage`. Make the raw view stream live too.

## Why it was excluded (current behavior)

S6b (#1975) streams the *aggregated* series by merging pushed `aggregateRows` by `windowStart` (`mergeAggregateHeadRows`). The raw view renders ungrouped per-row `rawRows` (a different shape), so streaming it needs a different mechanism — **prepend new raw rows** (like the Connection Events feed's `prependHead` dedup-by-id), not an aggregated head-merge.

## Scope

- Add a raw-rows live-edge path: subscribe `trafficUsage{bucket:'raw', …filters}` at the live edge (`until == null`), and **prepend** the pushed new raw rows into the page's raw-rows state (dedup by a stable row identity; never mutate paged history), showing the `<LiveBadge>` while streaming.
- Note the server side: `bucket=raw` currently returns `rawRows`; the ws push for a raw subscription should deliver the new raw rows for this to work (coordinate with #2040, which is changing raw handling — confirm whether the raw ws push carries rawRows or aggregated rows, and align the page accordingly).
- Relax the `bucket !== 'raw'` guard only for this rawRows-prepend path; keep the aggregated head-merge for 1m/10m/….

## Acceptance
- On the default (raw, live-edge) Traffic Usage view: a `<LiveBadge>` shows, a `subscribe trafficUsage` frame is sent, and new raw rows appear at the top live without a manual refresh; paged history is untouched; pinned past windows (`until != null`) stay GET-only.

Operator-requested 2026-06-28. **Blocked on the page first being usable (the companion HIGH bug — the page currently aborts its own requests).** Relates to #1975 (S6b), #2040, #1148.

sameerparekh:
Closes #2048.

## Problem

The Traffic Usage page defaults to `bucket='raw'` (ungrouped per-host `rawRows`), but `raw` was excluded from the live edge — `useWsTrafficUsageLiveEdge` gated `enabled = until == null && bucket !== 'raw'`. So the **default view** never showed a `<LiveBadge>` and never subscribed to `trafficUsage`; new rows only appeared on a manual reload.

## The rawRows-vs-aggregateRows nuance (resolved)

The `trafficUsage` push (`SpaPush.pushOneParamSet`) aggregated per the subscribed `groupBy` and shipped `rawRows = Nil`. The raw page view renders **rawRows**, not aggregated points — so a naive frontend prepend had nothing to prepend.

Verified against the GET (`UsageRoutes`): `bucket=raw` **without** a groupBy → `buildRaw` → `rawRows`; `bucket=raw` **with** a groupBy (the dashboard gauge's `groupBy=profile`, #2040) → aggregated. I made the ws push mirror that split exactly, keeping the GET and the push SSOT for the same params.

## Changes

**Server** (`SpaPush.scala`): an **ungrouped** `raw` param-set now reads the new raw rows for the ingest period via `trafficRepo.listRawInRange` over the real `[periodStart, periodEnd)` head window (`UsageTraffic.windowFor` for raw = the report period) and ships them as `rawRows`, built by the shared `UsageTraffic.buildRaw`. `raw`+groupBy (gauge) and the fixed buckets keep the existing aggregate path untouched.

**Client**:
- `wsCache.prependRawHead` — prepend new raw rows, dedup by `(periodStart, mac, host)`, no cap (history is paged via the GET, #862).
- `useWsTrafficUsageLiveEdge` now enables for raw at the live edge; raw subscribes **without** `groupBy` (so the server takes its rawRows path, matching the GET which drops groupBy for raw) and prepends into the page's `setRawRows`; fixed buckets keep the aggregated head-merge into `setAggRows`. A pinned `until` stays GET-only for both.
- The page passes both setters; the raw `<tr>` key is now the stable row identity so the prepend reconciles correctly.

**Docs**: updated the design-doc note that claimed a raw subscription is *always* aggregated (no longer true for the ungrouped case).

## Tests (TDD, RED first)

- `SpaWsS4Spec`: an ungrouped `raw` subscription's push carries `rawRows` byte-identical to the GET's `bucket=raw` body for the same window (and the existing `raw`+groupBy → aggregate test still passes). Full S4 suite green (9/9).
- vitest: `prependRawHead` (prepend/dedup/no-truncate); the hook subscribes raw (no groupBy) + prepends + dedups + doesn't touch agg state, stays GET-only on pinned `until`, and re-subscribes/flips push target on raw↔1m; page-level — default raw view subscribes `trafficUsage{bucket:raw}`, shows the `<LiveBadge>`, and prepends a pushed row above GET history; pinned `?until=` stays GET-only (no badge/subscribe). Full web suite green (514), lint + build clean.

### DIFF
--- a/api/src/routes/SpaPush.scala
+++ b/api/src/routes/SpaPush.scala
@@ -388,41 +388,74 @@ object SpaPush {
         // carry the true interval.
         val (headStart, headEnd) =
           UsageTraffic.windowFor(periodStart, periodEnd, parsed.bucket, parsed.zone)
-        val resolvedMacs = UsageTrafficQuery.resolveMacs(parsed.macs, parsed.profileIds, devices)
-        val filterEmpty  = parsed.filterRequested && resolvedMacs.isEmpty
-        val aggregate    =
-          if (filterEmpty) ZIO.succeed(List.empty)
-          else
-            UsageTrafficQuery.aggregate(
-              trafficRepo,
-              rollupRepo,
-              resolvedMacs,
-              headStart,
-              headEnd,
-              parsed.bucket,
-              parsed.groupBySet,
-              parsed.zone,
-              devByMac,
-              profNames,
-              appsByHost,
+        val resolvedMacs   = UsageTrafficQuery.resolveMacs(parsed.macs, parsed.profileIds, devices)
+        val filterEmpty    = parsed.filterRequested && resolvedMacs.isEmpty
+        // #2048: `raw` WITHOUT a groupBy is the Traffic Usage page's per-host inspector (its DEFAULT
+        // view) — it renders per-host `rawRows`, so its live edge must carry the NEW rawRows for the
+        // ingest period (the rows the page prepends), NOT aggregated points. This mirrors the
+        // `GET /api/usage/traffic` split exactly (`UsageRoutes`: `Bucket.Raw if effectiveGroupBy
+        // .isEmpty` → `buildRaw`, else aggregate), so the stream and the GET stay SSOT. `raw` WITH a
+        // groupBy (the dashboard gauge's `groupBy=profile`) keeps the aggregate path below.
+        val isUngroupedRaw = parsed.bucket == UsageTraffic.Bucket.Raw && parsed.groupBySet.isEmpty
+        val computeBody: Task[TrafficUsageResponse] =
+          if (isUngroupedRaw) {
+            // The head window for `raw` IS the real report period `[periodStart, periodEnd)`
+            // (`UsageTraffic.windowFor`), so reading raw rows over `[headStart, headEnd)` returns
+            // exactly the just-ingested period's rows. No cursor/limit: a single ingest period is
+            // naturally bounded by its distinct `(mac, host)` rows (the page pages OLDER history via
+            // the GET; the push only delivers the fresh head). Built via the shared `buildRaw` (SSOT).
+            val rawZ =
+              if (filterEmpty) ZIO.succeed(List.empty[wifihaven.api.usage.TrafficUsageDbRow])
+              else trafficRepo.listRawInRange(resolvedMacs, headStart, headEnd)
+            rawZ.map(rows =>
+              TrafficUsageResponse(
+                bucket = parsed.bucket.code,
+                groupBy = Nil,
+                from = headStart.toString,
+                to = headEnd.toString,
+                tz = parsed.zone.getId,
+                rawRows = UsageTraffic.buildRaw(rows, devByMac, profNames),
+                aggregateRows = Nil,
+                nextCursor = None,
+              ),
             )
-        aggregate.flatMap { rows =>
-          // `from`/`to` describe the head BUCKET window `[headStart, headEnd)` — not wall-clock
-          // "live edge time". `to` is therefore the bucket end and can sit slightly ahead of `now`
-          // for an in-progress bucket; it's metadata only (the client merges by `windowStart` and
-          // derives B/s from the bucket width, never from this `to`).
-          val body = TrafficUsageResponse(
-            bucket = parsed.bucket.code,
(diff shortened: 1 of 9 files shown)


sameerparekh:
## Independent PR review — #2048 raw-rows live edge

Read-only adversarial pass over `git diff origin/main...HEAD` (9 dimensions). Verified the load-bearing claims against source rather than the comments: `UsageTraffic.windowFor` returns `(periodStart, periodEnd)` for `raw` (UsageTraffic.scala:184-197), and the GET split `Bucket.Raw if effectiveGroupBy.isEmpty → buildRaw` is real (UsageRoutes.scala:1106). The push raw path reuses the same `UsageTraffic.buildRaw` the GET uses, and `SpaWsS4Spec` pins byte-identity to the GET — SSOT holds.

### BLOCKERS
(none)

### SHOULD-FIX
1. **History — RED-before-GREEN not visible** (process, dim 2). The change landed in a single commit, so there's no standalone failing-test commit showing red→green for an autonomous session. The tests *do* assert the new behavior and would fail against `origin/main` (the new `SpaWsS4Spec` asserts `pushed.rawRows.nonEmpty`, which old code — aggregating the empty-groupBy raw set into one `aggregateRows` point with `rawRows=Nil` — fails; the hook test asserts a raw subscribe that old code suppressed). So the gate is satisfied in substance; flagging only that the progression isn't in history. Non-gating.

### NITS
1. `web/src/pages/TrafficUsagePage.tsx:547` — the React row key `${periodStart}|${mac}|${host.value}` omits `host.type`, whereas `prependRawHead`'s dedup key (`wsCache.ts:181`) includes it. Harmless (a collision needs same periodStart+mac+host.value across two host types, which the dedup already prevents from coexisting), but the two identity strings could be unified.
2. `SpaPush.scala` raw path reads **all** raw rows for the period with no cursor/limit, unlike the aggregated GET (caps at `rawLimit`) and the `connectionEvents` push (`ConnEventHeadLimit=200`). This is **defensible and not a finding**: the raw push window is a single ~60s ingest period (bounded by that period's distinct `(mac,host)` rows), not the connectionEvents push's up-to-24h `since`-window that motivated *its* cap. Noting for the record; the client also has the #1959 1MiB frame-reassembly cap as a backstop.

- **SSOT (dim 1):** raw push and GET share `buildRaw` + `windowFor`; aggregated path untouched. No re-derivation.
- **Wire (dim 4):** populates the existing `TrafficUsageResponse.rawRows` field — additive behavior on an existing field, no new shape. The dashboard gauge (`raw`+`groupBy=profile`) keeps the aggregate path, confirmed unaffected by the S4 `raw`+groupBy test.
- **Metrics (dim 6):** the raw branch fans out through the already-instrumented `registry.fanOutTrafficUsage` (`spa_ws_push_total{op=trafficUsage}`); no new path needs its own series.
- **Magic constants (dim 9):** none introduced; the raw window is sourced from `windowFor`, not re-hardcoded. The design-doc note that previously claimed raw is *always* aggregated was corrected.
- **Security (dim 8):** existing role-gated registry (admin/adult-only, S4-tested); parameterized `listRawInRange`. No new route.

VERDICT: APPROVE
Cleanly mirrors the GET's raw-vs-aggregated split into the ws push and prepends new rawRows on the client behind a dedup, with the live badge and pinned-window guards intact. SSOT is pinned by a full-stack byte-identity test; web + Scala suites, lint, and build are green. No blockers; the only SHOULD-FIX is a non-gating note about commit history.

